PR #1, opened 2018-05-31
meson options to disable lc0 backends

borg323:
This patch can be used to disable specific lc0 backends using the meson command line. The original intention was to disable the TensorFlow backend, which can be done by specifying -Dtf=false, but the final version allows -Dopencl=false -Dcudnn=false and -Dblas=false as well.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -106,7 +106,7 @@ if get_option('build_backends')
   tf_protobuf_lib = cc.find_library('libprotobuf',
             dirs: tensorflow_libdir, required: false)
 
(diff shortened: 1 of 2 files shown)


borg323:
This is a resend of glinscott/leela-chess#697

mooskagh:
Fyi they are working on a nice feature in meson:
https://github.com/mesonbuild/meson/pull/3376

But it's not there yet.

mooskagh:
Thanks!

PR #5, opened 2018-06-01
lc0-cudnn : add support for fp16 network eval

ankan-ban:
slightly more than 2x speedup (for large batch sizes) on supported hardware, without much loss of precision.
 - currently fp16 is only supported on Volta based GPUs (TitanV, tesla V100, etc)

### DIFF
--- a/src/neural/network_cudnn.cu
+++ b/src/neural/network_cudnn.cu
@@ -111,10 +111,12 @@ class BaseLayer {
   size_t GetOutputSize(int N) const { return bpe_ * N * C * H * W; }
 
   // input2 is optional (skip connection)
-  virtual void Eval(int N, float *output, const float *input,
-                    const float *input2, float *scratch, cudnnHandle_t cudnn,
+  virtual void Eval(int N, void *output, const void *input,
+                    const void *input2, void *scratch, cudnnHandle_t cudnn,
                     cublasHandle_t cublas) = 0;
 
+  static void enableFp16() { fp16_ = true; bpe_ = sizeof(half); }
+  static bool isFp16Enabled() { return fp16_; }
  protected:
   static bool fp16_;
   static size_t bpe_;  // size of each element
@@ -130,9 +132,9 @@ class ConvLayer : public BaseLayer {
   ConvLayer(BaseLayer *ip, int C, int H, int W, int size, int Cin,
             bool relu = false, bool bias = false);
   ~ConvLayer();
-  void LoadWeights(float *pfilter, float *pBias = nullptr);
-  void Eval(int N, float *output, const float *input, const float *input2,
-            float *scratch, cudnnHandle_t cudnn,
+  void LoadWeights(float *pfilter, float *pBias , void *scratch);
+  void Eval(int N, void *output, const void *input, const void *input2,
+            void *scratch, cudnnHandle_t cudnn,
             cublasHandle_t cublas) override;
 
  private:
@@ -141,8 +143,8 @@ class ConvLayer : public BaseLayer {
   const bool use_relu_;
   const bool use_bias_;
 
-  float *biases = nullptr;
-  float *weights = nullptr;
+  void *biases = nullptr;
+  void *weights = nullptr;
 
   cudnnFilterDescriptor_t filter_desc_;
   cudnnConvolutionDescriptor_t conv_desc_;
@@ -157,8 +159,8 @@ class ConvLayer : public BaseLayer {
 class SoftMaxLayer : public BaseLayer {
  public:
   SoftMaxLayer(BaseLayer *ip);
-  void Eval(int N, float *output, const float *input, const float *input2,
-            float *scratch, cudnnHandle_t cudnn,
+  void Eval(int N, void *output, const void *input, const void *input2,
+            void *scratch, cudnnHandle_t cudnn,
             cublasHandle_t cublas) override;
 
  private:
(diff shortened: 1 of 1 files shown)


mooskagh:
Would it be possible instead of converting all `float*` to `void*` to have two template variants, for `float` and `__fp16`, similarly to what I did here: https://github.com/LeelaChessZero/lc0/blob/release/src/neural/network_tf.cc#L299  for tensorflow.
and then use ValueType* instead?  (maybe some better name for ValueType though.. `Float`?)

So, it would be
```cpp
template <typename ValueType>
class CudnnNetwork ...
```
and all helper classes/function would also become templates, and registered like:

```cpp
REGISTER_NETWORK("cudnn", CudnnNetwork<float>, 110)
REGISTER_NETWORK("cudnn-fp16", CudnnNetwork<__fp16>, 105)
```
The drawback though is that best configuration  won't be automatically chosen.
But on other hand, as cudnn-fp32 has better precision, everyone should be able to select it.

So, with this approach, if there is a function which needs completely different implementation, then template specialization is useful. If however there is difference in small details,  this is possible:
```cpp
if (std::is_same<float, ValueType>::value) {  // In C++17 (or 14?)  ::value is not necessary
   // Do something float-specific.
}
```


Also this file becomes bit, maybe at that point it deserves a separate `src/neural/cudnn/` directory and it is worth splitting into several files?

ankan-ban:
Volta Tensor cores (for FP16 math) have much higher throughput than fp32. 
E.g: see table in https://www.anandtech.com/show/12170/nvidia-titan-v-preview-titanomachy

Just slightly more than 2X (instead of 8X is because): 
1. we aren't getting full utilization yet (need even bigger batch size).
2. cudnn can't use winograd convolution algorithm with tensor cores.

jnewlin12345:
I'm looking at the anandtech link and see:

Single Precision: 13.8 TFLOPS
Half Precision: 27.6 TFLOPS

Where is the 8x coming from?

ankan-ban:
Tensor Performance(Deep Learning) | 110 TFLOPS

Error323:
Hi @ankan-ban are you using TensorRT? Could you push what you have? I'd like to dive in :)

ankan-ban:
Haven't tried tensorRT yet. It's linux only and I am on Windows right now :(. I think it's time for me to setup a linux partition and compare performance of TensorRT with cudnn.

I was trying out in8 with cudnn where the adjustment/calibration of weights has to be done manually. (I believe TensorRT does it for you?)

Error323:
That's correct, tensorrt allows you to calibrate automatically given a dataset. I did some preliminary testing here https://github.com/glinscott/leela-chess/issues/52#issuecomment-362252598

src/neural/network_cudnn.cu
```diff
@@ -329,54 +400,97 @@ __global__ void expandPlanes_kernel(float *output, const uint64_t *masks,
...
   // allocate memory for weights (filter tensor) and biases
-  size_t weightSize = bpe_ * Cin * C * filter_size_ * filter_size_;
+  size_t weightSize = sizeof(DataType) * Cin * C * filter_size_ * filter_size_;
   reportCUDAErrors(cudaMalloc(&weights, weightSize));
 
-  size_t biasSize = bpe_ * C;
+  size_t biasSize = sizeof(DataType) * C;
   reportCUDAErrors(cudaMalloc(&biases, biasSize));
 
+  bool fp16 = std::is_same<half, DataType>::value;
```
mooskagh: const

src/neural/network_cudnn.cu
```diff
@@ -154,64 +153,63 @@ class ConvLayer : public BaseLayer {
...
-  void Eval(int N, float *output, const float *input, const float *input2,
-            float *scratch, cudnnHandle_t cudnn,
+  void Eval(int N, DataType *output, const DataType *input, const DataType *input2,
+            void *scratch, cudnnHandle_t cudnn,
             cublasHandle_t cublas) override;
 
  private:
   const bool use_relu_;
+
+  // always in float irrespective of DataType
```
mooskagh: Sorry, for nitpicking, but consider changing all comments to proper English sentences. (starting with capital letter, ending with a period).

Also consider running `clang-format -I -style=google network_cudnn.cu`  but I can do that myself if you don't have clang installed.

src/neural/network_cudnn.cu
```diff
@@ -419,30 +536,55 @@ ConvLayer::ConvLayer(BaseLayer *ip, int C, int H, int W, int filter, int Cin,
...
 }
 
-void ConvLayer::Eval(int N, float *output, const float *input,
-                     const float *input2, float *scratch, cudnnHandle_t cudnn,
-                     cublasHandle_t cublas) {
+template<typename DataType>
+void ConvLayer<DataType>::Eval(int N, DataType *output, const DataType *input,
+                               const DataType *input2, void *scratch, cudnnHandle_t cudnn,
+                               cublasHandle_t cublas) {
+  bool fp16 = std::is_same<half, DataType>::value;
```
mooskagh: also const

src/neural/network_cudnn.cu
```diff
@@ -641,14 +944,17 @@ class CudnnNetworkComputation : public NetworkComputation {
...
+  CudnnNetwork<DataType> *network_;
 };
 
+template <typename DataType>
 class CudnnNetwork : public Network {
  public:
   CudnnNetwork(Weights weights, const OptionsDict &options) {
     gpuId_ = options.GetOrDefault<int>("gpu", 0);
 
+    //int tryFp16 = options.GetOrDefault<int>("fp16", 0);
```
mooskagh: not needed?

src/neural/network_cudnn.cu
ankan-ban: Don't have clang.

PR #6, opened 2018-06-01
GetOptions fix on OpenCL.

frpays:
This commit contains a minor fix on opencl (getoptions<bool>) along with a minor code cleanup (mostly removing auxiliaries variables) on both blas/opencl.

### DIFF
--- a/src/neural/network_blas.cc
+++ b/src/neural/network_blas.cc
@@ -61,7 +61,7 @@ class BlasComputation : public NetworkComputation {
       float value = plane.value;
       const uint64_t one = 1;
       for (int i = 0; i < 64; i++)
-        input_data_[index++] = ((plane.mask & (one << i)) == 0) ? 0 : value;
+        input_data_[index++] = plane.mask & (one << i) ? value : 0;
     }
 
     std::vector<float> policy_data(weights_.ip_pol_b.size());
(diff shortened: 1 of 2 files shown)


src/neural/network_blas.cc
```diff
@@ -61,7 +61,7 @@ class BlasComputation : public NetworkComputation {
       float value = plane.value;
       const uint64_t one = 1;
       for (int i = 0; i < 64; i++)
-        input_data_[index++] = ((plane.mask & (one << i)) == 0) ? 0 : value;
+        input_data_[index++] = plane.mask & (one << i) ? value : 0;
```
mooskagh: Actually, MS visual studio requires == 0, else it's warning:
https://msdn.microsoft.com/en-us/library/b6801kcy.aspx

It's silly, I agree.
I think it's fine to keep it with == 0, as there are plenty of MSVC warnings in other places and noone cleaned them up yet.

src/neural/network_opencl.cc
```diff
@@ -152,10 +132,9 @@ class OpenCLNetwork : public Network {
       : weights_(weights), params_(), opencl_(), opencl_net_(opencl_) {
     params_.gpuId = options.GetOrDefault<int>("gpu", -1);
     params_.verbose = options.GetOrDefault<bool>("verbose", false);
-    params_.force_tune = options.GetOrDefault<int>("force_tune", false);
-    params_.tune_only = options.GetOrDefault<int>("tune_only", false);
-    params_.tune_exhaustive =
-        options.GetOrDefault<int>("tune_exhaustive", false);
+    params_.force_tune = options.GetOrDefault<bool>("force_tune", false);
```
mooskagh: Alternatively, 
`options.GetOrDefault<string>("tune", "auto")`, with acceptable values "force", "only", "auto" and "exhaustive".

But I'm not convinced that it would be better.

frpays: Not too convinced too. You could want to force an exhaustive tune, or force a non-exhaustive tune. Almost all 2^3 combinations make sense here.

src/neural/network_blas.cc
frpays: Rolled back. Using the !=0 which is as efficient on assembly level.

PR #10, opened 2018-06-02
Restore transforms.* to neural, renamed with blas_ prefix

dubslow:


mooskagh:
Maybe then move all blas-related files into neural/blas?
I think all non-trivial (having more than not one file) backends deserve own directory.

(The same is true for cudnn backend, I think it needs a separate dir).

frpays:
Sorry, I just realised this PR existed.
Replying to your comment Alexander, I moved the transforms file in BLAS as part as rewiring the vendor-specific routines.
I did not meant to supersede Dubslow development again...

dubslow:
Closed in favor of #21

PR #12, opened 2018-06-02
Fix castling handling, update default params.

mooskagh:


src/mcts/search.cc
```diff
@@ -56,26 +56,31 @@ const int kSmartPruningToleranceMs = 200;
...
+
+  options->Add<IntOption>(kMiniBatchSizeStr, 1, 1024, "minibatch-size") = 1;
+  options->Add<IntOption>(kMaxPrefetchBatchStr, 0, 1024, "max-prefetch") = 32;
+  options->Add<FloatOption>(kCpuctStr, 0.0f, 100.0f, "cpuct") = 1.2f;
+  options->Add<FloatOption>(kTemperatureStr, 0.0f, 100.0f, "temperature") =
+      0.0f;
   options->Add<IntOption>(kTempDecayMovesStr, 0, 100, "tempdecay-moves") = 0;
   options->Add<BoolOption>(kNoiseStr, "noise", 'n') = false;
   options->Add<BoolOption>(kVerboseStatsStr, "verbose-move-stats") = false;
   options->Add<BoolOption>(kSmartPruningStr, "smart-pruning") = true;
```
Tilps: Probably not worth doing anything about, but I'll mention it anyway. This list of 'safe' defaults seems kind of targeted at matching training parameters, except without temperature and noise.  I'm not sure where smart-pruning should fall on that scale of one side or the other.

src/neural/CL/cl2.hpp
```diff
@@ -86,39 +86,39 @@
  * fixes in the new header as well as additional OpenCL 2.0 features.
```
Tilps: IIRC last time this file was autoformatted it broke the OpenCL backend.  Suggest reverting.

PR #21, opened 2018-06-02
OpenBLAS: verbose mode, set threads/procs, use specific vendor routines.

frpays:
This development completes the blas/opencl port.

Previously lczero called specific vendor (MKL, OpenBlas) routines to make the BLAS layer to use 1 thread.
This routine calls were still commented in the port. 

Now we can use 2 options:
* `verbose`, bool (default false)
* `blas_cores`, int, (default int)

When verbose is true, the backend displays the vendor is built against, and when available some specific informations, like the version or how many threads it can use.
Matrices are too small for the computing to be accelerated by multithreading. Maybe in the future if we can develop batch computing.

The development has been tested on:
- MacOS sierra
- Linux with MKL
- Linux with openblas

Finally, note : 
- that the meson build has been augmented with an option `mkl_include`
- that blas.h, transforms.cc, transform.hh have  been moved to neural/BLAS.

Here are the output with different BLAS vendors.

`

> ./lc0 --backend=blas  --backend-opts="verbose=true,blas_cores=6"
       _
|   _ | |
|_ |_ |_| built Jun  2 2018
go nodes 100
Found network file: ./weights.txt
Creating backend [blas]...
BLAS vendor: OpenBlas.
OpenBlas [NO_LAPACKE DYNAMIC_ARCH NO_AFFINITY Nehalem].
OpenBlas found 12 Nehalem core(s).
OpenBLAS using 6 core(s) for this backend.

> ./lc0 --backend=blas  --backend-opts="verbose=true,blas_cores=6"
       _
|   _ | |
|_ |_ |_| built Jun  2 2018
go nodes 100
Found network file: ./weights.txt
Creating backend [blas]...
BLAS vendor: MKL.
MKL Intel(R) Math Kernel Library Version 2018.0.3 Product Build 20180406 for Intel(R) 64 architecture applications.
MKL platform: Intel(R) 64 architecture, processor: Intel(R) Streaming SIMD Extensions 4.2 (Intel(R) SSE4.2) enabled processors.
MKL can use up to  12 thread(s).
MKL using 6 thread(s) for this backend.

> ./lc0 --backend=blas --backend-opts="verbose=true"
       _
|   _ | |
|_ |_ |_| built Jun  2 2018
go nodes 100
Found network file: ./aaa
Creating backend [blas]...
BLAS vendor: Apple vecLib
`


src/neural/BLAS/transforms.h
```diff
@@ -25,35 +25,32 @@ namespace lczero {
 
 class Transforms {
  public:
-  
-  
   static constexpr auto kWinogradAlpha = 4;
   static constexpr auto kWinogradTile = kWinogradAlpha * kWinogradAlpha;
```
dubslow: perhaps these can be moved to blas.h? (not a problem with current PR)

src/neural/CL/OpenCLTuner.cc
```diff
@@ -30,7 +30,7 @@
 #include "OpenCLParams.h"
 #include "OpenCLTuner.h"
 
-#include "utils/blas.h"
+#include "../BLAS/blas.h"
```
dubslow: Wouldn't this be better as `neural/BLAS/blas.h` ?

src/neural/network_blas.cc
```diff
@@ -16,9 +16,10 @@
  along with Leela Chess.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-#include "neural/CL/transforms.h"
-#include "neural/factory.h"
 #include "neural/network.h"
+#include "BLAS/blas.h"
+#include "BLAS/transforms.h"
+#include "neural/factory.h"
```
dubslow: These `neural/` leading parts can be dropped

src/neural/network_blas.cc
```diff
@@ -188,8 +188,11 @@ class BlasNetwork : public Network {
  public:
   virtual ~BlasNetwork(){};
 
-  BlasNetwork(const Weights& weights, const OptionsDict& /* options */)
+  BlasNetwork(const Weights& weights, const OptionsDict& options)
       : weights_(weights) {
+    bool verbose = options.GetOrDefault<bool>("verbose", false);
```
dubslow: I think I'd prefer this to be true by default, explicit is better than implicit, and letting users know exactly which backend is currently running is very useful imo

src/neural/network_blas.cc
```diff
@@ -220,23 +223,51 @@ class BlasNetwork : public Network {
       Transforms::InvertBatchNormStddev(conv2.bn_stddivs);
     }
 
-    Transforms::OffsetBatchNormMeans(weights_.policy.bn_means, weights_.policy.biases);
+    Transforms::OffsetBatchNormMeans(weights_.policy.bn_means,
+                                     weights_.policy.biases);
     Transforms::InvertBatchNormStddev(weights_.policy.bn_stddivs);
 
-    Transforms::OffsetBatchNormMeans(weights_.value.bn_means, weights_.value.biases);
+    Transforms::OffsetBatchNormMeans(weights_.value.bn_means,
```
dubslow: Is putting them on two lines really necessary? It fits just fine on one line.

src/neural/network_opencl.cc
```diff
@@ -94,8 +94,8 @@ class OpenCLComputation : public NetworkComputation {
     policy_data_.emplace_back(move(policy_data));
 
     // Now get the score
-    double winrate =
-        Transforms::Innerproduct(weights_.ip2_val_w, value_data_) + weights_.ip2_val_b[0];
+    double winrate = Transforms::Innerproduct(weights_.ip2_val_w, value_data_) +
+                     weights_.ip2_val_b[0];
```
dubslow: If splitting here, then I'd put the `+` at the beginning of the second line

frpays:
Sorry Dubslow, there is king of misunderstanding here.
This indent changes comes from the clang-format google that Alexander wish the files to be indented with.

dubslow:
Sure, I know they were made by the linter, but that doesn't mean we can't tweak the linter's output. And I think at least some of it is worth tweaking (though I suppose it doesn't really matter much)

mooskagh:
I weakly prefer `opencl/` and `blas/` over `CL/` and `BLAS/` too, but I don't insist, depending how lazy you feel at the moment.

src/neural/BLAS/transforms.h
frpays: Yes, they are both an acronyms (Basic Linear Algebra Sub-routine) without using google :)
 I would prefer  OpenCL than CL for the folder inside neural.

src/neural/CL/OpenCLTuner.cc
frpays: I don't know. There is 1 pro 1 cons for both possible methods.

src/neural/network_blas.cc
frpays: All the formatting is not mine... (clang-format)

src/neural/network_blas.cc
frpays: Are you sure? I also fear that it can disturb UCI clients.

frpays:
The CL can be renamed to OpenCL. I see no objection to it, and I would rather prefer.
But not now, as the other PRs could fail.

src/neural/CL/OpenCLTuner.cc
mooskagh: I'm not sure if it's better, but style guide https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes indeed says to have full include paths from root.

src/neural/network_blas.cc
```diff
@@ -220,23 +223,51 @@ class BlasNetwork : public Network {
...
 #ifdef USE_OPENBLAS
-// openblas_set_num_threads(1);
-// printf("BLAS Core: %s\n", openblas_get_corename());
+    int num_procs = openblas_get_num_procs();
+    blas_cores = std::min(num_procs, blas_cores);
+    openblas_set_num_threads(blas_cores);
+    if (verbose) {
+      const char* core_name = openblas_get_corename();
+      const char* config = openblas_get_config();
+      printf("BLAS vendor: OpenBlas.\n");
```
mooskagh: Please output that to stderr rather than stdout.
If it's during initialization only, it's fine to remove "verbose" parameter and output always.

Also consider using C++ `std::cerr` instead. (but I don't insist)

src/neural/network_blas.cc
mooskagh: That's because https://google.github.io/styleguide/cppguide.html#Line_Length sets line limit of 80 chars.
Very handy, that way I can have three source panes open side by side.

src/neural/network_opencl.cc
mooskagh: clang-format won't agree. :)

src/neural/CL/OpenCLTuner.cc
frpays: Pushed #include "neural/BLAS/blas.h"

src/neural/network_blas.cc
frpays: Done, thanks

src/neural/network_blas.cc
frpays: Yes, that has to be discussed. 
I have no opinion of stdout/stderr for initial verbose comments...
But if we decide something, let use change all the backends at the same time.

src/neural/network_opencl.cc
frpays: About the fact that thread=1 is better, but we still offer the parameters. I guess,
it enable users to check by themselves...

src/neural/network_blas.cc
frpays: sorry did get you there, "cleanup the printf laying around"

src/neural/network_blas.cc
dubslow: nah lczero has been outputting similar info and it hasn't ever caused problems

src/neural/network_blas.cc
dubslow: do `rgrep printf`, and replace everything it finds with standard C++ io

src/neural/network_blas.cc
mooskagh: Outputting to stdio is surely bad for uci. But to stderr is fine.

src/neural/network_blas.cc
frpays: I really dislike C++ io which is an abomination imho and I would stick with printf(stderr, ...)

frpays:
verbose is now true by default and information goes to stderr.

src/neural/network_opencl.cc
```diff
@@ -16,7 +16,9 @@
  along with Leela Chess.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-#include "neural/CL/transforms.h"
+#include "BLAS/blas.h"
```
mooskagh: `neural/BLAS/blas.h`

And maybe still `blas` and `opencl` instead of `BLAS` and `OPENCL`. But if you don't want to do that, I can do that myself later.

src/neural/network_blas.cc
```diff
@@ -220,23 +223,51 @@ class BlasNetwork : public Network {
...
 #ifdef USE_OPENBLAS
-// openblas_set_num_threads(1);
-// printf("BLAS Core: %s\n", openblas_get_corename());
+    int num_procs = openblas_get_num_procs();
+    blas_cores = std::min(num_procs, blas_cores);
+    openblas_set_num_threads(blas_cores);
+    if (verbose) {
+      const char* core_name = openblas_get_corename();
+      const char* config = openblas_get_config();
+      fprintf(stderr, "BLAS vendor: OpenBlas.\n");
```
mooskagh: It's still better to do it using `std::cerr`, but if I'm find with doing that myself later.

src/neural/network_blas.cc
```diff
@@ -16,9 +16,10 @@
  along with Leela Chess.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-#include "neural/CL/transforms.h"
-#include "neural/factory.h"
-#include "neural/network.h"
+#include "network.h"
```
mooskagh:
```
neural/network.h
neural/blas/blas.h
etc
```

src/neural/network_blas.cc
frpays: I said earlier that it was find to do, but after the merge of the 2 PRs.
Otherwise bad things (conflicts) may happens.

src/neural/network_blas.cc
frpays: Is that necessary or a suggestion?

frpays:
Can we move on and merge the thing?

src/neural/network_blas.cc
frpays: I let you do that change printf to cerr.
I my humble opinion, printf is superior or iostream and I always use printf on purpose, not because I don't know better.
iostream is a disaster of an api.

mooskagh: I'd say it's necessary (style guide is style guide..), and it's easy to do in this PR.
Other suggestions is fine to do in other PRs.

frpays:
I have done the includes fix for network_blas.

src/neural/BLAS/transforms.cc
```diff
@@ -18,7 +18,7 @@
 
 #include "transforms.h"
 
-#include "utils/blas.h"
+#include "blas.h"
```
mooskagh: Is it system header or from sources?
If system should be in angle brackets, if it's in sources, full path.
The same with transforms.h above.

src/neural/CL/OpenCLTuner.cc
```diff
@@ -30,7 +30,7 @@
 #include "OpenCLParams.h"
 #include "OpenCLTuner.h"
```
mooskagh: Full path, or square brackets, depending on what is meant.

frpays: Ok  - I do it now. But it's now pretty sure we will have a conflict between the 2 PR.

mooskagh: Hm, I don't see any changes, did you forget to push your commit?

mooskagh: Found them in another PR!

frpays:
Resolved the conflicts du to merging of 25.
This conflicts were avoidable. doing tree separate PR.

mooskagh:
@dubslow Please don't merge PRs with unresolved comments.

dubslow:
The only unresolved comment I see is yours saying `It's still better to do it using std::cerr, but if I'm find with doing that myself later.`, which based on the last part I'm not sure I would call "unresolved".

All the other comments still visible weren't actual issues with the code, only questions from me that were answered, or one comment by me specifically for future PRs.

mooskagh:
Could anyone make a PR with changing `fprint(stderr` to `std:cerr`?
I said initially that I would do that later, but for now it I really don't have time, and having inconsistent style is bad.

frpays:
We sould rather use fprintf for all the project.

-- blas uses printf
-- check uses printf ---c udnn uses printf
-- As c++ developer, I find stdio superior to iostream.

So I vote printf.


On Sun, Jul 15, 2018 at 9:05 PM, Alexander Lyashuk <notifications@github.com

dubslow:
Why should we use fprintf?

frpays:
I find it superior: easier to use. I don't like iostream: there a lot of
things so easy in printf that need obscure modifier in iostreeam.
For the moment very backend every backed has been coded with printf.

dubslow:
I agree on this dislike of iostream, it's not a very good design, but it is safer to use from an engineering perspective (even if not easier to use from a human perspective).

PR #22, opened 2018-06-02
Fix crash where uci command "go nodes" is entered.

frpays:
The UCI loop is try-catched but atoi (an old std-c routine **crashes**).

> ./lc0 --backend="blas"
       _
|   _ | |
|_ |_ |_| built Jun  2 2018
go nodes
libc++abi.dylib: terminating with uncaught exception of type std::invalid_argument: stoi: no conversion
Abort trap: 6

Now:

go nodes xyz
error invalid value xyz
go nodes
error expected value after nodes

### DIFF
--- a/src/chess/uciloop.cc
+++ b/src/chess/uciloop.cc
@@ -96,10 +96,18 @@ int GetNumeric(
   	throw Exception("Unexpected error");
   }
   std::string str=iter->second.c_str();
-  if (str.empty()) {
-    throw Exception("expected value after "+key);
+  try {
+    if (str.empty()) {
+      throw Exception("expected value after "+key);
+    }
+    return stoi(str);
   }
-  return stoi(str);
+  catch(std::invalid_argument& e){
+    throw Exception("invalid value "+str);
+  }
+ 
+  
+ 
 }
 
 
--- a/src/chess/uciloop.cc
+++ b/src/chess/uciloop.cc
@@ -88,28 +88,22 @@ std::string GetOrEmpty(
   return iter->second;
 }
 
-int GetNumeric(
-    const std::unordered_map<std::string, std::string>& params,
-    const std::string& key) {
+int GetNumeric(const std::unordered_map<std::string, std::string>& params,
+               const std::string& key) {
   auto iter = params.find(key);
   if (iter == params.end()) {
-  	throw Exception("Unexpected error");
+    throw Exception("Unexpected error");
   }
-  std::string str=iter->second.c_str();
+  std::string str = iter->second;
   try {
     if (str.empty()) {
-      throw Exception("expected value after "+key);
+      throw Exception("expected value after " + key);
     }
     return stoi(str);
+  } catch (std::invalid_argument& e) {
+    throw Exception("invalid value " + str);
   }
-  catch(std::invalid_argument& e){
-    throw Exception("invalid value "+str);
-  }
- 
-  
- 
 }
-
 
 bool ContainsKey(const std::unordered_map<std::string, std::string>& params,
                  const std::string& key) {
--- a/src/chess/uciloop.cc
+++ b/src/chess/uciloop.cc
@@ -94,12 +94,12 @@ int GetNumeric(const std::unordered_map<std::string, std::string>& params,
   if (iter == params.end()) {
     throw Exception("Unexpected error");
   }
-  std::string str = iter->second;
+  const std::string& str = iter->second;
   try {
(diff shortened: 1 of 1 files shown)


dubslow:
Wouldn't it be simpler to just improve the macro? Something like:

```
#define OPTION(x)                                        \
  if (ContainsKey(params, #x)) {                     \
    go_params.x = GetOrEmpty(params, #x)); \
    if (!go_params.x.empty()) {                        \
      go_params.x = std::stoi(go_params.x);    \
    } else {                                                       \
      throw Exception("expected value after "+key); // optional \
    }                                                                \
   }	   ```

**Comment by frpays** [comment]
My instinct is to simplify the macro to the minimum.
With a method, the code is actually shared (not inlined) and you can check-it with your preferred development environment.
In fact, we should look to remove the macro altogether. This is not a very modern construct.

File: src/chess/uciloop.cc [hunk 1]
```diff
@@ -88,6 +88,29 @@ std::string GetOrEmpty(
...
+    if (str.empty()) {
+      throw Exception("expected value after "+key);
+    }
+    return stoi(str);
+  }
+  catch(std::invalid_argument& e){
+    throw Exception("invalid value "+str);
+  }
+ 
+
```
mooskagh: Formatting. :)
Three empty lines?.. Why?

src/chess/uciloop.cc
```diff
@@ -88,6 +88,29 @@ std::string GetOrEmpty(
...
 }
 
+int GetNumeric(
+    const std::unordered_map<std::string, std::string>& params,
+    const std::string& key) {
+  auto iter = params.find(key);
+  if (iter == params.end()) {
+  	throw Exception("Unexpected error");
+  }
+  std::string str=iter->second.c_str();
```
mooskagh: why `.c_str()`?

src/chess/uciloop.cc
```diff
@@ -88,6 +88,29 @@ std::string GetOrEmpty(
...
+  try {
+    if (str.empty()) {
+      throw Exception("expected value after "+key);
+    }
+    return stoi(str);
```
mooskagh: std::stoi

frpays:
Changes requested addressed (the code was needlessly creating a temp std::strind).
Apparently, it is advised to use stoi over atoi. 
Finally I clang-formatted the function.

frpays:
> > ./lc0 
>        _
> |   _ | |
> |_ |_ |_| built Jun  2 2018
> go nodes 10
> Found network file: ./weights.txt
> Creating backend [opencl]...
> info depth 1 seldepth 2 time 747 nodes 2 score cp 2 hashfull 0 nps 2 pv e2e4 c7c5
> info depth 1 seldepth 3 time 863 nodes 5 score cp 8 hashfull 0 nps 5 pv e2e4 c7c5 c2c3
> info depth 1 seldepth 4 time 1057 nodes 13 score cp 10 hashfull 0 nps 12 pv e2e4 c7c5 g1f3 d7d6
> info depth 1 seldepth 4 time 1057 nodes 13 score cp 10 hashfull 0 nps 12 pv e2e4 c7c5 g1f3 d7d6
> bestmove e2e4
> go nodes
> error expected value after nodes
> go nodes ddsdsdfd
> error invalid value ddsdsdfd
>

src/chess/uciloop.cc
```diff
@@ -88,6 +88,23 @@ std::string GetOrEmpty(
   return iter->second;
 }
 
+int GetNumeric(const std::unordered_map<std::string, std::string>& params,
+               const std::string& key) {
+  auto iter = params.find(key);
+  if (iter == params.end()) {
+    throw Exception("Unexpected error");
+  }
+  std::string str = iter->second;
```
mooskagh: Actually,
```cpp
const std::string& str = iter->second;
```

src/chess/uciloop.cc
```diff
@@ -88,6 +88,23 @@ std::string GetOrEmpty(
...
+  try {
+    if (str.empty()) {
+      throw Exception("expected value after " + key);
+    }
+    return stoi(str);
```
mooskagh: `std::stoi`

src/chess/uciloop.cc
frpays: addressed.

src/chess/uciloop.cc
frpays: addressed.

PR #23, opened 2018-06-02
Change isready to call UpdateNetwork so backends initialize

killerducky:
I think UCI expects isready to block and do all long operations, so for us that should include backend initialization. From the examples I think all setoptions should be done before this so we will know the network the user wants by now.

I made engine::UpdateNetwork public instead of private to do this, let me know if you want to do it differently. I also called EnsureOptionsSent, not sure if we need that or other things.

(Originally https://github.com/glinscott/leela-chess/pull/686)

### DIFF
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -222,6 +222,8 @@ void EngineLoop::CmdUci() {
 }
 
 void EngineLoop::CmdIsReady() {
+  EnsureOptionsSent();
+  engine_.UpdateNetwork();
   engine_.EnsureReady();
   SendResponse("readyok");
 }
--- a/src/engine.h
+++ b/src/engine.h
@@ -43,6 +43,8 @@ class EngineController {
     search_.reset();
   }
 
+  void UpdateNetwork();
+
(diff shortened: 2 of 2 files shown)


mooskagh:
Actually, it's `ucinewgame`'s job, as per http://wbec-ridderkerk.nl/html/UCIProtocol.html:
```
   As the engine's reaction to "ucinewgame" can take some time the GUI should always send "isready"
   after "ucinewgame" to wait for the engine to finish its operation.
```

It doesn't feel right that `isready` triggers some processing while it only should wait for other commands to complete.

Also I've fixed tree release in `position` command and now it happens after `bestmove` is returned.

But if you insist on doing that in `isready`, then it's better to call UpdateNetwork() from EnsureReady() rather than from CmdIsReady().

killerducky:
Oh you're right about ucinewgame. I guess we can just wait to see if any GUI users have problems. I know Fritz was picky about timeouts and didn't work well with lzcero. 

For now I'll just close this.

PR #24, opened 2018-06-02
Moved OpenCL code in .opencl files.

frpays:
* created:

clsource/config.opencl
clsource/convolve1.opencl
clsource/convolve3.opencl

* made OpenCL.cc invulnerable to future clang-format

* clang-format-ted OpenCL.cc

The critical change enabling OpenCL.cc to be invulnerable to clang-format is to be to write:

`const std::string sourceCode_sgemm =
    sourceCode_blast_level3_common + sourceCode_blast_level3_xgemm_part1 +
    sourceCode_blast_level3_xgemm_part2 + sourceCode_blast_level3_xgemm_part3 +
    sourceCode_blast_level3_xgemm_batched;`

Instead of:

`const std::string sourceCode_sgemm = R"
#include ...
#include ...
`


src/neural/CL/OpenCL.cc
```diff
@@ -1,527 +1,183 @@
...
-  
+
   auto weightSize = size * sizeof(decltype(converted_weights)::value_type);
   m_layers.back().weights.emplace_back(
-                                       m_opencl.m_context,
-                                       CL_MEM_COPY_HOST_PTR | CL_MEM_READ_ONLY,
-                                       weightSize,
-                                       const_cast<net_t*>(converted_weights.data()));
+      m_opencl.m_context, CL_MEM_COPY_HOST_PTR | CL_MEM_READ_ONLY, weightSize,
+      const_cast<net_t*>(converted_weights.data()));
```
dubslow: In reference to my previous comment on different PR, if it *has* to be on more than one line, than only one argument per line is preferable IMO.

frpays: It's clang-format google.

src/neural/CL/OpenCL.cc
```diff
@@ -530,123 +186,87 @@ void OpenCL_Network::forward(const std::vector<net_t>& input,
...
-                inBuffer,
-                VBuffer,
-                MBuffer,
-                conv_weights,
-                nullptr,
-                bn_weights,
-                skip_in_trans, skip_next_in_trans, true);
+      convolve3(layer.channels, layer.outputs, inBuffer, inBuffer, VBuffer,
+                MBuffer, conv_weights, nullptr, bn_weights, skip_in_trans,
+                skip_next_in_trans, true);
```
dubslow: Undecided about which is better here...

frpays: again, It's clang-format google.

frpays: I did not indent/reindent this file by hand.

src/neural/CL/OpenCL.cc
frpays: The point of this PR is to pass clang-format on the OpenCL.cc file.

src/neural/CL/OpenCL.cc
```diff
@@ -53,19 +53,31 @@ const std::string sourceCode_convolve3 =
 #include "clsource/convolve3.opencl"
     ;
 
-// Important: keep the follow includes in the same orders.
-// Beware of using a formatting tool like clang-format
-// that could mess with this order.
-// The correct order is: common/part1/part2/part3/batched.
```
dubslow: This comment is incredibly useful to anyone reading this code for the first time. Perhaps remove the comment about linters, but the part about preserving the correct order is very important and should be kept IMO.

frpays: I removed it because I found a way around and it was no longer necessary.

frpays: See the "critical change" in the initial comment of the PR.

frpays: Now you can clang-format the OpenCL.cc file all you want.

frpays: Fair enough.

mooskagh: You can do
```
// clang-format off
around the code you don't want to be touched by clang-format
// clang-format on
```

PR #25, opened 2018-06-02
Revert cpscore to raw valued head output (scaled by 5)

dubslow:


mooskagh:
Why?

fli:
I like @dubslow's change better. You're probably already completely winning if you're 500 cp up. With the old formula that's only ~84%
Here are the 2 functions compared https://www.desmos.com/calculator/ifm065uyru (I scaled Q to 0-1 because it's more intuitive for human me). The bottom 2 functions in that graph are related below:

I found this old post (https://chessprogramming.wikispaces.com/Pawn+Advantage,+Win+Percentage,+and+ELO) which tries to match material values to winrate. Swapping x and y gives the 4th function in my linked graph and using 3.5 instead of 5 as a scaling factor gives almost a perfect fit to that function. **So I suggest we scale by 3.5.** If we really wanted to, we could repeat the analysis with Leela games, and maybe only use pawn material differences so that we make no assumptions about the values of other pieces and further tune the scaling factor.

dubslow:
I believe the current `tan` version was closely tuned to match Stockfish for close positions, i.e. |score| < 2, and it's that regime that the 5x scale matches well with old function.

Either way, I think that link is also good evidence that using `tan` isn't very useful either, at least relative to atanh/log (thanks for that btw).

killerducky:
How does the formula behave in the adjudicate win area? I think TCEC adjudicates around 650 centipawns. Other users may have their own tournaments setup to adjudicate at different thresholds. So there is a defacto standard for the centipawn scale.

Thanar2:
For the reasons killerducky stated, it is quite possible that lc0 would not be able to compete in TCEC 13 (or other such tournaments) without the ability to give some centipawn based eval output.

dubslow:
@Thanar2 ??? This is centipawn output.

dubslow:
I rebased on current master.

See here to compare: https://www.desmos.com/calculator/mdm2pu2ozx

fli:
If the goal is to match stockfish eval, I found this (albeit from 2014): https://chesscomputer.tumblr.com/post/98632536555/using-the-stockfish-position-evaluation-score-to. Looking at that graph, 250cp advantage is ~85% winrate whereas with a 7x scaling factor it is only ~67%.

Houdini also tunes its evaluation a bit more aggresively:

> The evaluations have again been calibrated to correlate directly with the win expectancy in the position. A +1.00 pawn advantage gives a 75% chance of winning the game against an equal opponent at blitz time control. At +1.50 the engine will win 90% of the time, and at +2.50 about 99% of the time. To win nearly 50% of the time, you need and advantage of about +0.60 pawn. (http://www.cruxis.com/chess/houdini.htm)

dubslow:
Well theoretically the scale of 5 matches the old formula, which was tuned to stockfish as well supposedly. So perhaps 7 is a good compromise between those two?

At any rate I think 7 is good enough for now. Maybe further use will indicate tweaks but this is a good first order guess, good enough to accept the PR I think and get the value head score as cpscore merged.

fli:
I would prefer 3-4, since the winrate matches analysis done by other people and the evals make more conceptual sense to me (>+5 should almost be a certain win), so I'd pick 5 over 7. But having said that, both 5 or 7 are reasonable. I'm fine with just picking something and then if people start complaining about Leela over/undervaluing positions vs other engines we can adjust as needed.

Here's some winrates at +6.50 at different scaling values for reference:

Scale | Winrate@6.5
--- | ---
2.5 | 99.5%
3 | 98.7%
3.5 | 97.6%
4 | 96.3%
4.5 | 94.7%
5 | 93.1%
5.5 | 91.4%
6 | 89.7%
6.5 | 88.1%
7 | 86.5%

dubslow:
Changed it back to 5.

PR #29, opened 2018-06-03
Per UCI protocol example, `isready` directly initializes network (fixes startup lag)

dubslow:
http://wbec-ridderkerk.nl/html/UCIProtocol.html

Note that it is ambiguous in the specification if options may be set/reset after isready, but it is clear that isready should finish all expensive initialization, options or not (quoting the link):

// the engine has sent all parameters and is ready
		uciok

// Note: here the GUI can already send a "quit" command if it just wants to find out
//       details about the engine, so the engine should not initialize its internal
//       parameters before here.
// now the GUI sets some values in the engine
// set hash to 32 MB
setoption name Hash value 32

// init tbs
setoption name NalimovCache value 1
setoption name NalimovPath value d:\tb;c\tb

// waiting for the engine to finish initializing
// this command and the answer is required here!
isready

// engine has finished setting up the internal values
		readyok

// now we are ready to go

// if the GUI is supporting it, tell the engine that is is
// searching on a game that is hasn't searched on before
ucinewgame

// if the engine supports the "UCI_AnalyseMode" option and the next search is supposted to
// be an analysis, the GUI should set "UCI_AnalyseMode" to true if it is currently
// set to false with this engine
setoption name UCI_AnalyseMode value true

// tell the engine to search infinite from the start position after 1.e4 e5
position startpos moves e2e4 e7e5
go infinite

### DIFF
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -146,6 +146,7 @@ void EngineController::SetCacheSize(int size) { cache_.SetCapacity(size); }
 
 void EngineController::EnsureReady() {
   GarbageCollectNodePool();
+  UpdateNetwork();
   std::unique_lock<RpSharedMutex> lock(busy_mutex_);
 }
 


fli:
The page is a bit ambiguous but I don't think _isready_ is where you should be doing initialising. My reading of it is that _isready_ is a cheap way to check if the engine is responsive (it even says you should be able to send it while calculating, so maybe garbage collection shouldn't be done in there either). I think a better option is to initialise the weights when `setoption name Network weights file path value {path}` is called, doing that is analogous to their example where they initialise the tbs after `setoption name NalimovPath value d:\tb;c\tb`. In the example, they send _isready_ afterwards to make sure that the engine is ready to respond to commands from the GUI and not blocked initialising.

> **isready**
	this is used to synchronize the engine with the GUI. When the GUI has sent a command or
	multiple commands that can take some time to complete,
	this command can be used to wait for the engine to be ready again or
	to ping the engine to find out if it is still alive.
	E.g. this should be sent after setting the path to the tablebases as this can take some time.
	This command is also required once before the engine is asked to do any search
	to wait for the engine to finish initializing.
	This command must always be answered with "readyok" and can be sent also when the engine is calculating
	in which case the engine should also immediately answer with "readyok" without stopping the search.

dubslow:
That of course assumes that all clients everywhere will send options, which is not the slightest bit true (command line options, endusers misconfiguring uci options, etc).

The text that describes `isready` is variously interpretable, but the example I think makes it quite clear what is "supposed" to happen, and that's certainly the only way we can expect endusers of various sorts to behave.

(Not sure about the garbage collection point, I'm more inclined to agree there.)

Basically, uci sucks :P

fli:
Hmm, you're right. Probably whichever of `isready` or `setoption name Network weights file path value {path}` comes first should initialise the weights. This change doesn't look like it really affects `isready`'s performance either once it's been initialised, so there's no real problem. I think a change for the future is to initialise from the `setoption` command too.

Haha, for sure! And the only real source is that yellow background, monospaced font website!

mooskagh:
Initializing network on `setoption` will not work either.
In order to initialize networks, it needs not only weights, but backend selection and configuration, so it has to wait all that to be defined first.

That's why there is (pretty ugly) `EnsureOptionsSent()`, which should be called as soon as there is pretty high confidence that no further `setoptions` will be there. (Currently not called on `isready`, btw).

PR #33, opened 2018-06-04
Get curve into time management

jjoshua2:
Should move full time until move 44 and then decay how long it wants to spend per move. Should allow increasing default kSlowMoverStr

### DIFF
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -76,7 +76,7 @@ void EngineController::PopulateOptions(OptionsParser* options) {
   defaults->Set<int>(Search::kAllowedNodeCollisionsStr, 32);  // Node collisions
 }
 
-SearchLimits EngineController::PopulateSearchLimits(int /*ply*/, bool is_black,
+SearchLimits EngineController::PopulateSearchLimits(int ply, bool is_black,
                                                     const GoParams& params) {
   SearchLimits limits;
   limits.visits = params.nodes;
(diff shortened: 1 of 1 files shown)


jjoshua2:
time curve
This time curve should be good starting point...

src/engine.cc
```diff
@@ -107,6 +107,7 @@ SearchLimits EngineController::PopulateSearchLimits(int /*ply*/, bool is_black,
   // reduce it.
   if (slowmover < 1.0 || this_move_time > kSmartPruningToleranceMs) {
     // Budget X*slowmover for current move, X*1.0 for the rest.
+    slowmover *= pow( (1 + exp((ply - 64.5) / 6.85)), -0.171) + DBL_MIN
```
mooskagh: std::pow
std::exp
it's better to add `f` suffix after all constants, hoping that everything is float instead of double, e.g. `64.5f`.
`slowmover` is float, so DBL_MIN should be FLOAT_MIN, or better std::numeric_limits<float>::min()

mooskagh: ...but actually that PR doesn't do what it should... It seems that it will work in opposite way.

During first moves, it doesn't "see" that later it will need less.. Hmm, need to think more about it, but i'm find with submitting this variant to start with.

jjoshua2:
I think so

Issue #34, opened 2018-06-04
Fix memory leak


mooskagh:
There are reports that the recent lc0 version leaks memory.

That's the issue for myself so that I don't forget about this issue.

dubslow:
Wasn't this fixed in 38?

mooskagh:
It was!

PR #38, opened 2018-06-04
Do GC in a separate thread.

mooskagh:
Memory leak from #34 was only half a bug.
* It was expected to call garbage collection after returning a best move, but didn't. (bug)
* It was expected to cal GC during "isready". That worked.

### DIFF
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -145,7 +145,6 @@ void EngineController::UpdateNetwork() {
 void EngineController::SetCacheSize(int size) { cache_.SetCapacity(size); }
 
 void EngineController::EnsureReady() {
-  GarbageCollectNodePool();
   UpdateNetwork();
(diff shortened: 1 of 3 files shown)


src/engine.cc
```diff
@@ -179,11 +178,6 @@ void EngineController::Go(const GoParams& params) {
   auto limits = PopulateSearchLimits(tree_->GetPlyCount(),
                                      tree_->IsBlackToMove(), params);
 
-  BestMoveInfo::Callback best_move_callback = [this](const BestMoveInfo& info) {
-    best_move_callback_(info);
-    GarbageCollectNodePool();
```
dubslow: Why wasn't this one working?

mooskagh: Because I passed `best_move_callback_`, not`best_move_callback`.
Removing underscore would fix it, but it's even better with GC thread.

PR #39, opened 2018-06-05
blas/opencl cleanup

dubslow:
1) a few renames to make things conceptually cleaner

2) remove unnecessary code/variables

### DIFF
--- a/src/neural/blas/transforms.cc
+++ b/src/neural/blas/transforms.cc
@@ -260,7 +260,7 @@ void Transforms::Convolve(size_t outputs, const std::vector<float>& input,
                           const std::vector<float>& weights,
                           const std::vector<float>& biases,
                           std::vector<float>& output) {
-  constexpr unsigned int filter_len = 1;
+  constexpr unsigned int filter_len = 1; // filter_size * filter_size;
   // fixed for 8x8
   constexpr unsigned int width = 8;
   constexpr unsigned int height = 8;
@@ -270,7 +270,17 @@ void Transforms::Convolve(size_t outputs, const std::vector<float>& input,
   assert(outputs * board_squares == output.size());
 
   std::vector<float> col(filter_dim * width * height);
-
+  //Im2Col<filter_size>(input_channels, input, col);
+  // The Im2Col function was implemented in lczero, but was never called
+  // except for its specialization to when <filter_size> was <1>, and that
+  // specialization was nothing more than the std::copy below, and that its
+  // only use; therefore the unused general implementation and the
+  // specialization were both removed, in favor of just the one inplace
+  // copy(). In the unlikely case that the network architecture changes in the
+  // future, the old code can be found either in the lczero repository, the
+  // the original Leela Zero repository, or lc0 commit
+  // 9aff76d9e4e10bb853715b532a59175589c22e62
+  // from which this comment originates.
   std::copy(begin(input), begin(input)+col.size(), begin(col));
 
   // Weight shape (output, input, filter_size, filter_size)
--- a/src/neural/blas/transforms.cc
+++ b/src/neural/blas/transforms.cc
@@ -44,7 +44,6 @@ std::vector<float> Transforms::ZeropadU(const std::vector<float>& U,
       }
     }
   }
-
   return Upad;
 }
 
@@ -83,7 +82,6 @@ std::vector<float> Transforms::WinogradTransformF(const std::vector<float>& f,
       }
     }
   }
-
   return U;
 }
 
@@ -258,22 +256,22 @@ void Transforms::WinogradConvolve3(const int outputs,
   WinogradTransformOut(M, output, outputs);
 }
 
-template <unsigned int filter_size>
 void Transforms::Convolve(size_t outputs, const std::vector<float>& input,
                           const std::vector<float>& weights,
                           const std::vector<float>& biases,
                           std::vector<float>& output) {
+  constexpr unsigned int filter_len = 1;
   // fixed for 8x8
   constexpr unsigned int width = 8;
   constexpr unsigned int height = 8;
   constexpr unsigned int board_squares = width * height;
-  constexpr unsigned int filter_len = filter_size * filter_size;
   const auto input_channels = weights.size() / (biases.size() * filter_len);
   const auto filter_dim = filter_len * input_channels;
   assert(outputs * board_squares == output.size());
 
   std::vector<float> col(filter_dim * width * height);
-  Im2Col<filter_size>(input_channels, input, col);
+
+  std::copy(begin(input), begin(input)+col.size(), begin(col));
 
   // Weight shape (output, input, filter_size, filter_size)
   // 96 22 3 3
@@ -320,10 +318,10 @@ void Transforms::Innerproduct(const std::vector<float>& inputs,
   }
 }
 
-template <size_t spatial_size>
 void Transforms::Batchnorm(size_t channels, std::vector<float>& data,
-                           const float* means, const float* stddivs,
+                           const std::vector<float>& means, const std::vector<float>& stddivs,
                            const float* eltwise) {
+  constexpr size_t spatial_size = 64;
   auto lambda_ReLU = [](float val) { return (val > 0.0f) ? val : 0.0f; };
 
   for (auto c = size_t{0}; c < channels; ++c) {
@@ -342,48 +340,6 @@ void Transforms::Batchnorm(size_t channels, std::vector<float>& data,
       auto res = &eltwise[c * spatial_size];
       for (auto b = size_t{0}; b < spatial_size; b++) {
         arr[b] = lambda_ReLU(res[b] + (scale_stddiv * (arr[b] - mean)));
-      }
-    }
-  }
-}
-
-template <unsigned long filter_size>
-void Transforms::Im2Col(const int channels, const std::vector<float>& input,
-                        std::vector<float>& output) {
-  constexpr unsigned int height = 8;
-  constexpr unsigned int width = 8;
-  constexpr unsigned int channel_size = height * width;
-
-  constexpr int pad = (filter_size / 2);
-  constexpr unsigned int output_h = height + 2 * pad - filter_size + 1;
-  constexpr unsigned int output_w = width + 2 * pad - filter_size + 1;
-
-  const float* data_im = input.data();
-  float* data_col = output.data();
-
-  for (int channel = channels; channel--; data_im += channel_size) {
-    for (unsigned int kernel_row = 0; kernel_row < filter_size; kernel_row++) {
-      for (unsigned int kernel_col = 0; kernel_col < filter_size;
-           kernel_col++) {
-        int input_row = -pad + kernel_row;
-        for (int output_rows = output_h; output_rows; output_rows--) {
-          if ((unsigned)input_row < height) {
-            int input_col = -pad + kernel_col;
-            for (int output_col = output_w; output_col; output_col--) {
-              if ((unsigned)input_col < width) {
-                *(data_col++) = data_im[input_row * width + input_col];
-              } else {
-                *(data_col++) = 0;
-              }
-              input_col++;
-            }
-          } else {
-            for (int output_cols = output_w; output_cols; output_cols--) {
-              *(data_col++) = 0;
-            }
-          }
-          input_row++;
-        }
       }
     }
   }
@@ -425,26 +381,4 @@ void Transforms::InvertBatchNormStddev(std::vector<float>& weights) {
   for (auto& w : weights) w = 1.0f / std::sqrt(w + EPSILON);
 }
 
-/* Template instantiations and specializations */
-
-template <>
-void Transforms::Im2Col<1>(const int channels, const std::vector<float>& input,
-                           std::vector<float>& output) {
-  constexpr unsigned int boardsize = 8;
-  auto outSize = size_t{channels * boardsize * boardsize};
-  assert(output.size() == outSize);
-  std::copy(begin(input), begin(input) + outSize, begin(output));
-}
-
-template void Transforms::Batchnorm<64>(size_t channels,
-                                        std::vector<float>& data,
-                                        const float* means,
-                                        const float* stddivs,
-                                        const float* eltwise);
-
-template void Transforms::Convolve<1>(size_t outputs,
-                                      const std::vector<float>& input,
-                                      const std::vector<float>& weights,
-                                      const std::vector<float>& biases,
-                                      std::vector<float>& output);
-}
+} // namespace lczero
--- a/src/neural/blas/transforms.h
+++ b/src/neural/blas/transforms.h
@@ -23,11 +23,12 @@
 
 namespace lczero {
 
-class Transforms {
- public:
-  static constexpr auto kWinogradAlpha = 4;
-  static constexpr auto kWinogradTile = kWinogradAlpha * kWinogradAlpha;
+static constexpr auto kWinogradAlpha = 4;
+static constexpr auto kWinogradTile = kWinogradAlpha * kWinogradAlpha;
 
+class Transforms {
+
+ public:
   static std::vector<float> ZeropadU(const std::vector<float>& U,
                                      const int outputs, const int channels,
                                      const int outputs_pad,
--- a/src/neural/blas/transforms.h
+++ b/src/neural/blas/transforms.h
@@ -53,7 +53,6 @@ class Transforms {
                                 std::vector<float>& V, std::vector<float>& M,
                                 std::vector<float>& output);
 
-  template <unsigned int filter_size>
   static void Convolve(size_t outputs, const std::vector<float>& input,
                        const std::vector<float>& weights,
                        const std::vector<float>& biases,
@@ -64,14 +63,9 @@ class Transforms {
                            const std::vector<float>& biases,
                            std::vector<float>& output, bool apply_relu = false);
 
-  template <size_t spatial_size>
   static void Batchnorm(size_t channels, std::vector<float>& data,
-                        const float* means, const float* stddivs,
+                        const std::vector<float>& means, const std::vector<float>& stddivs,
                         const float* eltwise = nullptr);
-
-  template <unsigned long filter_size>
-  static void Im2Col(const int channels, const std::vector<float>& input,
-                     std::vector<float>& output);
 
   static void Softmax(const std::vector<float>& input,
                       std::vector<float>& output);
--- a/src/neural/network_opencl.cc
+++ b/src/neural/network_opencl.cc
@@ -44,14 +44,14 @@ class OpenCLNetwork;
 struct OpenCLWeights {
   const std::vector<float> ip2_val_w;
   const std::vector<float> ip2_val_b;
-  const size_t num_output_policies;
-  const size_t num_value_channels;
+  const size_t num_policy_outputs;
+  const size_t num_value_outputs;
 
   OpenCLWeights(const Weights& weights)
       : ip2_val_w(weights.ip2_val_w),
         ip2_val_b(weights.ip2_val_b),
-        num_output_policies(weights.ip_pol_b.size()),
-        num_value_channels(weights.ip1_val_b.size()) {}
+        num_policy_outputs(weights.ip_pol_b.size()),
+        num_value_outputs(weights.ip1_val_b.size()) {}
 };
 
 class OpenCLComputation : public NetworkComputation {
@@ -61,7 +61,7 @@ class OpenCLComputation : public NetworkComputation {
       : opencl_net_(opencl_net),
         weights_(weights),
         input_data_(kInputPlanes * 64),
-        value_data_(weights_.num_value_channels),
+        value_data_(weights_.num_value_outputs),
         policy_data_(),
         q_value_(0) {}
 
@@ -85,7 +85,7 @@ class OpenCLComputation : public NetworkComputation {
         input_data_[index++] = (plane.mask & (one << i)) != 0 ? value : 0;
     }
 
-    std::vector<float> policy_data(weights_.num_output_policies);
+    std::vector<float> policy_data(weights_.num_policy_outputs);
     opencl_net_.forward(input_data_, policy_data, value_data_);
 
     // Get the moves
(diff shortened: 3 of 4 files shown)


src/neural/blas/transforms.h
```diff
@@ -63,15 +62,10 @@ class Transforms {
                            const std::vector<float>& biases,
                            std::vector<float>& output, bool apply_relu = false);
 
-  template <size_t spatial_size>
   static void Batchnorm(size_t channels, std::vector<float>& data,
```
frpays: I  kind I liked the const float* formulation, and at least, it was consistent with the 3rd argument.

src/neural/blas/transforms.h
```diff
@@ -23,11 +23,11 @@
 
 namespace lczero {
```
frpays: Delete the Transform constructor, so that nobody instanstiate by mistaque

src/neural/blas/transforms.cc
```diff
@@ -347,56 +345,14 @@ void Transforms::Batchnorm(size_t channels, std::vector<float>& data,
   }
 }
 
-template <unsigned long filter_size>
```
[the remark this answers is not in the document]
dubslow: Code that is not used, at all in any way, nor has any prospects of being used at all in the future, has no right to be in the codebase.

src/neural/blas/transforms.cc
```diff
@@ -258,22 +256,22 @@ void Transforms::WinogradConvolve3(const int outputs,
   WinogradTransformOut(M, output, outputs);
 }
 
-template <unsigned int filter_size>
```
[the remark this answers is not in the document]
dubslow: Why would the filter size ever change from 1?

src/neural/blas/transforms.cc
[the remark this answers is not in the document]
frpays: Im2Col is a cornerstone function is machine learning. It is of paramount use. It has all "right" to be in the codebase.

src/neural/blas/transforms.cc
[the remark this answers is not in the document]
dubslow: This code is very much dead. It has not been modified once since Gary first made the port from Leela Go. I'm not even sure it's changed in Leela Zero's history, ever.

src/neural/blas/transforms.cc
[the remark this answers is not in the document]
killerducky: Yes we discussed it before. I still vote to keep the template.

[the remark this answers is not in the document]
dubslow: This isn't a machine learning library, this is a chess engine. We should keep only the code we need and use. We don't use that code at all. It is totally useless to this codebase, "cornerstone of machine learning" or not.

mooskagh:
Review status: 3 of 21 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/neural/network_blas.cc, line 19 at r1](https://reviewable.io/reviews/leelachesszero/lc0/39#-LEGKvOlETUjB0UUjRh_:-LEGKvOlETUjB0UUjRha:bslnnax) ([raw file](https://github.com/leelachesszero/lc0/blob/906106725dc296625f611aaf516f3f3f0bf1073b/src/neural/network_blas.cc#L19)):*
> ```cpp
>  */
> 
> #include "neural/network.h"
> ```

Maybe move that file to neural/blas/ too? So that all files are together.
And then we can leave neural/ only only for single-file backends.

---

*[src/neural/network_opencl.cc, line 175 at r1](https://reviewable.io/reviews/leelachesszero/lc0/39#-LEGMCFu5LQAVsBKKfDX:-LEGMCFu5LQAVsBKKfDY:bls0a1v) ([raw file](https://github.com/leelachesszero/lc0/blob/906106725dc296625f611aaf516f3f3f0bf1073b/src/neural/network_opencl.cc#L175)):*
> ```cpp
>     // Residual tower
> 
>     for (auto& residual : weights.residual) {
> ```

`const auto&`
If I understand it correctly that `residual` isn't modified in the loop.

also in two lines below.

---

*[src/neural/network_opencl.cc, line 189 at r1](https://reviewable.io/reviews/leelachesszero/lc0/39#-LEGMZuf1i5FDi0BRB5q:-LEGMZuf1i5FDi0BRB5r:b-4s59z) ([raw file](https://github.com/leelachesszero/lc0/blob/906106725dc296625f611aaf516f3f3f0bf1073b/src/neural/network_opencl.cc#L189)):*
> ```cpp
>                                         m_ceil, m_ceil);
> 
>       std::vector<float> batchnorm_means_1 = conv1.bn_means;  // copy ctor
> ```

Could you add comment why it is necessary to have a copy?

---

*[src/neural/blas/transforms.h, line 65 at r1](https://reviewable.io/reviews/leelachesszero/lc0/39#-LEGFtdIBa_-mf5qUs0n-r1-65:-LEGLYlpC1_0AIOn8dt0:b-6zizey) ([raw file](https://github.com/leelachesszero/lc0/blob/906106725dc296625f611aaf516f3f3f0bf1073b/src/neural/blas/transforms.h#L65)):*
<details><summary><i>Previously, frpays (François Pays) wrote…</i></summary><blockquote>

I  kind of liked the const float* formulation, and at least, it was consistent with the 3rd argument.
</blockquote></details>

Could you add comment what eltwise really means?
It would also be nice to add comments to all public functions, but I'm not sure whether we decided to require code in blas/ and opencl/ to be styleguide-compliant or it's "semi-external" for now.

---

*[src/neural/blas/transforms.cc, line 370 at r1](https://reviewable.io/reviews/leelachesszero/lc0/39#-LEGNA6GEtJSZmLwJADs:-LEGNA6GEtJSZmLwJADt:b-8bwmo2) ([raw file](https://github.com/leelachesszero/lc0/blob/906106725dc296625f611aaf516f3f3f0bf1073b/src/neural/blas/transforms.cc#L370)):*
> ```cpp
> }
> 
> void BlasTransforms::OffsetBatchNormMeans(std::vector<float>& bn_means,
> ```

Actually, non-const reference parameters are disallowed by style guide.
If function changes argument, it should be pointer.

(I'm also fine with declaring neural/blas/ a styleguild-incompliant zone)

---

*[src/neural/opencl/cl2.hpp, line 1 at r1](https://reviewable.io/reviews/leelachesszero/lc0/39#-LEGQijR0OxuMjXsNwS5:-LEGQijR0OxuMjXsNwS6:blk7l8u) ([raw file](https://github.com/leelachesszero/lc0/blob/906106725dc296625f611aaf516f3f3f0bf1073b/src/neural/opencl/cl2.hpp#L1)):*
> ```cpp
> /*******************************************************************************
> ```

I thinks this should go from src/neural/opencl/cl2.hpp to third_party/CL/.
All code which we are not willing to touch ourselves should be there.

We can switch compiler to ignore all warnings in that code, and also add include there as -isystem rather than -I

Fine to do in separate PR, or I can do that myself later.

---


*Comments from [Reviewable](https://reviewable.io/reviews/leelachesszero/lc0/39)*

mooskagh:
I didn't look into details of the conversation, but we should delete all unused code. If we ever need it, there is github history.

---

Review status: all files reviewed at latest revision, 8 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io/reviews/leelachesszero/lc0/39#-:-LEGVD7Y11pyQ9tYtSMT:b-32ejqs)*

dubslow:
Most of those comments should have been directed at the original porting PR. They can be addressed of course, but this PR isn't the appropriate place to discuss them.

The "removing unused but possibly useful-in-future code" discussion is the one where I'm most interested in your opinion. Obviously there's strong feelings on both sides of that particular discussion.

killerducky:
Vote-wise it is 2:2 on the template thing. So crem needs to either decide to use himself as a tiebreaker, or get a 5th opinion and use that.

---

Review status: all files reviewed at latest revision, 8 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io/reviews/leelachesszero/lc0/39#-:-LEGyRdfBBxHRfH-KClk:bmavhs4)*

src/neural/network_opencl.cc
```diff
@@ -134,111 +134,102 @@ class OpenCLNetwork : public Network {
...
-    /*
-     static constexpr int NUM_VALUE_INPUT_PLANES = 32;
-     static constexpr int NUM_POLICY_INPUT_PLANES = 32;
-     static constexpr int NUM_OUTPUT_POLICY = 1858;
-     static constexpr int NUM_VALUE_CHANNELS = 128;
-     */
+    ////////////////////////////////////////////////////////////////////////////
+    // Input convolution
 
-    int NUM_VALUE_INPUT_PLANES = weights.value.bn_means.size();
```
frpays: The naming scheme of these variables was not correct but you could have let them live. 
Materializing the quantities was important for someone who as reading/writing the code.
Now, with these 4 temporary variables gone, the code is not faster bu much less unreadable.
It will be even harder to add batching or alike, without undoing this change, and make these useful variables appear again.

```
int num_value_input_planes = weights.value.bn_means.size();
int num_policy_input_planes= weights.policy.bn_means.size();
int num_output_policy = weights.ip_pol_b.size();
int num_value_channels = weights.ip1_val_b.size();
```

src/neural/network_opencl.cc
```diff
@@ -134,111 +134,102 @@ class OpenCLNetwork : public Network {
     params_.tune_exhaustive =
         options.GetOrDefault<bool>("tune_exhaustive", false);
 
-    const int inputChannels = kInputPlanes;
     const int channels = weights.input.biases.size();
-    const size_t residual_blocks = weights.residual.size();
 
-    /*
```
frpays: This comment block was meant to remind the people working of this backend, the usual values they were dealt with.
```
/*
     static constexpr int NUM_VALUE_INPUT_PLANES = 32;
     static constexpr int NUM_POLICY_INPUT_PLANES = 32;
     static constexpr int NUM_OUTPUT_POLICY = 1858;
     static constexpr int NUM_VALUE_CHANNELS = 128;
*/
```

src/neural/network_opencl.cc
```diff
@@ -134,111 +134,102 @@ class OpenCLNetwork : public Network {
...
-    for (auto i = size_t{0}; i < residual_blocks; i++) {
-      auto& residual = weights.residual[i];
+    ////////////////////////////////////////////////////////////////////////////
+    // Residual tower
+
+    for (auto& residual : weights.residual) {
       auto& conv1 = residual.conv1;
       auto& conv2 = residual.conv2;
 
       std::vector<float> conv_weights_1 =
```
frpays: Watch the code you changed. Previous version versus newest Version.
And ask you the question: which one the more easy to read?

 `Transforms::WinogradTransformF(conv1.weights, channels, channels);`

or 

 `BlasTransforms::WinogradTransformF(conv1.weights, conv1.biases.size(), conv1.biases.size());`

mooskagh:
Less code is bettet. Even if there is a large chance of using the code, like 70% (which is totally not the case here), I'd still say it's better to remove it.

In this particular case I'd remove it and left a comment in the code with a link to github review where original code could be found.

---

Review status: all files reviewed at latest revision, 9 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io/reviews/leelachesszero/lc0/39#-:-LELq3ph-BlWCNQmKm4u:b-2t9oqz)*

dubslow:
I'll make a revised version without the second commit (leaving `Transforms` name as is) and with the final comment leaving behind a lengthy comment and pointer to removed code.

src/neural/network_opencl.cc
dubslow: The latter is far easier to read, since it doesn't have pointless redirection and obfuscation.

If `channels` had some actual calculation going on, such that it differed from `biases.size()` in some way, then I'd say sure, lets make a separate variable. But if nothing is actually changing, then adding a separate variable only adds confusion.

src/neural/network_opencl.cc
dubslow: What do the "usual" values matter? What matters from a code perspective is that the size of the blocks/data must be correct, and if we can read those sizes from the `Weights` struct, much better to do that than to think about the actual underlying numbers. Especially since those numbers are subject to change, and in fact *did* change in the v1 -> v2 transition. Reliance on the global constants is the only reason that horrific "input planes bug" was even possible in the first place. Using dynamic sizes from the `Weights` struct makes that class of bug entirely impossible, and adding back the actual numbers, even in a comment, serves no purpose.

src/neural/network_opencl.cc
dubslow: See my other recent comments; I very much disagree that it's more readable, I think it's far more confusing to have separate variables that add no calculation. And referring to block weight sizes would have made far more sense to me when I was reading lczero code the first time than to random "num input planes" or "num policy channels" or whatever, which make no reference to which weights they are about.

src/neural/blas/transforms.h
dubslow: I don't see any?

src/neural/blas/transforms.h
dubslow: That whole function left me feeling icky. We're using C++, there shouldn't be any pointer arithmetic anyways. blegh. Would you be satisfied if I changed the `eltwise` one to be a vector of floats as well?

dubslow:
@frpays @killerducky I've left a comment block explaining the dead code situation, including a pointer to the commit that removed the code in question so that we may find it back as necessary. Please re-review.

src/neural/network_opencl.cc
mooskagh: Actually, that 1858 is used over all places in the code.
I think since long ago that it should be replaced with a constant in network.h, but didn't have good enough occasion to do that. So if anyone will edit somewhere there, that would be good thing to change too.

dubslow: Perhaps it's used in other places, but it's not used in network_{opencl,blas}.cc, and has no reason to be when the *actual* weights size is available. Agreed it's good to consolidate the other uses where the actual weights aren't available.

src/neural/network_opencl.cc
frpays: No, of course not. And you probably know yourself that this is not true. For someone you is not familiar with the code, and even for me, the first version is far much more readable.
Do you realise that you oppose every comment that is made to you? You are not doing any single compromise. What the point of reviewing?

frpays:
@dubslow could you please change back BlasTransform back to Transform. This is one something you agreed on.

frpays:
Review status: 17 of 21 files reviewed at latest revision, 9 unresolved discussions.

---

*[src/neural/network_blas.cc, line 88 at r3](https://reviewable.io/reviews/leelachesszero/lc0/39#-LEazWekCx4GBpu8ZFhi:-LEazWekCx4GBpu8ZFhj:b-i1so3j) ([raw file](https://github.com/leelachesszero/lc0/blob/5a1818cd76270a91b8346826d5572ebb0f96d652/src/neural/network_blas.cc#L88)):*
> ```cpp
> 
>     int NUM_VALUE_INPUT_PLANES = weights_.value.bn_means.size();
>     int NUM_POLICY_INPUT_PLANES = weights_.policy.bn_means.size();
> ```

The following to this function much more readable is you insert two variables:
 `int num_value_input_planes = weights_.value.bn_means.size();
 int num_policy_input_planes = weights_.policy.bn_means.size();`
and leave the substitution below.

How many people realise instantly  that `weights_.policy.bn_means.size()` is the number of output policies (1858)... come on.

---

*[src/neural/network_blas.cc, line 94 at r3](https://reviewable.io/reviews/leelachesszero/lc0/39#-LEb005v--ut1yDS8Ewe:-LEb005v--ut1yDS8Ewf:b-hneezm) ([raw file](https://github.com/leelachesszero/lc0/blob/5a1818cd76270a91b8346826d5572ebb0f96d652/src/neural/network_blas.cc#L94)):*
> ```cpp
> 
>     // Calculate output channels
>     const auto output_channels = weights_.input.biases.size();
> ```

Same here, leave `output_channels`. Because  when `weights_.input.biases.size()`, you don't understand that this is the number of layers of your res blocks.

---

*[src/neural/network_opencl.cc, line 189 at r1](https://reviewable.io/reviews/leelachesszero/lc0/39#-LEGMZuf1i5FDi0BRB5q:-LEb0SZD1sdp1tmEXVI_:b-roxdbr) ([raw file](https://github.com/leelachesszero/lc0/blob/906106725dc296625f611aaf516f3f3f0bf1073b/src/neural/network_opencl.cc#L189)):*
<details><summary><i>Previously, mooskagh (Alexander Lyashuk) wrote…</i></summary><blockquote>

Could you add comment why it is necessary to have a copy?
</blockquote></details>

I agree, that needs a comment block at the beginning. 
This has to be converted to proper english:
 
 // Here we are initialising the OpenCL network.
// We essentially passing the content of the 
// Weights structure to the underlying OpenCL layer.
//
// It is important to note that the underlying OpenCL layer 
// will copy everything we will pass  to individual clBuffers
// in the network initialisation.
//
// Hence, contrary to BLAS network, The OpenCL Network does
//  not keep a copy of // the Weights structure. This is why here,
//  we referencing the original const struct, not a copy.
//
// Occasionally when we do modify the content of a Weights vector
// before passing it to the underlying OpenCL layer.
// This is when we need to copy some vectors.

---

*[src/neural/network_opencl.cc, line 206 at r3](https://reviewable.io/reviews/leelachesszero/lc0/39#-LEb3xGP9v3bNYRjYcho:-LEb3xGP9v3bNYRjYchp:bfnm7s4) ([raw file](https://github.com/leelachesszero/lc0/blob/005d0ef0ac69b32d2bff826e7f66bcc27624b03f/src/neural/network_opencl.cc#L206)):*
> ```cpp
>     }
> 
>     ////////////////////////////////////////////////////////////////////////////
> ```

The comments are good, but the ascii art above is debatable.

---

*[src/neural/network_opencl.cc, line 244 at r3](https://reviewable.io/reviews/leelachesszero/lc0/39#-LEb4COi3j-2vzz1DMRp:-LEb4COi3j-2vzz1DMRq:b-5tb46l) ([raw file](https://github.com/leelachesszero/lc0/blob/5a1818cd76270a91b8346826d5572ebb0f96d652/src/neural/network_opencl.cc#L244)):*
> ```cpp
>                            bn_val_means, bn_val_stddivs, weights.ip1_val_w,
>                            weights.ip1_val_b);
>   }
> ```

Overall, in network_blas and network_opencl, do not replace the few variables remaining that comment the code.
You should just change the case of  `NUM_VALUE_INPUT_PLANES` and the like to make them compliant. 

---


*Comments from [Reviewable](https://reviewable.io/reviews/leelachesszero/lc0/39)*

frpays:
*[src/neural/blas/transforms.h, line 24 at r1](https://reviewable.io/reviews/leelachesszero/lc0/39#-LEGFtdIBa_-mf5qUs0n-r1-24:-LEb6Sag-zjr4Ye-SWf9:bgjwwvu) ([raw file](https://github.com/leelachesszero/lc0/blob/906106725dc296625f611aaf516f3f3f0bf1073b/src/neural/blas/transforms.h#L24)):*
<details><summary><i>Previously, dubslow wrote…</i></summary><blockquote>

I don't see any?
</blockquote></details>

It's a special c++ syntax. 
`Transform() = delete;`
Fancy, isn't it?

---


*Comments from [Reviewable](https://reviewable.io/reviews/leelachesszero/lc0/39)*

frpays:
*[src/neural/blas/transforms.h, line 65 at r1](https://reviewable.io/reviews/leelachesszero/lc0/39#-LEGFtdIBa_-mf5qUs0n-r1-65:-LEb7EmZ-TEzyabz3HaZ:b-mt7toz) ([raw file](https://github.com/leelachesszero/lc0/blob/906106725dc296625f611aaf516f3f3f0bf1073b/src/neural/blas/transforms.h#L65)):*
<details><summary><i>Previously, dubslow wrote…</i></summary><blockquote>

That whole function left me feeling icky. We're using C++, there shouldn't be any pointer arithmetic anyways. blegh. Would you be satisfied if I changed the `eltwise` one to be a vector of floats as well?
</blockquote></details>

No please leave it like that. 

This is almost C level here. Low level algebra use a lot of "pointer arithmetic" because the pointer carries an implicit offset. I suspect that the while Transform.h interface should use const float* and float* back. 

Why ? because when we will add batching, we will need to allocate a unique vector for the batch and the we will need to pass x+offset, y+offset. You can do that with pointer (implicit offset), but not with vectors.

@Crem the `eltwise` is the skip connection of the residual block. No idea why it is spelt *`elwise`*.

---


*Comments from [Reviewable](https://reviewable.io/reviews/leelachesszero/lc0/39)*

src/neural/network_opencl.cc
frpays: "far more confusing". 

The code is written for the machine, but also for humans to read it back later.

This is not because `int num_output_policy = weights.ip_pol_b.size();` add *no calculation* that you can remove it.

No only the remaining portion of code is more readable but it is shorter.  Because everywhere you will need `num_output_policy, you won't have to write `weights.ip_pol_b.size()`.

It's very difficult to understand you disagreeing here, as it is pretty much common sense.

frpays:
I finished the new benchmarks

https://docs.google.com/spreadsheets/d/1eU2gb67V69_7yXFzIaF2WI4djF1pXGAih-rwhrstmg0


Platform | Threads | master | PR87 (512 batch size) | Speed-up
-- | -- | -- | -- | --
MacOS/vecLib | 1 | 43 | 114 | 2.65
MacOS/vecLib | 2 | 58 | 172 | 2.97
MacOS/vecLib | 4 | 89 | 222 | 2.49
MacOS/vecLib | 6 | 100 | 240 | 2.40
MacOS/vecLib | 8 | 98 | 239 | 2.44
MacOS/vecLib | 10 | 110 | 239 | 2.17
Linux/MKL 1-core | 1 | 24 | 38 | 1.58
Linux/MKL 1-core | 2 | 47 | 74 | 1.57
Linux/MKL 1-core | 4 | 89 | 134 | 1.51
Linux/MKL 1-core | 6 | 127 | 192 | 1.51
Linux/MKL 1-core | 8 | 161 | 245 | 1.52
Linux/MKL 1-core | 10 | 192 | 285 | 1.48
Linux/MKL 2-core | 1 | 33 | 60 | 1.82
Linux/MKL 2-core | 2 | 72 | 115 | 1.60
Linux/MKL 2-core | 4 | 132 | 213 | 1.61
Linux/MKL 2-core | 6 | 160 | 275 | 1.72
Linux/MKL 2-core | 8 | 98 | 250 | 2.55
Linux/MKL 2-core | 10 | 52 | 250 | 4.81
Linux/OpenBLAS 1-core | 1 | 24 | 39 | 1.63
Linux/OpenBLAS 1-core | 2 | 48 | 76 | 1.58
Linux/OpenBLAS 1-core | 4 | 89 | 140 | 1.57
Linux/OpenBLAS 1-core | 6 | 127 | 202 | 1.59
Linux/OpenBLAS 1-core | 8 | 160 | 257 | 1.61
Linux/OpenBLAS 1-core | 10 | 199 | 289 | 1.45
Linux/OpenBLAS 2-core | 1 | 29 | 61 | 2.10
Linux/OpenBLAS 2-core | 2 | 59 | 114 | 1.93
Linux/OpenBLAS 2-core | 4 | 102 | 207 | 2.03
Linux/OpenBLAS 2-core | 6 | 111 | 276 | 2.49
Linux/OpenBLAS 2-core | 8 | 100 | 256 | 2.56
Linux/OpenBLAS 2-core | 10 | 97 | 231 | 2.38

src/neural/blas/transforms.cc
```diff
@@ -320,10 +328,10 @@ void Transforms::Innerproduct(const std::vector<float>& inputs,
   }
 }
 
-template <size_t spatial_size>
 void Transforms::Batchnorm(size_t channels, std::vector<float>& data,
-                           const float* means, const float* stddivs,
+                           const std::vector<float>& means, const std::vector<float>& stddivs,
                            const float* eltwise) {
```
mooskagh: Any idea what `eltwise` is?

src/neural/blas/transforms.cc
```diff
@@ -258,22 +256,32 @@ void Transforms::WinogradConvolve3(const int outputs,
...
+  // except for its specialization to when <filter_size> was <1>, and that
+  // specialization was nothing more than the std::copy below, and that its
+  // only use; therefore the unused general implementation and the
+  // specialization were both removed, in favor of just the one inplace
+  // copy(). In the unlikely case that the network architecture changes in the
+  // future, the old code can be found either in the lczero repository, the
+  // the original Leela Zero repository, or lc0 commit
+  // 9aff76d9e4e10bb853715b532a59175589c22e62
+  // from which this comment originates.
+  std::copy(begin(input), begin(input)+col.size(), begin(col));
```
mooskagh: It's not obvious that input has at least col.size() elements. Maybe
```cpp
assert(col.size() == input.size());
```
?

dubslow:
Superseded by 87

PR #41, opened 2018-06-06
Add support for no-op resign

gonzalezjo:
The premise is to:

1) make it easier for resign to be added 

2) Make it safe to test resign from `lczero`.


mooskagh:
Only fully implemented features should go into the repo.
You can do the development in your own fork and then pull the implemented feature.

dubslow:
Well it would have been full implementation of "don't crash under latest selfplay server settings" :)

But anyways, this can be closed in favor of #47 right?

gonzalezjo:
That was the purpose, @dubslow. I was under the impression that for lczero to get resign, lc0 would at least not die when receiving --resignpct. That said, #47 renders this PR unnecessary.

PR #43, opened 2018-06-06
Expreimental support for cuDNN 6

borg323:
With this patch I am able to compile and run lc0 on linux using cuDNN version 6 together with CUDA 9. This can probably be refactored, but I don't know CUDA so at this stage I am interested in comments on code correctness.

### DIFF
--- a/src/neural/network_cudnn.cu
+++ b/src/neural/network_cudnn.cu
@@ -471,7 +471,16 @@ void ConvLayer::Eval(int N, float *output, const float *input,
           out_tensor_desc_, output));
     }
     else {
-      throw Exception("use_bias_ without use_relu_ not supported");
+      reportCUDNNErrors(cudnnConvolutionForward(
+          cudnn, &alpha, in_tensor_desc_, input, filter_desc_, weights,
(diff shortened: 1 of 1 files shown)


mooskagh:
Why would you want to support cuDNN 6? Are there GPUs that work on 6 but not 7?

borg323:
The cuDNN library is already installed in colab so I expect this would simplify using it. It would require downloading a pre-compiled binary but this is doable. 
I am exploring options at the moment, so I don't think this should be merged, at least not in the current form.

borg323:
Found a better workaround, closing

radha:
@borg323   could you describe the workaround for cudnn < 7.1. I am stuck with cuddn 7.0 for another project.

borg323:
@radha the workaround is for colab which has old nvidia drivers (384.111) - According to https://stackoverflow.com/questions/30820513/what-is-version-of-cuda-for-nvidia-304-125/30820690#30820690 only up to cuda 9.0 is compatible. However cuda installation will try to upgrade several other packages so they need to be held to version 384.111: nvidia-384, nvidia-opencl-icd-384, libcuda1-384, nvidia-384-dev, cuda-drivers

PR #46, opened 2018-06-06
Switch to params, suggested by @ASilver.

mooskagh:
I know that's controversial (especially as it diverges from a0 paper, namely in policy softmax=2.2), but it shown good results.
Feel free to reject. :)

### DIFF
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -71,8 +71,9 @@ void EngineController::PopulateOptions(OptionsParser* options) {
   auto defaults = options->GetMutableDefaultsOptions();
 
   defaults->Set<int>(Search::kMiniBatchSizeStr, 256);    // Minibatch = 256
-  defaults->Set<float>(Search::kFpuReductionStr, 0.2f);  // FPU reduction = 0.2
-  defaults->Set<float>(Search::kCpuctStr, 3.1f);         // CPUCT = 3.1
+  defaults->Set<float>(Search::kFpuReductionStr, 0.9f);  // FPU reduction = 0.9
+  defaults->Set<float>(Search::kCpuctStr, 3.4f);         // CPUCT = 3.4
+  defaults->Set<float>(Search::kPolicySoftmaxTempStr, 2.2f);  // Psoftmax = 2.2
   defaults->Set<int>(Search::kAllowedNodeCollisionsStr, 32);  // Node collisions
 }
 


mooskagh:
That commits chain is due to rebase. I don't know what squashing is.

ASilver:
When the PUC and FPU values were changed from 1.7 and 0.0 (lc0 equivalent) to 1.2 and 0.2, I complained this reduced the exploration and was not ideal for training. My complaints were regularly dismissed since playing performance was all that mattered, even in training. I recently ran a deep CLOP on three settings to either challenge or confirm them and came up with very different values. Tests have shown them to be superior in all areas, in matches and tactics (quite spectacularly). However with one significant caveat I will share openly. Please read to the end. Here are some results reported:

**CCLS:** **[In tests they showed a 30 Elo increase](https://docs.google.com/spreadsheets/d/1XSJiCcQpCLv0fNwrUn7jXjdkZFU63YFEWpdXv6dSSg0/edit#gid=0)** over default settings against a variety of opponents. See May26 (default) vs June4 with id315. 

**@jjoshua2** Showed a much improved result in long TC against Andscacs.

**Kai Laskos,** a statistician, also reported the first [large improvements over Komodo 10.2 at LTC](http://talkchess.com/forum3/viewtopic.php?p=764710#p764710) 
and then tested it at 1m+1s [against Houdini 1.5a, and reported a 60 Elo improvement](http://talkchess.com/forum3/viewtopic.php?p=764750#p764750) vs default settings.

I also tested it myself and came up with a [63 Elo improvement over the default settings](http://talkchess.com/forum3/viewtopic.php?p=764619#p764619) against Protector 1.9. 

Finally, I also tested this against default in my corrected WAC set, at 10s/position, id369, and default scored 109/200, while the new settings scored 159/200.

However, the caveat to this is that while the much more exploratory settings seem to favor almost all conditions, they lose badly to the default settings when pitted against itself. At least in any very fast or relatively fast TC that I checked.

In the end, i think that any change will come down more to one of training philosophy: exploration vs exploitation. the modified settings GCP produced clearly favor exploitation against itself, but it is not at all clear they favor increased learning or a stronger NN in the end.

killerducky:
Maybe we can wait until things are more stable to test this change in the training feedback loop? So change defaults for matches as you have here, but also override to the previous values for training? Or only increase puct back to 0.85*2=1.7 instead of 0.6*2=1.2?

If we wait on this change during the lc0 reboot we can see results for the current settings minus the bugs. I'm hoping that the only reason we need higher softmax for matches on the current net is because it's damaged from earlier bugs.

dubslow:
Your CCLS link points instead to CCRL.

And for lczero's lower nodes counts, increasing cpuct actually reduced performance in CCLS. Though I still fully support raising cpuct again in training.

As for self play tests, I think given how hilariously overfitted the network both was and is, selfplay results should *not* be trusted in any way. It's not surprising to me that such an overfitted network would benefit from *less* exploration against itself while benefiting from more exploration against literally any other opponent.

I wholeheartedly support resetting training cpuct from 0.6/1.2 to 0.85/1.7 or maybe even 1.0/2.0. I think lowering to 0.6 was a major mistake in retrospect. (FPU Reduction also was a mistake, and was immediately reverted; even though it was immediately reverted, we still did not see full reversion of playing strength, adding to the very large and multi-sourced pile of evidence that the 0.85 -> 0.6 change was bad -- maybe very bad -- for the network.)

And crem, squashing is one of several operations available in `git rebase -i`. Suggest you `git checkout experimental; git rebase -i master`, read the instructions that appear in your text editor, and leave only one commit on top of master.

mooskagh:
This is a change for play only, training settings are all untouched (and are the same as lczero.exe still has).

killerducky:
Oh since it's play engine only no problem.

PR #53, opened 2018-06-07
implement ponder

exoticorn:
- add "Ponder" uci option, doing nothing but signalling that
  ponder is supported.
- add ponder token to bestmove reply
- add ponder token to go command
- add ponderhit command


exoticorn:
Some notes:

This ponders on the position on the board (after Leela's move) instead after the expected move by the opponent. Once it is Leela's move again, she can just tree-reuse whatever move was actually played.

Since the UCI protocol sends the position after the ponder move before starting pondering, in the "go ponder ..." case, the last move is stripped from the position.

Since setting the tree to the position after the ponder move and then going back a move again would discard the whole tree and effectively disabling tree-reuse, I modified `SetPosition` to only store the position for later and only update the tree once the search is started.

In the `info` lines while pondering, UCI expects the `pv` to start from the position after the ponder move, so if the `pv` starts with that move, it is stripped from the `pv`, otherwise the whole `pv` is removed.

src/engine.cc
```diff
@@ -66,6 +66,8 @@ void EngineController::PopulateOptions(OptionsParser* options) {
   options->Add<FloatOption>(kSlowMoverStr, 0.0f, 100.0f, "slowmover") = 2.2f;
   options->Add<IntOption>(kMoveOverheadStr, 0, 10000, "move-overhead") = 100;
```
mooskagh: Why empty line? :)

src/engine.cc
```diff
@@ -155,11 +157,20 @@ void EngineController::NewGame() {
   cache_.Clear();
   search_.reset();
   tree_.reset();
+  current_position_ = optional<CurrentPosition>();
```
mooskagh: It's better to add reset()
http://en.cppreference.com/w/cpp/utility/optional/reset

to our poor-man's std::optional and use it here.

src/engine.h
```diff
@@ -87,6 +96,9 @@ class EngineController {
   std::string network_path_;
   std::string backend_;
   std::string backend_options_;
+
+  optional<CurrentPosition> current_position_;
```
mooskagh: What does current_position_ really contain, and how is it different from tree_.GetPositionHistory()? Could you add a comment?

src/engine.cc
exoticorn: Err, no idea? I'll remove it.

src/engine.h
exoticorn: That's a very good question. It made me realize that a) this should really be called something like `next_search_position_`, especially since b) this really needs to be reset when doing a normal search, otherwise repeated `go` commands will just search the same position instead of playing a full game.

I'll push a commit to address your comments soon.

exoticorn: Well, looks like I misremembered how UCI works and the implementation as is was actually fine. (repeated `go` commands should in fact just search the same position over and over.) So I kept the name as is and only added a comment to document that this position is the same as the position in the search tree, except during ponder search.

exoticorn:
In this review fix commit I also snuck in a small bugfix:

Before, when doing `go ponder nodes 100`, the search would terminate and print a `bestmove` after 100 nodes instead of pondering until further notice. I don't think that using ponder with fixed nodes is a practical use-case, but this should now be more correct.

This also changes `go infinite nodes 100` to be equivalent to `go infinite` instead of `go nodes 100`.

mooskagh:
Could you please resolve the conflicts, I'll do another round of review then!

jjoshua2:
I read a MCTS pondering paper for Go and they found no elo improvement over what you're doing for only pondering the best move or 5 or other N best moves over just letting MCTS ponder all of them and have tree reuse. Which makes sense on average they are the same just less chance to get lucky for a extra good hit.
However, they did find an improvement for on a ponder hit to use less time on the followup move.

dubslow:
1) I seem to recall not really liking this PR on first read through, it seemed somehow messy and complicated to me

2) about the tests vs go, I'd argue that it's somewhat unnatural for Leela to actually focus on one move, so maybe lets start with the natural way, then we can implement and test the unnatural way to see if it is actually better. One major difference between Chess and Go is that we have computer opponents with totally different style of search, unlike Go which only has other mcts+nn implementations to compare L0 against.

exoticorn:
@jjoshua2: I agree that pondering on the expected move, vs. all moves might not make a significant difference on average. I actually started out meaning to implement pondering on the expected move first (and compare pondering on all moves later), because the UCI protocol is designed with that approach in mind and I hoped that this would keep the code simpler. However, that way conflicts with tree-reuse, because searching on ex. `1. e4 e5` discards the trees for all other potential black replies. This could of course worked around with a special case in the tree code, or might not be a real issue because of the NN cache, but this was my primary reason for going this way.

@mooskagh: I can totally see why the code change in the PR does not look overly attractive to you. If you have an idea how to better resolve the dissonance between the UCI protocol and leelas search I'd be happy to rework it. For now, I can only try to justify my decisions in a UCI session:

```
> position startpos
> go wtime 100000
< info [...] pv e2e4 e7e5
< bestmove e2e4 ponder e7e5
```

So far nothing exciting except that leela now returns a ponder move.

`> position startpos moves e2e4 e7e5`

(while black is still deciding on their move)

First problem: actually resetting the tree to this position discards the sub-trees for all other black replies. That's why I changed `EngineController::SetPosition` to only store the position in `current_position_` for now until we know whether that position is meant for a normal or a ponder search.

`> go ponder wtime 1000000`

Starting an infinite ponder search, but at the same time giving the time limits in case of a `ponderhit`, so I have to store these in `go_params_`.
Instead of starting a search on the given position this code searches the position one ply before (which should be the actual position on the board). In addition to feeling more natural, this has the advantage of not discarding the sub-trees for all the non-ponder move replies by black.

`< info [...] pv g1f3`

The UCI protocol assumes a search on the position after the ponder move and therefore expects the pv to start from there, so I added a custom info callback that adjusts the pv accordingly.

Now we either get:

`ponderhit`

Black played the expected move e7e5 and leela must now start a regular search from the position `1. e4 e5` with the time limits given in the `go ponder` command.

Or:

```
> position startpos moves e2e4 e7e6
> go wtime 1000000
```

The `position` command should stop the ponder search with no further output and then we get a normal `go` command.

In any case, I have merged the current `master` and am looking forward to your feedback.

mooskagh:
I cannot really understand the uci protocol related to ponders.
Could you provide some short snippet of that would be example of uci conversation?
Also, why is there both `go ponder` option and uci param "Ponder"?

killerducky:
@mooskagh See the part below about time management and ponder:
```
		*  = Ponder, type check
			this means that the engine is able to ponder.
			The GUI will send this whenever pondering is possible or not.
			Note: The engine should not start pondering on its own if this is enabled, this option is only
			needed because the engine might change its time management algorithm when pondering is allowed.
```

exoticorn:
@mooskagh See my comment from 15 days ago above for a ponder session, I'll add another (fictional) snippet here without prose (and without `info` lines for brevity):

```
> position startpos
> go wtime 100000
< bestmove e2e4 ponder e7e5
> position startpos moves e2e4 e7e5
> go ponder wtime 100000
// some time passes, opponent does indeed move e7e5
> ponderhit
< bestmove f2f4 ponder d7d6
> position startpos moves e2e4 e7e5 f2f4 d7d6
> go ponder wtime 100000
// some time passes, opponent choses a different move than we expected
> position startpos moves e2e4 e8e5 f2f4 e5f4
< bestmove g1f3 ponder g7g5
```

As Killerducky mentioned, the UCI parameter is basically just there to inform the engine that it will be allowed to ponder, while the `go ponder <time options>` command actually starts a ponder search. (Engines are not allowed to do any search (including ponder) without a `go` command.)

The reason why the UCI protocol works the way it does for ponder is that A/B engines can start a search on the expected move and then when they get a `ponderhit` command, just continue the search they are already performing. Restarting the search would lose quite a bit of progress.

dubslow:
```
< bestmove e2e4 ponder e7e5
> position startpos moves e2e4 e7e5
> go ponder wtime 100000
> ponderhit
< bestmove f2f4 ponder d7d6
```

This seems to me like a goddamn moronic design. Why put the ponder move in the `position` command when it's not actually on the board, as opposed to in the `go ponder` command, like `go ponder <move>`?

mooskagh:
Ok, I've read the uci protocol description and think that understand what it's supposed to do.

How about implementing it this way:
1. `go ponder` is the same as `go infinite` (will some changes: never output bestmove, even when checkmate etc, and last move of position is moves and stored somewhere)
2. `ponderhit` does the following: stops the search (bestmove is not output as search was "ponder" search), redoes the "position" with the stored move, and restarts with the original params with the only exception that `go ponder` part is removed (but `wtime`, `binc` etc are reset).
3. Ensure that `position` command while pondering stops previous search (but doesn't send `bestmove`)

exoticorn:
I'm not sure how exactly this proposal differs from the PR as is.

The only thing not mentioned in the proposal are the changes related to `setPosition` (storing the position instead of applying it directly) which are necessary so that "`go ponder`: last move of position is moves (removed?) and stored somewhere" doesn't break tree re-use.

src/engine.cc
```diff
@@ -210,20 +220,57 @@ void EngineController::SetPosition(const std::string& fen,
...
+  go_params_ = params;
+
+  ThinkingInfo::Callback info_callback(info_callback_);
+
+  if(current_position_) {
+    if (params.ponder && !(*current_position_).moves.empty()) {
+      std::vector<std::string> moves((*current_position_).moves);
+      std::string ponder_move = moves.back();
+      moves.pop_back();
+      SetupPosition((*current_position_).fen, moves);
```
mooskagh: The problem with that approach is that first when position is set up, useful (for tree reuse) subtrees are chopped.
And then when `go ponder` comes it, it turns out that those subtrees were useful, and how the entire node is useless too, so it's trimmed.
Shortly speaking, ponder kills tree reuse.
I'm trying to come up with elegant-ish way to solve that, ideas are welcome. We can move all that deallocation to `go` command, but then why are those `isready` there at all..

exoticorn: Where exactly are useful subtrees chopped? The `SetupPosition` should set the position currently  on the board (the uci ponder position minus the last move). That's the whole point of the deferred position setup.
The unmodified `SetPosition` function would indeed have chopped useful subtrees.

exoticorn: Yes, I know. That's why I made sure that the new position is never shorter than the old one. (At least I think I did, and the node counts when testing this PR suggested that indeed the tree was correctly re-used.)

borg323:
@exoticorn, we have had a modification to the license after you started doing this work (see #184). Are you OK with this?

exoticorn:
@borg323 Yes, I'm ok with the updated license.

src/engine.cc
```diff
@@ -193,11 +194,20 @@ void EngineController::NewGame() {
...
   tree_.reset();
+  current_position_.reset();
   UpdateNetwork();
 }
 
 void EngineController::SetPosition(const std::string& fen,
                                    const std::vector<std::string>& moves_str) {
+  current_position_ = CurrentPosition { fen, moves_str };
+
+  SharedLock lock(busy_mutex_);
```
mooskagh: It's better to keep that lock in the beginning of the function.
It's not for synchronization (where we need to make it as short as possible), but only for `isready` command. When it's locked, `isready` waits before responding `readyok`.

src/engine.cc
```diff
@@ -210,20 +220,57 @@ void EngineController::SetPosition(const std::string& fen,
 }
 
 void EngineController::Go(const GoParams& params) {
-  if (!tree_) {
-    SetPosition(ChessBoard::kStartingFen, {});
+  go_params_ = params;
+
+  ThinkingInfo::Callback info_callback(info_callback_);
+
+  if(current_position_) {
```
mooskagh: It seems that it should work in this scenario, but could you double-check?
```
position startpos  # sets up the position
go nodes 1000      # thinks some
go nodes 5000      # thinks more on the same position
```

src/engine.cc
```diff
@@ -210,20 +220,57 @@ void EngineController::SetPosition(const std::string& fen,
...
+    if (params.ponder && !(*current_position_).moves.empty()) {
+      std::vector<std::string> moves((*current_position_).moves);
```
mooskagh: Consider implementing `operator->` in `utils/optional.h`
(Which will exactly the same implementation as `operator*`)

src/engine.cc
mooskagh: Yeah, I thought that you reset tree in setposition too, sorry for the confusion!

src/engine.cc
```diff
@@ -85,6 +85,7 @@ void EngineController::PopulateOptions(OptionsParser* options) {
                             "time-curve-left-width") = 82.0f;
   options->Add<FloatOption>(kTimeCurveRightWidth, 0.0f, 1000.0f,
                             "time-curve-right-width") = 74.0f;
+  options->Add<BoolOption>("Ponder", "ponder") = false;
```
mooskagh: Maybe add a comment that we actually don't use it, but GUIs require that to exist?

exoticorn: Will do. Actually, I don't even know which GUIs, if any, require that option to exist, but the UCI spec has the comment: "this means that the engine is able to ponder." next to the option without making it really clear whether that comment is meant for the engine (ie. if the option is true, the engine knows it is able to ponder) or the GUI (ie. if the option exists, the GUI knows the engine is able to ponder). I thought it was safer to offer the option for GUIs that chose the second interpretation.

src/mcts/search.cc
```diff
@@ -383,6 +383,10 @@ std::pair<Move, Move> Search::GetBestMoveInternal() const
 
   Move ponder_move;  // Default is "null move" which means "don't display
                      // anything".
+  if (best_node.HasNode() && best_node.node()->HasChildren()) {
+   ponder_move =
```
dubslow: indentation is off here

Issue #32, opened 2018-06-03
Implement power decay averaging


mooskagh:
https://github.com/Videodr0me/leela-chess-experimental/wiki#power-decay-averaging-tree-search---backpropagate-gamma075

Very simple idea, easy to implement, shown to increase strength.

Not to be implemented immediately, but I'm filing this issue in order it not to be forgotten.

mooskagh:
Implemented in https://github.com/LeelaChessZero/lc0/pull/56

Issue #13, opened 2018-06-02
Reduce memory footprint (Node structure is too large)


mooskagh:
_From @mooskagh on April 16, 2018 9:33_

Plan is:
- Remove ChessBoard from Node structure
- Make leaves take less space
- Consider swtiching to stockfish movegen, if result of two previous steps is slow.

_Copied from original issue: glinscott/leela-chess#331_

mooskagh:
I removed chessboards from nodes, size went from 112 to 56. Others ideas to do:
* Make move 2 bytes, not 4  (-2 bytes)
* Don't store Q, as it's W/N (-4 bytes)
* Don't store V, it's only needed right after NN eval (-4 bytes)
* Make N-in-flight 2 bytes instead of 4 (-2 bytes)
* Make depth and full-depth 1 bytes each instead of 2, or possibly don't store them at all (-2..-4 bytes)
* Don't store parent (-8 bytes)
* Store is_terminal not as a separate bool, but e.g. as highest bit of N. (-1 byte)
* Instead of 64-bit pointer to child, use 32-bit and resolve through node-pool (-4 bytes)
* Store siblings in array rather than linked list (-7 bytes)

All in all, it will give 56->20 size reduce per node if all is implemented. Not everything from the list is feasible as it can slow things down.

Also leaves only really need 6 bytes (move+P), it may be worthwhile to explore using separate structs for leaves vs internal node. Although some bookkeeping to distinguish between them will require at least one additional byte.
Probably there's no feasible solution with replacing linked list with an array in that case, but we still can have node translation optimization, so it will be ~10 bytes per leaf, and 24 bytes per internal node.

dubslow:
One million nodes times 100 bytes is... 100MB. That's not much on modern systems. That's less than a modern webbrowser. And removing the chess board is probably good enough, down to 50MB. And a million nodes is a *really* large tree, like 60-120 minute time controls on a very good GPU.

mooskagh:
Actually it's not that good.

I on my old GTX 970 have 8800 nps, it takes less than 2 minutes to get 1000000 visits.

More importantly, tree contains not only visits (which are counted for user) but also unvisited leaf nodes (but they do contain some useful information, namely P value and move itself). Unvisited nodes increase number of nodes by factor of 35 (from startpos at least).

That means that even on my GTX 970, memory growth is
```
8800 * 35 * 60 * 56 = 1GB per minute
```
which is not that negligible at all.

[...]

PR #56, opened 2018-06-09
 Node goes on a diet (#13). Remove V and W, use power decay averaging (#32)

mooskagh:
Also remove extra virtual loss.


mooskagh:
Game board was long time removed (end of April), it made it go 112 -> 56 bytes.

Removing unexpanded nodes is what I'm working on actively right now.
(proof: stats of total children and visited children after doing 80000 nodes from startpos:
https://docs.google.com/spreadsheets/d/1LoBEjxZ0xqwNdV9OFP36RsxyB2_suGNdySq8Hr4IDRU/edit#gid=0, generated this morning)

But just as I was low energy and sleepy in the morning, wanted to do something small and easy which required very little concentration.

mooskagh:
Tests whether it affects strength of the gameplay (on default settings) are still in progress.

Videodr0me:
If you check recent testing on my experimental site, power decay averaging does not work well for high visit searches (looses elo). I am still investigating this, but i would not have very high hopes for this. But i have some new data on some other improvements soon, which look very good and scale well. Thats also motivation for the following comment which is not directly related to this PR:

@mooskagh Please do not remove unexpanded children just yet.  They might become extremely useful and even though they have some computational and space overhead these may potentially be offset by some real elo gains. Let me explain (this is not advocating against removing them, but maybe do that a little later):

1) I rearranged the code to test for terminal nodes in the unexpanded children creation loop. This allows Leela to see mates and other terminal positions one visit earlier as the result can be backpropagated to the parent node right away.

2) This plays well with certainty propagation another small elo gain (more so if TB get integrated). That might be interesting for official LC0 in the future (not now obiously)

3) A new method called tree-balancing relies on some stats calculated there and stores it in these nodes. This methods inital testing is very promising (and scales well with high visit searhes). More info soon.

Just, wait a little so see how all these tests turn out in the end.

killerducky:
@mooskagh @Videodr0me Is there an update here? The last comment by Videodr0me says this makes the engine worse at high visit counts. I'm not sure which version of the code is in this PR, but I see a table that says for 10K visits, "Elo: -11.12 LOS: 4.94%"

Was there new test data?

mooskagh:
It only makes a difference when values of those parameters are not 1.0.
With the default values of 1.0 it should (in theory) not make a difference. I'll do a proper test though today.

As for @Videodr0me's comment, he meant that with non-1.0 settings it was expected to improve strength but it doesn't.

dubslow:
I'm thinking, between this and the dynamic temp thing, that we should make it possible to add "silent" options, which are options which exist but aren't displayed by `--help`. Then these two backpropogate options, and the two evaltemp options can remain usable for those interested without unnecessarily confusing people in need of `--help`.

mooskagh:
We should aim to reduce code complexity, not --help complexity. I've never heard anyone complaining about --help being too long so far.

Therefore I'm for aggressive removal of all options of non-proved usefulness. If needed later, we can add that later.
Unused parameters really slow down code understanding and adding of new features.

Videodr0me:
@mooskagh Agree, couldn't make it work, doubtful it ever will, get rid of it.

PR #57, opened 2018-06-09
Add no tree reuse option for tournaments.

### LINKED ISSUE #26: Disable tree reuse in training

Tilps:
To close #26.

This option should work regardless of kShareTrees - but in case share trees is also disabled I made cache sharing automatic if networks are equal even if share trees is not true. It looks like a safe performance improvement for that scenario... I think.

### DIFF
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -370,7 +370,7 @@ V3TrainingData Node::GetV3TrainingData(GameResult game_result,
   return result;
 }
 
-void NodeTree::MakeMove(Move move) {
+void NodeTree::MakeMove(Move move, bool no_tree_reuse) {
   if (HeadPosition().IsBlackToMove()) move.Mirror();
 
   Node* new_head = nullptr;
@@ -382,6 +382,10 @@ void NodeTree::MakeMove(Move move) {
   }
   gNodePool.ReleaseAllChildrenExceptOne(current_head_, new_head);
   current_head_ = new_head ? new_head : current_head_->CreateChild(move);
+  if (no_tree_reuse) {
+    gNodePool.ReleaseChildren(current_head_);
+    current_head_->ResetStats();
+  }
   history_.Append(move);
 }
 
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -370,7 +370,7 @@ V3TrainingData Node::GetV3TrainingData(GameResult game_result,
   return result;
 }
 
-void NodeTree::MakeMove(Move move, bool no_tree_reuse) {
+void NodeTree::MakeMove(Move move) {
   if (HeadPosition().IsBlackToMove()) move.Mirror();
 
   Node* new_head = nullptr;
@@ -382,11 +382,12 @@ void NodeTree::MakeMove(Move move, bool no_tree_reuse) {
   }
   gNodePool.ReleaseAllChildrenExceptOne(current_head_, new_head);
   current_head_ = new_head ? new_head : current_head_->CreateChild(move);
-  if (no_tree_reuse) {
-    gNodePool.ReleaseChildren(current_head_);
-    current_head_->ResetStats();
-  }
   history_.Append(move);
+}
+
+void NodeTree::TrimTreeAtHead() {
+  gNodePool.ReleaseChildren(current_head_);
+  current_head_->ResetStats();
 }
 
 void NodeTree::ResetToPosition(const std::string& starting_fen,
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -195,6 +195,8 @@ class NodeTree {
   ~NodeTree() { DeallocateTree(); }
   // Adds a move to current_head_.
   void MakeMove(Move move);
+  // Resets the current head to ensure it doesn't carry over details from a
+  // previous search.
   void TrimTreeAtHead();
   // Sets the position in a tree, trying to reuse the tree.
   // If @auto_garbage_collect, old tree is garbage collected immediately. (may
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -194,7 +194,8 @@ class NodeTree {
  public:
   ~NodeTree() { DeallocateTree(); }
   // Adds a move to current_head_.
-  void MakeMove(Move move);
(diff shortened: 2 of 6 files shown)


src/mcts/node.cc
```diff
@@ -382,6 +382,10 @@ void NodeTree::MakeMove(Move move) {
   }
   gNodePool.ReleaseAllChildrenExceptOne(current_head_, new_head);
   current_head_ = new_head ? new_head : current_head_->CreateChild(move);
+  if (no_tree_reuse) {
+    gNodePool.ReleaseChildren(current_head_);
+    current_head_->ResetStats();
+  }
```
dubslow: Think this function can be written cleaner. There should only be one Release() call for either branch

src/mcts/node.cc
```diff
@@ -370,7 +370,7 @@ V3TrainingData Node::GetV3TrainingData(GameResult game_result,
   return result;
 }
 
-void NodeTree::MakeMove(Move move) {
+void NodeTree::MakeMove(Move move, bool no_tree_reuse) {
```
mooskagh: Tree reuse has little to do with making a move, to this parameter doesn't seem like a good fit for this function.

Instead, I'd do a separate function NodeTree::TrimTreeAtHead() and would call it when I need to kill a tree.

src/selfplay/tournament.cc
```diff
@@ -51,6 +52,7 @@ void SelfPlayTournament::PopulateOptions(OptionsParser* options) {
   options->AddContext("player2");
 
   options->Add<BoolOption>(kShareTreesStr, "share-trees") = false;
+  options->Add<BoolOption>(kNoTreeReuseStr, "disable-tree-reuse") = false;
```
mooskagh: It's better to replace no-tree-reuse=false  to reuse-tree=true.
(make this parameter mean the opposite, with opposite default.  But actually disabling tree reuse by default is probably right thing to do).

Also the same for variables. not no_tree_reuse, but reuse_tree

src/selfplay/tournament.cc
```diff
@@ -140,7 +143,7 @@ SelfPlayTournament::SelfPlayTournament(const OptionsDict& options,
   // Initializing cache.
   cache_[0] = std::make_shared<NNCache>(
       options.GetSubdict("player1").Get<int>(kNnCacheSizeStr));
-  if (kShareTree) {
+  if (kShareTree || networks_[0] == networks_[1]) {
```
mooskagh: That's wrong.
Often when they have the same network, they still don't want to have common tree. For example, they may have different puct, time to think, any parameter really.

Tilps: This doesn't make them share a tree - it makes them share a NNCache.  Obviously sharing a tree is not okay, but I think sharing the cache should be fine.

src/selfplay/tournament.cc
Tilps: Done.

src/mcts/node.cc
Tilps: Done.

src/mcts/node.h
```diff
@@ -195,6 +195,7 @@ class NodeTree {
   ~NodeTree() { DeallocateTree(); }
   // Adds a move to current_head_.
   void MakeMove(Move move);
+  void TrimTreeAtHead();
```
mooskagh: Public functions should have a comment.

src/selfplay/game.cc
```diff
@@ -69,7 +69,15 @@ void SelfPlayGame::Play(int white_threads, int black_threads) {
     // Add best move to the tree.
     Move move = search_->GetBestMove().first;
     tree_[0]->MakeMove(move);
-    if (tree_[0] != tree_[1]) tree_[1]->MakeMove(move);
+    if (!reuse_tree_) {
```
mooskagh: Actually, I think it makes sense to make it per player options, match "with tree reuse  vs no tree reuse" would totally make sense.

Also, I'd call TrimTreeAtHead() before the search, and only for the player which is about to start thinking.

src/selfplay/tournament.cc
mooskagh: Things like cache size, history length to include into cache etc, may be configured differently for different players.
Initially I thought about having share-cache parameter also, but then decided that share-tree would be enough.

Now maybe it's time to introduce it. Feel free to do that if you want.
Another sensible option would be not to share cache ever.

src/selfplay/game.cc
Tilps: I have no clue how to make a per-player option :)

src/mcts/node.h
Tilps: Done.

src/selfplay/tournament.cc
Tilps: Reverting under the assumption we'll just use kShareTree for training games.
We definitely want to share cache in training games, since they will have tree reuse disabled, and its a massive performance hit to not have a cache in that scenario.

src/selfplay/tournament.cc
```diff
@@ -28,6 +28,7 @@ namespace lczero {
 
 namespace {
 const char* kShareTreesStr = "Share game trees for two players";
+const char* kReuseTreeeStr = "Reuse the node statistics between moves.";
```
mooskagh: No dot in the end.
(It's not usual to have them in UCI, and --help adds dot by itself).

src/selfplay/game.cc
mooskagh: It's already in `options_[idx].uci_options.Get<bool>("Reuse the node statistics between moves.")`, which is per player, but the question is how to share that string constant "Reuse the node statistics between moves." nicely between tournament.cc and game.cc.

The cleanest way would be to have static function `SelfPlayGame::PopulateOptions`, and call it from `SelfPlayTournament::PopulateOptions`, and define that constant in game.cc rather than tournament.cc. But feel free to solve it differently if you wish.

Tilps: I think I've worked it out, maybe...

src/selfplay/tournament.cc
Tilps: Done.

mooskagh:
Thanks!

PR #58, opened 2018-06-09
build if BLAS includes are missing but not required

borg323:
The BLAS include directories were added in the build unconditionally if the MKL or OpenBLAS library was detected. However the libraries may be present without the include directories, and the build would fail even when building only the cuDNN backend.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -137,23 +137,25 @@ if get_option('build_backends')
 
   openblas_lib = cc.find_library('openblas', required: false)
 
-  if mkl_lib.found() 
-    add_project_arguments('-DUSE_MKL', language : 'cpp')
-    includes += include_directories(get_option('mkl_include'))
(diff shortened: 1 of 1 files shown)


frpays:
Another comment I made on Discord, is: sometimes, you have a problem with specifically mkl or openblas, a result of a strange installation or a strange system. You may want to shutdown *mkl* or *openblas*, or *accelerate* individually. Also for the purpose of changing the order of preference in meson. (For example on Mac systems, if you want to test if mkl is faster than accelerate).

borg323:
@frpays your suggestions will be the subject of a new PR improving BLAS build system usability.

dubslow:
Merged with the understanding that 1) this is strictly better than status quo, and 2) despite that, at least one potential problem remains, to be the subject of a future PR

PR #59, opened 2018-06-09
Add adjustable peaking time curve

fli:
Weighs the time to give each move by a parameterised peaking function.

This has the effect of changing the time curve to adjustably spend less time on earlier moves and more time in the midgame.

The weightings are parameterised by:

- Peak: The ply to peak the weightings given to each move
- Width: How wide the time curve is (I think this is approximately how far after the peak the inflection point will be)

Note: The parameters change the shape of the weights given to each move, not the time curve directly.

Because `Scale thinking time`effectively changes the time curve, I've just removed it since you should be able to get similar effects by changing parameters of the weighting function.

- Example of weights given with _peak=30_ and _width=30_
  screen shot 2018-06-07 at 7 04 45 am
- Example of weights given with _peak=45_ and _width=60_
   screen shot 2018-06-07 at 7 32 04 am

Here are different move times between the current time curve vs SF's time curve (adjusted to try to offset smart-pruning) vs the proposed time curve (with peak=28, width=25).
tpm3 2

The shape of the peaking function was determined by analysing training games and weighing moves in proportion to _average_winrate_loss_<sub>ply</sub> * _p_game_undecided_<sub>ply</sub>. Where  _p_game_undecided_ adjudicated games if both players saw >85% winrate.

I'm still in the process of tuning peak and width but the default values given should provide an ELO boost. Here are the results of peak=28, width=28 (I meant to do 25) vs the current time curve.

```
Score of lc0-newtm vs lc0-oldtm: 109 - 32 - 69  [0.683] 210
Elo difference: 133.61 +/- 39.97
```

Both players have temp-decay-moves=25 for variety and are playing with 15s+.1s.

### DIFF
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -89,7 +89,7 @@ void EngineController::PopulateOptions(OptionsParser* options) {
 
 double move_weight(int ply, float peak, float left_width, float right_width) {
   const float width = ply > peak ? right_width : left_width;
-  constexpr float width_scaler = 2 / std::log(2 + std::sqrt(3));
+  constexpr float width_scaler = 1.518651485; // 2 / log(2 + sqrt(3))
   return std::pow(std::cosh((ply - peak) / width / width_scaler), -2);
 }
 
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -44,6 +44,16 @@ const char* kTimeCurveRightWidth = "Time weight curve width right of peak";
 const char* kTimeCurveLeftWidth = "Time weight curve width left of peak";
 
 const char* kAutoDiscover = "<autodiscover>";
+
+float ComputeMoveWeight(int ply, float peak, float left_width,
+                        float right_width) {
+  // Inflection points of the function are at ply = peak +/- width.
+  // At these points the function is at 2/3 of its max value.
+  const float width = ply > peak ? right_width : left_width;
+  constexpr float width_scaler = 1.518651485f;  // 2 / log(2 + sqrt(3))
+  return std::pow(std::cosh((ply - peak) / width / width_scaler), -2.0f);
+}
+
 }  // namespace
 
 EngineController::EngineController(BestMoveInfo::Callback best_move_callback,
@@ -87,12 +97,6 @@ void EngineController::PopulateOptions(OptionsParser* options) {
   defaults->Set<int>(Search::kAllowedNodeCollisionsStr, 32);  // Node collisions
 }
 
-double move_weight(int ply, float peak, float left_width, float right_width) {
-  const float width = ply > peak ? right_width : left_width;
-  constexpr float width_scaler = 1.518651485; // 2 / log(2 + sqrt(3))
-  return std::pow(std::cosh((ply - peak) / width / width_scaler), -2);
-}
-
 SearchLimits EngineController::PopulateSearchLimits(int ply, bool is_black,
                                                     const GoParams& params) {
   SearchLimits limits;
@@ -120,13 +124,13 @@ SearchLimits EngineController::PopulateSearchLimits(int ply, bool is_black,
                time + increment * (movestogo - 1) - move_overhead * movestogo);
 
   constexpr int kSmartPruningToleranceMs = 200;
-  float this_move_weight = move_weight(
+  float this_move_weight = ComputeMoveWeight(
       ply, time_curve_peak, time_curve_left_width, time_curve_right_width);
   float other_move_weights = 0.0f;
   for (int i = 1; i < movestogo; ++i)
     other_move_weights +=
-        move_weight(ply + 2 * i, time_curve_peak, time_curve_left_width,
-                    time_curve_right_width);
+        ComputeMoveWeight(ply + 2 * i, time_curve_peak, time_curve_left_width,
+                          time_curve_right_width);
   // Compute the move time without slowmover.
   float this_move_time = total_moves_time * this_move_weight /
                          (this_move_weight + other_move_weights);
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -17,6 +17,7 @@
 */
 
 #include <algorithm>
+#include <cmath>
 #include <functional>
 
 #include "engine.h"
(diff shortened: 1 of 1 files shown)


dubslow:
This conflicts with #33, please compare. Also I much prefer equal time on early and midgame moves, early moves can make all the difference

fli:
I've been talking with @jjoshua2 about it, so we'll see if he comes and comments. Even though the curve in the comparison graph peaks at about ~16 the curve is effectively shifted even more left due to smart-pruning.

This is the output of the training game analysis where the y-axis is the `median winrate loss * probability game is undecided` and the x-axis is the ply. It shows where leela is most error-prone.
screen shot 2018-06-10 at 2 41 40 am

jjoshua2:
I think it makes sense, and will probably be better than my PR. I asked @fli to run some test games against it to see. I'm testing @fli live now on my streaming. It wasn't using enough time early on, so I changed peak to 25 and left width at 25, and it seems to be quite good at 6+6 for what I imagine the time curve should be. Faster than andsacs at first, and then in the endgame less time than andsacs (but enough increment to get by hopefully, and to have a good position by then). Initial settings did a poor move 8-9 and couldn't recover in middlegame when it had more time, and blundered in time pressure at end.

fli:
Against @PR #33 it is +28-12=32 ELO: 78.52 +/- 60.68

jjoshua2:
I adjusted peak move to 20, and width to 23, and I feel it does about what I want in 6+6 games against andsacs and fizbo. Seems to have as good of time management as them by eyeballing in my gauntlet now.

fli:
Here's my plan for this PR:

- [ ] Readd `Scale thinking time` and move the multiplier out of the move budgeting line. I think the purpose of `Scale thinking time` is really to offset smart-pruning, so I want to untangle smart-pruning and the time curve
- [ ] Finish tuning `peak` and `width` at fast TC. I'll let the tuning run for another half-day or so, current tuned values are around `peak=34.5`, `width=23.5`
- [ ] Change #33 to scale the time inside a for-loop (like this PR). Then tune `Scale thinking time` for #33 and use it as a baseline. This is essentially Stockfish's time curve so it is a safe baseline.
- [ ] Compare this tuned time curve vs baseline at different time controls. If it is a clear improvement then I would recommend this be merged.

If you want to change the time curve soon I, or someone else, can do the 3rd step and merge that. I think this or #33 (once it uses a for-loop and `Scale thinking time` is retuned) are clear winners over the current time curve either way.

fli:
I changed it so that you can set the width left and right of the peak. The default values are set to imitate SF's time curve. Tuning it is quite difficult (but I'm trying), so I encourage others to try some values and then maybe they can be set as default. Scale thinking time is back and moved out of the slow mover budgeting stuff so it doesn't change the time curve. I recommend that it is set to offset smart pruning.

GBeauregard:
Build for me fails on macOS 10.13.5. I think this is a C++ standards thing (?) - it doesn't like you using the math functions in a constexpr.

```
FAILED: lc0@exe/src_engine.cc.o
c++ -Ilc0@exe -I. -I../.. -I../../src -I/opt/intel/mkl/include -Isubprojects/zlib-1.2.11 -I../../subprojects/zlib-1.2.11 -I/System/Library/Frameworks/OpenCL.framework/Headers -I/opt/local/include -I/usr/local/include -Xclang -fcolor-diagnostics -DNDEBUG -pipe -Wall -Winvalid-pch -Wnon-virtual-dtor -std=c++14 -O3 -Wthread-safety -Wextra -pedantic -march=native -mtune=native -DUSE_MKL  -MD -MQ 'lc0@exe/src_engine.cc.o' -MF 'lc0@exe/src_engine.cc.o.d' -o 'lc0@exe/src_engine.cc.o' -c ../../src/engine.cc
../../src/engine.cc:92:19: error: constexpr variable 'width_scaler' must be initialized by a constant expression
  constexpr float width_scaler = 2 / std::log(2 + std::sqrt(3));
                  ^              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../../src/engine.cc:92:38: note: non-constexpr function 'log' cannot be used in a constant expression
  constexpr float width_scaler = 2 / std::log(2 + std::sqrt(3));
                                     ^
/usr/include/math.h:387:15: note: declared here
extern double log(double);
              ^
1 error generated.
[34/116] Compiling C++ object 'lc0@exe/src_mcts_search.cc.o'.
ninja: build stopped: subcommand failed.
```

fli:
@GBeauregard I precalculated that constant, should compile now

src/engine.cc
```diff
@@ -77,7 +87,13 @@ void EngineController::PopulateOptions(OptionsParser* options) {
   defaults->Set<int>(Search::kAllowedNodeCollisionsStr, 32);  // Node collisions
 }
 
-SearchLimits EngineController::PopulateSearchLimits(int /*ply*/, bool is_black,
+double move_weight(int ply, float peak, float left_width, float right_width) {
```
mooskagh: `ComputeMoveWeight` instead of `move_weight`.
Also it's better to move that function into anonymous namespace.
Also it makes sense for it to return float rather than double.

src/engine.cc
```diff
@@ -77,7 +87,13 @@ void EngineController::PopulateOptions(OptionsParser* options) {
...
+  const float width = ply > peak ? right_width : left_width;
+  constexpr float width_scaler = 1.518651485; // 2 / log(2 + sqrt(3))
```
mooskagh: Nitpicking: 1.518651485**f**.
Also if you run  clang-format, it will put two spaces before the comment rather than one. :)

src/engine.cc
```diff
@@ -77,7 +87,13 @@ void EngineController::PopulateOptions(OptionsParser* options) {
...
+  return std::pow(std::cosh((ply - peak) / width / width_scaler), -2);
```
mooskagh: `-2.0f` to ensure all arguments are float and there are no double <-> float conversion.

fli:
Made the changes, but I think there's one more thing to fix. The time given per move drops off a cliff at some point with increment. This is because the budgeting ahead for 50 moves doesn't take into account that there will be a maximum time to calculate when the clock time gets low enough. So I need to add something to consider that so it smooths out.
chart-2

PR #62, opened 2018-06-09
Implement dynamic temperature as a function of current eval

dubslow:
Also converts current linear movetemp decay to logarithmic, like lczero.

This is the followup to https://github.com/glinscott/leela-chess/pull/420 on the old repo.

I used @jkiliani's logarithm-of-polynomial formula, though with a power of 4 rather than 2 to keep close games at high temperature.

I'll soon include an example selfplay game with debug output from the commented out print statement here, for comparison.


killerducky:
Can you clarify if this is for matches, self-play, or both?

dubslow:
I added the functionality and command line options. What people use (for either matches or selfplay) is up to them.

I do hope to use it for selfplay though, in lieu of resign, in the same manner as jhellis3 proposed. I have some commented out selfplay settings that can act as a starting point for discussing it. Theoretically, by tuning the eval decay parameter, we can get game-length reductions equivalent to the 4% resign analyses done by Chad.

dubslow:
Here's an example game played with those commented out default selfplay params:

https://pastebin.com/a9kaPxYb (lichess pgn https://lichess.org/aFAnpMD7)

Note that sufficiently high temperature still allows lead changes, but when it gets to more than 5 pawns apart the temperature drops and the game ends shortly thereafter, shorter, I daresay, than with t=1.

src/mcts/search.cc
```diff
@@ -65,9 +67,14 @@ void Search::PopulateUciParams(OptionsParser* options) {
   options->Add<IntOption>(kMiniBatchSizeStr, 1, 1024, "minibatch-size") = 1;
   options->Add<IntOption>(kMaxPrefetchBatchStr, 0, 1024, "max-prefetch") = 32;
   options->Add<FloatOption>(kCpuctStr, 0.0f, 100.0f, "cpuct") = 1.2f;
-  options->Add<FloatOption>(kTemperatureStr, 0.0f, 100.0f, "temperature") =
-      0.0f;
-  options->Add<IntOption>(kTempDecayMovesStr, 0, 100, "tempdecay-moves") = 0;
+  options->Add<FloatOption>(kMoveTempStr, 0.0f, 100.0f, "movetemperature") = 0.0f;
+  options->Add<FloatOption>(kMoveTempDecayStr, 1.0f,
+                            std::numeric_limits<float>::max(),
```
mooskagh: It's better to set limits to something a little bit above reasonable rather than to totally unreasonable.
That way it's easier to see in --help and in uci UI (although for float it's not the case) what is the sensible range.
Also, it looks nicer in --help text.

src/mcts/search.cc
```diff
@@ -135,56 +144,72 @@ void ApplyDirichletNoise(Node* node, float eps, double alpha) {
...
+  if (movetemp < floor) movetemp = 0.0f;
+
+  // evaltemp ~ 1/log(eval^4*decay)
+  float eval = -root_node_->GetQ(0, 0);
+  float evaltemp = kEvalTemp /
+                   (1 + std::log(1 + kEvalTempDecay * std::pow(eval, 4)));
+  if (evaltemp < floor) evaltemp = 0.0f;
+
+  float temperature = movetemp + evaltemp;
+  //std::cout << "move: " << move << " movetemp: " << movetemp << " eval: " << eval << " evaltemp: " << evaltemp << " temp: " << temperature << "\n";
```
mooskagh: Remove this line.

src/mcts/search.cc
```diff
@@ -135,56 +144,72 @@ void ApplyDirichletNoise(Node* node, float eps, double alpha) {
...
-    }
-    std::tuple<int, float, float> val(node->GetN(), node->GetQ(-10.0, 0.0),
-                                      node->GetP());
-    if (val > best) {
-      best = val;
-      best_node = node;
+float Search::CalculateTemperature() const {
+  // Put a floor on temperature to limit powers computed. Temps below the floor
+  // are rounded to 0. We round each component separately for better control.
+  float floor = 0.10f; // semi-arbitrary
```
mooskagh: It would be nice to estimate what should be the real floor. I remember that for 10000000 total visits (reasonable realistic maximum), underflow happened at ~0.02.
Also, const.
Also, better name.

src/mcts/search.cc
```diff
@@ -135,56 +144,72 @@ void ApplyDirichletNoise(Node* node, float eps, double alpha) {
...
+
+  // movetemp ~ 1/log(move*decay)
+  int move = played_history_.Last().GetGamePly() / 2;
```
mooskagh: move_number?

src/mcts/search.cc
```diff
@@ -135,56 +144,72 @@ void ApplyDirichletNoise(Node* node, float eps, double alpha) {
...
+  float movetemp = kMoveTemp / (1 + std::log(1 + kMoveTempDecay * move));
+  if (movetemp < floor) movetemp = 0.0f;
+
+  // evaltemp ~ 1/log(eval^4*decay)
+  float eval = -root_node_->GetQ(0, 0);
+  float evaltemp = kEvalTemp /
+                   (1 + std::log(1 + kEvalTempDecay * std::pow(eval, 4)));
+  if (evaltemp < floor) evaltemp = 0.0f;
```
mooskagh: Do you have any graphs to see what that decay looks like?

src/mcts/search.cc
```diff
@@ -135,56 +144,72 @@ void ApplyDirichletNoise(Node* node, float eps, double alpha) {
...
+
+  // movetemp ~ 1/log(move*decay)
+  int move = played_history_.Last().GetGamePly() / 2;
+  float movetemp = kMoveTemp / (1 + std::log(1 + kMoveTempDecay * move));
+  if (movetemp < floor) movetemp = 0.0f;
+
+  // evaltemp ~ 1/log(eval^4*decay)
+  float eval = -root_node_->GetQ(0, 0);
+  float evaltemp = kEvalTemp /
+                   (1 + std::log(1 + kEvalTempDecay * std::pow(eval, 4)));
```
mooskagh: Are you sure that that's what you want for negative eval?
If yes, please add the comment with reasoning.

src/mcts/search.cc
```diff
@@ -135,56 +144,72 @@ void ApplyDirichletNoise(Node* node, float eps, double alpha) {
...
+
+  return temperature;
+}
+
+// Returns a child with most visits, if temp==0, or child chosen randomly
+// according to (visits_fraction)^(1/temp)
+Node* Search::GetBestChild(Node* parent,
+                           float temperature) const {
+
+  if (temperature <= 0.0f || parent->GetN() <= 1) {
```
mooskagh: Maybe factor out this to a separate function?
This function is too long and really does two things.

Also getting best child doesn't always involve temperature (e.g. for pv building and ponder).

src/mcts/search.cc
```diff
@@ -135,56 +144,72 @@ void ApplyDirichletNoise(Node* node, float eps, double alpha) {
...
+  if (evaltemp < floor) evaltemp = 0.0f;
+
+  float temperature = movetemp + evaltemp;
+  //std::cout << "move: " << move << " movetemp: " << movetemp << " eval: " << eval << " evaltemp: " << evaltemp << " temp: " << temperature << "\n";
+
+  return temperature;
+}
+
+// Returns a child with most visits, if temp==0, or child chosen randomly
+// according to (visits_fraction)^(1/temp)
```
mooskagh: Period at the end of the sentence.

src/mcts/search.cc
```diff
@@ -135,56 +144,72 @@ void ApplyDirichletNoise(Node* node, float eps, double alpha) {
...
+Node* Search::GetBestChild(Node* parent,
+                           float temperature) const {
+
+  if (temperature <= 0.0f || parent->GetN() <= 1) {
+    // temp shouldn't be < 0 ever, but whatever
```
mooskagh: Capitalize first word, use full word "temperature" unless it would cause the comment to break into multiple lines, period in the end.

src/mcts/search.cc
```diff
@@ -135,56 +144,72 @@ void ApplyDirichletNoise(Node* node, float eps, double alpha) {
...
+    Node* best_node = nullptr;
+    // break visit count ties with search-eval, and secondarily with policy
```
mooskagh: The same.
Also it requires some concentration to understand that comment. I'd change it to something more "human machine readable", e.g.:

```
Best node is selected based on:
* Largest node count.
* If it's equal and 0, then based on node priors.
* If it's equal and non-0, then based on node evals
```

src/mcts/search.cc
```diff
@@ -135,56 +144,72 @@ void ApplyDirichletNoise(Node* node, float eps, double alpha) {
...
+    return best_node;
   }
-  return best_node;
-}
 
-Node* Search::GetBestChildWithTemperature(Node* parent,
-                                          float temperature) const {
-  std::vector<float> cumulative_sums;
-  float sum = 0.0;
+  float coldness = 1 / temperature;
```
mooskagh: Const.

src/mcts/search.h
```diff
@@ -93,9 +96,19 @@ class Search {
...
-  // Returns a child with most visits.
-  Node* GetBestChild(Node* parent) const;
-  Node* GetBestChildWithTemperature(Node* parent, float temperature) const;
+  // Returns a random child according to (visit_fraction)^(1/temp), which
+  // simplifies to child with most visits when temp==0
+  Node* GetBestChild(Node* parent, float temperature = 0.0f) const;
+  // calculates current temperature based on ply, eval, and options passed
+  float CalculateTemperature() const;
+
+  bool NodeInSearchmoves(Node* node) const {
```
mooskagh: The function is too large for header file.
Also, if it's a predicate, should start with `Is`
Also, some few words comment would be beneficial, it's not that trivial.

src/mcts/search.cc
```diff
@@ -135,56 +144,72 @@ void ApplyDirichletNoise(Node* node, float eps, double alpha) {
...
+  std::vector<float> partial_sums;
+  float sum = 0.0f;
   const float n_parent = parent->GetN();
 
   for (Node* node : parent->Children()) {
-    if (parent == root_node_ && !limits_.searchmoves.empty() &&
-        std::find(limits_.searchmoves.begin(), limits_.searchmoves.end(),
-                  node->GetMove()) == limits_.searchmoves.end()) {
+    if (parent == root_node_ && NodeInSearchmoves(node)) {
```
mooskagh: I would add comment explaining what's going on here. Something like "If we are in UCI searchmoves mode, don't look into moves not listed there".

src/mcts/search.h
```diff
@@ -21,6 +21,7 @@
 #include <functional>
 #include <shared_mutex>
 #include <thread>
+#include <algorithm>
```
mooskagh: Not sorted alphabetically. :)
But clang-format will do that for you.

src/selfplay/tournament.cc
```diff
@@ -70,7 +70,12 @@ void SelfPlayTournament::PopulateOptions(OptionsParser* options) {
   auto defaults = options->GetMutableDefaultsOptions();
   defaults->Set<int>(Search::kMiniBatchSizeStr, 32);     // Minibatch size
   defaults->Set<bool>(Search::kSmartPruningStr, false);  // No smart pruning
-  defaults->Set<float>(Search::kTemperatureStr, 1.0);    // Temperature = 1.0
+  defaults->Set<float>(Search::kMoveTempStr, 1.0);    // Temperature = 1.0
+  defaults->Set<float>(Search::kMoveTempDecayStr, 0.0); // no temp decay
+  //defaults->Set<float>(Search::kMoveTempStr, 1.0);
```
mooskagh: What are those lines for?

src/mcts/search.cc
dubslow: Yea that's one thing I knew wasn't optimal. The problem is, what's a suitably large max? Being logarithmic, values like 1e6 or 1e8 are perfectly reasonable. 1e10? 1e15?

src/mcts/search.cc
dubslow: 0.10 is what lczero used, but more importantly, it's easy for humans to remember. I had slightly different floors in testing based on semiarbitrary default values for movedecay, but it was so arbitrary I decided that "humanly memorable" was better than "calculated according to some arbitrary criterion".

And you mean underflow happened with temperature 0.02? The main reason to have a floor is to avoid large values of `coldness`, not necessarily underflow, though that is another good reason to have a floor.

src/mcts/search.cc
dubslow: I'm kinda disinclined to change that, it has the same number of characters as `eval`, but also I don't think move_number is any clearer?

src/mcts/search.cc
dubslow: I'm going to change it back to 2 I think, but yea I can post graphs

src/mcts/search.cc
dubslow: Well if you view the example game, note that the eval for white's first move is positive. Without this negative sign, the eval for white's first move is negative, which is of course nonsense. (Basically, all evals throughout the game were flipped, compared against lichess-stockfish to be absolutely sure.) It doesn't *truly* matter since we want evaltemp to be symmetric around 0 anyways, but it made that commented print statement much easier to understand.

I will add a comment to this effect.

src/mcts/search.cc
dubslow: Well I asked you about it on discord yesterday, I guess my opinion is that this is an API function, and users on the outside (pv building, move selection, ponder) don't really care about the finer details of tiebreaking with no temperature vs with temperature. All they care about is whether or not temperature is 0 or not, and I don't see the point of having two API functions to do the same thing where the only difference is the value of the argument... that's the whole point of arguments. And in search.h you'll see that temperature has default of 0, so pv building and ponder use cases don't look any different than they did before (that is, they don't need to even consider temperature, since it's default off for this function).

If you want though I guess we could make private-with-temperature function be a slave to the main method or something, but I'm definitely of the opinion that from a public perspective, there should only be one function.

src/mcts/search.cc
dubslow: As best as I understood though, this comment was inaccurate anyways. "Node count", which I take to mean visit count, should never be less than 1 anyways, and furthermore the comparison tuple you make doesn't actually make any reference to visit counts being 0.

I guess I can expand/clarify the replacement comment though

src/mcts/search.cc
dubslow: This should have been addressed in the PR that created that code then :P

src/selfplay/tournament.cc
dubslow: These are my current suggested defaults for selfplay. I guess I can remove them for PR purposes

src/mcts/search.cc
dubslow: Ah, I misunderstood this original comment of yours. I'll tweak it slightly.

dubslow:
Here's another two games played with a squared dynamic eval: https://ghostbin.com/paste/z37m4

dubslow:
@mooskagh I didn't change everything you requested, but many things I did. The max option value is now 1e15, but unfortunately that still is one character too long, apparently: 


```./lc0 --help
       _
|   _ | |
|_ |_ |_| built Jun 10 2018
Usage: ./lc0 [<mode>] [flags...]

Available modes. A help for a mode: ./lc0 <mode> --help
  uci        (default) Act as UCI engine
  selfplay   Play games with itself
  debug      Generate debug data for a position

Allowed command line flags for current mode:
  -h,  --help                         Show help and exit.
  -w,  --weights=STRING               Network weights file path.
                                      (default: <autodiscover>)
  -t,  --threads=1..128               Number of worker threads.
                                      (default: 2  min: 1  max: 128)
       --nncache=0..999999999         NNCache size.
                                      (default: 200000  min: 0  max: 999999999)
       --backend=CHOICE               NN backend to use.
                                      (default: opencl  values: opencl,blas,check,random,multiplexing)
       --backend-opts=STRING          NN backend parameters.
       --slowmover=0..100             Scale thinking time.
                                      (default: 2.2  min: 0  max: 100)
       --move-overhead=0..10000       Move time overhead in milliseconds.
                                      (default: 100  min: 0  max: 10000)
       --minibatch-size=1..1024       Minibatch size for NN inference.
                                      (default: 256  min: 1  max: 1024)
       --max-prefetch=0..1024         Max prefetch nodes, per NN call.
                                      (default: 32  min: 0  max: 1024)
       --cpuct=0..100                 Cpuct MCTS option.
                                      (default: 3.4  min: 0  max: 100)
       --movetemperature=0..100       Initial move-decay temperature.
                                      (default: 0  min: 0  max: 100)
       --movetemperature-decay=1..1e+15 Move-decay temperature decay rate.
                                      (default: 100  min: 1  max: 1e+15)
       --evaltemperature=0..100       Initial eval-decay temperature.
                                      (default: 0  min: 0  max: 100)
       --evaltemperature-decay=1..1e+15 Eval-decay temperature decay rate.
                                      (default: 5  min: 1  max: 1e+15)
  -n,  --[no-]noise                   Add Dirichlet noise at root node.
                                      (default: false)
       --[no-]verbose-move-stats      Display verbose move stats.
                                      (default: false)
       --[no-]smart-pruning           Enable smart pruning.
                                      (default: true)
       --virtual-loss-bug=-100..100   Virtual loss bug.
                                      (default: 0  min: -100  max: 100)
       --fpu-reduction=-100..100      First Play Urgency Reduction.
                                      (default: 0.9  min: -100  max: 100)
       --cache-history-length=0..7    Length of history to include in cache.
                                      (default: 7  min: 0  max: 7)
       --extra-virtual-loss=0..100    Extra virtual loss.
                                      (default: 0  min: 0  max: 100)
       --policy-softmax-temp=0.1..10  Policy softmax temperature.
                                      (default: 2.2  min: 0.1  max: 10)
       --allowed-node-collisions=0..1024 Allowed node collisions, per batch.
                                      (default: 32  min: 0  max: 1024)
  -l,  --debuglog=STRING              Do debug logging into file.
```

Would it be acceptable to you to abbreviate the actual option names to `*temp` instead of `*temperature`?

dubslow:
Oh and here's a plot of various choices for evaltemp decay, currently the default constant is 5 and the power is 2 https://www.wolframalpha.com/input/?i=plot+1%2F(1%2Blog(1%2B5*x%5E2)),+1%2F(1%2Blog(1%2B5*x%5E4)),+1%2F(1%2Blog(1%2B10*x%5E4)),++1%2F(1%2Blog(1%2B8*%7Cx%7C%5E3)),+-1%3Cx%3C1

dubslow:
Rebased on current master, less changes to search. Please re-review.

src/mcts/search.cc
```diff
@@ -36,8 +36,10 @@ namespace lczero {
 const char* Search::kMiniBatchSizeStr = "Minibatch size for NN inference";
 const char* Search::kMaxPrefetchBatchStr = "Max prefetch nodes, per NN call";
 const char* Search::kCpuctStr = "Cpuct MCTS option";
-const char* Search::kTemperatureStr = "Initial temperature";
-const char* Search::kTempDecayMovesStr = "Moves with temperature decay";
+const char* Search::kMoveTempStr = "Initial move-decay temperature";
```
mooskagh: That's fine to have that zoo of params for testing, but after few days we should reduce it.
Was it tested that having eval-decay actually helps with anything, or it's purely theoretical speculation?

src/mcts/search.cc
mooskagh: This comment doesn't seem to be addressed.

src/mcts/search.cc
dubslow: Well we need *some* eval-decay if/when we ever try this on main net. It's like resign threshold, the correct value may change. (And at any rate, "actual" testing hasn't happened yet. This would need to be merged first, followed by a run on the test pipeline before we can make any conclusions about reasonable eval-decays are.)

mooskagh:
Is this still actual?

dubslow:
Eh, I'm less interested in this for now. Maybe someone else can take it up again later.

PR #64, opened 2018-06-10
Print out a version string so the server can do version checks.

jnewlin12345:


Tilps:
Heh, I didn't see this PR before I wrote mine.

I don't mind either way - I'll be happy to have mooskagh make the call here.

Issue #66, opened 2018-06-10
Random backend only includes mask in hash.


Tilps:
It should also include value so that rule 50 plane affects the random output.

PR #68, opened 2018-06-11
Mix in the plane value when calculating random backend hash.

Tilps:
Quick hack to address #66.

### DIFF
--- a/src/neural/network_random.cc
+++ b/src/neural/network_random.cc
@@ -31,11 +31,8 @@ class RandomNetworkComputation : public NetworkComputation {
     std::uint64_t hash = 0;
     for (const auto& plane : input) {
       hash = HashCat({hash, plane.mask});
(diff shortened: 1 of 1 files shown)


src/neural/network_random.cc
```diff
@@ -31,6 +31,12 @@ class RandomNetworkComputation : public NetworkComputation {
     std::uint64_t hash = 0;
     for (const auto& plane : input) {
       hash = HashCat({hash, plane.mask});
+      // This assumes value is not large and not very granular, which should
+      // be a reasonable assumption for a NN input.
+      std::uint64_t value_hash =
```
mooskagh: What about `reinterpret_cast<std::uint32_t>(plane.value)`? :)

Tilps: Sounds better - I'll do that.

Tilps: Done.

Issue #61, opened 2018-06-09
Black moves pieces towards h1 corner


killerducky:
http://testserver.lczero.org/match/28
http://testserver.lczero.org/match_game/27988 (white is test31, black is test27).
This match was probably played with the T=1 bug, which is why it lasted so long in the first place. But another question is why does black tend to move its pieces towards h1? It's probably because moves towards h1 for black (h8 for white) have the largest encoded values. So possibly something is overflowing, and all results are tied, and the final tie-break is the size of the move encoding.

So far though I'm not able to reproduce that, search looks normal, it wants to play 150.Bb7.
```
[...]

PR #69, opened 2018-06-11
Avoid all zero policy which can behave badly.

Tilps:
Might help with #61.

This is not identical to lczero, in that I've forced Dirichlet noise because equal priors are also not a great option due to the way that UCTSearch selects the next child to visit.

But its a large improvement for the priors to just not be zero, so I'm happy to revert the force_noise path
(Or rework it with a smaller noise amount) if you prefer.

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -786,6 +786,7 @@ void SearchWorker::FetchNNResults() {
     node->SetV(-computation_->GetQVal(idx_in_computation));
     // Populate P values.
     float total = 0.0;
+    int total_children = 0;
     for (Node* n : node->Children()) {
       float p =
           computation_->GetPVal(idx_in_computation, n->GetMove().as_nn_index());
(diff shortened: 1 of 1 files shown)


src/mcts/search.cc
```diff
@@ -794,14 +795,26 @@ void SearchWorker::FetchNNResults() {
...
+      // deterministicly biased towards earlier children rather than later.
+      float value = 1.0f / static_cast<float>(total_children);
+      for (Node* n : node->Children()) n->SetP(value);
+      force_noise = true;
     }
-    // Add Dirichlet noise if enabled and at root.
-    if (search_->kNoise && node == search_->root_node_) {
+    // Add Dirichlet noise if enabled and at root or forced to avoid bias
+    // from uniform priors.
+    if (search_->kNoise && node == search_->root_node_ || force_noise) {
```
dubslow: Consider adding parentheses, since I think most people don't know the order of ands vs ors off the the top of their head. I certainly don't

Tilps: Sure, I can do that - but I'll wait until someone else weighs in on whether the force_noise path is a bad idea or not.

killerducky:
I wouldn't do force noise. Even if policy is uniform, the value can steer the search. So I don't think it's necessary.

Without this bug fix though in lczero P=0 cancels even the exploring part of the U term, so the search put 100% of moves into the first move it picked. Then value cannot steer the search at all.

src/mcts/search.cc
```diff
@@ -786,6 +786,7 @@ void SearchWorker::FetchNNResults() {
     node->SetV(-computation_->GetQVal(idx_in_computation));
     // Populate P values.
     float total = 0.0;
+    int total_children = 0;
```
mooskagh: node is guaranteed to have children here, so no need to count them.
Instead `assert(node->HasChildren());` would be useful.

`force_noise` may be unneeded too, `total == 0.0f` (or even calling ApplyDirichletNoise() right after filling the vector with values) may work, although I'm not sure.

..and actually, no need to initialize it. Passing 1 as `eps` param to ApplyDirichletNoise() probably does exactly the same. Or maybe not exactly, but still good enough for sure.

Tilps: total children is being counted so we can create a correct size uniform distribution - the fact that I use it in the if statement is just defensive programming.

We could use eps of 1 instead of counting - it will make the distribution even more random though.

Tilps:
The problem which adding noise aims to solve is that in a scenario where your policy is so bad that it can't even give you a single valid move - your value head is also likely to be so bad that it almost always gives 0 or 1.
Thus value head steering is unlikely to be effective at hiding the bias that the first move explored is always the first move in the in move list, in the case of a uniform distribution.  Adding noise ensures that bias does not exist, so does not need to be hidden by value steering.

src/mcts/search.cc
Tilps: Reason why I'm not advocating for more random is because the only reason I want any random at all is to ensure that the policy priors are not all *exactly* equal.
That being said I can't justify why more random would necessarily be bad.

Tilps: Happy to make changes here - but is there a decision about what to?

mooskagh:
Do you still want to have this in code?
If not, probably better to close this PR.

Tilps:
I think the code is potentially valid, but only if we ever stop bootstrapping off of random backend data.  So, yeah could just close for now.  Feel free to do so if you agree.

PR #71, opened 2018-06-11
Add version number to id name.  Also send it for non-interactive tournaments.

### LINKED ISSUE #70: lc0 needs a version number so server can perform limiting to reject versions with known bugs or lacking features.
Current output includes the build date, but this doesn't tell us much about the code.

My thought is to add a version number to the id name string like lczero does. Whenever there are training affecting changes committed to the source control everyone uses to build it, this version number should be increased.

Tilps:
To fix #70.


src/chess/uciloop.cc
```diff
@@ -200,6 +200,11 @@ void UciLoop::SendResponse(const std::string& response) {
   std::cout << response << std::endl;
 }
 
+void UciLoop::SendId() {
+  SendResponse("id name The Lc0 chess engine. v0.13");
```
jnewlin12345: I kind of would like to change the version string to be xx.yy.zz  major/minor/patch.  I have server side change pending to accept either.

I would just suggest one thing here, move the version into it's own var (maybe file)
static const version[] = "v0.13"  or "0.13.0"

SendResponse(StrCat("id name..", version));

jnewlin12345: Hmm also can you print the version at startup like:
https://github.com/LeelaChessZero/lc0/pull/64/files

Or will that screw up uci engines?  We need the version in self-play mode as well.

Tilps: I centralised the id printing to this one place, and made self play mode invoke it as well.

PR 64 is simpler and probably fine.

The server side shouldn't need a change to accept either I would have thought, the version parser should parse either already?

dubslow:
I'm in favor of this PR, being more flexible and more useful, but I agree that the version shouldn't be embedded in the string. Perhaps put it in `src/version.h` where anyone (including external tools) can get at it?

src/chess/uciloop.cc
mooskagh: I'd also suggest to put version into a separate file, something like `version.inc` (which would be included into .cc file but is not really a header).
When we switch to auto-push of a new version, this file will be automatically generated, so it should be simple.

Also, I think we should switch to "v0.13.00" format of the version.

dubslow:
The version number should not have a "v" in it (let the string interpolation add that), and only one '0' (so 0.11.0), but otherwise agreed. I like the header-that-isn't-a-header idea too.

Tilps:
Ok, I'll make the changes in a few hours when I'm free.

mooskagh:
I like that sa bit more:
```cpp
constexpr int kLc0Version = 120305;  // means v12.03.05
```
But it doesn't really have any difference as they can be converted back and forth easily.

Tilps:
I moved version in to its own file.  Let the complaints about the fact that I like preprocessor macros for these things commence :)

src/version.inc
```diff
@@ -0,0 +1,9 @@
+#define LC0_VERSION_MAJOR 0
+#define LC0_VERSION_MINOR 13
+#define LC0_VERSION_PATCH 0
+#define STR_INNER(a) #a
+#define STR(a) STR_INNER(a)
+#define LC0_VERSION_STRING     \
```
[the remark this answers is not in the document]
Tilps: ## doesn't work between string constant and a numeric constant.

I moved the logic to the one place that needs it rather than have STR having to escape a header in order for LC0_VERSION_STRING to be useful.

I changed LC0_VERSION_PATCH to be 0 padded as requested, although I don't think its a great idea.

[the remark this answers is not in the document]
Tilps: Actually pushed the change to be 0 padded this time... heh.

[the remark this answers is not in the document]
dubslow: I don't like "0.01.02" because it's highly nonstandard, and is liable to confuse a lot of code that assumes the standard form.

As for the macros, `version.inc` should only contain the defines/constexprs whatever they are. How the version information is formatted into a string is unrelated to the actual version numbers themselves, and so version.inc should not have any string formatting code, or indeed any code at all as crem said.

(Data: https://en.wikipedia.org/wiki/Software_versioning#Schemes or https://semver.org/)

src/chess/uciloop.cc
```diff
@@ -200,6 +201,21 @@ void UciLoop::SendResponse(const std::string& response) {
   std::cout << response << std::endl;
 }
 
+#define STR_INNER(a) #a
+#define STR(a) STR_INNER(a)
```
dubslow: There's still a good chance, in general, that you're accidentally overwriting some other macro with this (and #undef doesn't help, since it can't restore *previous* macros). Though I guess it's not too much of a problem for a non-library like lc0.

jnewlin12345: personally don't care but seems like a lot of macro magic, why not just
const static char version[] = "0.13.0";  or whatever

if you really want the macros maybe just add:
#ifdef STR_INNER
#error "oh no"
#endif 

etc

src/chess/uciloop.cc
Tilps: This comment was addressed. Commenting to convince github of that.

Error323:
My personal strategy for dealing with versioning is to extract the version from git through e.g. `git describe --tags --always`. This way a version bump is fully automated. The `version.inc` could be generated by meson and git for this.

If that's not desirable then I approve of this PR :)

Tilps:
I'm not volunteering to modify the meson to do that - although it sounds like a decent idea.

dubslow:
Although I think no one is especially happy with the current state of affairs, it works and we need to get on with lc0 promotion to main pipeline. I expect there will be future PRs both to remove the macros and make it on the whole more extensible.

PR #72, opened 2018-06-11
Add an option to do at most one visit per child.

### LINKED ISSUE #27: Create a PR585 type variant for lc0
Either as a configuration parameter, or just as a never-to-submit PR which has it hard coded, implement a search option which gives every root child exactly one visit.
This should be able to be batched very efficiently, which might allow Aloril's spreadsheet to update PR585 columns much faster.

(As a stretch goal, support arbitrary depth alpha-beta search using value head, with policy head suggesting which child to search first to optimize the alpha-beta.  This however requires a different get_best_child implementation and probably doesn't provide a massive increase in value over the simple depth1 case for testing purposes.)

Tilps:
To fix #27.
This visit limiting in picking results in tiebreaking based on eval - effectively creating the depth1 mode from PR585 for lczero.

Unlike PR585 this doesn't ensure that every child gets a visit, But if your visits limit is greater then maximum children it should work the same.


Tilps:
This PR doesn't work yet... apparently my analysis was incomplete.

Tilps:
Okay, fixed the infinite background loop consuming all the universes memory after it returned the answer.  Also made it not run forever for go infinite if this flag is in play, since that seems silly.  I think this is probably okay to use now.

mooskagh:
Hm, it doesn't look nice to introduce debug flag just for some non-generic use case..
Maybe there's a way to make `go depth 1` work like this.. And have some similar support of `go depth 2` and so on.. Not sure then how they should work. Will think!

Tilps:
What I really want is an alternative search which uses A/B instead of mcts and flag switch between them :P

dubslow:
`go depth` should not change the behavior/nature of MCTS.

mooskagh:
I'm all for multiple search implementations, feel free to do. :-P
Instead of A/B, for debug purposes, something simpler like "fixed depth" may be worthwhile, or even fixed depth 1.

It can be a separate lc0 mode. E.g. `lc0 exhaustive` instead of `lc0 [uci]`.

Tilps:
Merged master into this so its possible to do depth1 testing with SE nets.  However I'm lazy and there was a lot of conflicts, so I removed the command line flag to make my life simpler - this PR is thus *definitely* not suitable for submission as is.

src/mcts/search.cc
```diff
@@ -860,6 +863,11 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend(
     int possible_moves = 0;
     const float fpu = GetFpu(params_, node, is_root_node);
     for (auto child : node->Edges()) {
```
mooskagh: It crashes when it happens that all moves have `GetN() > 0` (and `possible_moves == 0`).
Then during the next iteration `best_edge` is used, but it doesn't point to anything useful.

Tilps: So, maybe workaround would be to pass the collision event max flag to 1 (like it used to be?)

Tilps: Seems collisions event max of 1 was not sufficient.  But out of order eval being disabled is (or maybe is in combination with setting collisions max to 1).

Tilps: actually none of that was sufficient - so I made some more fixes.
Along the way I think I found and fixed a performance issue.

But still needs threads=1 to be correct I think, otherwise there is a chance that the least desirable child node won't be evaluated before the result is returned.  Could be fixed by changing the flag to only be set if there are 0 possible moves, rather than 1 or less.

Tilps:
I found what I think is another potential bug - which could cause it to take a terminal draw or loss in preference to any other move by virtue of terminal hits potentially increasing the visit count over 1.
Added exceptions to crash out if the visit count is not exactly 1 for every child at end when move is selected.

Tilps:
Merged master to pick up the cuda bug fix so this can be used for T37/T40.

Tilps:
Merged latest master into this PR - so it can support T50 networks (assuming the merge didn't go badly).

Tilps:
Updated branch one_visit to latest master - should now survive networks with negative gammas.

Naphthalin:
Is the problem this PR tries to solve still relevant, or do we avoid this now with a different batching logic?

AlexGreason:
This isn't really "solving a problem" - it's used to get value head strength independent of the policy head, by getting value head output for all immediate children of root and choosing a move based on that, which is where the "value head ordo" measurement comes from.

mooskagh:
I think it's "never to be merged but kept in sync with master" branch to run value head tests.

Tilps:
Yeap - its not been used much recently since Aloril stopped doing value head tests vs SF.  I use value_tournament branch instead which is *much* faster for doing value head testing of lc0 vs lc0.
But in theory this still has value, it just needs to be synced and conflicts resolved.

Tilps:
Closing in preference for my newer attempt to do this.

Issue #20, opened 2018-06-02
Implement resign in lc0


mooskagh:
_From @mooskagh on May 27, 2018 16:13_

Resign is not yet implemented in lc0. Implement that.

_Copied from original issue: glinscott/leela-chess#674_

mooskagh:
Best done via adding `bool want_resign;` into `BestMoveInfo`

gonzalezjo:
Do we want cfg_min_resign_moves to be changeable from the command line? That wasn't the case in lczero, but the naming conventions suggest that it was intended.

mooskagh:
Why would we ever need to limit minimum number of moves before resign at all... I'd not implement this functionality unless there is justification.

As for resign threshold, that surely should be configurable.

gonzalezjo:
No idea, but it was in lczero. I'll leave that out, then.

killerducky:
Just for history: It was originally in LZGo so it doesn't instantly resign handicap games without at least trying. They later modified it with more stuff like scaled based on handicap stones.

dubslow:
Closed by #80, I believe.

PR #80, opened 2018-06-12
Implement resign in selfplay games

dubslow:
Per discussion in #47 (which this replaces), using the design that crem and I both came up with. I've tested that it in fact works: https://ghostbin.com/paste/75mj5

### DIFF
--- a/src/selfplay/game.cc
+++ b/src/selfplay/game.cc
@@ -80,7 +80,7 @@ void SelfPlayGame::Play(int white_threads, int black_threads) {
         GameResult::UNDECIDED, tree_[idx]->GetPositionHistory()));
 
     const float resignpct =
-            options_[idx].uci_options->Get<float>(kResignPctStr);
+            options_[idx].uci_options->Get<float>(kResignPctStr) / 100;
     if (resignpct > 0) {
       float eval = search_->GetBestEval();
       eval = (eval + 1) / 2;
--- a/src/selfplay/game.cc
+++ b/src/selfplay/game.cc
@@ -25,10 +25,12 @@ namespace lczero {
 
 namespace{
 const char* kReuseTreeeStr = "Reuse the node statistics between moves";
+const char* kResignPctStr = "Resign when win percentage drops below n";
 }
 
 void SelfPlayGame::PopulateUciParams(OptionsParser* options) {
   options->Add<BoolOption>(kReuseTreeeStr, "reuse-tree") = true;
+  options->Add<FloatOption>(kResignPctStr, 0, 100, "resignpct", 'r')  = 0.0f;
 }
 
 SelfPlayGame::SelfPlayGame(PlayerOptions player1, PlayerOptions player2,
@@ -77,8 +79,17 @@ void SelfPlayGame::Play(int white_threads, int black_threads) {
     training_data_.push_back(tree_[idx]->GetCurrentHead()->GetV3TrainingData(
         GameResult::UNDECIDED, tree_[idx]->GetPositionHistory()));
 
+    auto move_data = search_->GetBestMove();
+    auto move = std::get<0>(move_data);
+
+    // Resign if practical.
+    if (std::get<2>(move_data)) {
+      game_result_ = blacks_move ?
+        GameResult::WHITE_WON : GameResult::BLACK_WON;
+      break;
+    }
+
     // Add best move to the tree.
-    Move move = search_->GetBestMove().first;
     tree_[0]->MakeMove(move);
     if (tree_[0] != tree_[1]) tree_[1]->MakeMove(move);
     blacks_move = !blacks_move;
--- a/src/selfplay/game.cc
+++ b/src/selfplay/game.cc
@@ -25,12 +25,13 @@ namespace lczero {
 
 namespace{
 const char* kReuseTreeStr = "Reuse the node statistics between moves";
-const char* kResignPctStr = "Resign when win percentage drops below n";
+const char* kResignPercentageStr = "Resign when win percentage drops below n";
 }
 
 void SelfPlayGame::PopulateUciParams(OptionsParser* options) {
   options->Add<BoolOption>(kReuseTreeStr, "reuse-tree") = false;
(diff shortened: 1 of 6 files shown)


dubslow:
I believe this also closes #20? It's not a full implementation, using BestMoveInfo, but it does what the project needs?

killerducky:
Still needed: A way to measure what the resign threshold should be. A 2nd command line switch to set % of games where resign is disabled completely (AGZ paper used 10% for this).

gonzalezjo:
@killerducky 


Wasn't that supposed to be implemented on the server? If I understand what you're referring to, then the PR that enables resign on the server does that here: https://github.com/glinscott/leela-chess/pull/465/files

src/selfplay/game.cc
```diff
@@ -25,10 +25,12 @@ namespace lczero {
 
 namespace{
 const char* kReuseTreeeStr = "Reuse the node statistics between moves";
+const char* kResignPctStr = "Resign when win percentage drops below n";
 }
 
 void SelfPlayGame::PopulateUciParams(OptionsParser* options) {
   options->Add<BoolOption>(kReuseTreeeStr, "reuse-tree") = true;
+  options->Add<FloatOption>(kResignPctStr, 0, 100, "resignpct", 'r')  = 0.0f;
```
mooskagh: Maybe `--resign-percentage`? `pct` is a but cryptic, and generally we separate words with hyphens (although tempdecay-moves is an exception)

src/selfplay/game.cc
```diff
@@ -25,10 +25,12 @@ namespace lczero {
 
 namespace{
 const char* kReuseTreeeStr = "Reuse the node statistics between moves";
```
mooskagh: Maybe while you here, also change Treee to Tree? I missed it in previous review.

src/selfplay/game.cc
```diff
@@ -45,7 +47,7 @@ SelfPlayGame::SelfPlayGame(PlayerOptions player1, PlayerOptions player2,
   }
 }
 
-void SelfPlayGame::Play(int white_threads, int black_threads) {
+void SelfPlayGame::Play(int white_threads, int black_threads, bool use_resign) {
```
mooskagh: For my taste `allow_resign` name would fit better, but I'm ok with keeping it.

src/selfplay/game.cc
```diff
@@ -73,10 +75,24 @@ void SelfPlayGame::Play(int white_threads, int black_threads) {
...
 
-    // Append training data.
+    // Append training data. The GameResult is later overwritten.
     training_data_.push_back(tree_[idx]->GetCurrentHead()->GetV3TrainingData(
         GameResult::UNDECIDED, tree_[idx]->GetPositionHistory()));
 
+    if (use_resign) {
+      const float resignpct =
+              options_[idx].uci_options->Get<float>(kResignPctStr) / 100;
+      if (resignpct > 0) {
```
mooskagh: This `if` is not really needed.

dubslow:
Seems to work as intended https://ghostbin.com/paste/9nfo9

src/selfplay/tournament.cc
```diff
@@ -65,6 +66,8 @@ void SelfPlayTournament::PopulateOptions(OptionsParser* options) {
       "multiplexing";
   options->Add<StringOption>(kNnBackendOptionsStr, "backend-opts");
   options->Add<BoolOption>(kVerboseThinkingStr, "verbose-thinking") = false;
+  options->Add<FloatOption>(kResignPlaythroughStr, 0, 100, "resign-playthrough")
```
mooskagh: 0.0f and 100.0f (one of compilers shows warning otherwise, don't remember which but probably MSVC)

src/selfplay/tournament.cc
```diff
@@ -40,6 +40,7 @@ const char* kTrainingStr = "Write training data";
 const char* kNnBackendStr = "NN backend to use";
 const char* kNnBackendOptionsStr = "NN backend parameters";
 const char* kVerboseThinkingStr = "Show verbose thinking messages";
+const char* kResignPlaythroughStr = "What percentage of games ignore resign";
```
mooskagh: It's not very consistent to have question rather than noun phrase. Maybe `Percentage of games with disabled resign`?

src/selfplay/tournament.cc
```diff
@@ -233,8 +237,13 @@ void SelfPlayTournament::PlayOneGame(int game_number) {
   }
   auto& game = **game_iter;
 
+  bool use_resign = true;
+  if (kResignPlaythrough > 0) {
+    use_resign = Random::Get().GetFloat(100) >= kResignPlaythrough;
```
mooskagh: 100.0f

src/selfplay/tournament.cc
```diff
@@ -233,8 +237,13 @@ void SelfPlayTournament::PlayOneGame(int game_number) {
   }
   auto& game = **game_iter;
 
+  bool use_resign = true;
+  if (kResignPlaythrough > 0) {
```
mooskagh: `if` is not really needed. (and creation of a game is a rare event so we don't have to care about performance)

src/selfplay/game.cc
dubslow: Perhaps `disable_resign`? Since what we're doing is an override, the variable should be true when we're taking positive action

src/selfplay/tournament.cc
dubslow: Done

src/selfplay/tournament.cc
dubslow: Done (though others nearby don't have the `f`)

src/selfplay/game.cc
dubslow: Done, ish, please re-review

src/selfplay/game.cc
dubslow: Done

src/selfplay/game.cc
dubslow: Done

PR #83, opened 2018-06-13
Random seed option for random backend

### LINKED ISSUE #67: Random backend isn't random enough.
I think that for generating a large initial sets of positions each game should be played between random backends each of which have a different randomly selected seed to their hash.
(Can still have a default of 0 seed for producing predictable output.)

Tilps:
To fix #67.


src/neural/network_random.cc
```diff
@@ -26,9 +26,9 @@ namespace lczero {
 
 class RandomNetworkComputation : public NetworkComputation {
  public:
-  RandomNetworkComputation(int delay) : delay_ms_(delay) {}
+  RandomNetworkComputation(int delay, int seed) : delay_ms_(delay), seed_(seed) {}
```
Error323: Maybe give seed a default value of 0?

`RandomNetworkComputation(int delay, int seed = 0) : delay_ms_(delay), seed_(seed) {}`

PR #87, opened 2018-06-15
Batch computation for BLAS.

frpays:
### x2 overall speedup.

I was able to re-order the dimension of V/M matrices, so that I could join the inner sgemm of the Winograd convolve3  into on single sgemm for every tile. Since the weights-tile matrix does not change, this leads to about x2 speed-up.
I also made various optimisations into the scatter/gather methods of the Winograd convolve3. I also reduced significantly the allocations and the pressures on new/delete. Now the vector allocations are once for all the computation. I also removed an unnecessary copy that dealt with the skip-connection.
Finally, I reduced the memory footprint of the computation, removing the input buffer and relying on 3 rotating buffers.

Benchmarks:

https://docs.google.com/spreadsheets/d/1eU2gb67V69_7yXFzIaF2WI4djF1pXGAih-rwhrstmg0

(less complete but formerly https://docs.google.com/spreadsheets/d/1vjPisPj2mbjc2Ic481E57h9O8o7ClyqYVXgU4HWoADQ/)

References:

https://arxiv.org/pdf/1509.09308.pdf
https://ai.intel.com/winograd/
https://ai.intel.com/winograd-2/

### Details

The initial non-batched algorithm looks like that:

```
foreach batch in batches 
   foreach tile in tiles
       V := scatter (input, batch)
       M[tile] := W[batch][time] x V[tile]
       output := gather(M, batch)
```

You can make M and V `batch_size times` larger and batch `scatter`and `gather`.

```
foreach tile in tiles
   V := scatter (input)
   foreach tile in tiles 
       M[batch][tile] := W[tile] x V[batch][tile]
   output := gather(M)
``` 

With the proper reordering of M and V, the inner multiplication can be contiguous *along the columns* of M and V. Note that W is fixed and does not change along the batch. So the *tiles* inner matrix multiplications are joined in one  multiplication. The input/output columns are *batch_size* times larger. 

```
foreach tile in tiles
   V := scatter (input)
    M[batch]:= W[tile] x V[batch
   output := gather(M)
```
   **We reuse W[batch] for all the tiles, this is what gives us the speedup**.

### Known problems

This change is known to be affected by https://github.com/LeelaChessZero/lc0/issues/98, but is not responsible for.


dubslow:
That sounds like these optimizations can be pushed back to Leela Go, correct?

frpays:
> That sounds like these optimizations can be pushed back to Leela Go, correct?

If they still use the CPU-based Winograd convolution3x3 algorithm, yes. But with a lot of care since ours is mildly specialised to a 8x8 board. But the general principle can be applied.
It's the following:

The convolution buffer U and M have shape [ kWinogradTiles,  kTiles,  channels]. 
For 8x8 board, kWinogradTiles and kTiles are both 16.


When you batch the convolution, they have shape [ batch, kWinogradTiles, kTiles, channels ]

gsobala:
I've had a segfault:6 and a 
```
lc0-pr85(9019,0x70000810b000) malloc: *** error for object 0x7ff0b118a128: incorrect checksum for freed object - object was probably modified after being freed.
*** set a breakpoint in malloc_error_break to debug
Abort trap: 6
```
when running this on macos with ```lc0-pr85 --backend=multiplexing "--backend-opts=a(backend=blas,threads=8),b(backend=opencl,threads=3)" -t 11 --minibatch-size=64``` compiled against the Acclerate framework (no mkl)

frpays:
> I've had a segfault:6 

Which network ?

src/neural/network_blas.cc
```diff
@@ -17,270 +17,314 @@
...
-                                          weights_.policy.bn_stddivs.data());
+    Convolution<1>::Forward(
+        batch_size, output_channels, num_policy_input_planes, conv_out,
+        weights_.policy.weights.data(), weights_.policy.biases.data(),
+        policy_buffer.data());
 
-    Transforms::Batchnorm<width * height>(NUM_VALUE_INPUT_PLANES, value_data,
-                                          weights_.value.bn_means.data(),
-                                          weights_.value.bn_stddivs.data());
+    Convolution<1>::Forward(batch_size, output_channels, num_value_input_planes,
```
dubslow: The template argument is *always* `1`, correct?

dubslow:
Reading the code, do I understand that this batching is only effective for MKL, not OpenBLAS or Accelerate?

frpays:
@dubslow I answer only to your 2). for the moment.

We need pointer, and there is a big reason for it.

As we do batching, we need to apply the same calculation in the same buffer at different offset. We need to allocate std::vector v(batch_size*size). and then call foo(v&[0]) and then foo(V&[ 0 * size]) ,    foo(V&[ 1 * size]),  foo(V&[ 2 * size])… Vector works with foo(). But already if foo is foo(const float*), and you realise that in foo() calls bar(…) the signature of bar will not be vectors. You could get away, with temporary vector all over the place, but using pointer is the right way to go.

So we use pointer because:

- As we do batching:

  1. the array length of the argument, *has no meaning*. You *always* needs to be given the size of the job, as parameters. You should *never* rely on the size.
  2. we need to call methods with methods with offset:  &array[+offset]

Here you can see that vectors are not fit for the job since:

  1. vectors convey a length parameter, pointers don’t
  2. pointers convey an implicit offset, vectors don’t

- this is very low level almost c-like routine, shielded from the rest of the program,

- all BLAS routines accept float* and we will need to do a lot of pointer arithmetics.

- Pointer are not bad. It’s not “archaic” it depends on where and why you utilise them.

frpays:
@dubslow , here I address preliminary the rest of your points (in the sense that I tell you my understanding of the problem, without actually doing something for the moment).

1. You are right. 

I would like to do that, but without **coupling*** the classes or methods that have nothing in common. For the moment I think that redefining `width=8, height=8` in every class, is better that coupling the classes. In definitive, all it says is: "This class is doing its job for a 8x8 board".

For the winograd constants, you have a point, and I would definitely want to do that in a clean way.

"blas.h" is absolutely and definitely related to BLAS, the Basic Algebra Linear Subroutines. It's here to abstract the different vendors. We can definitely envisage a "blas.c", in the future, to complete the missing batched sgemm routines that are not yet supported by openBlas or MKL.

frpays:
@dubslow , here I address preliminary your point (without actually doing something for the moment).

3. Ok let's get rid of it. 

But in some sense, that's very sad, because you are killing an general purpose batched convolve nxn routines that we can need in the future.  

But okay.

src/neural/blas/batchnorm.cc
```diff
@@ -0,0 +1,68 @@
...
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU General Public License for more details.
+
+ You should have received a copy of the GNU General Public License
+ along with Leela Chess.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "batchnorm.h"
+
+#include "cmath"
```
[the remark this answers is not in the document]
frpays: Yes. Will be addressed.

src/neural/blas/batchnorm.h
```diff
@@ -0,0 +1,45 @@
...
+#include <vector>
+#include <cstddef>
+
+namespace lczero {
+
+class Batchnorm {
+ public:
+  static void Apply(const int batch_size, const int channels, float* data,
+                    const float* means, const float* stddivs,
+                    const float* eltwise = nullptr);
```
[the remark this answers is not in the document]
frpays: Pointers are a very assumed choice here and we need it. See my lengthy explanation.

src/neural/blas/batchnorm.h
```diff
@@ -0,0 +1,45 @@
...
+
+  static void OffsetMeans(std::vector<float>& bn_means,
+                          const std::vector<float>& biases);
+
+  static void InvertStddev(std::vector<float>& weights);
+
+ private:
+  static constexpr float kEpsilon = 1e-5f;
```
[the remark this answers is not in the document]
frpays: Yes, I would like too. I some point I wanted to move them in methods of Weights. But I guess there are ok-ish there.

src/neural/blas/batchnorm.cc
```diff
@@ -0,0 +1,68 @@
...
+        // BN + residual add
+        auto arr = &data[c * kSquares];
+        auto res = &eltwise[c * kSquares];
+        for (int b = 0; b < kSquares; b++) {
+          float val = res[b] + (scale_stddiv * (arr[b] - mean));
+          arr[b] = val > 0 ? val : 0;
+        }
+      }
+    }
+    data += channels * kSquares;
```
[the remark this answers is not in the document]
frpays: Here again, we have no choice. All the point in batching is to repeat some computations in the vector. Hopefully, all of methods do batching, but at some point, some method did not accept it, and the batching was external:

``
// non batched computation
for (int i=0; i<batch_size; i++)
    do_some_computation(vect1+length1, vect2+length2);
// batched computation
do_some_batched_computation(batch_size, vect1, vect2);
``

src/neural/blas/batchnorm.cc
```diff
@@ -0,0 +1,68 @@
...
+    if (eltwise != nullptr) eltwise += channels * kSquares;
+  }
+}
+
+void Batchnorm::OffsetMeans(std::vector<float>& bn_means,
+                            const std::vector<float>& biases) {
+  // Biases are not calculated and are typically zero but some networks might
+  // still have non-zero biases.
```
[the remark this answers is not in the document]
frpays: Yes, we need to get rid of it, good catch.

src/neural/blas/convolution.h
```diff
@@ -0,0 +1,39 @@
...
+#include <vector>
+
+namespace lczero {
+
+template <unsigned int filter_size>
+class Convolution {
+ public:
+  static void Forward(const int batch_size, const int input_channels,
+                      const int output_channels, const float* input,
+                      const float* weights, const float* biases, float* output);
```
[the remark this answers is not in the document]
frpays: Please Let's have this conversation once for all.

src/neural/blas/convolution.cc
```diff
@@ -0,0 +1,161 @@
...
+ along with Leela Chess.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "neural/blas/convolution.h"
+#include "neural/blas/blas.h"
+
+namespace lczero {
+
+template <unsigned int filter_size>
+
```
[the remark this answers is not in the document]
frpays: Every blank line or things like that will be handled by clang-format at the end.

src/neural/blas/fully_connected_layer.cc
```diff
@@ -0,0 +1,121 @@
...
+                output_size, batch_size, input_size, 1.0f,
+                // A     lda
+                weights, input_size,
+                // B    ldb   beta,
+                inputs, input_size, 0.0f,
+                // C   ldc
+                outputs, output_size);
+  }
+  for (int i = 0; i < batch_size; i++) {
+    if (apply_relu) {
```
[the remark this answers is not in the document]
frpays: It does not make sense to leave it in the inner loop. I would say, as long as it's in one outer loop that you know is large enough, it's ok.

But ok. I will address it.

src/neural/blas/fully_connected_layer.h
```diff
@@ -0,0 +1,46 @@
...
+class FullyConnected {
+ public:
+  // From input_size to output_size (batched)
+  static void Forward(const int batch_size, const int input_size,
+                      const int output_size, const float* input,
+                      const float* weights, const float* biases,
+                      bool apply_relu, float* output);
+
+  // No batched, from input_size to scalar
+  static float ToScalar(const int input_size, const float* input,
```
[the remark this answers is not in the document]
frpays: It definitely belongs in FullyConnectedLayer since it's a fully-connected layer computation (all-values to one value).

But I would like a better name. Suggest.

src/neural/network_blas.cc
```diff
@@ -17,270 +17,314 @@
...
+  void EncodePlanes(const InputPlanes& sample, float* buffer);
 
-    std::vector<float> policy_data(weights_.ip_pol_b.size());
-    forward(input_data_, policy_data, value_data_);
+  static constexpr int kWidth = 8;
+  static constexpr int kHeight = 8;
+  static constexpr int kSquares = kWidth * kHeight;
+  static constexpr int kTiles = kSquares / 4;
+  static constexpr int kWinogradAlpha = 4;
+  static constexpr int kWinogradTile = kWinogradAlpha * kWinogradAlpha;
```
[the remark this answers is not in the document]
frpays: Yes. I would like them in place too. But not in the global scope. 
Be aware that this is values for a 8x8 board, a 8x8 board *only*.

src/neural/network_blas.cc
```diff
@@ -17,270 +17,314 @@
...
   }
 
-  void forward(std::vector<float>& input, std::vector<float>& output_pol,
-               std::vector<float>& output_val) {
+ private:
+  // A cap on the max batch size since it's consume a lot of memory
+  static constexpr int kMaxBatchSize = 2048;
+
+  Weights weights_;
+  int max_batch_size_;
```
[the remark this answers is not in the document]
frpays: There a max batch size fixed by the user, and max-max batch size that the layer accepts for the user max batch size value. Please suggest names.

src/neural/network_blas.cc
```diff
@@ -17,270 +17,314 @@
...
+
+  // max_channels is the maximum number of input channels of any
+  // convolution.
+  // Residual blocks are identical, but the first convolution might be bigger
+  // when the network has very few filters
+  const int max_channels = std::max(output_channels, kInputPlanes);
+
+  // Determine the largest batch for allocations
+  const int plane_count = static_cast<int>(planes_.size());
+  const int largest_batch_size = std::min(max_batch_size_, plane_count);
```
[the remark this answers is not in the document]
frpays: Well, ... ok. Let's do that.

src/neural/network_blas.cc
frpays: Please leave the general discussions (pointer use and convolve<> code suppression) out of the line-by-line review, (except when it makes sense of course).

frpays:
> Reading the code, do I understand correctly that this batching is only effective for MKL, not OpenBLAS or Accelerate?

No Accelerate/VecLib is concerned too. What did you think that MKL is not concerned by the batching? (It may be a misleading comment by me).

Ah, I see. I put MacOS instead of Accelerate/VecLib in the PR header. I can change it.

The thing is there are 3 or 4 names.

1. What every end-user care: **Apple** or **MacOS**
2. What we put in the code and in meson: **Accelerate**
3. What it's really is: **VecLib**, a part of the **Accelerate** framework.

frpays:
Sunday 17,  2018.  A quick update on investigating the memory corruption.

Here are the facts:

- Linux platform, MKL and OpenBAS are not affected. You can run batched BLAS computations for hours.

- MacOS pure BLAS ( --backend="blas") is not affected.

- I made I bound-checked version a BLAS computation (using a custom T* wrapper class), and there is no out-of-bounds access.

Only the following command-line on MacOS / Accelerate crashes after sometimes more that a minute of computations.

`lc0-pr85 --backend=multiplexing "--backend-opts=a(backend=blas,threads=8),b(backend=opencl,threads=3)" -t 11 --minibatch-size=64`

So it looks very very tricky.

src/neural/blas/batchnorm.cc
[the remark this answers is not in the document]
dubslow: `std::vector.data()` gives you the underlying pointer when you need it.

src/neural/blas/fully_connected_layer.cc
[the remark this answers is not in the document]
dubslow: It just looks funny to me from a style perspective to have it "in between". If you'd rather not have it in the inner loop, then you probably shouldn't have it in any loop at all?

src/neural/blas/fully_connected_layer.h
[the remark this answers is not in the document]
dubslow: `InnerProduct` was fine, imo!

src/neural/network_blas.cc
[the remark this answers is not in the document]
dubslow: `blas.h` is perfect for these, no?

src/neural/network_blas.cc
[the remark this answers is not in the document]
dubslow: Maybe `max_batch_size_` and `kHardMaxBatchSize`? Or perhaps `kMaxBatchSize` and simply `batch_size_`? Maybe crem has a better idea

dubslow:
2) About pointers, you can always use `std::vector.data()` to get a pointer when you need it for the underlying C libs. But, IMO, you should use it *only* for translation to clib calls, and not in our C++ code. And `std::vector::iter` will allow you to do implicit offset computation as necessary, and I believe you can do `std::vector::iter.data()` as well, when necessary?

And I'm not blocking this PR! I didn't submit the review as "changes requested". Crem may decide at any time to ignore my qualms, approve, and merge.

1) That still leaves a bad taste in my mouth, to redefine the board size n times over, but I guess I understand your point of view... maybe if we had a separate header file in the blas folder that wasn't named blas.h? `blas_common.h" or something :P

4) A matter of preference it may be, but it's certainly true that all the `.size()` methods return `size_t`s, and when you do all those `static_cast`s to `int`s, it certainly *looks* like you have some coding reason to use `int`s over `size_t`s, and makes everyone else wonder why it's so important. So this should only be done if it actually *is* important, and so if it's not important, just use the `size_t`s. I'd be happy to make the patch myself if you'd prefer, it's certainly a pretty boring patch to create.

frpays:
2. I  would like you to understand why pointers are better fitted for the job *here*.

The problem is simply that passing a vector is meaningless when all you can do is take the pointer out of it, when vector.size() has no meaning. Take the ToScalar (that will be renamed DoProduct) method:

Here is the implementation in float*

```float FullyConnected::ToScalar(const int size, const float* x, const float* y) {
  return cblas_sdot(size, x, 1, y, 1);
}```

Here is the implementation in vector

```float FullyConnected::ToScalar(const int size, const vector& x, const &vector y) {
  // return cblas_sdot(x.size(), x, 1, y, 1); wrong !!
  // return cblas_sdot(y.size(), x, 1, y, 1); wrong !!
    return cblas_sdot(size, x.data(), 1, y.data(), 1); 
}```

The first reason why we choose here the const float* signature, is to emphasise on the fact that you vector size cannot be trusted.

The second reason, is that sometimes, there is simply no choice.

Look at the ToScalar caller. We have a very long vector of size batch_size * num_value_channel and we want to to dotproduct from [0...num_value_channel-1] then [num_value_channel... num_value_channel-1] etc...

Look at the actual code:
 ``` 
    for (int j = 0; j < batch_size; j++) {
      std::vector<float> policy(num_output_policy);

      // Get the moves
      FullyConnected::Softmax(
          num_output_policy, &output_pol[j * num_output_policy], policy.data());

      policies_.emplace_back(std::move(policy));

      // Now get the score
      double winrate = FullyConnected::ToScalar(
                           num_value_channels, weights_.ip2_val_w.data(),
                           &output_val[j * num_value_channels]) +
                       weights_.ip2_val_b[0];

      q_values_.emplace_back(std::tanh(winrate));
    }
``` 

How do change the code so that you can have ToScalar accept vector from such signature ? 

```
 // We already know that input.size() and weights.size() are meaningless.
 // but we cannot even call this signature!
//
  static float ToScalar(const int input_size, const std::vector<float>& input, 
                       const std::vector<float>& weights);

```

You cannot.

There is no possibiltiy to create from previously allocated memory at a specified offset. 

```double winrate = FullyConnected::ToScalar2(
                           num_value_channels, weights_.ip2_val_w,
                           output_val[j * num_value_channels]  // error we need a vector ?!?
                           ) +                       weights_.ip2_val_b[0];```

**Comment by dubslow** [comment]
1) If the size of the vector is wrong, then the code is wrong (and I suspect there's a better way to manage memory than to allocate for the maximum batch size and softcap it at actual batchsize... could be wrong,  but I'll look into it)

2) I already said, implicit offsets can be done just fine with `std::vector::iterator`

**Comment by frpays** [comment]
1. constants

-  Iwould like first you see that's this PR is a big improvement over the current master branch. In the master, width, height, tiles, are redefined in every function WinogradTransformIn, WinogradTransformOut, etc...

Now there are only defined once in each class.

- I think it's very ok to redefine width=8, height=8 in the 3 classes. "It's really a statement: this classes are convolution on 8x8 board.". It's not a redefinition, it's a statement. You cannot change this values, all the code is based on it.


- The Winograd constants are a slightly different matters. But they are only defined twice, once in WinogradConvolution3.h and network_blas.cc. And the good news is, it's no more needs in network_blas.cc. The reason, is it was needed for U/V allocation sizing. I wanted the allocation to be encapsulated with the actual use. So now U/V are allocated in the WinogradConvolution3() constructor. (That was all the point).
 
So your point will be addressed on the Winograd constants. But not on the width/heights. See with=8 and height=8 as a statement "this class is for a 8x8 board".

**Comment by frpays** [comment]
2. size_t.

Ok I will change the int in size_t.

**Comment by dubslow** [comment]
Not so sure I agree about the width/height constants, but I at least understand it, and it is acceptable. Perhaps adding a comment of the nature of "this class works on 8x8 boards independently of everything else" will help other readers understand why they're separate.

**Comment by frpays** [comment]
> If the size of the vector is wrong, then the code is wrong (and I suspect there's a better way to manage memory than to allocate for the maximum batch size and softcap it at actual batchsize... could be wrong, but I'll look into it)

No. the size of the vector is longer that the actual size of the computation, because we are doing batch computation. Please read closer.


> I already said, implicit offsets can be done just fine with std::vector::iterator

You really want the signature to be:
```
  static float ToScalar2(const int input_size, const std::vector<float>::iterator x,
                         const std::vector<float>::iterator weights);
````

Please, you are off limit there.

**Comment by frpays** [comment]
> Not so sure I agree about the width/height constants, but I at least understand it, and it is acceptable. Perhaps adding a comment of the nature of "this class works on 8x8 boards independently of everything else" will help other readers understand why they're separate.

Sure. Thanks for agreeing.

**Comment by frpays** [comment]
> I guess an inner product in some sense acts on all its inputs, but to my mind, "fully connected layer" implies as many outputs as inputs, or at least a full 2D matrix of outputs, unlike a simple 1D-1D inner product. Could we put InnerProduct, Softmax, and kWidth/kHeight into a seperate utils.h or something inside the blas folder? (Since you prefer to not have them in blas.h, and I understand that point if not necessarily share that opinion)

No, fully connected layer means there is weight for every input. 

So the two current method Forward and ToScalar are actually methods of a fully connected methods.
It's really the same method indeed even if it does not look like.

Forward: 
- use a batch_size
- is used when the output layer is of size>1
- apply biases
- apply relu

ToScalar:
- does not use a batch_size
- is used when the output layer is of size==1
- does not apply relu

So My proposal is the followinng:

- Rename Forward and ToScalar back into InnerProduct and DoProduct.
- Move Softmax into Activation.

**Comment by dubslow** [comment]
To me, the word "layer" implies a sense of 2-dimensionality, hence why I don't think a single inner product counts as FCL, because it doesn't count as the L in FCL. Perhaps they can be `MatrixProduct` and `VectorProduct`? (I'm beginning to lean away from the "inner product" term, because it implies certain mathematical properties that aren't really meaningful for neural networks.)

**Comment by frpays** [comment]
> About the libs, the call to sgemm_batched or whatever it was, was only under an #ifdef USE_MKL, so that's why I asked if it's MKL only. Does Accelerate/VecLib #define USE_MKL??

OK, understood.. I will explain that.

The batching of computation is that we can regroup the 16*batch_size sgemm (matrix-matrix multiplications) into only 16 sgemm.  This is what gives us a 2x speed up.

There is not a single point in the code, where we do that. This is done all over TransformIn and TransformOut. We changed the shape of U/V, we inverted dimension 1&2 and 3&4 so that the U/V tile matrices are joinable.

But at the end, we have 16 sgemm to do.

VecLib and OpenBLAS will make them in sequence. But MKL offer us the possibility to make these multiplication in one call. There is nothing more MKL can do that computing the SGEMM in sequence since the matrices are all different. But we can expect a very very very tiny gain. So why not. performance But this really give almost any gain. I did not measure, but if you change #ifdef MKL  in   #ifdef NODEFINED you will no drop much in performance, if any.

Actually, we could remove the while #ifdef MKL completely as I suppose it does not bring anything.

**Comment by frpays** [comment]
> To me, the word "layer" implies a sense of 2-dimensionality, hence why I don't think a single inner product counts as FCL, because it doesn't count as the L in FCL. Perhaps they can be MatrixProduct and VectorProduct? (I'm beginning to lean away from the "inner product" term, because it implies certain mathematical properties that aren't really meaningful for neural networks.)

These are one dimensional layers.

In the policy/value head, all the dimensions are flattened after the convolution<1>, and come fully connected layers. There are one dimensional layers (vector).

But since we have a batch, this count as another dimension. (column of vectors). So when you go from a fully-connected-layer to another layer, this should be Matrix-vector products. But since you have a batch, this is a matrix-matrix product.

I hope I am clear.

**Comment by frpays** [comment]
> To me, the word "layer" implies a sense of 2-dimensionality, hence why I don't think a single inner product counts as FCL, because it doesn't count as the L in FCL. Perhaps they can be MatrixProduct and VectorProduct? (I'm beginning to lean away from the "inner product" term, because it implies certain mathematical properties that aren't really meaningful for neural networks.)

You are right InnerProduct is a deep mathematical term (but you looked so confident).

InnerProduct: does a multiplication matrix-vector (matrix-matrix with batch_size>1) followed by adding a bias for every output with rely.
ToScalar(): does a vector/vector doProduct.

I don't know how to rename InnerProduct, but ToScalar can be renamed DoProduct.

InnerProduct is really: SGMMAddBiasRelu....

File: src/neural/blas/fully_connected_layer.h (hunk 10 above)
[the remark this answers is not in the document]
Comment (frpays) [reply src/neural/blas/fully_connected_layer.h]: But's it's really a Matrix-Matrix multiplication (which is not an inner-product mathematically) followed by adding the bias, plus a Relu.

As I recall (and without wikipedia please) mathematically  an inner product is when you have xT M x, with M positive definite. This is an inner product.

File: src/neural/network_blas.cc (hunk 11 above)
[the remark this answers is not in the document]
Comment (frpays) [reply src/neural/network_blas.cc]: don't touch this.

File: src/neural/network_blas.cc (hunk 12 above)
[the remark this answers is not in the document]
Comment (frpays) [reply src/neural/network_blas.cc]: That's gonna be kSomething because I will be kCrem compliant.

``kHardMaxBatchSize``  would be my pick if anything else.

**Comment by frpays** [comment]
Sunday 17, 2018. A another quick update on investigating the memory corruption.

Linux/OpenBLAS is confirmed to be unaffected by the crash.

 I run successfully for 6 hours:

 ``./lc0 --backend="blas" --backend-opts="batch_size=256,blas_cores=4" -t 4```


``info depth 3 seldepth 32 time 18951411 nodes 2483814 score cp 19 hashfull 1000 nps 131 pv c2c4 g8f6 d2d4 e7e6 b1c3 d7d5 c4d5 e6d5 c1g5 c7c6 d1c2 c8e6 e2e3 h7h6 g5h4 f8d6 f1d3 e8g8 g1f3 b8d7 e1g1 f8e8 h4g3 d6g3 h2g3```

And:

``./lc0 --backend=multiplexing "--backend-opts=a(backend=blas,threads=8),b(backend=opencl,threads=3)" -t 11 --minibatch-size=64```

``
info depth 3 seldepth 39 time 19090649 nodes 11890453 score cp 20 hashfull 1000 nps 622 pv c2c4 g8f6 d2d4 c7c6 g1f3 d7d5 b1c3 e7e6 e2e3 b8d7 d1c2 f8e7 f1d3 b7b6 e1g1 c8b7 f1d1 e8g8 e3e4 d5e4 c3e4 d8c7 e4f6 d7f6 c4c5 f8d8 b2b4 a7a5 b4a5 b6c5 
``


Whatever it is, it is limited to either MacOS or VecLib.

**Comment by dubslow** [comment]
Ah I see. I think that's worthy of a comment in the code :) Can't really assume that people reading the code will understand the details of the matrices involved -- I certainly don't!

And I'm fine for leaving it as is, because it leaves open the possibility for MKL to introduce further optimizations (maybe, one day, it can do all 16 sgemms in parallel with SIMD instructions!). And if VecLib/OpenBLAS introduce a batched lib call, then we can just rearrange the defines.

**Comment by frpays** [comment]
We found a first occurance of the same kind of crash on the master.

> Here's my crash with the latest master, run with ./lc0 weights.txt --backend=multiplexing '--backend-opts=a(backend=blas,threads=4),b(backend=opencl,threads=3)' -t 7 --minibatch-size=64. This is not using the PR. https://ghostbin.com/paste/5vxbg

The most probable hypothesis is a bug/concurrency problem is the master branch. PR87 is making the problem more frequent.

**Comment by frpays** [comment]
Here is a batch commits:

- The winograd constants are now declared only once
- Convolve<n> and Im2Col are gone
- I used size_t everywhere it's possible. Careful that it's possible to go fatally in bridge to far in TransformIn Transform Out where there  is critical `int` arithmetics going on. (size_t minus size_t is size_t).
 - Ichanged constant name kMaxBatchSize to kHardMaxBatchSize
- Fixed the FullyConnectedLayer name, and changed the forward methods into Forward1D and Forward0D.
- I merged upstream to resolve the conflict introduced by PR103. I also fixed the corresponding spell errors in the merge.
- I reapplied `clang-format -i -style=google`

**Comment by borg323** [comment]
Some nps results on a i5-4590 CPU @ 3.30GHz with 4 cores.
* net337

|                        | PR87 |   | master |   |
|------------------------|----------:|-----------:|----------:|-----------:|
|                        | mkl |  openblas | mkl |  openblas |
| blas_cores 1 threads 4 | 304      | 265       | 203      | 138       |
| blas_cores 2 threads 2 | 228      | 192       | 128      | 94        |
| blas_cores 4 threads 1 | 146      | 120       | 80       | 52        |
| blas_cores 4 threads 4 | 130      | 67        | 67       | 3         |


* test1 net27

|                        | PR87 |   | master |   |
|------------------------|----------:|-----------:|----------:|-----------:|
|                        | mkl |  openblas | mkl |  openblas |
| blas_cores 1 threads 4 | 1205     | 1234      | 935      | 725       |
| blas_cores 2 threads 2 | 1196     | 891      | 685      | 537       |
| blas_cores 4 threads 1 | 877      | 671       | 397      | 328       |
| blas_cores 4 threads 4 | 424      | 134       | 238      | 389       |

**Comment by frpays** [comment]
Here is a new benchmarks suite, that goes along all dimensions (blas_core, search threads and batch_size).

https://docs.google.com/spreadsheets/d/1eU2gb67V69_7yXFzIaF2WI4djF1pXGAih-rwhrstmg0


Platform | Threads | master | PR87 (512 batch size) | Speed-up
-- | -- | -- | -- | --
MacOS/vecLib | 1 | 43 | 114 | 2.65
MacOS/vecLib | 2 | 58 | 172 | 2.97
MacOS/vecLib | 4 | 89 | 222 | 2.49
MacOS/vecLib | 6 | 100 | 240 | 2.40
MacOS/vecLib | 8 | 98 | 239 | 2.44
MacOS/vecLib | 10 | 110 | 239 | 2.17
Linux/MKL 1-core | 1 | 24 | 38 | 1.58
Linux/MKL 1-core | 2 | 47 | 74 | 1.57
Linux/MKL 1-core | 4 | 89 | 134 | 1.51
Linux/MKL 1-core | 6 | 127 | 192 | 1.51
Linux/MKL 1-core | 8 | 161 | 245 | 1.52
Linux/MKL 1-core | 10 | 192 | 285 | 1.48
Linux/MKL 2-core | 1 | 33 | 60 | 1.82
Linux/MKL 2-core | 2 | 72 | 115 | 1.60
Linux/MKL 2-core | 4 | 132 | 213 | 1.61
Linux/MKL 2-core | 6 | 160 | 275 | 1.72
Linux/MKL 2-core | 8 | 98 | 250 | 2.55
Linux/MKL 2-core | 10 | 52 | 250 | 4.81
Linux/OpenBLAS 1-core | 1 | 24 | 39 | 1.63
Linux/OpenBLAS 1-core | 2 | 48 | 76 | 1.58
Linux/OpenBLAS 1-core | 4 | 89 | 140 | 1.57
Linux/OpenBLAS 1-core | 6 | 127 | 202 | 1.59
Linux/OpenBLAS 1-core | 8 | 160 | 257 | 1.61
Linux/OpenBLAS 1-core | 10 | 199 | 289 | 1.45
Linux/OpenBLAS 2-core | 1 | 29 | 61 | 2.10
Linux/OpenBLAS 2-core | 2 | 59 | 114 | 1.93
Linux/OpenBLAS 2-core | 4 | 102 | 207 | 2.03
Linux/OpenBLAS 2-core | 6 | 111 | 276 | 2.49
Linux/OpenBLAS 2-core | 8 | 100 | 256 | 2.56
Linux/OpenBLAS 2-core | 10 | 97 | 231 | 2.38

**Comment by frpays** [comment]
What the benchmarks show is:

- We should use a high search thread count to get the largest BLAS throughput.
- We should stick to blas_cores=1, we can get 290 nps/core is sufficient search_threads.

File: meson.build [hunk 14]
```diff
@@ -163,7 +163,10 @@ if get_option('build_backends')
   if get_option('blas') and has_blas
 
     blas_files = [
-    'src/neural/blas/transforms.cc',
+    'src/neural/blas/winograd_convolution3.cc',
```
mooskagh: Let's sort the files in a list. :)

src/neural/blas/batchnorm.cc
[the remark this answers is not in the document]
mooskagh: I agree that if it's possible to make vectors, then we should use vectors.
Alternatively, we can have something like std::dynarray (never got into standard, so we have to implement that outselves) in utils/ and use it as a wrapper

src/neural/blas/batchnorm.cc
```diff
@@ -0,0 +1,64 @@
...
+          arr[b] = val > 0 ? val : 0;
+        }
+      }
+    }
+    data += channels * kSquares;
+    if (eltwise != nullptr) eltwise += channels * kSquares;
+  }
+}
+
+void Batchnorm::OffsetMeans(std::vector<float>& bn_means,
```
mooskagh: Output parameters should be pointers, not references. And be the last in the list of arguments.

src/neural/blas/batchnorm.cc
```diff
@@ -0,0 +1,64 @@
...
+                            const std::vector<float>& biases) {
+  for (size_t i = 0; i < bn_means.size(); i++) bn_means[i] -= biases[i];
+}
+
+void Batchnorm::InvertStddev(std::vector<float>& weights) {
```
mooskagh: Pointers, not references.

src/neural/blas/batchnorm.h
```diff
@@ -0,0 +1,45 @@
...
+#pragma once
+
+#include <cstddef>
+#include <vector>
+
+namespace lczero {
+
+class Batchnorm {
+ public:
+  static void Apply(const size_t batch_size, const size_t channels, float* data,
```
mooskagh: Public functions should have a comment.

src/neural/blas/batchnorm.h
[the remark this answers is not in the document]
mooskagh: There are many things like this already (like policy head size). Maybe have `network-common.h` ?

src/neural/blas/convolution1.cc
```diff
@@ -0,0 +1,71 @@
...
+ Leela Chess is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU General Public License for more details.
+
+ You should have received a copy of the GNU General Public License
+ along with Leela Chess.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "neural/blas/convolution1.h"
```
mooskagh: sorted includes. :)
But clang-format would do that for you.

src/neural/blas/convolution1.cc
```diff
@@ -0,0 +1,71 @@
...
+    // lda The size of the first dimension of matrix A; if you are
+    // passing a matrix A[m][n], the value should be m.
+    //    cblas_sgemm(CblasRowMajor, TransA, TransB, M, N, K, alpha, A, lda, B,
+    //                ldb, beta, C, N);
+
+    //             C                               A                     B
+    //
+    //           outputs             :=          weights        x      input
+    //
+    //   cols:  kSquares (N)                 input_channels (K)      kSquares
```
mooskagh: Comment lost formatting due to line break. Also consider reducing whitespace in comments.

Also in other files.

src/neural/blas/convolution1.cc
```diff
@@ -0,0 +1,71 @@
...
+    for (size_t o = 0; o < output_channels; o++) {
+      const auto bias = biases[o];
+      for (size_t b = 0; b < kSquares; b++) {
+        batch_output[index++] += bias;
+      }
+    }
+  }
+}
+
+}  // namespace lczerp
```
mooskagh: clang-format would fix that mistype :)

src/neural/blas/convolution1.cc
```diff
@@ -0,0 +1,71 @@
...
+    //   cols:  kSquares (N)                 input_channels (K)      kSquares
+    //   (N)
+    //
+    //   rows:  output_channels (M)         output_channels (M)
+    //   input_channels (K)
+
+    const float* batch_input = input + i * kSquares * input_channels;
+    float* batch_output = output + i * kSquares * output_channels;
+
+    cblas_sgemm(CblasRowMajor, CblasNoTrans, CblasNoTrans,
```
mooskagh: I have no idea what's going on. :) That's fine though.
Consider having one parameter per line, and comment to the right of it, rather then above.

The same in other places.

src/neural/blas/convolution1.h
```diff
@@ -0,0 +1,37 @@
...
+#pragma once
+
+#include <cstddef>
+#include <vector>
+
+namespace lczero {
+
+class Convolution1 {
+ public:
+  static void Forward(const size_t batch_size, const size_t input_channels,
```
mooskagh: Comment to a public function. Also comment to a class would be useful (why it is Convolution1 and not simply Convolution?)

src/neural/blas/fully_connected_layer.cc
```diff
@@ -0,0 +1,120 @@
...
+                                  float* output) {
+  auto alpha = *std::max_element(input, input + size);
+  
+  auto denom = 0.0f;
+  for (size_t i = 0; i < size; i++) {
+    auto val = std::exp(input[i] - alpha);
+    output[i] = val;
+    denom += val;
+  }
+  for (size_t i = 0; i < size; i++) {
```
mooskagh: If it all was dynarrays or vectors, we could use nice range-based loops. :)

src/neural/blas/winograd_convolution3.cc
```diff
@@ -0,0 +1,337 @@
...
+        }
+      }
+    }
+  }
+}
+
+void WinogradConvolution3::Sgemm(const size_t batch_size, const float* weights,
+                                 const size_t input_channels,
+                                 const size_t output_channels) {
+#ifdef USE_MKL
```
mooskagh: Why is there ifdef? Is there a conditional compilation?

src/neural/blas/winograd_convolution3.cc
```diff
@@ -0,0 +1,337 @@
...
+
+  /*
+   void cblas_sgemm_batch (const CBLAS_LAYOUT Layout, const CBLAS_TRANSPOSE*
```
mooskagh: What is this comment?

src/neural/blas/batchnorm.cc
dubslow: I still don't understand that style guidance. The whole point of references was to be able to make inplace modifications of this nature without the risk of munging the address caller-side. (The easier syntax doesn't hurt either.)

src/neural/blas/batchnorm.cc
frpays: Will be addressed. 

But don't you want the Weights structure to host InvertStddev and OffsetMeans?

src/neural/blas/batchnorm.h
frpays: Comments on every public functions. OK.

src/neural/blas/convolution1.cc
frpays: Yes.

But I cannot replicate the ascii graphics above and else where.
I helped me to figure ou what the sgemm what doing.
On comment a line will not help you/me.  You will have comments like K, M, N, ld(A), A.
I will not help you understand what algebraic operation is going on.

src/neural/blas/winograd_convolution3.cc
frpays: There is a conditional compilation going on, still MKL offers an API that we can use to accelerate the computations even further. But we can remove the USE_MKL block and make MKL behaves the same way as vecLib and OpenBLAS.

I made use of the USE_MKL, and USE_OPENBLAS, they are defines that were defined in lczero to be able to include the correct includes, because there are slight differences and mostly because of specific functions for every vendor. For example, the API for choosing the number of threads is not defined is vecLib, has a given signature in openblas and another on in MKL .

frpays: Look at blas.h in neural/blas for reference.

src/neural/blas/winograd_convolution3.cc
frpays: This comment is the commentated signature of the batched sgemm api in MKL.

You won't find this signature on your platform unless you installed MKL. We can get rid of this comment, but  anyone who will have to maintain this method will have to read it at some point.

src/neural/blas/convolution1.h
frpays: 1 is the spacial parameter. It's a 1x1 convolution.  I will make this comment:

```
// Convolution 1x1
class Convolution1 {
  public:
  static void Forward(const size_t batch_size, const size_t input_channels,
```

src/neural/blas/batchnorm.cc
frpays: OffsetMeans and InvertStddev are special cases that I did not touch.

No we cannot use vectors as the length parameter would be meaningless.

src/neural/blas/convolution1.cc
frpays: I used clang-format a few times on the files.

src/neural/blas/convolution1.cc
frpays: Yes, to prevent clang to do that, I have to shorter the ascii art.

src/neural/blas/convolution1.cc
frpays: > There are many things like this already (like policy head size). Maybe have network-common.h ?
Actually no. There are not many things like this. We figure network parameters using weights vector sizes. Maybe the number of input planes is the exception? Even the number of policies is computed:

```
const auto num_output_policy = weights_.ip_pol_b.size();
``

Comment (frpays) [reply src/neural/blas/convolution1.cc]: > sorted includes. :)
> But clang-format would do that for you.

I will pass clang-format on the final result.

**Comment by frpays** [comment]
@mooskagh  I addressed your comments and pushed back changes
Except changing the signatures of OffsetMeans/InvertMeans that would really complicate things.
Please review and consider merging.

File: src/neural/blas/batchnorm.h [hunk 26]
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+#pragma once
+
+#include <cstddef>
+#include <vector>
+
+namespace lczero {
+
+// Batch mormalization related methods
```
GBeauregard: mormalization -> normalization

src/neural/blas/batchnorm.h
```diff
@@ -0,0 +1,52 @@
...
+class Batchnorm {
+ public:
+  Batchnorm() = delete;
+
+  // Apply batch mormalization, along with a Relu and add a possible skip
```
GBeauregard: mormalization -> normalization

src/neural/blas/winograd_convolution3.h
```diff
@@ -0,0 +1,84 @@
...
+  static constexpr auto kHeight = 8;
+  static constexpr auto kSquares = kWidth * kHeight;
+
+  static constexpr auto kWtiles = (kWidth + 1) / 2;  // 4
+  static constexpr auto kTiles = kWtiles * kWtiles;  // 16
+
+  static constexpr auto kWinogradAlpha = 4;
+  static constexpr auto kWinogradTile = kWinogradAlpha * kWinogradAlpha;
+
+  std::vector<float> V_;
```
mooskagh: Why uppercase V_ and M_?

src/neural/blas/winograd_convolution3.h
```diff
@@ -0,0 +1,84 @@
...
+// https://arxiv.org/abs/1509.09308
+//
+// https://ai.intel.com/winograd/
+// https://ai.intel.com/winograd-2/
+
+// Convolution 3x3 using the Winograd algorithm
+class WinogradConvolution3 {
+ public:
+  // Create the zero-padded U matrix
+  static std::vector<float> ZeropadU(const std::vector<float>& U,
```
mooskagh: Why uppercase U?

src/neural/blas/winograd_convolution3.h
```diff
@@ -0,0 +1,84 @@
...
+                                     const size_t outputs,
+                                     const size_t channels,
+                                     const size_t outputs_pad,
+                                     const size_t channels_pad);
+
+  // Transform the weights
```
mooskagh: Sorry for nitpicking but as there is another round of changes anyway..
Comments are sentences and should end with a period. (also in fully_connected_layer.h)
Also, in general, function comments should be in 3rd person (what function does), e.g. "Transforms the weights.".

Also from the comment it's not clear how exactly it transforms it.

src/neural/blas/winograd_convolution3.h
```diff
@@ -0,0 +1,84 @@
...
+  static std::vector<float> TransformF(const std::vector<float>& f,
+                                       const size_t outputs,
+                                       const size_t channels);
+
+  // Allocate for the largest batch size
```
mooskagh: Allocate what? And why is function called "WinograteConvolution" and not "Allocate<Something>"?

src/neural/blas/winograd_convolution3.cc
```diff
@@ -0,0 +1,339 @@
...
+#include <array>
+
+namespace lczero {
+
+std::vector<float> WinogradConvolution3::ZeropadU(const std::vector<float>& U,
+                                                  const size_t outputs,
+                                                  const size_t channels,
+                                                  const size_t outputs_pad,
+                                                  const size_t channels_pad) {
+  // Fill with zeroes
```
mooskagh: Period in the end of the sentence.

src/neural/blas/fully_connected_layer.h
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+#pragma once
+
+#include <cstddef>
+#include <vector>
+
+namespace lczero {
+
+// Fully connected layer
```
mooskagh: No need for the comment if it's just the class name.

src/neural/blas/batchnorm.cc
mooskagh: @dubslow 
Many style guides (including google styleguide that we happen to use) disallow mutable references because otherwise it can lead to surprises at calling point (that's a habit from C language, where there's no references). E.g. (extreme example):
```cpp
int x = 32;
ComputeSomethingXTimes(x);  // Unexpected that x may change.
DoSomethingElseXTimes(x);  // x is 0 now.

void ComputeSomethingXTimes(int& x) {
   while (x--) ComputeSomething();
}
```

src/neural/blas/winograd_convolution3.h
frpays: matrices often get a single char uppercase name in code dealing with algebra. The original code used uppercase, and it was helping me to read the code.
As a maintener, I would vote to keep the uppercase.

src/neural/blas/winograd_convolution3.h
frpays: Same answer.

src/neural/blas/winograd_convolution3.h
frpays: It's a class that does the forward inference of a convolution 3x3 layer.
The matrices M and U are memory resources that are allocated once for all for the duration of the computation. The batch size of the computation may be larger than the actual largest internal batch.  M and U are allocated for the largest batch size.

So the class name is WinogradConvolution3. But the comment can be enhanced.

src/neural/blas/winograd_convolution3.h
frpays: It's very common, through algebra software.
Altenatively we can use matrixU_, but U_ is fine.

src/neural/blas/winograd_convolution3.h
frpays: Sorry I cannot find anything better than "[building the] filter Transform" in the literature.
https://ai.intel.com/winograd-2/

src/neural/blas/winograd_convolution3.h
frpays: I settled for 
```
  // The instance will allocate memory resources for the
  // largest batch size, and the largest input and output
  // layers.
```

src/neural/network_blas.cc
```diff
@@ -17,270 +17,321 @@
...
-                     weights_.ip2_val_b[0];
-    q_value_.emplace_back(std::tanh(winrate));
+  std::unique_ptr<NetworkComputation> NewComputation() override {
+    return std::make_unique<BlasComputation>(weights_, max_batch_size_);
   }
 
-  void forward(std::vector<float>& input, std::vector<float>& output_pol,
-               std::vector<float>& output_val) {
+ private:
+  // A cap on the max batch size since it's consume a lot of memory
```
mooskagh: "it consumes", period at the end.

src/neural/blas/convolution1.h
```diff
@@ -0,0 +1,41 @@
...
+#include <vector>
+
+namespace lczero {
+
+// Convolution 1x1
+class Convolution1 {
+ public:
+  Convolution1() = delete;
+
+  // Batched forward inference
```
mooskagh: Period.

src/neural/blas/winograd_convolution3.h
```diff
@@ -0,0 +1,86 @@
...
+
+  // Create the filter transform matrix.
+  static std::vector<float> TransformF(const std::vector<float>& f,
+                                       const size_t outputs,
+                                       const size_t channels);
+
+  // The instance will allocate memory resources for the
+  // largest batch size, and the largest input and output
+  // layers.
+  WinogradConvolution3(const size_t max_batch_size,
```
mooskagh: I've just realized it's a constructor!
Could you move that declaration to be the first in the class?

src/neural/blas/batchnorm.cc
mooskagh: Still not fixed.

src/neural/blas/batchnorm.cc
mooskagh: Hm... The Weights from network.h? No, probably not.
But something added into utils/ is possible.

frpays:
@mooskagh I did not see you comment about Weights. So I moved the methods there.
It really helps, the network_blas and network_opencl really benefit from that.
It's now so clear and simple:

```

    conv1.OffsetMeans();
    conv2.OffsetMeans();

    conv1.InvertStddev();
    conv2.InvertStddev();
````

Please review.

**Comment by frpays** [comment]
**Please consider the newly gained clarity that I would like to keep.**


Previously the code was: 
```
      Transforms::OffsetBatchNormMeans(conv1.bn_means, conv1.biases);
      Transforms::OffsetBatchNormMeans(conv2.bn_means, conv2.biases);

      Transforms::InvertBatchNormStddev(conv1.bn_stddivs);
      Transforms::InvertBatchNormStddev(conv2.bn_stddivs);
    }

    Transforms::OffsetBatchNormMeans(weights_.policy.bn_means,
                                     weights_.policy.biases);
    Transforms::InvertBatchNormStddev(weights_.policy.bn_stddivs);

    Transforms::OffsetBatchNormMeans(weights_.value.bn_means,
                                     weights_.value.biases);
    Transforms::InvertBatchNormStddev(weights_.value.bn_stddivs);

```

or

```
      std::vector<float> batchnorm_means_1 = conv1.bn_means;  // copy ctor
      Transforms::OffsetBatchNormMeans(batchnorm_means_1, conv1.biases);

      std::vector<float> batchnorm_means_2 = conv2.bn_means;  // copy ctor
      Transforms::OffsetBatchNormMeans(batchnorm_means_2, conv2.biases);

      std::vector<float> batchnorm_stddivs_1 = conv1.bn_stddivs;  // copy ctor
      Transforms::InvertBatchNormStddev(batchnorm_stddivs_1);

      std::vector<float> batchnorm_stddivs_2 = conv2.bn_stddivs;  // copy ctor
      Transforms::InvertBatchNormStddev(batchnorm_stddivs_2);
```

**Comment by frpays** [comment]
@mooskagh The PR is ok to merge in this state.

But a question. I have seen this in loader.cc.

```
void PopulateLastIntoVector(FloatVectors* vecs, Weights::Vec* out) {
  *out = std::move(vecs->back());
  vecs->pop_back();
}

void PopulateConvBlockWeights(FloatVectors* vecs, Weights::ConvBlock* block) {
  PopulateLastIntoVector(vecs, &block->bn_stddivs);
  PopulateLastIntoVector(vecs, &block->bn_means);
  PopulateLastIntoVector(vecs, &block->biases);
  PopulateLastIntoVector(vecs, &block->weights);
}
}  // namespace

Weights LoadWeightsFromFile(const std::string& filename) {
  FloatVectors vecs = LoadFloatsFromFile(filename);
```

Does it mean I can do the same, use non-const ptr *  ?

**Comment by mooskagh** [comment]
Yes, non-const ptr is what should be used instead of non-const reference!
That way at the calling place it's clear that it may be modified.

**Comment by mooskagh** [comment]
As for methods in `Network` class, that's intended as const-only pure data class (eventually to be replaced with protobufs most likely).
As for where to put them.. `src/neural/blas/utils.{h,cc}`? E.g.
```cpp
void OffsetMeans(Network::ConvBlock* conv_block);
```

frpays:
I really dislike `util` referencing `neural`.
I'd rather put them back in `Batchnorm`.
Would you agree?

frpays:
@mooskagh Put it back in Backnorm with correct signature.

frpays:
I was tempted to fix the InvertStddev formula. I think it's not completely correct.
But it has been copied in tf and cudnn and I think if the formula has to be fixed, then it has to be fixed on the 4 backends. So I rolled back.

I have nothing to add to this PR for the moment. Please review.

Ttl:
InvertStddev formula is correct. That's how it's defined in the batch normalization paper (https://arxiv.org/abs/1502.03167) and calculated in tensorflow during the training.

frpays:
> InvertStddev formula is correct. That's how it's defined in the batch normalization paper (https://arxiv.org/abs/1502.03167) and calculated in tensorflow during the training.

Alright.  That looked suspicious, thanks for making it clear.

PR #94, opened 2018-06-19
Remove v_ from Node, take 2

dubslow:
Half for reference relative to the bug fix, half serious to get that max mem savings. As far as debug usefulness goes, is there a debug macro defined by meson that we can use to conditionally include V outputs? (By which I mean we could have a conditionally-compiled std::vector<evals> in the `Search` for printing purposes)

### DIFF
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -79,18 +79,16 @@ class Node {
   // Returns p / N, which is equal to U / (cpuct * sqrt(N[parent])) by the MCTS
   // equation. So it's really more of a "reduced U" than raw U.
   float GetU() const { return p_ / (1 + n_ + n_in_flight_); }
-  // Returns value of Value Head returned from the neural net.
-  float GetV() const { return v_; }
   // Returns value of Move probability returned from the neural net
   // (but can be changed by adding Dirichlet noise).
   float GetP() const { return p_; }
   // Returns whether the node is known to be draw/lose/win.
   bool IsTerminal() const { return is_terminal_; }
+  // Unlike GetQ, this is works when n_ == 0
+  float GetTerminalNodeValue() const { return q_; }
   uint16_t GetFullDepth() const { return full_depth_; }
   uint16_t GetMaxDepth() const { return max_depth_; }
 
-  // Sets node own value (from neural net or win/draw/lose adjudication).
-  void SetV(float val) { v_ = val; }
   // Sets move probability.
   void SetP(float val) { p_ = val; }
   // Makes the node terminal and sets it's score.
@@ -145,9 +143,6 @@ class Node {
   // Root node contains move a1a1.
   Move move_;
 
-  // Q value fetched from neural network. It's not strictly necessary to have in
-  // Node, but it's useful for debug output.
-  float v_;
   // Average value (from value head of neural network) of all visited nodes in
   // subtree. For terminal nodes, eval is stored.
   float q_;
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -84,7 +84,6 @@ class Node {
   float GetP() const { return p_; }
   // Returns whether the node is known to be draw/lose/win.
   bool IsTerminal() const { return is_terminal_; }
-  // Returns q_ unconditionally (unlike GetQ).
   float GetTerminalNodeValue() const { return q_; }
   uint16_t GetFullDepth() const { return full_depth_; }
   uint16_t GetMaxDepth() const { return max_depth_; }
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -84,7 +84,7 @@ class Node {
   float GetP() const { return p_; }
   // Returns whether the node is known to be draw/lose/win.
   bool IsTerminal() const { return is_terminal_; }
-  // Unlike GetQ, this is works when n_ == 0
+  // Returns q_ unconditionally (unlike GetQ).
   float GetTerminalNodeValue() const { return q_; }
   uint16_t GetFullDepth() const { return full_depth_; }
   uint16_t GetMaxDepth() const { return max_depth_; }
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -215,9 +215,12 @@ void Search::SendMovesStats() const {
         << std::setw(2) << node->GetNInFlight() << ") ";
 
     oss << "(V: ";
-    // Handrolled optional only has a few overloads available, simple usage only
-    optional<float> v = GetCachedFirstPlyValue(node);
-    if (!v && node->IsTerminal()) v = node->GetTerminalNodeValue();
+    optional<float> v;
+    if (node->IsTerminal()) {
+      v = node->GetTerminalNodeValue();
+    } else {
+      v = GetCachedFirstPlyValue(node);
+    }
     if (v) {
       oss << std::setw(6) << std::setprecision(2) << *v * 100;
     } else {
@@ -244,10 +247,8 @@ void Search::SendMovesStats() const {
 
 optional<float> Search::GetCachedFirstPlyValue(const Node* node) const {
   assert(node->GetParent() == root_node_);
-  // This function serves a specific purpose, fetching a NN value from a first
-  // ply move, but it would be relatively straightforward to generalize this
-  // to fetch a complete NN evaluation (policy+value) for an abitrary move.
-  // But, there isn't yet a usecase for that, so we leave this simple for now.
+  // It would be relatively straightforward to generalize this to fetch a
+  // complete NN evaluation (policy+value) for an abitrary move.
   optional<float> retval;
   PositionHistory history(played_history_); // Is it worth it to move this
   // initialization to SendMoveStats, reducing n memcpys to 1? Probably not.
@@ -499,13 +500,13 @@ void SearchWorker::ExecuteOneIteration() {
   // 4. Run NN computation.
   RunNNComputation();
 
-  // 5. Populate computed nodes with results of the NN computation.
+  // 5. Retrieve NN computations (and terminal values) into nodes.
   FetchMinibatchResults();
 
-  // 6. Update nodes.
+  // 6. Propogate a new node's information to all its parents in the tree.
   DoBackupUpdate();
 
-  // 7. Update status/counters.
+  // 7. Update the Search's status and progress information.
   UpdateCounters();
 }
 
@@ -820,21 +821,21 @@ void SearchWorker::RunNNComputation() {
   if (computation_->GetBatchSize() != 0) computation_->ComputeBlocking();
 }
 
-// 5. Populate computed nodes with results of the NN computation.
+// 5. Retrieve NN computations (and terminal values) into nodes.
 // ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 void SearchWorker::FetchMinibatchResults() {
-  // Copy NN results, or cached/duplicate results, into nodes.
-  // The latter is much simpler.
+  // Populate NN/cached results, or terminal results, into nodes.
   int idx_in_computation = 0;
   for (auto& node_to_process : nodes_to_process_) {
     Node* node = node_to_process.node;
-    if (!node_to_process.nn_queried) {
+    if (!node_to_process.nn_queried) { // Terminal nodes are quite simple!
       node_to_process.v = node->GetTerminalNodeValue();
       continue;
     }
-    // Populate V value.
+    // For NN results, we need to populate policy as well as value.
+    // First the value...
     node_to_process.v = -computation_->GetQVal(idx_in_computation);
-    // Populate P values.
+    // ...and secondly, the policy data.
     float total = 0.0;
     for (Node* n : node->Children()) {
       float p =
@@ -845,7 +846,7 @@ void SearchWorker::FetchMinibatchResults() {
       total += p;
       n->SetP(p);
     }
-    // Scale P values to add up to 1.0.
+    // Normalize P values to add up to 1.0.
     if (total > 0.0f) {
       float scale = 1.0f / total;
       for (Node* n : node->Children()) n->SetP(n->GetP() * scale);
@@ -858,7 +859,7 @@ void SearchWorker::FetchMinibatchResults() {
   }
 }
 
-// 6. Update nodes.
+// 6. Propogate a new node's information to all its parents in the tree.
 // ~~~~~~~~~~~~~~
 void SearchWorker::DoBackupUpdate() {
   // Update nodes.
@@ -908,7 +909,7 @@ void SearchWorker::DoBackupUpdate() {
   }
 }
 
-// 7. UpdateCounters()
+// 7. Update the Search's status and progress information.
 //~~~~~~~~~~~~~~~~~~~~
 void SearchWorker::UpdateCounters() {
   search_->UpdateRemainingMoves();  // Updates smart pruning counters.
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -214,20 +214,6 @@ void Search::SendMovesStats() const {
     oss << " N: " << std::right << std::setw(7) << node->GetN() << " (+"
         << std::setw(2) << node->GetNInFlight() << ") ";
 
-    oss << "(V: ";
-    optional<float> v;
-    if (node->IsTerminal()) {
-      v = node->GetTerminalNodeValue();
-    } else {
-      v = GetCachedFirstPlyValue(node);
-    }
-    if (v) {
-      oss << std::setw(6) << std::setprecision(2) << *v * 100;
-    } else {
-      oss << " --.--";
-    }
-    oss << "%) ";
-
     oss << "(P: " << std::setw(5) << std::setprecision(2) << node->GetP() * 100
         << "%) ";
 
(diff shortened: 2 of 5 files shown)


mooskagh:
I suggest to keep it for now.
I returned it back after removal because someone asked where it's gone and I also thought it useful for debug.
While I think it should be removed from Node, at the same time printing V from other source should be added.
Either store Vs for first-level (from root) nodes in Search class.
Or query them explicitly when outputting --verbose-move-stats stats.

dubslow:
> Either store Vs for first-level (from root) nodes in Search class.

This is what I suggested in the OP comment here. I'm not sure what your other option means.

mooskagh:
Other option would be to create NetworkComputation, stuff it with positions after first move, do ComputeBlocking, and display the result. (not that I like that option).

dubslow:
Yich. I'll stick to a conditional variable in `Search`.

dubslow:
Okay I've implemented looking up first ply values via the cache, per discussion on Discord, and it works like a charm except for this little problem:

```
bestmove f4c7
quit
lc0: ../../src/utils/cache.h:44: lczero::LruCache<K, V>::~LruCache() [with K = long unsigned int; V = lczero::CachedNNRequest]: Assertion `allocated_ == 0' failed.
Aborted
```

This is the final assert in the cache destructor located at utils/cache.h:44. I don't see how doing a simple lookup of data in cache could lead to this problem. That lookup does "pin" the entries, but even if that's suboptimal performance wise, it certainly should not cause asserts to start failing just by using one single public method of the class. This would seem to indicate there's a bug in the cache implementation, perhaps specifically in the pin-handling; not coincidentally, this might be able to explain the other intermittent crashes seen, if they only happen when the cache is full or otherwise under severe stress.

dubslow:
Of course it could just be me misunderstanding how to use cache in someway (though I'm doubtful of that), so please review my commit carefully as well.

dubslow:
Fixing the idiocy still leaves the cache problem, and it's trivially reproducible. Just compile this PR, launch lc0, and run `go nodes 2` followed by `quit`. `go nodes 1` doesn't cause the assertion to fail, so it is definitely something about the `cache_->Lookup()` call in `GetCachedFirstPlyValue`.

dubslow:
The output formatting was tweaked slightly, that can be reverted easily:

```
info string g2g4  (378 ) N:       0 (+ 0) (V:  --.--%) (P:  1.81%) (Q: -0.84576) (U: 0.86935) (Q+U:  0.02360) 
info string f2f3  (346 ) N:       0 (+ 0) (V:  --.--%) (P:  2.10%) (Q: -0.84576) (U: 1.00607) (Q+U:  0.16031) 
info string g1h3  (161 ) N:       0 (+ 0) (V:  --.--%) (P:  2.15%) (Q: -0.84576) (U: 1.03141) (Q+U:  0.18565) 
info string b1a3  (34  ) N:       0 (+ 0) (V:  --.--%) (P:  2.20%) (Q: -0.84576) (U: 1.05324) (Q+U:  0.20748) 
info string h2h4  (403 ) N:       2 (+ 1) (V:  -5.07%) (P:  2.24%) (Q: -0.06295) (U: 0.26915) (Q+U:  0.20620) 
info string f2f4  (351 ) N:       4 (+ 0) (V:  -3.35%) (P:  2.49%) (Q: -0.04708) (U: 0.23916) (Q+U:  0.19208) 
info string a2a4  (207 ) N:       4 (+ 0) (V:  -3.17%) (P:  2.55%) (Q: -0.03219) (U: 0.24473) (Q+U:  0.21254) 
info string b2b4  (234 ) N:       5 (+ 0) (V:  -1.97%) (P:  2.80%) (Q: -0.01099) (U: 0.22413) (Q+U:  0.21314) 
info string b1c3  (36  ) N:       6 (+ 0) (V:  -0.39%) (P:  3.14%) (Q: -0.01811) (U: 0.21505) (Q+U:  0.19694) 
info string h2h3  (400 ) N:       6 (+ 0) (V:  -1.08%) (P:  3.26%) (Q: -0.01374) (U: 0.22330) (Q+U:  0.20956) 
info string a2a3  (204 ) N:       6 (+ 0) (V:  -0.30%) (P:  3.23%) (Q: -0.00840) (U: 0.22099) (Q+U:  0.21260) 
info string d2d3  (288 ) N:       6 (+ 0) (V:  -1.31%) (P:  3.29%) (Q: -0.00640) (U: 0.22513) (Q+U:  0.21873) 
info string b2b3  (230 ) N:       7 (+ 0) (V:  -0.14%) (P:  3.40%) (Q: -0.00785) (U: 0.20373) (Q+U:  0.19588) 
info string c2c3  (259 ) N:      10 (+ 0) (V:   1.35%) (P:  4.78%) (Q:  0.00886) (U: 0.20854) (Q+U:  0.21739) 
info string e2e3  (317 ) N:      11 (+ 0) (V:   2.34%) (P:  5.06%) (Q:  0.01103) (U: 0.20219) (Q+U:  0.21322) 
info string g2g3  (374 ) N:      12 (+ 1) (V:   1.35%) (P:  5.85%) (Q:  0.00998) (U: 0.20025) (Q+U:  0.21023) 
info string e2e4  (322 ) N:      23 (+ 0) (V:   3.23%) (P:  9.94%) (Q:  0.02342) (U: 0.19872) (Q+U:  0.22214) 
info string g1f3  (159 ) N:      27 (+ 0) (V:   3.56%) (P: 11.02%) (Q:  0.03202) (U: 0.18871) (Q+U:  0.22073) 
info string d2d4  (293 ) N:      34 (+ 1) (V:   4.34%) (P: 14.16%) (Q:  0.03185) (U: 0.18868) (Q+U:  0.22053) 
info string c2c4  (264 ) N:      36 (+ 1) (V:   3.81%) (P: 14.53%) (Q:  0.03436) (U: 0.18346) (Q+U:  0.21781)```

File: src/mcts/search.cc [hunk 1]
```diff
@@ -187,11 +187,20 @@ void Search::SendMovesStats() const {
...
+  const float puct_u_coeff =
+      kCpuct * std::sqrt(std::max(root_node_->GetChildrenVisits(), 1u));
+
   for (Node* child : root_node_->Children()) {
     nodes.emplace_back(child);
   }
   std::sort(nodes.begin(), nodes.end(),
-            [](const Node* a, const Node* b) { return a->GetN() < b->GetN(); });
+            [&parent_q, &puct_u_coeff](const Node* a, const Node* b) {
+              auto an = a->GetN(), bn = b->GetN();
```
mooskagh:
```cpp
   return std::tie(a->GetN(), a->GetQ(parent_q) + a->GetU() * puct_u_coeff) < 
          std::tie(b->GetN(), b->GetQ(parent_q) + b->GetU() * puct_u_coeff);
```
is more idiomatic, and probably more readable (debatable).

src/mcts/search.cc
```diff
@@ -216,16 +224,29 @@ void Search::SendMovesStats() const {
...
-                   std::sqrt(
-                       std::max(node->GetParent()->GetChildrenVisits(), 1u))
-        << ") ";
+        << node->GetQ(parent_q) + puct_u_coeff * node->GetU() << ") ";
     info.comment = oss.str();
     info_callback_(info);
   }
 }
 
+std::string Search::GetCachedFirstPlyValue(const Node* node) const {
```
mooskagh: Function "Get node from cache, but only if it's first ply move, extract V, and convert to string, as percentage with precision of 2" does a bit too much (inspired by [this](http://php.net/manual/en/function.gzgetss.php)? :) ).

What I suggest to do:
1. Make it return NNCacheLock(), so that caller could extract either V or P if needed. (you'll need `return std::move(nneval)`).
    If you object doing that, return at least numeric value of V (as `optional<float>`).
2. Keep all string formatting in SendMovesStats().
3. It seems that it would be useful also to support cases other than first ply depth, and it's doable.
   (whole path from node to root_node_ can be restored, then applied to history).
   It's not necessary to implement that (but go ahead if you want) but I think "FirstPly" part should go from the function name. If you won't implement that, keep the assert and maybe add a short comment that it can be implemented by restoring path from root node to node if needed.
(that item is also debatable)

4. With all those changes, function should be called something like `GetCachedNodeRequestForDebug`, or something less clumsy. The `ForDebug` part (also debatable) is to hint people that it's not optimized and slow.

src/mcts/search.cc
```diff
@@ -515,13 +536,20 @@ void SearchWorker::GatherMinibatch() {
 
     // If node is already known as terminal (win/loss/draw according to rules
     // of the game), it means that we already visited this node before.
-    if (node->IsTerminal()) continue;
+    // Furthermore, we must pass this termination value to NodeToProcess::v,
+    // both here and after ExtendNode.
+    if (node->IsTerminal()) {
+      nodes_to_process_.back().v = node->GetTerminalNodeValue();
```
mooskagh: It feels to be wrong place to fill `v`.
Kind of wrong that v comes from different places for terminal and non terminal nodes. It's easy to remove one and forget another.

`FetchNNResults` really feels the place "where V is populated", but then it probably should be renamed to `PopulateEvalsIntoNodes` or something like that.

Also in this function, code duplication..

src/utils/cache.h
```diff
@@ -84,8 +87,8 @@ class LruCache {
 
   // Looks up and pins the element by key. Returns nullptr if not found.
   // If found, brings the element to the head of the queue (makes it last to
-  // evict).
-  V* Lookup(K key) {
+  // evict); furthermore, a call to Unpin must be made for each such element.
```
mooskagh: That cache pinning is pretty standard thing, so it sounds almost like "malloc(), allocates memory; futhermore, a call to free() must be made for those elements".

I think what would be useful is usage example in comment on the top of the file, and for LookupAndPin() and Unpin() functions comments that `LruCacheLock` is much more recommended.

src/mcts/search.cc
dubslow: Ah yes, I like that better than arbitrarily long chains of comparisons

src/utils/cache.h
dubslow: I'd never heard of cache pinning before, it's certainly not a thing in Python land. Maybe it's only a thing in compiled/optimized languages?

And yes "`malloc` allocates and requires `free`" is absolutely crucial documentation to malloc. Even though "[everyone knows](https://xkcd.com/1053/)" it, we must never assume such knowledge and always document it. Can you imagine glibc maintainers accepting a change to malloc documentation which removed the explanation that a call to `free` is always required?

With sufficiently good documentation, I think an example is unnecessary, and I think that adding more references to `LruCacheLock` as you suggest is sufficient to avoid needing an example.

src/mcts/search.cc
dubslow: Your thoughts mirror mine. Perhaps instead of `FetchNNResults` it could be `FetchMinibatchResults`, including both actually-network'd nodes and nodes that weren't.

src/mcts/search.cc
dubslow: I guess I broadly fall in the category of "[we aren't gonna need it](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)" and "do the simplest thing that could possibly work" (DTSTTCPW).

Although it may be useful in the future, I see no reason to implement things we don't even have a usecase for yet. I'm fine with designing the *current* use to be forward compatible (easy to modify, when needed), but I don't see the sense in actually implementing otherwise unused code. I guess I'm saying that "FirstPly" should stay in the function name, since we don't actually use it otherwise.

So I guess I'll go the `return optional<float>` route for now (not the slightest bit surprised that the string return value wasn't approved (:    ), and if we later need to generalize it, then sure we can do that relatively easily.

src/mcts/search.cc
```diff
@@ -187,45 +187,77 @@ void Search::SendMovesStats() const {
...
+        << node->GetQ(parent_q) + U_coeff * node->GetU() << ") ";
+
     info.comment = oss.str();
     info_callback_(info);
   }
 }
 
+optional<float> Search::GetCachedFirstPlyValue(const Node* node) const {
+  assert(node->GetParent() == root_node_);
+  // This function serves a specific purpose, fetching a NN value from a first
```
mooskagh: This comment will probably not be discovered, as it's not expected that function with this name will have a comment about something else.
I'd either delete that comment completely or stated in one line.

src/mcts/search.cc
```diff
@@ -187,45 +187,77 @@ void Search::SendMovesStats() const {
...
+  // ply move, but it would be relatively straightforward to generalize this
+  // to fetch a complete NN evaluation (policy+value) for an abitrary move.
+  // But, there isn't yet a usecase for that, so we leave this simple for now.
```
mooskagh: Useless sentence. :)

src/mcts/node.h
```diff
@@ -79,18 +79,16 @@ class Node {
...
   // equation. So it's really more of a "reduced U" than raw U.
   float GetU() const { return p_ / (1 + n_ + n_in_flight_); }
-  // Returns value of Value Head returned from the neural net.
-  float GetV() const { return v_; }
   // Returns value of Move probability returned from the neural net
   // (but can be changed by adding Dirichlet noise).
   float GetP() const { return p_; }
   // Returns whether the node is known to be draw/lose/win.
   bool IsTerminal() const { return is_terminal_; }
+  // Unlike GetQ, this is works when n_ == 0
```
mooskagh: It's better to write what it does, not only how it differs. Also, "is" is not needed. Also, period in the end. :)

Also, maybe it's better to remove this function and make GetQ() check for `is_terminal_`?

src/mcts/search.cc
```diff
@@ -187,45 +187,77 @@ void Search::SendMovesStats() const {
...
-    oss << "(V: " << std::setw(6) << std::setprecision(2) << node->GetV() * 100
-        << "%) ";
+
+    oss << " N: " << std::right << std::setw(7) << node->GetN() << " (+"
+        << std::setw(2) << node->GetNInFlight() << ") ";
+
+    oss << "(V: ";
+    // Handrolled optional only has a few overloads available, simple usage only
+    optional<float> v = GetCachedFirstPlyValue(node);
+    if (!v && node->IsTerminal()) v = node->GetTerminalNodeValue();
```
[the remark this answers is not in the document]
dubslow: Well it's needed only if we assume terminal nodes are never cached, which while true, isn't necessarily obvious to the reader at this particular part of the code.

I guess I'll remove it though.

src/mcts/search.cc
```diff
@@ -790,16 +822,18 @@ void SearchWorker::RunNNComputation() {
 
 // 5. Populate computed nodes with results of the NN computation.
 // ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-void SearchWorker::FetchNNResults() {
-  if (computation_->GetBatchSize() == 0) return;
-
-  // Copy NN results into nodes.
+void SearchWorker::FetchMinibatchResults() {
+  // Copy NN results, or cached/duplicate results, into nodes.
+  // The latter is much simpler.
```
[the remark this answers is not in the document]
dubslow: Describes why we have one loop with a shortcircuit `break`, as opposed to something fancier like two different loops or something

src/mcts/search.cc
```diff
@@ -187,45 +187,77 @@ void Search::SendMovesStats() const {
...
-    oss << std::right << std::setw(7) << node->GetN() << " (+" << std::setw(2)
-        << node->GetNInFlight() << ") ";
-    oss << "(V: " << std::setw(6) << std::setprecision(2) << node->GetV() * 100
-        << "%) ";
+
+    oss << " N: " << std::right << std::setw(7) << node->GetN() << " (+"
+        << std::setw(2) << node->GetNInFlight() << ") ";
+
+    oss << "(V: ";
+    // Handrolled optional only has a few overloads available, simple usage only
```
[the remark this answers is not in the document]
dubslow: You know, I tried it in a slightly different way the first time, but it didn't work quite right because of the limited functionality of the handrolled optional, so I changed it to this which works.

But looking again, why didn't I just do this?
```
optional<float> v;
if (node->IsTerminal()) {
  v = node->GetTerminalNodeValue();
} else { 
  v = GetCachedFirstPlyValue(node);
}
```

?

I think the answer is that I was trying to combine that ^ set of conditionals with the ones below about the actual string, but now I'm thinking this is best.

src/mcts/node.h
dubslow: You're the one who added the function :P I kinda like having a separate function for terminal nodes though, it may help prevent human errors regarding conditionals like `if (node->IsTerminal())`.

src/utils/cache.h
```diff
@@ -26,7 +26,10 @@
 
 namespace lczero {
 
-// Generic LRU cache. Thread-safe.
+// Generic LRU cache. Thread-safe. Takes ownership of all values, which are
+// deleted upon eviction; thus, using values stored requires pinning them, which
+// in turn requires Unpin()ing them before cache destruction. The use of
```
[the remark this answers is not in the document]
dubslow: Well I thought the "before destruction" part is more fundamentally important than the "after use" part, since failing to unpin after use won't cause crashes (maybe it will degrade performance), whereas failing to unpin before destruction obviously does cause crashes. None the less, updated, as this way is perfectly sensible too.

src/mcts/search.cc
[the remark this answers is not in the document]
mooskagh: Looks good.

src/mcts/search.cc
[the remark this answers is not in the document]
mooskagh: To code reader (in this case, me), it's not clear that it describes that, so I'd remove that.
Instead, it may make sense to add one sentence comment in the code in lines 831-834 where it's actually happening.
Something like:
```cpp
  // For terminal nodes the V value is know without need to request it from NN.
```

src/mcts/search.cc
```diff
@@ -467,13 +500,13 @@ void SearchWorker::ExecuteOneIteration() {
   // 4. Run NN computation.
   RunNNComputation();
 
-  // 5. Populate computed nodes with results of the NN computation.
-  FetchNNResults();
+  // 5. Retrieve NN computations (and terminal values) into nodes.
+  FetchMinibatchResults();
 
-  // 6. Update nodes.
+  // 6. Propogate a new node's information to all its parents in the tree.
```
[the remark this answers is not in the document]
dubslow: And also "a" -> "the". You're correct that the function operates on the entire batch all at once

src/mcts/search.h
```diff
@@ -108,9 +109,12 @@ class Search {
   void UpdateRemainingMoves();
   void MaybeTriggerStop();
   void MaybeOutputInfo();
-  void SendMovesStats() const;
   void SendUciInfo();  // Requires nodes_mutex_ to be held.
 
+  void SendMovesStats() const;
+  // This function serves only as a helper to SendMovesStats for debug output.
```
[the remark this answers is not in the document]
dubslow: Better to keep inane comments than remove useful comments IMO.

PR #95, opened 2018-06-19
build fixes and instructions for BLAS/OpenCL in windows

borg323:
Makes OpenCL header directory configurable and adds a boost dependency, both needed for building OpenCL under windows. Also a build script is provided together with step-by-step instructions.

### DIFF
--- a/windows_build.md
+++ b/windows_build.md
@@ -1,4 +1,4 @@
-## Windows MKL/OpenCL
+## Windows BLAS/OpenCL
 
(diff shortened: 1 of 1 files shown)


meson.build
```diff
@@ -204,6 +204,9 @@ if get_option('build_backends')
         opencl_files += 'src/neural/blas/transforms.cc'
       endif
 
+      deps += dependency('boost', required: true)
```
dubslow: This will need to be platform dependent, since boost is not required on Linux

borg323: It was required, but only some headers and they were already in /usr/include so the build worked. However the latest version removes the boost dependency completely.

Issue #98, opened 2018-06-19
Crash in multiplexed setup with blas, opencl and heavy threading


frpays:
The testing of PR https://github.com/LeelaChessZero/lc0/pull/87 did bring out a crash on MacOS (and MacOS only) under Multiplexed condition with OpenCL and BLAS, and heavy threading.

Only command line was causing the crash (between a few seconds and a few minutes) is the following:
``
./lc0 --backend=multiplexing "--backend-opts=a(backend=blas,threads=8),b(backend=opencl,threads=3)" -t 11 --minibatch-size=64
``

The crash was not happening every time immediately. You had to run up to 100k nodes (possibly in different runs) to experience it. Sometimes it went it in less a 1K nodes. But it was very replicable, but very random.

Today Gabby experienced same problem happened **on master**:

> Here's my crash with the latest master, run with ./lc0 weights.txt --backend=multiplexing '--backend-opts=a(backend=blas,threads=4),b(backend=opencl,threads=3)' -t 7 --minibatch-size=64. This is not using the PR. https://ghostbin.com/paste/5vxbg

[He] was also running the 256x20 net by @logosgg at the time

frpays:
This crash looks very similar to the problem encountered in testing PR87.  

The https://github.com/LeelaChessZero/lc0/pull/87

- If the problem are related, the PR87 makes it happen much more often.

- The problem happens ** only on MacOS **

- The problem happens both with vecLib or Mkl

Here are a few stack traces or the crash happening under PR87.

https://pastebin.com/2R224eaz
https://pastebin.com/ez9CPy7r
https://pastebin.com/3AQ6s0iz
https://pastebin.com/mGVPttHS
https://pastebin.com/3iJAnxcM 
https://pastebin.com/2DRQtUbt 
https://pastebin.com/1fRHzAAG 
https://pastebin.com/wFMzgT0F
https://pastebin.com/xTaVzXxM
https://pastebin.com/yMfYj7V1

mooskagh:
If someone wants to try to reproduce that on linux+clang+llvm stl (it's suspected to be llvm stl specific), here's what to do:

In meson.build, add:
```
  add_project_arguments('-stdlib=libc++', language : 'cpp')
...
  deps += [
    cc.find_library('c++'),
    cc.find_library('c++abi'),
    ]
```

Generate project like this:
```
[...]

PR #99, opened 2018-06-20
fix data race condition

GBeauregard:
This PR fixed a data race condition that was discovered when investigating #98. The fix is by @mooskagh, who analyzed the thread sanitizer output. It currently appears the race condition is unrelated to the crashes that issue pertains to.

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -724,7 +724,7 @@ int SearchWorker::PrefetchIntoCache(Node* node, int budget) {
   // n = 0 and n_in_flight_ > 0, that means the node is being extended.
   if (node->GetN() == 0) return 0;
   // The node is terminal; don't prefetch it.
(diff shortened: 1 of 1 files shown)


src/mcts/search.cc
```diff
@@ -721,7 +721,9 @@ int SearchWorker::PrefetchIntoCache(Node* node, int budget) {
     return 1;
   }
 
-  // If it's a node in process of expansion or is terminal, don't prefetch it.
+  // n = 0 and n_in_flight_ > 0, that means the node is being extended.
+  if (node->GetN() == 0) return 0;
+  // The node is terminal; don't prefetch it.
   if (!node->HasChildren()) return 0;
```
mooskagh: After some more thinking, I think this line as also better to change into
```cpp
  if (node->IsTerminal()) return 0;
```

Does the same, but kind of matches the comment.

PR #106, opened 2018-06-23
Track false positive resign thresholds for non-resign tournament games.

Tilps:
This is untested right now as I'm still making the changes to the client.

### DIFF
--- a/src/selfplay/game.cc
+++ b/src/selfplay/game.cc
@@ -120,7 +120,6 @@ float SelfPlayGame::GetWorstEvalForWinnerOrDraw() const {
   return std::min(min_eval_[0], min_eval_[1]);
 }
 
-
 void SelfPlayGame::Abort() {
   std::lock_guard<std::mutex> lock(mutex_);
   abort_ = true;
--- a/src/selfplay/game.cc
+++ b/src/selfplay/game.cc
@@ -50,6 +50,7 @@ SelfPlayGame::SelfPlayGame(PlayerOptions player1, PlayerOptions player2,
 
 void SelfPlayGame::Play(int white_threads, int black_threads, bool enable_resign) {
   bool blacks_move = false;
+  min_eval_[0] = min_eval_[1] = 1.0f;
 
   // Do moves while not end of the game. (And while not abort_)
   while (!abort_) {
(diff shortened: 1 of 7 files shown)


dubslow:
LGTM at first readthrough

src/selfplay/loop.cc
```diff
@@ -75,6 +75,15 @@ void SelfPlayLoop::CmdStart() {
...
 
 void SelfPlayLoop::SendGameInfo(const GameInfo& info) {
+  // Send separate resign report before gameready as client gameready parsing
+  // will easily get confused by adding new parameters as both training file
+  // and move list potentially contain spaces.
+  if (info.min_false_positive_threshold) {
+    std::string resign_res = "resign_report";
+    resign_res += 
+        " fp_threshold " + std::to_string(*info.min_false_positive_threshold);
+    SendResponse(resign_res);
```
Tilps: I think I should be using a SendResponses at end of function and a vector of strings instead - so the output mutex is held across both outputs to prevent interleaving.
I'll do that unless someone else has a better idea before I make the change :P

src/selfplay/game.cc
```diff
@@ -50,6 +50,7 @@ SelfPlayGame::SelfPlayGame(PlayerOptions player1, PlayerOptions player2,
 
 void SelfPlayGame::Play(int white_threads, int black_threads, bool enable_resign) {
   bool blacks_move = false;
+  min_eval_[0] = min_eval_[1] = 1.0f;
```
mooskagh: Maybe init that in constructor? Or even at definition:
```cpp
   float min_eval_[2] = {1.0f, 1.0f};
```

src/selfplay/loop.cc
```diff
@@ -75,6 +75,16 @@ void SelfPlayLoop::CmdStart() {
 }
 
 void SelfPlayLoop::SendGameInfo(const GameInfo& info) {
+  std::vector<std::string> responses;
+  // Send separate resign report before gameready as client gameready parsing
+  // will easily get confused by adding new parameters as both training file
+  // and move list potentially contain spaces.
+  if (info.min_false_positive_threshold) {
+    std::string resign_res = "resign_report";
```
mooskagh: `gameready` is a fake uci command, why to add separate resign_report command rather than add additional parameter to gameready?

Tilps: I was trying to make the change in a way which won't break current clients.
Because both file name and and moves have spaces in them, the parsing code for gameready is messy.

The first version will crash out if there are any spaces in the file name and also won't work if you add any new parameters.
The second version with the fix for spaces on further reflection looks like its possible to insert new parameters between game id and moves, by making the code even more messy.
Breaking the first version is probably fine I guess, but its probably simpler to have this optional aspect be in its own line.
Let me know if you feel strongly and I'll change it and make the hacks to the client to match.

mooskagh: Maybe it makes sense to fix the client then, because uci format is really "split string using known tokens", not "split using spaces".
We had problems with spaces in training data already because of that interpretation. Also having to output two lines atomically is not really nice.

But if you don't feel too energetic to implement that, current approach is kind of ok too.

Tilps: The code I wrote for parsing the new line, splits by spaces and walks looking for tokens to take args from following index.  But it only works when token arguments aren't free form text (which I was happy to think should be a safe assumption for a resign_report).
If we want to 'fix it properly' (which yes I'm not feeling energetic to do) the filename parameter in gameready would need to be passed in a form which can be recognized as a token - maybe quoted with escape sequence support?
Such a change would need to be added as optional support in the client, left for a while (say at least a few days?) then the corresponding change made in lc0.  

I think having this tracking in place in the shorter term has immediate value. So am happy to push forward if you aren't going to push back too hard :P

Issue #112, opened 2018-06-24
Meson: handle better headers include directories.


frpays:
Right now, for tensor_flow, openblas, mkl, cudnn, a unique include path is specified in `meson_options` If it's different from the platform, the build fails and the user has to edit `meson_options`.

But honestly we should do better here.

For all include headers, we could have a list of possible directories, and add the first one that exists, or the first one that exists and contains the header. If someone knows how to code that in meson. That would be a few options less, or some options that you almost never have to change.

And I point out that when you found the library, the include path is most often derived from it.

frpays:
This PR addresses https://github.com/LeelaChessZero/lc0/issues/112 I guess.

mooskagh:
Do we kind of have that now?

ddobbelaere:
Although it's a bit hacky, #113 does the trick for openblas (add first found include dir out of list of options).

PR #113, opened 2018-06-24
Meson: added search for OpenBLAS include directories

ddobbelaere:
Search for OpenBLAS headers in array `get_option('openblas_include')`. Add the first directory that contains `cblas.h` to `includes`.

This fixes following build meson build error on Ubuntu 18.04 with `libopenblas-dev` installed

> meson.build:155:6: ERROR:  Include dir /usr/include/openblas/ does not exist.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -153,22 +153,22 @@ if get_option('build_backends')
     elif get_option('openblas') and openblas_lib.found()
       add_project_arguments('-DUSE_OPENBLAS', language : 'cpp')
 
-      openblas_headers_found = false
-      if cc.has_header('cblas.h')
-        openblas_headers_found = true
-      else
+      if not cc.has_header('openblas_config.h')
+        openblas_headers_found = false
+        
         foreach d : get_option('openblas_include')
           if not openblas_headers_found
-            if cc.has_header('cblas.h', args: '-I' + d)
-      	      includes += include_directories(d)
+            if cc.has_header('openblas_config.h', args: '-I' + d)
+              # add the first valid include directory
+              includes += include_directories(d)
               openblas_headers_found = true
             endif
           endif
         endforeach
-      endif
-      
-      if not openblas_headers_found
-        error('Openblas headers are required but not found!')
+        
+        if not openblas_headers_found
+          error('Failed to detect OpenBLAS headers. Did you install libopenblas-dev?')
+        endif
       endif
 
       deps += [ openblas_lib ]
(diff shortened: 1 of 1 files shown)


borg323:
What about something like:
`if not cc.has_header('cblas.h')`
`   includes += include_directories(get_option('openblas_include'))`
`endif`

ddobbelaere:
That would also have worked in my specific case (Ubuntu 18.04, as `cblas.h` is in system include directory). The suggested solution also work for multiple possible include directories. It searches for the first match and only adds this directory to `includes`.

frpays:
The proposal of @borg323 is to be examined. If it works as well on every know platform, it's simpler and maybe more in the spirit.

borg323:
My suggestion can be found here: https://github.com/borg323/lc0/tree/openblas_inc

Regarding openblas_dll, it is searched first because the windows build also include a libopenblas.a that is for mingw and doesn't work with MSVS.

meson.build
```diff
@@ -152,7 +152,25 @@ if get_option('build_backends')
 
     elif get_option('openblas') and openblas_lib.found()
       add_project_arguments('-DUSE_OPENBLAS', language : 'cpp')
-      includes += include_directories(get_option('openblas_include'))
+
+      openblas_headers_found = false
```
mooskagh: That's fine if that's the easiest way to implement that, but maybe there is another way.
I suggest you to show this code in #mesonbuild channel on irc.freenode.org, maybe they will suggest something.
Or I can do that myself tomorrow.

frpays:
Please tell me if got it right.

- In the borg version, 'openblas_include' is used only if the blas header is not found in standard directories. If this case, it has to exist, and contain the blas headers otherwise the build fails.

- In the original PR version, you could actually fill the 'openblas_include' with all the possible paths when you can find the openblas headers on all platform. Then the for loop will chose the first one containing the blas headers. If non contains the blas headers, the build fails.

borg's version is simpler but asks some users to do some research on fix the meson_options.
orignal version has the advantage that once you gathered all possible paths, there's no need for anyone to change this part on meson_options. It's more "magical".

If I not mistaking, then I can of like the original version.

ddobbelaere:
You are right, this was my intention from the beginning (look for the first valid include directory and only add this to includes, in line with issue https://github.com/LeelaChessZero/lc0/issues/112). This gives us the flexibility to add an extra include directory to the option `openblas_include`. I'm glad you like it.

I like @borg323's suggestion to look for the presence of `openblas_config.h`, I've changed this. I also tried to make the code a bit more compact. Sorry for the commit mess, I am a bit new to pull requests, but it should be alright now.

PR #118, opened 2018-06-26
Update time tuning to tune on proper exe

jjoshua2:
Tune after 833 games. Better than tune with wrong exe. Probably comparable to original values.


jjoshua2:
zz tested my 500 game tune vs the original and it was slightly worse at 48s. But at least it isn't the serious regression this is, so I'll commit but make note we need more tuning and testing.
```   # PLAYER                                      :  RATING  ERROR   GAMES
   1 LC0_Id9106_SM1.8_TCP41_TCLW1000_TCRW39.5    :       0   ----      50
   2 LC0_Id9106_SM1.99_TCP24_TCLW55_TCRW70       :     -28     51      50```

PR #119, opened 2018-06-26
Rename resign options (shorter, more descriptive)

dubslow:


src/selfplay/tournament.cc
```diff
@@ -67,8 +67,8 @@ void SelfPlayTournament::PopulateOptions(OptionsParser* options) {
       "multiplexing";
   options->Add<StringOption>(kNnBackendOptionsStr, "backend-opts");
   options->Add<BoolOption>(kVerboseThinkingStr, "verbose-thinking") = false;
-  options->Add<FloatOption>(kResignPlaythroughStr, 0.0f, 100.0f,
-                            "resign-playthrough") = 0.0f;
+  options->Add<FloatOption>(kResignPlaythruStr, 0.0f, 100.0f,
+                            "resign-playthru") = 0.0f;
```
mooskagh: allow-resign-percentage
and 
"The percentage of games which are allowed to resign"
?

dubslow: That would change the value to 1-value, you'd have to take that up with Discord.

mooskagh:
I've discussed that several times on #dev and there's really no conclusion.
I don't like the current flag names, but I'm not fond of the proposed ones too.

"playthrough" (and variants) doesn't really tell anything to people who don't have any context.
Some variants like "only-allow-resign-in-this-fraction-of-games" would be better but I cannot come up with a short name which that meaning.

Also from "threshold" it's not clear why it should be between 0 and 100. 
`--resign-winrate` proposal is fine (and make it between -1 and 1)..
`--ignore-resign-percentage`... Meh.

At this point it feels that keeping the current names is the best thing to do. Not perfect but at least people already use them.

PR #120, opened 2018-06-27
Clop time managment update

jjoshua2:
2267 games update. 
left peak 65. right peak 80. peak ply 27. scale 1.95 was tested around 1900 games, first result of actually beating the initial time management according to ZZ.
```   # PLAYER                                      :  RATING  ERROR   GAMES
   1 LC0_Id9149_SM1.95_TCP27_TCLW65_TCRW80       :      11     43     100
   2 LC0_Id9149_SM1.8_TCP41_TCLW1000_TCRW39.5    :       0   ----     100
```
11 elo ahead instead of 28 behind for a gain of about 40 elo since last commit. at 48s+0.2


dubslow:
The commits don't actually match the posted values? lol (I see 67.0/76.0/26.0/1.93 in git)

PR #124, opened 2018-06-30
Depth redux: remove from Node, reimplement

dubslow:
This is a to-be-finished PR which I'm starting now so that I can record my "data collection" as far as implementing depth is concerned. (It also depends on 94, since they both affect tons of similar code. Only the final of the commits listed is relevant to the PR.)

I've implemented in this commit 3 tentative choices:

1) The simple average depth of all nodes in the tree

2) An exponential moving average of the depth of recently-added nodes, with 1/e-life of 50 playouts

3) The simple average depth of all *leaf* nodes in the tree

Number 1) was my "obvious" choice, with 3) having been suggested to me which I considered a decent alternative; there was some concern, which I partially shared, that recent nodes added to the tree might be significantly deeper than the average depth of the entire tree, hence the inclusion of option 2).

Here is some data I collected on the matter. First two positions come from Kramnik vs Leko 2004, famous game 8.

First, where the best test9 6block network is unable to spot Qd3!! (in other words, for our testing purposes, this search is highly concentrated, with the "best" move getting 97% of playouts): https://pastebin.com/raw/TE3GNtm7

Notice several features of the various depths: a) 2) is generally 2 ply above 1), but it is wildly non-monotonic,  with a several-ply error-spread around its average (including *below* 1) on occasion) and is generally, I believe, unfit for use as is. It could be smoothed a lot more by increasing the 1/e-life, but this would also serve to bring it much closer to 1) in value. b) I'm fascinated by how closely 3) matches 1). They are *consistently* separated by 0.8 ply, with error bars of perhaps ±0.1 ply.

Second position, right after Qd3 is on the board: https://pastebin.com/raw/DLtcTFah The 6 block network actually took *way* more nodes to locate the trouble than did 45x; 45x realizes by 10K-20K nodes that white is in deep trouble (including switching away from a7 pretty quickly), but here it takes until 45K-50K before test9 goes "oh shit", so I gave it 200K nodes instead of 100K (and even so stays stuck on a7 far, far more than 45x). Note that after going "oh shit", the depth never gets above 13 as the network explores various ways to try and get out of the hole. In this case the search is a little more spread out, so on the whole option 2) isn't really different from options 1) or 3). (What continues to fascinate me is how steadily 3) mirrors 1) with that gap of 0.85 ply, except in the middle where the net begins to change its mind, where 3) is 1.0 ply above 1), but before and after changing its mind, 3) is firmly, like glue, stuck on 1) + 0.85 ply.)

Here's a third position, semi-conincidentally also Kramnik vs Leko 2004, this time game 14 with fewer pieces on the board. https://pastebin.com/raw/FBVjUpEC I see more or less the same as before. And finally, to make sure lots of terminal nodes are hit in the search, here's test9 confirming that 40. c7 is a good move: https://pastebin.com/raw/P9m4hK2u all seems well there too. All 3 options give very similar results.

Given those tests, I also submit the following subjective criteria: 1) is simplest to understand for users, and it is also not coincidentally the simplest to implement. 2) is nearly as simple to implement, but is not intuitive and quite arbitrary, and doesn't even seem to give very different results from 1) anyways. 3) is substantially more complicated to implement (not truly hard, but by comparison, more so), and as per the test data doesn't really give any different information than 1) either. So I propose we select the simplest, most understandable options, since it gives no less information than the others: number 1).

Thoughts? I want to have a choice down before I clean this up and make it compatible with tree reuse, since that depends on the depth implementation.


Videodr0me:
Good work! Thanks for actually trying all this. Completely agree with your conclusion. The only minor detail to really consider is: Do we want depth to be monotonically increasing or do we allow depth to fluctuate somewhat? We talked about this already and you do not mind the fluctuations, while i believe this might throw some guis off and lead to a lot of user inquieries. So i would still advocate for a avg_new > avg_old check.

On the topic of tree-reuse. I would seriously advocate for seperating the concepts of search depth and tree depth. This would enable the simplest and most bug free implementation, for something which is essentially a cosmetic or semantic issue. All other approaches of storing depth info in first or even second ply nodes are somewhat shortsighted - as tree-reuse might evolve to enable reusing over several moves. I find it completely intuitive and in line with user expectations if search depth is newely calculated for each new search even in reused trees.

dubslow:
Superseded by #160.

PR #130, opened 2018-07-02
Bug fix: biases used twice in conv1x1 in network_blas

frpays:
This bug existed since the port of BLAS.


dubslow:
Basically, Batchnorm::OffsetMeans handles this already in weights initialization right?

frpays:
We use the batch normalization process to add the biases. So the biases are
added to the batchnorm means at init.

PR #132, opened 2018-07-02
cuDNN windows build for AppVeyor

borg323:
This enables building cuDNN lc0 executables for windows, in addition to openblas and OpenCL ones. Also the openblas and OpenCL executables are packaged with the appropriate dlls in respective zip files.


jjoshua2:
Sounds nice. But we might not want to include dll with every update since they are the same? Maybe they should be released on github with versions in case we ever update blas

mooskagh:
CUDA EULA seems to allow bundling .dlls with other software that uses it, but doesn't allow distributing them separately.
So we have to do one of:
1. Include .dlls into lc0 packages.
2. Not include, and write instructions how to get there from NVidia site.
3. Have two versions built, (1) and (2).

We surely don't want to have a separate package only with .dlls.

borg323:
It is trivial to add the dlls in a zip together with the executalbe as done for opencl and blas. Should I do it here or a separate PR? Also I agree with @jjoshua2 about creating the zip files only on releases/tags and not with every commit which is done with the latest commit for blas and opencl

borg323:
With the latest commit a zip file will be created containing lc0 together with the cuda/cudnn dlls.

PR #137, opened 2018-07-03
Update engine.cc time management to slower CLOP tune

jjoshua2:
Updated to 1.95 and 26.2 from over 5000 games of clop at 30+.17. retuned left and right for close to 1000 games at 60+1. And got these values. ZZ tested this compared to equal with less increment as well.

```
TC 192s+0.8s/move match running, 28/100 games played
   # PLAYER                                      :  RATING  ERROR   GAMES
   1 LC0_Id9149_SM1.95_TCP26.2_TCLW82_TCRW74     :       0     35      28
   2 LC0_Id9149_SM1.8_TCP41_TCLW1000_TCRW39.5    :       0   ----      28`
```

This shows even though I tuned for a 60/1 ratio (wheras tcec is 180 ratio) it works just as good as original even with a 240 ratio. (Although likely would be stronger saving a tiny bit more time for end compared to tune)

### DIFF
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -77,14 +77,14 @@ void EngineController::PopulateOptions(OptionsParser* options) {
   options->Add<ChoiceOption>(kNnBackendStr, backends, "backend") =
       backends.empty() ? "<none>" : backends[0];
   options->Add<StringOption>(kNnBackendOptionsStr, "backend-opts");
-  options->Add<FloatOption>(kSlowMoverStr, 0.0f, 100.0f, "slowmover") = 1.93f;
+  options->Add<FloatOption>(kSlowMoverStr, 0.0f, 100.0f, "slowmover") = 1.95f;
   options->Add<IntOption>(kMoveOverheadStr, 0, 10000, "move-overhead") = 100;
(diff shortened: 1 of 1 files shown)


jjoshua2:
previous parameters got lucky with +11 and retesting showed
test with 11+/-43 Elo repeated (TC 48s+0.2s/move):
  # PLAYER                                      :  RATING  ERROR   GAMES
   1 LC0_Id9149_SM1.8_TCP41_TCLW1000_TCRW39.5    :       0   ----     100
   2 LC0_Id9149_SM1.95_TCP27_TCLW65_TCRW80       :     -21     42     100

jjoshua2:
CCLS verified that leela was using less time at beginning than other engines. This makes it more in line with other engines, and has been tested on my stream, and by Redshift and other streamers.

jjoshua2:
In the future I want to retun slow_mover with the aggressive pearlyexit hopefully at 2+1 tc to balance lots of nodes with a regular sized increment. I think this commit will get just about all the elo there is in time though, and keep it looking pretty, until someone does some serious fail low spend more time on important positions type of work.

PR #138, opened 2018-07-03
Reduced Move from 4 bytes to 2 bytes

AlexGreason:
Changed the internal format of Move from four 1-byte values to packed into one 16-bit value


src/chess/bitboard.h
```diff
@@ -190,17 +190,29 @@ class Move {
...
+  }
   Move(const std::string& str, bool black = false);
   Move(const char* str, bool black = false) : Move(std::string(str), black) {}
 
-  BoardSquare from() const { return from_; }
-  BoardSquare to() const { return to_; }
-  Promotion promotion() const { return promotion_; }
-  bool IsCastling() const { return castling_; }
-  void SetCastling() { castling_ = true; }
+  BoardSquare to() const      { return (BoardSquare)((data_ & 0b0000000000111111)); }
```
DanielUranga: possibly would be good to define constants for this magic numbers?

src/chess/bitboard.cc
```diff
@@ -275,46 +275,47 @@ const int kQueenCastleIndex =
...
 
-uint16_t Move::as_packed_int() const {
-  if (promotion_ == Promotion::Knight) {
-    return from_.as_int() * 64 + to_.as_int();
-  } else {
-    return static_cast<int>(promotion_) * 64 * 64 + from_.as_int() * 64 +
-           to_.as_int();
-  }
-}
+    uint16_t Move::as_packed_int() const {
```
dubslow: Given the fuss you're going to to maintain just such a packed int, I'd say this can be simplified to just `return data_`, since this function is considered as a black box from the outside.

src/chess/bitboard.h
```diff
@@ -190,17 +190,29 @@ class Move {
...
+  Move(BoardSquare from, BoardSquare to) {
+    data_ = 0;
+    data_ |= to.as_int();
+    data_ |= from.as_int() << 6;
+  }
+  Move(BoardSquare from, BoardSquare to, Promotion promotion){
+    data_ = 0;
+    data_ |= to.as_int();
+    data_ |= from.as_int() << 6;
+    data_ |= (uint8_t)promotion << 12;
```
dubslow: The actual bit twiddling should be nowhere except in the getters and setters. Use those here for these two constructors.

src/chess/bitboard.h
dubslow: Yes, absolutely.

Something like this:

```
private:
  enum MASKS_ : uint16_t {
    TO_MASK_     = 0b0000000000111111;
    FROM_MASK_   = 0b0000111111000000;
    PROMO_MASK_  = 0b0111000000000000;
    CASTLE_MASK_ =  0b1000000000000000;
  }
```

Then you can do e.g. `void SetFrom(BoardSquare from) { data_ = (data_ & ~FROM_MASK_) | (from.as_int() << 6); }`

src/chess/bitboard.h
```diff
@@ -190,17 +190,29 @@ class Move {
...
   Move(const char* str, bool black = false) : Move(std::string(str), black) {}
 
-  BoardSquare from() const { return from_; }
-  BoardSquare to() const { return to_; }
-  Promotion promotion() const { return promotion_; }
-  bool IsCastling() const { return castling_; }
-  void SetCastling() { castling_ = true; }
+  BoardSquare to() const      { return (BoardSquare)((data_ & 0b0000000000111111)); }
+  BoardSquare from() const    { return (BoardSquare)((data_ & 0b0000111111000000) >> 6); }
+  Promotion promotion() const { return (Promotion)  ((data_ & 0b0111000000000000) >> 12); }
```
dubslow: I suspect, but am not sure, that crem would request that these be at least capitalized, like all other member functions, and possibly prefixed with `Get`, like the rest of the classes in this codebase.

src/chess/bitboard.h
```diff
@@ -190,17 +190,29 @@ class Move {
...
+  bool IsCastling() const     { return (bool)       ((data_ & 0b1000000000000000) >> 15); }
+  void SetCastling() { data_ |= 0b1000000000000000; }
+
+  void SetTo(BoardSquare to){ data_ =             (data_ & 0b1111111111000000) | to.as_int(); }
+  void SetFrom(BoardSquare from){ data_ =         (data_ & 0b11110000001111111) | (from.as_int() << 6); }
+  void SetPromotion(Promotion promotion){ data_ =(data_ & 0b1000111111111111) | ((uint8_t)promotion << 12); }
```
dubslow: Space between `)` and `{`, and between `=` and `(` please

src/chess/bitboard.h
```diff
@@ -209,21 +221,25 @@ class Move {
...
 
   void Mirror() {
-    from_.Mirror();
-    to_.Mirror();
+    BoardSquare tmpfrom = from();
+    BoardSquare tmpto = to();
+    tmpfrom.Mirror();
+    tmpto.Mirror();
+    SetFrom(tmpfrom);
+    SetTo(tmpto);
```
dubslow: Wouldn't this be simpler as 
```
SetFrom(from().Mirror());
SetTo(to().Mirror());
```
?

Or am I missing something obvious about that?

src/chess/bitboard.h
```diff
@@ -190,17 +190,29 @@ class Move {
...
-  bool IsCastling() const { return castling_; }
-  void SetCastling() { castling_ = true; }
+  BoardSquare to() const      { return (BoardSquare)((data_ & 0b0000000000111111)); }
+  BoardSquare from() const    { return (BoardSquare)((data_ & 0b0000111111000000) >> 6); }
+  Promotion promotion() const { return (Promotion)  ((data_ & 0b0111000000000000) >> 12); }
+  bool IsCastling() const     { return (bool)       ((data_ & 0b1000000000000000) >> 15); }
+  void SetCastling() { data_ |= 0b1000000000000000; }
+
+  void SetTo(BoardSquare to){ data_ =             (data_ & 0b1111111111000000) | to.as_int(); }
+  void SetFrom(BoardSquare from){ data_ =         (data_ & 0b11110000001111111) | (from.as_int() << 6); }
```
dubslow: Maybe we should add a comment to the BoardSquare class that its value is guaranteed to have only the lowest 6 bits set?

Tilps:
Might want to do some performance checks with the random backend to see how much this costs in performance for the potential memory usage improvement.

Tilps:
Also, rather than all this bit fiddling you could try declaring them as bitfields - might perform better depending on compiler optimization tricks.

src/chess/bitboard.cc
AlexGreason: I'd planned on that, but the current code is not equivalent to that. If it ignored the promotion when it was Promotion::None, that'd be equivalent to just masking off the castling bit, but as it is it ignores knight promotions, and adding that logic with the packed int wouldn't be much simpler than what's already there.

src/chess/bitboard.h
AlexGreason: Done

src/chess/bitboard.h
AlexGreason: Done

src/chess/bitboard.h
AlexGreason: Done

src/chess/bitboard.h
AlexGreason: Done

AlexGreason:
Checked with random backend:
With the modified version, I get 1554756 nodes on a go movetime 10000 from the startpos
With the original version, I get 1554182 nodes on a go movetime 10000 from startpos.
So no detectable performance degradation.

src/chess/bitboard.h
AlexGreason: Done

src/chess/bitboard.cc
AlexGreason: I've pinged crem asking if it's supposed to be ignoring knight promotions, if that's unintended behaviour then I agree that it'd be better to just return data.

src/chess/bitboard.h
```diff
@@ -257,6 +257,13 @@ class Move {
 
  private:
    uint16_t data_; //first bit castling, next 3 promotion, next 6 fromsquare, next 6 tosquare
```
dubslow: You can declare this `= 0;` and remove those initializations from the constructor. 

Also, according to the style guide in use, all comments must be capitalized and end in a period (or other grammatical marker).

AlexGreason: Done

src/chess/bitboard.cc
AlexGreason: Done

src/chess/bitboard.h
```diff
@@ -190,40 +190,64 @@ class Move {
...
-      : from_(from), to_(to), promotion_(promotion) {}
-  Move(const std::string& str, bool black = false);
-  Move(const char* str, bool black = false) : Move(std::string(str), black) {}
-
-  BoardSquare from() const { return from_; }
-  BoardSquare to() const { return to_; }
-  Promotion promotion() const { return promotion_; }
-  bool IsCastling() const { return castling_; }
-  void SetCastling() { castling_ = true; }
+  Move(BoardSquare from, BoardSquare to) {
```
mooskagh: This constructor and the next one are quite performance critical and `SetTo()` functions do unnecesary job (clearing bits, for example).
I'd suggest to write directly to data_ in constructor:
`: data_(to.as_int() + (from.as_int() << 6)){}`

src/chess/bitboard.h
```diff
@@ -190,40 +190,64 @@ class Move {
...
+  }
+  Move(BoardSquare from, BoardSquare to, Promotion promotion) {
+    SetTo(to);
+    SetFrom(from);
+    SetPromotion(promotion);
+  }
+  Move(const std::string &str, bool black = false);
+  Move(const char *str, bool black = false) : Move(std::string(str), black) {}
+
+  BoardSquare GetTo() const { return (BoardSquare)((data_ & TO_MASK_)); }
```
mooskagh: Modern way is to write
`{data_ & TO_MASK_}` instead of `(BoardSquare)((data_ & TO_MASK_))`.

src/chess/bitboard.h
```diff
@@ -190,40 +190,64 @@ class Move {
...
+  BoardSquare GetFrom() const {
+    return (BoardSquare)((data_ & FROM_MASK_) >> 6);
+  }
+  Promotion GetPromotion() const {
+    return (Promotion)((data_ & PROMO_MASK_) >> 12);
+  }
+  bool IsCastling() const { return (bool)((data_ & CASTLE_MASK_) >> 15); }
```
mooskagh: I think Visual Studio will complain on this with its "performance warning". Also, `>> 15` is really redundant here.
To make everyone happy, `return (data & CASTLE_MASK_) != 0;`.

src/chess/bitboard.h
```diff
@@ -190,40 +190,64 @@ class Move {
...
+  void SetCastling() { data_ |= CASTLE_MASK_; }
+
+  void SetTo(BoardSquare to) { data_ = (data_ & ~TO_MASK_) | to.as_int(); }
+  void SetFrom(BoardSquare from) {
+    data_ = (data_ & ~FROM_MASK_) | (from.as_int() << 6);
+  }
+  void SetPromotion(Promotion promotion) {
+    data_ = (data_ & ~PROMO_MASK_) | ((uint8_t)promotion << 12);
```
mooskagh: `static_cast<uint8_t>(promotion)`

src/chess/bitboard.h
```diff
@@ -190,40 +190,64 @@ class Move {
...
   uint16_t as_packed_int() const;
 
   // 0 .. 1857, to use in neural networks.
   uint16_t as_nn_index() const;
 
-  bool operator==(const Move& other) const {
-    return from_ == other.from_ && to_ == other.to_ &&
-           promotion_ == other.promotion_;
+  bool operator==(const Move &other) const {
+    return GetFrom() == other.GetFrom() && GetTo() == other.GetTo() &&
```
mooskagh: Indeed it should be `data_ == other.data_`, castlings are expected to be different moves.

src/chess/bitboard.h
```diff
@@ -190,40 +190,64 @@ class Move {
...
+           GetPromotion() == other.GetPromotion();
   }
 
-  bool operator!=(const Move& other) const { return !operator==(other); }
-  operator bool() const { return from_.as_int() != 0 || to_.as_int() != 0; }
+  bool operator!=(const Move &other) const { return !operator==(other); }
+  operator bool() const {
+    return GetFrom().as_int() != 0 || GetTo().as_int() != 0;
```
mooskagh: `data_ != 0`
"Empty" move only happens when it's initialized with a default constructor.

src/chess/bitboard.h
```diff
@@ -190,40 +190,64 @@ class Move {
...
+  }
 
   void Mirror() {
-    from_.Mirror();
-    to_.Mirror();
+    BoardSquare tmpfrom = GetFrom();
```
mooskagh: That's a performance critical function, consider doing that manually here:
```
data_ ^= 0x111000111000;
```

src/chess/bitboard.h
```diff
@@ -240,10 +264,15 @@ class Move {
   }
 
  private:
-  BoardSquare from_;
-  BoardSquare to_;
-  Promotion promotion_ = Promotion::None;
-  bool castling_ = false;
+  uint16_t data_ = 0;  // First bit castling, next 3 promotion, next 6
```
mooskagh: (regarding the comment)
bits are counted from lowest to highest, from zero.
```cpp
// Move, using the following encoding:
// bits 0..5 "to"-square
// bits 6..11 "from"-square
// bits 12..14 promotion value
// bit 15 whether move is a castling
```

src/chess/bitboard.h
```diff
@@ -240,10 +264,15 @@ class Move {
...
+  // fromsquare, next 6 tosquare.
+
+  enum MASKS_ : uint16_t {
+    TO_MASK_ = 0b0000000000111111,
```
mooskagh: https://google.github.io/styleguide/cppguide.html#Enumerator_Names

Enum names should be written as constants: `kToMask`. Like-macro `TO_MASK` is also acceptable, but without trailing underscore.

src/chess/bitboard.h
```diff
@@ -240,10 +264,15 @@ class Move {
...
 
  private:
-  BoardSquare from_;
-  BoardSquare to_;
-  Promotion promotion_ = Promotion::None;
-  bool castling_ = false;
+  uint16_t data_ = 0;  // First bit castling, next 3 promotion, next 6
+  // fromsquare, next 6 tosquare.
+
+  enum MASKS_ : uint16_t {
```
mooskagh: `MASKS_` is a type name, should be mixed case: `Masks` and `BitMasks` etc.

src/chess/board.cc
```diff
@@ -337,8 +337,8 @@ MoveList ChessBoard::GeneratePseudolegalMoves() const {
 }
 
 bool ChessBoard::ApplyMove(Move move) {
-  const auto& from = move.from();
-  const auto& to = move.to();
+  const auto& from = move.GetFrom();
```
mooskagh: I think it's better to make `from` and `to` not references now. (I doubt it was a good idea before, either).

src/chess/board.cc
```diff
@@ -502,8 +502,8 @@ bool ChessBoard::IsUnderAttack(BoardSquare square) const {
 }
 
 bool ChessBoard::IsLegalMove(Move move, bool was_under_check) const {
-  const auto& from = move.from();
-  const auto& to = move.to();
+  const auto& from = move.GetFrom();
```
mooskagh: Also remove references.

src/chess/bitboard.h
dubslow: I had suggested exactly the opposite, which is why it's this way now. Is it really that important? There are already data with the current code that shows no performance difference.

src/chess/bitboard.h
dubslow: As with the other one, performance testing shows no slow down with code as is.

src/chess/bitboard.h
dubslow: That's my bad, thanks for the correction

src/chess/board.cc
AlexGreason: Done

src/chess/board.cc
AlexGreason: Done

src/chess/bitboard.h
AlexGreason: Done

src/chess/bitboard.h
AlexGreason: Done

src/chess/bitboard.h
AlexGreason: Done

src/chess/bitboard.h
AlexGreason: Done

src/chess/bitboard.h
AlexGreason: Done

src/chess/bitboard.h
AlexGreason: Done

src/chess/bitboard.h
AlexGreason: Done

src/chess/bitboard.h
AlexGreason: Done

src/chess/bitboard.h
AlexGreason: That worked for the BoardSquares, but threw an error for the Promotion

src/chess/bitboard.h
AlexGreason: Done

src/chess/bitboard.h
```diff
@@ -190,40 +190,46 @@ class Move {
...
-  Move(const char* str, bool black = false) : Move(std::string(str), black) {}
+      : data_(to.as_int() + (from.as_int() << 6) +
+              (static_cast<uint8_t>(promotion) << 12)) {}
+  Move(const std::string &str, bool black = false);
+  Move(const char *str, bool black = false) : Move(std::string(str), black) {}
+
+  BoardSquare GetTo() const { return {data_ & kToMask}; }
+  BoardSquare GetFrom() const { return {(data_ & kFromMask) >> 6}; }
+  Promotion GetPromotion() const {
+    return (Promotion)((data_ & kPromoMask) >> 12);
```
mooskagh: `{(data_ & kPromoMask) >> 12}`

AlexGreason: doesn't work, as I commented on the earlier one
../../src/chess/bitboard.h:204:39: error: cannot convert ‘<brace-enclosed initializer list>’ to ‘lczero::Move::Promotion’ in return
     return {(data_ & kPromoMask) >> 12};

PR #139, opened 2018-07-04
[WIP don't merge yet] Add "HorizontallyMirror" method to the ChessBoard class

DanielUranga:
The idea is to horizontally mirror the board previous to send to the NN if: no castlings are avaliable and the king of the side to move is on the left side of the board (a, b, c, or d columns).
This way the NN won't need to understand as many possible possitions.

### DIFF
--- a/src/chess/bitboard.h
+++ b/src/chess/bitboard.h
@@ -50,6 +50,9 @@ class BoardSquare {
   // Row := 7 - row.  Col remains the same.
   void Mirror() { square_ = square_ ^ 0b111000; }
 
+  // Col := 7 - col.  Row remains the same.
+  void HorizontallyMirror() { square_ = square_ ^ 0b000111; }
+
   // Checks whether coordinate is within 0..7.
   static bool IsValidCoord(int x) { return x >= 0 && x < 8; }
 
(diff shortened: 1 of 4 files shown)


dubslow:
Uh... pretty sure this violates the Zero principle, at least DeepMind's conception of it. Exploiting symmetries is exploiting domain specific knowledge. Also, won't that make it harder for the NN to generalize the non-HMirrorable positions?

DanielUranga:
On the DeepMind's paper it can be read:
> Go is well suited to the neural network architecture used in AlphaGo
because the rules of the game are translationally invariant (matching the weight sharing structure of convolutional networks),  are defined in terms of liberties corresponding to the adjacencies between points on the board (matching the local structure of convolutional networks), **and are rotationally and
reflectionally symmetric (allowing for data augmentation and ensembling)**

This is the same idea as https://github.com/glinscott/leela-chess/issues/25

dubslow:
Sure, they did exploit domain symmetries in earlier AlphaGo versions, but not in either AlphaGo Zero or AlphaZero, since that isn't Zero.

DanielUranga:
The HMirrorable positions will be in the mid to late game. The idea is to have lc0 to understand less positions, since they're the same.
I agree that this could be considered exploiting domain specific knowledge, but I think it is debatable.
Will post this on Discord to see what other thinks.

DanielUranga:
@Error323 so this should be hypothetically implemented in both lc0 "playing" and "training" part, right? I was going to modify this function: https://github.com/LeelaChessZero/lc0/blob/master/src/neural/encoder.cc#L30

dubslow:
> It won't work across all domains, but I don't think this is important for us?

That's what we mean when we call the project "Leela Chess Zero". I'd be fine with this is we took out the zero from the name.

DanielUranga:
Should I close this as "Won't do" then?

mooskagh:
Do you still have it in your branch? It shows that PR comes from "unknown repository".

DanielUranga:
oh.. I deleted my old repo and forked again, but I can make this PR again if required.

mooskagh:
I think it may be worth trying when we have easy way to experiment, but for now we don't.
This PR cannot be reused anyway, and experimenting framework will be ready not earlier than in a month (I guess). So for now I'm closing this.

PR #147, opened 2018-07-08
Multiplexing mutex optimization for improvement of multi-GPU scalability

dje-dev:
Please see attached text document for full description of this mutex optimization for multi-GPU scaling
[mutex_optimization_comment.txt](https://github.com/LeelaChessZero/lc0/files/2173304/mutex_optimization_comment.txt)

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -570,10 +570,12 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend() {
   Node* node = search_->root_node_;
   // Initialize position sequence with pre-move position.
   history_.Trim(search_->played_history_.GetLength());
+
+  SharedMutex::Lock lock(search_->nodes_mutex_);
+
   // Fetch the current best root node visits for possible smart pruning.
   int best_node_n = 0;
   {
-    SharedMutex::Lock lock(search_->nodes_mutex_);
     if (search_->best_move_node_)
       best_node_n = search_->best_move_node_->GetN();
   }
@@ -583,7 +585,6 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend() {
   while (true) {
     // First, terminate if we find collisions or leaf nodes.
     {
-      SharedMutex::Lock lock(search_->nodes_mutex_);
       // n_in_flight_ is incremented. If the method returns false, then there is
       // a search collision, and this node is already being expanded.
       if (!node->TryStartScoreUpdate()) return {node, true};
@@ -592,8 +593,6 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend() {
       // If we fall through, then n_in_flight_ has been incremented but this
       // playout remains incomplete; we must go deeper.
     }
-
-    SharedMutex::SharedLock lock(search_->nodes_mutex_);
     float puct_mult =
         search_->kCpuct * std::sqrt(std::max(node->GetChildrenVisits(), 1u));
     float best = -100.0f;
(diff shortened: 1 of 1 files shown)


dubslow:
Contents of the text file:

```Adjust SearchWorker::PickNodeToExtend() to take lock on the nodes_mutex_ only once, improving multi-GPU scalability

It has been noted that with the current multiplexing logic, multi-GPU scaling is poor (acutely sublinear or possibly even sub 1.0x). See for example:
 https://github.com/glinscott/leela-chess/issues/687

This simple adjustment to SearchWorker::PickNodeToExtend() appears to dramatically improves multi-GPU performance. We simply take the lock on the nodes_mutex_ only once (at beginning of the method) and hold it, rather than three separate acquisitions/releases.

The following tests appear to confirm the value of this change:
-- the "nodes per second" performance in various configurations shown below (using "go movetime 20000" from the starting position) shows that fp16 performance scales by 1.96x in a 3-GPU configuration (from 53,001 to 103,178 nps) compared to the current scaling of 0.82 (from 48,140 to 39,236). In the case of fp32, the improvement is even better, with 3-GPU yielding scaling of 2.95x).
-- a test match of 40 games at 2 seconds per move, the 3-GPU version scored 4 wins and 0 losses (using fp16 for both)


[  NPS  ] [COMMAND LINE]
[ 25,488] lc0_org -t 6 --minibatch-size=1024 --backend=multiplexing "--backend-opts=(backend=cudnn,gpu=0)"
[ 48,140] lc0_org -t 6 --minibatch-size=1024 --backend=multiplexing "--backend-opts=(backend=cudnnfp16,gpu=0)"
[ 25,460] lc0_mutex -t 6 --minibatch-size=1024 --backend=multiplexing "--backend-opts=(backend=cudnn,gpu=0)"
[ 53,001] lc0_mutex -t 6 --minibatch-size=1024 --backend=multiplexing "--backend-opts=(backend=cudnnfp16,gpu=0)"

[ 46,024] lc0_org -t 6 --minibatch-size=1024 --backend=multiplexing "--backend-opts=(backend=cudnn,gpu=0),(backend=cudnn,gpu=1),(backend=cudnn,gpu=2)"
[ 39,236] lc0_org -t 6 --minibatch-size=1024 --backend=multiplexing "--backend-opts=(backend=cudnnfp16,gpu=0),(backend=cudnnfp16,gpu=1),(backend=cudnnfp16,gpu=2)"
[ 75,047] lc0_mutex -t 6 --minibatch-size=1024 --backend=multiplexing "--backend-opts=(backend=cudnn,gpu=0),(backend=cudnn,gpu=1),(backend=cudnn,gpu=2)"
[103,178] lc0_mutex -t 6 --minibatch-size=1024 --backend=multiplexing "--backend-opts=(backend=cudnnfp16,gpu=0),(backend=cudnnfp16,gpu=1),(backend=cudnnfp16,gpu=2)"

NOTES:
  - lc0_org is Windows binary compiled from codebase as of 7 July (source unmodified)
  - lc0_mutex is Windows binary compiled from codebase as of 7 July (with one modification of the mutex optimization) in PickNodeToExtend
  - all 3 GPUS are Titan V (Volta)
  - the weights file is from network 395
  - the performance (nps) reported above are from a single run, but these tests were all performed three times and the timings were highly stable
```

mooskagh:
The idea was that most of the code runs behind shared mutex, to multiple thereads can do most of stuff simultaneously, but apparently locking overhead is too large when there is contention.

Would you mind also checking whether replacing `search_->nodes_mutex_` to Mutex (rather than SharedMutex) and locking it exclusively everywhere would not slow down things? That way we could get rid of shared mutex completely..

dje-dev:
As suggested I ran an experiment to replace search_->nodes_mutex_ to Mutex and locked exclusively everywhere. Performance tests showed clearly inferior results (easily statistically significant) in the 3-GPU case. For example, the last two reported NPS above decreased to 71,512 and 84,833 (from 75,047 and 103,178).

PR #150, opened 2018-07-08
Code formatting of network_cudnn.cu

mooskagh:
* Ran clang-format
* Added periods at the end of comments.
* Renamed variables from thisFormat to this_format
  (in 2/3 of the file).
No other changes (other than formatting).

### DIFF
--- a/src/neural/network_cudnn.cu
+++ b/src/neural/network_cudnn.cu
@@ -33,7 +33,7 @@
 namespace lczero {
 namespace {
 
-void cudnnError(cudnnStatus_t status, const char *file, const int &line) {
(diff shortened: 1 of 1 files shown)


src/neural/network_cudnn.cu
```diff
@@ -358,25 +357,25 @@ __global__ void batchNormForward_kernel(T *output, const T *input,
   output[index] = (T)el;
 }
 
-// each thread processes single element
+// Every thread processes single element.
```
dubslow: I would think that "each" is more appropriate than "every"?

(also "processes a single element")

PR #160, opened 2018-07-10
Remove depth from Node (-4 bytes in Node, depth now in NodeTree)

dubslow:
This also replaces the "full" depth (which was useless) with the average depth of every node in the tree.
(This is considered superior to "average depth of recently added nodes" and "average depth of leaf nodes", per the testing data recorded in the now-superseded #124.)

In addition to the memory savings, this is also computationally faster (though negligibly so, on the whole). This removes O(N log(N)) operations from pseudo-critical `Search`/`go ...` code, and replaces it with O(N) operations in `Search`/`go ...` code in addition to O(N) non-critical `NodeTree`/`position` code.

Non-regression-of-performance is supported by the following, two tests of each command:

```
time ./lc0-edge-node selfplay --games=100 --parallelism=8 --movetime=30 --backend=random --weights=/home/bill/leela/weights/weights_t9_9155.txt.gz --reuse-tree --share-trees
real    1m17.833s
user    8m14.924s
sys     0m8.156s

real    1m17.675s
user    7m59.168s
sys     0m7.876s


time ./lc0-depth selfplay --games=100 --parallelism=8 --movetime=30 --backend=random --weights=/home/bill/leela/weights/weights_t9_9155.txt.gz --reuse-tree --share-trees
real    1m15.590s
user    8m8.440s
sys     0m8.264s

real    1m9.253s
user    7m16.568s
sys     0m7.152s
```

### DIFF
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -33,6 +33,7 @@
 #include <cstring>
 #include <iostream>
 #include <sstream>
+#include <stack>
 #include <thread>
 #include "neural/encoder.h"
 #include "neural/network.h"
@@ -193,24 +194,6 @@ void Node::FinalizeScoreUpdate(float v) {
   --n_in_flight_;
 }
 
-void Node::UpdateMaxDepth(int depth) {
-  if (depth > max_depth_) max_depth_ = depth;
-}
-
-bool Node::UpdateFullDepth(uint16_t* depth) {
-  // TODO(crem) If this function won't be needed, consider also killing
-  //            ChildNodes/NodeRange/Nodes_Iterator.
-  if (full_depth_ > *depth) return false;
-  for (Node* child : ChildNodes()) {
-    if (*depth > child->full_depth_) *depth = child->full_depth_;
-  }
-  if (*depth >= full_depth_) {
-    full_depth_ = ++*depth;
-    return true;
-  }
-  return false;
-}
-
 Node::NodeRange Node::ChildNodes() const { return child_.get(); }
 
 void Node::ReleaseChildren() { gNodeGc.AddToGcQueue(std::move(child_)); }
@@ -322,12 +305,66 @@ void NodeTree::MakeMove(Move move) {
   history_.Append(move);
 }
 
+// This function examines the tree to recalculate the already searched depth. At
+// first glance this might seem excessive, but, even aside from the memory
+// savings of not caching depths in Nodes, this is actually in sum a
+// computational savings as well. The simple reason is that for the old per-Node
+// depths, they are recalculated d times for *every* node, where d is the depth.
+// This recalculates *once* per node (and only the reused nodes at that), so
+// this replaces O(N log(N)) ops with 2*O(N) ops (N is the number of nodes,
+// d ~ log(N)). And, as if that wasn't cool enough, this actually moves some of
+// the computation out of the otherwise-critical search code/cpu time, and moves
+// it into the pre-search initialization which doesn't count against move-search
+// time. The only drawback is that this computational effort is changed from
+// spread-over-the-search to all-at-once-in-initializtion, but it's a very
+// simple and efficient loop, and even with dozens of millions of nodes, only
+// takes on the order of milliseconds (and UCI sends an `isready` command
+// between `position` and `go` anyways).
+void NodeTree::RecalculateDepth() {
+  // This is the iterative translation of the most obvious and simple recursive
+  // implementation of a traversal of the tree. Basically the wo choices for
+  // this are depth first and breadth first; the former is far more obvious from
+  // a recursive standpoint, and also has the advantage that the intermediate
+  // storage need only be logarithmically long relative to the number of nodes.
+  // (The latter, using a queue rather than a stack, would have an intermediate
+  // length much closer to linear in the number of nodes rather than
+  // logarithmic.) The recursive implementation looks like this:
+  // void RecalculateDepth(node, depth) {
+  //   Process(node, depth);
+  //   for (child : node) {
+  //     RecalculateDepth(child, depth+1);
+  //   }
+  // }
+  std::stack<Node*> stack;
+  cumulative_depth_ = max_depth_ = 0;
+
+  stack.push(current_head_);
+
+  while (!stack.empty()) {
+    auto& node = stack.top(); // stack::top returns a reference!
+    if (node) {
+      if (node->GetN() > 0) {
+        auto depth = stack.size() - 1;
+        cumulative_depth_ += depth;
+        if (max_depth_ < depth) max_depth_ = depth;
+
+        stack.push(node->child_.get());
+      }
+      node = node->sibling_.get(); // Modifies the stack entry inplace!
+    } else {
+      stack.pop();
+    }
+  }
+}
+
 void NodeTree::TrimTreeAtHead() {
   auto tmp = std::move(current_head_->sibling_);
   // Send dependent nodes for GC instead of destroying them immediately.
   gNodeGc.AddToGcQueue(std::move(current_head_->child_));
   *current_head_ = Node(current_head_->GetParent(), current_head_->index_);
   current_head_->sibling_ = std::move(tmp);
+
+  cumulative_depth_ = max_depth_ = 0;
 }
 
 void NodeTree::ResetToPosition(const std::string& starting_fen,
@@ -361,6 +398,8 @@ void NodeTree::ResetToPosition(const std::string& starting_fen,
   if (!seen_old_head) {
     assert(!current_head_->sibling_);
     TrimTreeAtHead();
+  } else {
+    RecalculateDepth();
   }
 }
 
@@ -370,6 +409,23 @@ void NodeTree::DeallocateTree() {
   gNodeGc.AddToGcQueue(std::move(gamebegin_node_));
   gamebegin_node_ = nullptr;
   current_head_ = nullptr;
+  cumulative_depth_ = max_depth_ = 0;
+}
+
+// Not thread safe!
+void NodeTree::UpdateDepth(uint16_t new_node_depth) {
+  if (new_node_depth > max_depth_) max_depth_ = new_node_depth;
+  cumulative_depth_ += new_node_depth;
+}
+
+uint16_t NodeTree::GetMaxDepth() const {
+  return max_depth_;
+}
+
+uint16_t NodeTree::GetAverageDepth() const {
+  float ret = static_cast<float>(cumulative_depth_) /
+              static_cast<float>(current_head_->n_ - 1); // Exclude root node.
+  return static_cast<uint16_t>(std::ceil(ret));
 }
 
 }  // namespace lczero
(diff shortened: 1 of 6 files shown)


dubslow:
```
bill@Gravemind ~/leela/lc0-new/build/debug $ ninja
[2/2] Linking target lc0.
bill@Gravemind ~/leela/lc0-new/build/debug $ ./lc0 --minibatch-size=1 --max-prefetch=0 --backend=blas -t 1 --verbose-move-stats -w ../../../weights/id483-2daee779 --fpu-reduction=0 --policy-softmax-temp=1 --cpuct=4.8 --no-smart-pruning
       _
|   _ | |
|_ |_ |_| built Jul 12 2018
Edge size 8 Node size 64 for total estimated bytes/Node of 344
quit
bill@Gravemind ~/leela/lc0-new/build/debug $ git stash
Saved working directory and index state WIP on master: 79de494 Use vs2015 compiler on appveyor to fix cuda and update cuda to latest (#163)
HEAD is now at 79de494 Use vs2015 compiler on appveyor to fix cuda and update cuda to latest (#163)
bill@Gravemind ~/leela/lc0-new/build/debug $ git checkout depth-re-redux 
Switched to branch 'depth-re-redux'
Your branch is up-to-date with 'mine/depth-re-redux'.
bill@Gravemind ~/leela/lc0-new/build/debug $ git stash pop
On branch depth-re-redux
Your branch is up-to-date with 'mine/depth-re-redux'.
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   ../../src/main.cc

no changes added to commit (use "git add" and/or "git commit -a")
Dropped refs/stash@{0} (aec85f561acc0ce85497a3d0d7d9208e7134182d)
bill@Gravemind ~/leela/lc0-new/build/debug $ ninja
[22/22] Linking target hashcat_test.
bill@Gravemind ~/leela/lc0-new/build/debug $ ./lc0 --minibatch-size=1 --max-prefetch=0 --backend=blas -t 1 --verbose-move-stats -w ../../../weights/id483-2daee779 --fpu-reduction=0 --policy-softmax-temp=1 --cpuct=4.8 --no-smart-pruning
       _
|   _ | |
|_ |_ |_| built Jul 12 2018
Edge size 8 Node size 56 for total estimated bytes/Node of 336
```

So although it's actually saving 8 bytes per node due to alignment issues, it still only saves 2% memory because the Edges dominate the Nodes.

Patch:

```bill@Gravemind ~/leela/lc0-new/build/debug $ git diff
diff --git a/src/main.cc b/src/main.cc
index 6888806..8678eec 100644
--- a/src/main.cc
+++ b/src/main.cc
@@ -20,12 +20,14 @@
 #include "engine.h"
 #include "selfplay/loop.h"
 #include "utils/commandline.h"
+#include "mcts/node.h"
 
 int main(int argc, const char** argv) {
   std::cerr << "       _" << std::endl;
   std::cerr << "|   _ | |" << std::endl;
   std::cerr << "|_ |_ |_| built " << __DATE__ << std::endl;
   using namespace lczero;
+  std::cerr << "Edge size " << sizeof(Edge) << " Node size " << sizeof(Node) << " for total estimated bytes/Node of " << sizeof(Node) + 35*sizeof(Edge) << std::endl;
   CommandLine::Init(argc, argv);
   CommandLine::RegisterMode("uci", "(default) Act as UCI engine");
   CommandLine::RegisterMode("selfplay", "Play games with itself");
```

src/mcts/node.h
```diff
@@ -428,13 +413,24 @@ class NodeTree {
...
   void DeallocateTree();
+
+  void RecalculateDepth(); // Brute force re-examines the entire tree
+
   // A node which to start search from.
   Node* current_head_ = nullptr;
   // Root node of a game tree.
   std::unique_ptr<Node> gamebegin_node_;
   PositionHistory history_;
+  uint16_t max_depth_ = 0;
```
mooskagh: Could you remind me why it is not in `Search`?

src/mcts/node.cc
```diff
@@ -318,12 +301,67 @@ void NodeTree::MakeMove(Move move) {
...
+  //   process node at depth
+  //   for each child of node:
+  //     RecalculateDepth(child, depth+1)
+  // }
+  std::stack<Node*> stack;
+  cumulative_depth_ = max_depth_ = 0;
+
+  stack.push(current_head_);
+
+  while (!stack.empty()) {
```
mooskagh: Traversing the whole tree just to recalculate depth? No, that's not a way to go.
Just reset all records to zero at the beginning of the search and that's it.

dubslow: It's less work than is done in current code, and it's not in any time managed code, and takes milliseconds. Why not? See my stream, where I'm running this PR now.

I strongly believe that making the UCI printed depth ignore the `NodeTree` will lead to inconsistent and confusing output. It would mean that the depth printed for a given position analysis changes wildly depending on the order of commands entered, even if the exact nodes searched are literally identical.

src/mcts/node.h
dubslow: Because the max depth of the `Search` need not be the max depth of the `NodeTree`.

src/mcts/node.cc
dubslow: Here's another way to do it: if it's not okay for my code to do O(N) work in this PR, then why is it okay for your code *currently in master* to do O(N log N) work? This PR quite literally reduces the total operations relating to depth computation.

dubslow: I guess I'd be okay with adding an option to disable depth computation of reused trees, but I'd recommend for TCEC games that reused trees must be included, otherwise the output will be grossly inconsistent and confusing. All those 2-second-insta-moves due to smart pruning on reused trees will indicate a depth of ~0, which is obviously incorrect when 2M nodes are reported.

dubslow: As an example, from my test which is ongoing:

```
1. b3 book b6 book 
2. Bb2 book Bb7 book 
3. c4 +0.12/16 453s c5 -0.05/15 622s 
4. Nf3 +0.11/17 716s Nf6 -0.07/16 1179s 
5. e3 +0.14/18 749s e6 -0.05/18 1403s 
6. Be2 +0.15/17 163s Be7 -0.05/18 338s 
7. O⁠-⁠O +0.18/17 1076s O⁠-⁠O -0.07/18 1634s 
8. Nc3 +0.17/15 1284s d5 -0.08/19 1028s 
9. cxd5 +0.10/17 2726s Nxd5 -0.08/17 161s 
10. Nxd5 +0.10/16 75s Bxd5 -0.08/16 73s 
11. d4 +0.28/19 3098s Nd7 -0.08/20 3118s 
12. Rc1 +0.28/20 2567s h6 -0.08/19 186s 
13. Bc4 +0.40/19 2754s Be4 -0.14/18 2639s 
14. Qe2 +0.34/19 2251s Qc7 -0.14/17 232s 
15. Rfd1 +0.53/16 2308s Rad8 -0.13/17 315s 
16. dxc5 +0.49/18 2244s bxc5 -0.27/17 2237s 
17. h3 +0.59/16 2012s Nb6 -0.26/15 58s 
18. Be5 +0.63/21 2301s Qc8 -0.22/19 733s 
19. Ba6 +0.66/21 2281s Rxd1+ -0.25/20 774s 
20. Rxd1 +0.72/21 1339s Qc6 -0.26/19 166s 
21. Bb5 +0.75/21 617s Qb7 -0.28/19 484s 
22. Ba6 +0.77/21 514s Qc6 -0.32/19 621s 
23. Bg3 +0.84/22 2642s Qa8 -0.51/17 2725s 
24. Ne5 +0.88/22 2403s Rd8 -0.54/17 1228s 
25. Rxd8+ +0.86/23 745s Qxd8 -0.69/21 2076s 
26. Qg4 +0.88/24 776s Bd5 -0.66/21 2455s 
27. e4 +1.08/27 2463s Ba8 -0.70/25 2533s 
28. Qh5 +1.08/25 26s Qe8 -0.69/23 63s 
29. Bb5 +1.03/25 1345s Qf8 -0.70/22 223s 
30. Qf3 +1.04/23 54s Bd6 -0.73/22 405s 
31. Qd3 +1.03/23 316s Bc7 -0.73/20 73s 
32. a4 +1.04/21 48s Qe7 -0.66/19 3348s 
33. Ba6 +1.19/23 1002s Qg5 -0.66/27 2379s 
34. Nxf7 +1.33/33 2275s ```

Notice how the depths are slightly different for each move, reflecting the shape of the tree (all moves are at fixed 2M nodes, smart pruning disabled)? Now notice the tree reuse that causes sometimes super fast moves (by comparison) at 10 times or greater faster than the longest move. Those tree reuse moves would have depths of 5 or 10 instead of 18-25, and we would no longer be able to compare the depths of each and every move.

Comment (dubslow) [reply src/mcts/node.cc]: Here's a recording of several large tree reuses in a row.

https://clips.twitch.tv/VenomousFunWormDoggo

**Comment by RedDenver** [comment]
Instead of traversing the tree, why not just keep a running average? It's only for informational purposes anyway, so having the average be off slightly after each move isn't a big deal. As search continues, it'll converge on the actual average.

File: src/mcts/node.cc (hunk 2 above)
Comment (dubslow) [reply src/mcts/node.cc]: Cscuile on Discord today confirmed that there is not any noticeable added time in starting searches, even with trees with several million nodes.

> Cscuile: @Dubslow I'm not noticing any significant differences.

> Dubslow: but it doesn't lag before starting its search? that's the part im worried about
Cscuile: Nope it doesnt for me

**Comment by mooskagh** [comment]
Ok, I fetched that PR and measured rescan time, it's 80ms per million nodes for me.
(232223us for 2.93M nodes)

That's really what author of changes should measure though, instead of asking around "do you feel any slowdown?".

**Comment by dubslow** [comment]
`while chess clock already ticking` is patently untrue.

I will not support option B for this reason: https://github.com/LeelaChessZero/lc0/issues/86#issuecomment-408179482

I would be okay with adding +2 bytes per node to implement the same depth as done here, since Edge-Node separation is now in place.

**Comment by RedDenver** [comment]
"Depth" is a concept that makes sense for A/B engines but is poorly defined for MCTS. I propose we just stop reporting "depth" since it's both misleading and only for user info.

**Comment by dubslow** [comment]
I disagree there. Depth is no less well defined for MCTS than for A/B engines. Both deal with searching at/around/on average "near" certain ply distance from root search, and both wander around the tree in a nonmonotonic way (proof: there are multiple AB engines I've seen that report nonmonotonic depths on CCLS stream, and besides it's easy to see that AB can back its search up as it classifies one branch at deeper depth then uses that to prune and search other branches at shallower depth).

**Comment by RedDenver** [comment]
"Depth" being poorly defined even for A/B engines makes my argument stronger. Let's stop reporting "depth".

**Comment by mooskagh** [comment]
```
`while chess clock already ticking` is patently untrue.
```
Did you check?


_Hint: I did. In cutechess, Arena and in lichess bot wrapper (timer runs everywhere, not waiting for readyok). I only don't understand why I have to spend time to check assumptions of your PR, and not you._

Just to conclude (again): option with rescanning tree before move will not go to the master.

dubslow:
I'm closing this for now to allow more community focus on the more diplomatic PR222. I will come back and reinvestigate the workings of cutechess specifically and uci more generally at some point.

PR #165, opened 2018-07-12
Actually sleep if no work happened.

Tilps:


Tilps:
I only tested this compiles nothing more.

This is a bugfix, but whether it actually is going to fix anything, or may actually result in a regression in nps, is less clear.

src/mcts/search.cc
```diff
@@ -920,8 +920,17 @@ void SearchWorker::UpdateCounters() {
...
   search_->MaybeTriggerStop();
 
-  if (nodes_to_process_.empty()) {
-    // If this thread had no work, sleep for some milliseconds.
+  // If this thread had no work, sleep for some milliseconds.
+  // Collisions don't count as work, so have to enumerate to find out if there
+  // was anything done.
+  bool work_done = false;
+  for (NodeToProcess& node_to_process : nodes_to_process_) {
+    if (!node_to_process.is_collision) {
```
mooskagh: Maybe just `if (is_collision) return`?

PR #166, opened 2018-07-12
gzipped protobuf compressed neural networks

Error323:
Tested on testserver network 10068 `lc0 -w 10068.pb.gz --verbose-move-stats --threads=1`

```
32bit txt
info string c2c4  (264 ) N:    1239 (+ 0) (P: 12.50%) (Q:  0.02736) (U: 0.03301) (Q+U:  0.06037) (V: -0.0142) 
info string d2d4  (293 ) N:    1451 (+ 0) (P:  9.75%) (Q:  0.03747) (U: 0.02198) (Q+U:  0.05945) (V: -0.0142) 
info string e2e4  (322 ) N:    3001 (+ 0) (P: 14.43%) (Q:  0.04315) (U: 0.01574) (Q+U:  0.05889) (V: -0.0142) 
bestmove e2e4

16bit protobuf
info string c2c4  (264 ) N:    1234 (+ 0) (P: 12.50%) (Q:  0.02733) (U: 0.03314) (Q+U:  0.06047) (V: -0.0142) 
info string d2d4  (293 ) N:    1431 (+ 0) (P:  9.75%) (Q:  0.03749) (U: 0.02229) (Q+U:  0.05978) (V: -0.0142) 
info string e2e4  (322 ) N:    3024 (+ 0) (P: 14.43%) (Q:  0.04306) (U: 0.01562) (Q+U:  0.05868) (V: -0.0142) 
bestmove e2e4
```

Loads network ~10x faster


DanielUranga:
Need to install the Protobuf dependency on both CircleCI and Appveyor, I volunteer to do it tonight/tomorrow.

src/version.inc
```diff
@@ -1,3 +1,3 @@
-#define LC0_VERSION_MAJOR 0
-#define LC0_VERSION_MINOR 14
-#define LC0_VERSION_PATCH 1
+#define LC0_VERSION_MAJOR 1
+#define LC0_VERSION_MINOR 0
+#define LC0_VERSION_PATCH 0
```
GBeauregard: can this be in a separate PR/commit?

Error323:
Sounds good @DanielUranga. Note that submodules also need initialization now.

mooskagh:
Protobuf can/should be taken from wrapdb, as we do with gtest/zlib.
https://wrapdb.mesonbuild.com/

No need to install them separately in the system.

src/version.inc
dubslow: Agreed, versions should always be separate commits. And besides that, any reason you're bumping to 1.0.0 instead of 0.15.0? We're very far from being even the slightest bit stable, imo, certainly from that stability which is implied by a 1.0.0

borg323:
@DanielUranga you can find my first effort for an Appveyor build here: <https://github.com/borg323/lc0/commits/protobuf>. I tried to use a meson submodule as a fallback for when protobuf is not installed, but it eventually fails with a compilation error. So probably your idea of installing protobuf is better, at least for Appveyor (and will probably also build faster). However I think some of the other parts of this will be useful and this will probably get you going on CircleCI quickly.

src/version.inc
Error323: I was being optimistic ;)

src/neural/loader.cc
```diff
@@ -46,19 +50,83 @@ FloatVectors LoadFloatsFromFile(const std::string& filename) {
...
+  for (int i = 0; i < n; i++) {
+    v[i] = data[i] / float(0xffff);
+    v[i] *= layer.max_val() - layer.min_val();
+    v[i] += layer.min_val();
+  }
+  vecs.emplace_back(v);
+}
+
+void DenormConvBlock(const lc0::Weights_ConvBlock &conv, FloatVectors &vecs) {
+  DenormLayer(conv.bn_stddivs(), vecs);
```
Tilps: Shouldn't this order be reversed?  PopulateConvBlockWeights expects to pop off the end in this order, so we should pop them on in reverse I think.

Error323: No :)

.clang-format
```diff
@@ -0,0 +1,117 @@
+---
+Language:        Cpp
+# BasedOnStyle:  Google
```
Tilps: Please use BasedOnStyle: Google - then specify the overriding additionals rather than specifying the entire default style. As new versions of clang-format come out we don't want to have to update this.

src/neural/loader.cc
```diff
@@ -46,17 +50,81 @@ FloatVectors LoadFloatsFromFile(const std::string& filename) {
...
+  for (int i = 0; i < n; i++) {
+    v[i] = data[i] / float(0xffff);
+    v[i] *= layer.max_val() - layer.min_val();
+    v[i] += layer.min_val();
+  }
+  vecs.emplace_back(v);
+}
+
+void DenormConvBlock(const lc0::Weights_ConvBlock& conv, FloatVectors& vecs) {
+  DenormLayer(conv.bn_stddivs(), vecs);
```
Tilps: Just noting again that these appear to be in the wrong order. In case you missed my previous PR comment because I accidentally applied it to an old commit.

.clang-format
```diff
@@ -0,0 +1,117 @@
+---
+Language:        Cpp
+BasedOnStyle:  Google
+AccessModifierOffset: -1
```
Tilps: I meant to delete all of the lines that were already a match with google style, not just uncomment this line.

src/neural/loader.cc
```diff
@@ -46,17 +50,81 @@ FloatVectors LoadFloatsFromFile(const std::string& filename) {
...
+  DenormLayer(conv.biases(), vecs);
+  DenormLayer(conv.weights(), vecs);
+}
+
+FloatVectors LoadFloatsFromPbFile(const std::string& filename) {
+  auto buffer = DecompressGzip(filename);
+  auto net = lc0::Net();
+  FloatVectors vecs;
+  net.ParseFromString(buffer);
+
```
Tilps: Should throw exception as unsupported if net.format is not linear16.

src/neural/loader.cc
```diff
@@ -91,12 +160,13 @@ void PopulateConvBlockWeights(FloatVectors* vecs, Weights::ConvBlock* block) {
...
 Weights LoadWeightsFromFile(const std::string& filename) {
-  FloatVectors vecs = LoadFloatsFromFile(filename);
+  FloatVectors vecs;
+  auto buffer = DecompressGzip(filename);
 
-  if (vecs.size() <= 19)
-    throw Exception("Weights file " + filename +
-                    " should have at least 19 lines");
-  if (vecs[0][0] != 2) throw Exception("Weights version 2 expected");
+  if (buffer[0] == '2' && buffer[1] == '\n')
```
Tilps: Crem suggested treating 1 and \n as an error case - in case of v1 weights.

src/neural/loader.h
```diff
@@ -28,12 +28,18 @@ namespace lczero {
 using FloatVector = std::vector<float>;
 using FloatVectors = std::vector<FloatVector>;
 
+// Read from protobuf.
+FloatVectors LoadFloatsFromPbFile(const std::string& buffer);
+
 // Read space separated file of floats and return it as a vector of vectors.
-FloatVectors LoadFloatsFromFile(const std::string& filename);
+FloatVectors LoadFloatsFromFile(std::string& buffer);
```
Tilps: Google style is to use pointers instead of mutable references.

DanielUranga:
Please review https://github.com/LeelaChessZero/lc0/pull/172
I'm getting this error:
```In file included from lc0@exe/lc0net.pb.cc:5:
lc0@exe/lc0net.pb.h:17:2: error: This file was generated by an older version of protoc which is
#error This file was generated by an older version of protoc which is
 ^
lc0@exe/lc0net.pb.h:18:2: error: incompatible with your Protocol Buffer headers.  Please
#error incompatible with your Protocol Buffer headers.  Please
 ^
lc0@exe/lc0net.pb.h:19:2: error: regenerate this file with a newer version of protoc.```

See https://circleci.com/gh/DanielUranga/lc0/55

File: meson.build [hunk 9]
```diff
@@ -39,6 +39,18 @@ files = []
...
+if not protoc.found()
+  message('protoc will be built from the subproject')
+  protoc = subproject('protobuf').get_variable('protoc')
+endif
+
+gen = generator(protoc, \
+  output    : ['@BASENAME@.pb.cc', '@BASENAME@.pb.h'],
+  arguments : ['--proto_path=@CURRENT_SOURCE_DIR@/libs/lczero-common', '--cpp_out=@BUILD_DIR@', '@INPUT@'])
+
+files += gen.process('libs/lczero-common/lc0net.proto')
```
mooskagh: Is `lc0net.proto` located in a root of `lczero-common` repo?
It's better to move it into proto/ subdir, as there may be other things in lczero-common.

Also, maybe it's better to generate it that way so they are also included like:
```cpp
#include "proto/lc0.pb.h"
```
?

I like that more but maybe it's an overkill as we don't expect large tree of protos..
But if you'd like to do that, then:
1. Don't modify --proto-path   (keep it ending at `/libs/lczero-common` rather than `/libs/lczero-common/proto`)
2. in gen.process add parameter:
```
  files += gen.process('libs/lczero-common/proto/lc0net.proto',
  preserve_path_from : meson.current_source_dir() + '/libs/lczero-common/
)

File: meson.build [hunk 10]
```diff
@@ -39,6 +39,18 @@ files = []
...
 has_backends = false
 
+deps += dependency('protobuf', fallback: ['protobuf', 'protobuf_dep'], required: true)
+protoc = find_program('protoc', required: false)
+if not protoc.found()
+  message('protoc will be built from the subproject')
+  protoc = subproject('protobuf').get_variable('protoc')
+endif
+
+gen = generator(protoc, \
```
mooskagh: Is the backslash at the end of the file really needed?

src/neural/loader.cc
```diff
@@ -46,25 +50,89 @@ FloatVectors LoadFloatsFromFile(const std::string& filename) {
...
-      buffer.push_back('\n');
       break;
     }
   }
   gzclose(file);
 
+  return buffer;
+}
+
+void DenormLayer(const lc0::Weights_Layer& layer, FloatVectors& vecs) {
```
mooskagh: Everything is in namespace `lczero`, maybe move proto also there rather than `lc0`?

Style guide doesn't allow non-const references! :)
Could you replace it with pointers?

src/neural/loader.cc
```diff
@@ -91,12 +160,15 @@ void PopulateConvBlockWeights(FloatVectors* vecs, Weights::ConvBlock* block) {
 }  // namespace
 
 Weights LoadWeightsFromFile(const std::string& filename) {
-  FloatVectors vecs = LoadFloatsFromFile(filename);
+  FloatVectors vecs;
+  auto buffer = DecompressGzip(filename);
```
mooskagh: Check for file size of at least 2.

src/neural/loader.cc
```diff
@@ -46,25 +50,89 @@ FloatVectors LoadFloatsFromFile(const std::string& filename) {
...
+  DenormConvBlock(w.value(), vecs);
+  DenormLayer(w.ip1_val_w(), vecs);
+  DenormLayer(w.ip1_val_b(), vecs);
+  DenormLayer(w.ip2_val_w(), vecs);
+  DenormLayer(w.ip2_val_b(), vecs);
+
+  return vecs;
+}
+
+FloatVectors LoadFloatsFromFile(std::string* buffer) {
```
mooskagh: It's better to make buffer parameter a const reference.
It's a really weird side effect of a function that it adds '\n'.

Probably better to remove that and add after the loop:
```cpp
  if (!line.empty()) {
       result.emplace_back();
       result.back().swap(line);
  }
```

src/neural/loader.cc
```diff
@@ -46,25 +50,89 @@ FloatVectors LoadFloatsFromFile(const std::string& filename) {
...
+    throw Exception("Weights require at least lc0 version: " + min_version);
+  if (net.min_version().minor() > LC0_VERSION_MINOR)
+    throw Exception("Weights require at least lc0 version: " + min_version);
+  if (net.min_version().patch() > LC0_VERSION_PATCH)
+    throw Exception("Weights require at least lc0 version: " + min_version);
+
+  if (net.format().weights_encoding() != lc0::Format::LINEAR16)
+    throw Exception("Invalid weight encoding");
+
+  auto& w = net.weights();
```
mooskagh: `const auto&`
It's const anyway, but I think it's useful to be explicit.

src/neural/loader.cc
```diff
@@ -46,25 +50,89 @@ FloatVectors LoadFloatsFromFile(const std::string& filename) {
...
+
+  DenormConvBlock(w.input(), vecs);
```
mooskagh: If `DenormConvBlock()` returned `FloatVector`, then I think
```cpp
  vecs.emplace_back(DenormConvBlock(w.input());
```
would move it magically without copying the memory and would be more readable.

src/neural/loader.h
```diff
@@ -28,12 +28,18 @@ namespace lczero {
...
+
 // Read space separated file of floats and return it as a vector of vectors.
-FloatVectors LoadFloatsFromFile(const std::string& filename);
+FloatVectors LoadFloatsFromFile(std::string* buffer);
 
 // Read v2 weights file and fill the weights structure.
 Weights LoadWeightsFromFile(const std::string& filename);
 
+// Read gzip compressed data into buffer.
+std::string DecompressGzip(const std::string& filename);
```
mooskagh: I think this function (as well as `LoadFloats*` functions) are not really used outside of loader.

In that case I suggest removing them from .h file and putting into anonymous namespace in .cc file.

DanielUranga:
CircleCI is passing on https://github.com/LeelaChessZero/lc0/pull/172, lots of warnings though.

src/neural/loader.cc
Error323: DenormConvBlock fills 4 layers of FloatVectors. So I don't see how this would work. DenormLayer() could certainly return a FloatVector.

src/neural/loader.cc
Error323: Where exactly? Check if the buffer is >= 2? Or the gzip file itself?

Tilps: I think you can just check the buffer size.  If the file itself is less than 2 bytes, gzopen is going to just assume its not compressed and give you a <2 byte buffer.

meson.build
Error323: Agreed, that would be nicer.

src/neural/loader.cc
mooskagh: Indeed, checking buffer size is what's needed.

src/neural/loader.cc
mooskagh: Indeed. Then just converting non-const reference parameter to pointer is what's needed.

src/neural/loader.cc
Error323: If I move proto there, it will clash with current lczero::Weights definition. So keeping as is for now.

src/neural/loader.cc
Error323: I don't see an easy fix for this. LoadFloatsFromFile is altering the buffer in various ways.

mooskagh: Which ways? I only see it adds `"\n"` to the end of the buffer and doesn't do anything else.
4 lines that I wrote in my previous comment is everything that is needed to remove that.
Just remove that `+= "\n"`, and add those 4 lines after `for` loop, and I think it will work.

src/neural/loader.cc
```diff
@@ -24,20 +24,35 @@
...
 #include <sstream>
+#include <string>
 #include "utils/commandline.h"
 #include "utils/exception.h"
 #include "utils/filesystem.h"
+#include "proto/net.pb.h"
+#include "version.inc"
 
-namespace lczero {
+namespace {
```
mooskagh: anonymous namespace should be inside named:
```cpp
namespace {
namespace lczero {

// functions

} // namespace

// public functions

} // namespace lczero
```

That way they don't have to explicitly do `lczero::` everywhere.

src/neural/loader.cc
```diff
@@ -126,13 +201,13 @@ Weights LoadWeightsFromFile(const std::string& filename) {
 }
 
 std::string DiscoveryWeightsFile() {
-  const int kMinFileSize = 10000000;  // 10 MB
+  const int kMinFileSize = 5000000;  // 5 MB
```
mooskagh: I think that there may be some experimental networks like (32x2), which may compress well.
Maybe reduce that further?.. 1MB?.. What is the compressed size of only policy+value head without residual tower weights?

src/neural/loader.cc
Error323: It also adds \0 back into the buffer in some cases I guess.

mooskagh: Indeed! Fine to keep it then.

PR #169, opened 2018-07-13
Fix two minor bugs in best_move_edge_ tracking

### LINKED ISSUE #167: PV sometimes incorrect with tree reuse
While doing some deep analysis of several positions, I've noticed that lc0 will sometimes show an incorrect PV under the following conditions:

1) A search from some position is started
2) After some time, the analysis is stopped, and lc0's top move is played
3) From the resulting position, analysis is restarted

Several times in these conditions, I've seen the PV show a first move that was clearly not the top move at any point in the current search.

Here's an example of output from one such search:

```
 5/37	00:00	 5,190k	0	+0.01	Bf1-e2 c5xd4 e3xd4 Bc8-f5 c2-c3 e7-e6 Qd1-b3 Qd8-c8 Ng1-f3 Bf8-e7 Nb1-d2 O-O O-O Nb8-c6 Rf1-e1 h7-h6 h2-h3 Rf8-d8 a2-a4 Be7-d6
 5/37	00:05	 5,232k	8k	+0.01	Bf1-e2 c5xd4 e3xd4 Bc8-f5 c2-c3 e7-e6 Qd1-b3 Qd8-c8 Ng1-f3 Bf8-e7 Nb1-d2 O-O O-O Nb8-c6 Rf1-e1 h7-h6 h2-h3 Rf8-d8 a2-a4 Be7-d6
 5/37	00:10	 5,274k	8k	+0.01	Bf1-e2 c5xd4 e3xd4 Bc8-f5 c2-c3 e7-e6 Ng1-f3 Bf8-d6 Bf4xd6 Qd8xd6 Nb1-d2 h7-h6 O-O O-O Rf1-e1 Qd6-b6 Qd1-b3 Qb6-c7 a2-a4 Nb8-c6
 5/37	00:12	 5,298k	9k	+0.01	Bf1-e2 c5xd4 e3xd4 Bc8-f5 c2-c3 e7-e6 Ng1-f3 Bf8-d6 Bf4xd6 Qd8xd6 Nb1-d2 h7-h6 O-O O-O Rf1-e1 Qd6-b6 Qd1-b3 Qb6-c7 a2-a4 Nb8-c6
f1e2 (130 ) N: 817009 (+264) (P: 2.91%) (Q: 0.00275) (U: 0.00028) (Q+U: 0.00302) (V: -.----) 
c2c3 (259 ) N: 856749 (+ 0) (P: 16.04%) (Q: -0.00077) (U: 0.00147) (Q+U: 0.00070) (V: -.----) 
g1f3 (159 ) N: 3150189 (+ 0) (P: 11.83%) (Q: 0.00032) (U: 0.00029) (Q+U: 0.00061) (V: -.----) 
```

Every PV showed Be2, even though Nf3 has clearly always been the top move in this search (it has 2.3 million more visits than the next two moves, while the current search only incremented visits by about 100,000).

lc0 will still play the correct top move; it seems to be just the PV display that is incorrect.

When I posted this example in Discord, @killerducky noted that the stopped analysis in the previous position was likely mostly exploring Be2 (note its substantially higher Q and Q+U), so that may well be relevant.

I originally saw the behavior from lc0-win-20180701-cuda92-cudnn714 (which @mooskagh correctly pointed out in Discord was rather old by now), and then reproduced with lc0-win-20180711-cuda92-cudnn71 (the example above was from the more recent build).

dubslow:
These two were visible in the UCI PV in certain situtations, where the putative PV differed from what was later reported as the best move.
This commit fixes #167, which is the corresponding bug report.

Both bugs arose from circumstances where best_move_edge_ differed from what GetBestChildNoTemperature(root_node_) would report.

One such circumstance was that, after tree reuse, when starting a second Search, best_move_edge_ was null, which means if the subsequent search failed to playout the previous/actual best node (from the reused tree), then by the updates in DoBackupNode, best_move_edge_ would be incorrectly set to recently searched nodes, which didn't include the previous most searched node, and thus disagree with GBCNT.
(Such a circumstance can arise either by contrived FPUR hacks, or in certain super long tactical searches, each as reported in the issue by apleasantillusion.)

The other circumstance was from the updates in DoBackupUpdates: if the "best" move had the *same* number of visits as the "current" move, no further updating was done, which meant that if the "current" move had better tiebreakers in GBCNT, it would be chosen by GBCNT even though best_move_edge_ wasn't updated to "current".

The former circumstance could lead to incorrect search behavior (especially relating to smart pruning), and the latter technically could, though I expect it makes no practical difference, and frankly even the former circumstance is quite rare anyways. Even so, it's best to correct these behaviors.

The implementation of fixes is up for debate. The tiebreaker updates in DBU could be simplified to always use GBCNT, even without temperature, but that's more CPU time (but is it trivial cpu time in exchange for cleaner code?). It might also be possible to optimize away certain boolean checks of best_move_edge_ as well, though for now I decided that fewer changes is better. LMKWYT.

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -401,6 +401,7 @@ EdgeAndNode Search::GetBestChildNoTemperature(Node* parent) const {
 // Returns a child chosen according to weighted-by-temperature visit count.
 EdgeAndNode Search::GetBestChildWithTemperature(Node* parent,
                                                 float temperature) const {
+  assert(parent->GetN() > 1);
   std::vector<float> cumulative_sums;
   float sum = 0.0;
   const float n_parent = parent->GetN();
@@ -573,8 +574,7 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend() {
   SharedMutex::Lock lock(search_->nodes_mutex_);
 
   // Fetch the current best root node visits for possible smart pruning.
-  int best_node_n = 0;
-  if (search_->best_move_edge_) best_node_n = search_->best_move_edge_.GetN();
+  int best_node_n = search_->best_move_edge_.GetN();
 
   // True on first iteration, false as we dive deeper.
   bool is_root_node = true;
@@ -901,12 +901,10 @@ void SearchWorker::DoBackupUpdate() {
       if (full_depth_updated)
         full_depth_updated = n->UpdateFullDepth(&cur_full_depth);
       // Best move.
-      if (n->GetParent() == search_->root_node_) {
-        if (!search_->best_move_edge_ ||
-            search_->best_move_edge_.GetN() < n->GetN()) {
-          search_->best_move_edge_ =
-              EdgeAndNode(search_->root_node_->GetEdgeToNode(n), n);
-        }
+      if (n->GetParent() == search_->root_node_ &&
+          search_->best_move_edge_.GetN() <= n->GetN()) {
+            search_->best_move_edge_ =
+                search_->GetBestChildNoTemperature(search_->root_node_);
       }
     }
     ++search_->total_playouts_;
--- a/src/mcts/search.h
+++ b/src/mcts/search.h
@@ -101,6 +101,8 @@ class Search {
   std::pair<Move, Move> GetBestMoveInternal() const;
 
   // Returns a child with most visits, with or without temperature.
+  // NoTemperature is safe to use on non-extended nodes, while WithTemperature
+  // accepts only nodes with at least 1 visited child.
   EdgeAndNode GetBestChildNoTemperature(Node* parent) const;
   EdgeAndNode GetBestChildWithTemperature(Node* parent,
                                           float temperature) const;


src/mcts/search.cc
```diff
@@ -902,10 +905,15 @@ void SearchWorker::DoBackupUpdate() {
         full_depth_updated = n->UpdateFullDepth(&cur_full_depth);
       // Best move.
       if (n->GetParent() == search_->root_node_) {
-        if (!search_->best_move_edge_ ||
-            search_->best_move_edge_.GetN() < n->GetN()) {
+        auto best_n = search_->best_move_edge_.GetN();
+        auto this_n = n->GetN();
+        if (!search_->best_move_edge_ || best_n < this_n) {
```
Tilps: This logic was written this way originally I suspect because best_move_node was a pointer, now that best_move_edge_ is a struct its GetN will be 0, and so this logic can be simplified?

dubslow: Yes that's what I wondered in the last sentence of my commit message, now that it's initialized by the constructor, it should never be null-equivalent any more. But there are several such checks around the code, and like I said, I'm not willing to claim that removing those checks won't introduce further bugs somehow, not at the moment at least. (That is to say, I think it's more likely to introduce bugs than my currently waiting depth PR.)

What we *really* ought to do is get a proper test suite operational, like I opined about way back when on Discord (around the time of the overfitting debacle I think). Unfortunately, for now, test pipe promo takes precedence.

Tilps: My comment was actually that even if it is null equivalent, GetN() returns 0, so at worst it will fall in to the GetBestChildNoTemperature path.

dubslow: Ah, now I understand. I'll think about it (eventually).

dubslow:
With prodding from Tilps, I further simplified away two boolean checks for `best_move_edge_` which were rendered unnecessary after the edge node split, since `best_move_edge_` is now a EdgeAndNode object (not Node* pointer) which reports 0 visits even for an uninitialized object. (Other such boolean checks aren't merely for visit counts, and so are left as is.)

I also added a comment to GBCNT which mentions that it's suitable for non-populated nodes (unlike GBCWT), as necessary for its new use in the `Search` constructor.

src/mcts/search.cc
```diff
@@ -401,6 +404,7 @@ EdgeAndNode Search::GetBestChildNoTemperature(Node* parent) const {
 // Returns a child chosen according to weighted-by-temperature visit count.
 EdgeAndNode Search::GetBestChildWithTemperature(Node* parent,
                                                 float temperature) const {
+  assert(parent->GetN() > 1); // Require at least one visited child
```
mooskagh: Maybe `parent->GetChildrenVisits() >= 1`?  It's the same, but function name matches comment better.
Also comment in that case is not needed (as it just repeats what statement says), but if you want to keep the comment, then period at the end of sentence. :)

dubslow: In theory, yes, but in practice, per my comment on Discord, `GetChildrenVisits` returns `1` when `n_` is 0, so is not suitable for this purpose.

I suspect there are many other such uses of `GetN` scattered about the codebase that can be so changed, only after we fix GetChildrenVisits, which I think is inappropriate for *this* PR. (There are several other minor/inconsequential blemishes I've been thinking about touching up, and fixing GCV is one of many things I'll do all at once.)

mooskagh: Yes, that GetChildrenVisits() returning 1 looks like a bug, so maybe also fix it in this PR?

All places calling that function seems to expect that it returns 0 sometimes.

dubslow:
Okay, since the two commits are really unrelated, can I be allowed to merge this as a rebase rather than squash?

src/mcts/search.cc
```diff
@@ -111,7 +111,10 @@ Search::Search(const NodeTree& tree, Network* network,
       kPolicySoftmaxTemp(options.Get<float>(kPolicySoftmaxTempStr)),
       kAllowedNodeCollisions(options.Get<int>(kAllowedNodeCollisionsStr)),
       kBackPropagateBeta(options.Get<float>(kBackPropagateBetaStr)),
-      kBackPropagateGamma(options.Get<float>(kBackPropagateGammaStr)) {}
+      kBackPropagateGamma(options.Get<float>(kBackPropagateGammaStr)) {
+        // Depends on the backpropagate options, so we can't use an initializer.
```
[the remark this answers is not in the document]
dubslow: That would solve the second circumstance, but would leave the first circumstance uncorrected. After tree reuse, `best_move_edge_` must be initialized independent of the actual search -- we cannot assume that the search will ever actually examine the reused-tree-best-node. See the original bug report for an example of this kind of tree reuse.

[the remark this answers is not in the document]
dubslow: And yes, I initially proposed they should be separate PRs, but you wanted the change in this PR specifically. I would rather move it to a separate PR than see two unrelated changes be squashed into one commit.

(I would see my rebase suggestion less as a violation of merge policy, and more of a violation of one change per PR.)

PR #171, opened 2018-07-13
Reduce Move size from 4 bytes to 2 (take two)

AlexGreason:
Turns out that the reason for the extraneous queen promotions was because Mirror was using 0x instead of 0b for the bitmask, so it was setting the wrong bits

### DIFF
--- a/src/chess/bitboard.cc
+++ b/src/chess/bitboard.cc
@@ -275,26 +275,26 @@ const int kQueenCastleIndex =
 
 Move::Move(const std::string& str, bool black) {
   if (str.size() < 4) throw Exception("Bad move: " + str);
-  from_ = BoardSquare(str.substr(0, 2), black);
-  to_ = BoardSquare(str.substr(2, 2), black);
+  SetFrom(BoardSquare(str.substr(0, 2), black));
+  SetTo(BoardSquare(str.substr(2, 2), black));
   if (str.size() != 4) {
     if (str.size() != 5) throw Exception("Bad move: " + str);
     switch (str[4]) {
       case 'q':
       case 'Q':
-        promotion_ = Promotion::Queen;
+        SetPromotion(Promotion::Queen);
         break;
       case 'r':
       case 'R':
-        promotion_ = Promotion::Rook;
+        SetPromotion(Promotion::Rook);
         break;
       case 'b':
       case 'B':
-        promotion_ = Promotion::Bishop;
+        SetPromotion(Promotion::Bishop);
         break;
       case 'n':
       case 'N':
-        promotion_ = Promotion::Knight;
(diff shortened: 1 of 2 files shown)


src/chess/bitboard.cc
```diff
@@ -303,17 +303,17 @@ Move::Move(const std::string& str, bool black) {
...
-    return from_.as_int() * 64 + to_.as_int();
+  if (promotion() == Promotion::Knight) {
+    return from().as_int() * 64 + to().as_int();
   } else {
-    return static_cast<int>(promotion_) * 64 * 64 + from_.as_int() * 64 +
-           to_.as_int();
+    return static_cast<int>(promotion()) * 64 * 64 + from().as_int() * 64 +
+           to().as_int();
   }
 }
```
RedDenver: Since the knight promotion is alone on bit 14 (0bxNxxxxxxxxxxxxxx, and none, queen, rook, and bishop are on bits 12 and 13) and the castling is on bit 15, I suggest:
```
uint16_t Move::as_packed_int() const {
  return data_ & 0b0011111111111111;
}
```

AlexGreason: There's another change which fully alters the as_packed_int and as_nn_index to just return data_, and also modifies BuildMoveIndecies accordingly, but I'd rather keep this change as a single PR and do the stuff modifiying those functions in a separate PR

RedDenver: Sounds good.
Will those changes require changes to the NN policy head output as well?

AlexGreason: No, the only changes needed are to bitboard.cc. It changes Build_move_indices so that kMoveToIdx[data_] returns the same thing as the current function, so that the NN sees the same inputs and returns the same outputs

src/chess/bitboard.h
```diff
@@ -190,40 +190,43 @@ class Move {
...
   uint16_t as_packed_int() const;
 
   // 0 .. 1857, to use in neural networks.
   uint16_t as_nn_index() const;
 
-  bool operator==(const Move& other) const {
-    return from_ == other.from_ && to_ == other.to_ &&
-           promotion_ == other.promotion_;
-  }
+  bool operator==(const Move& other) const { return data_ == other.data_; }
```
Tilps: Any idea why the old equality didn't compare the castling bit?

AlexGreason: no, but on the previous version of this PR Crem said "Indeed it should be data_ == other.data_, castlings are expected to be different moves."

dubslow:
He said "sounds good" in response to OP's "I'd rather do that in separate PR". I say merge.

RedDenver:
I'm new to github work flow, and I don't see a way to approve it.

PR #172, opened 2018-07-14
[CircleCI] Install protobuf dependency on the docker image

DanielUranga:
This is a PR to be applied on top of https://github.com/LeelaChessZero/lc0/pull/166


DanielUranga:
Will install a newer protobuf.

Tilps:
I did some searching about the CircleCI error here - which suggests we have a missmatch between proto runtime being used and protoc being used.  Specifically the suggestion was that protoc might be 3.5.1 or earlier and the runtime 'newer'.

Maybe downloading 3.6.0 instead of 3.5.1 would do the trick?

DanielUranga:
It was using the Ubuntu default one, which is 3.0.0. Using 3.5.1 now like the rest of the builds.

Issue #173, opened 2018-07-14
NPS slowdown issue in TCEC bonus games


killerducky:
Here is a log for TCEC of bonus games: http://paste.ubuntu.com/p/srK8KS7rq7/

The following section shows very low NPS (time 11s, nps 0 and time 14s, nps 18). In the end, g2g4 is picked with only 5822 visits after around 14s of thinking.

```
>position startpos moves d2d4 d7d5 c2c4 e7e6 b1c3 g8f6 c4d5 e6d5 c1g5 c7c6 e2e3 f8e7 f1d3 e8g8 d1c2 b8d7 g1e2 f8e8 e1c1 b7b5 h2h4 h7h6 g5f4 d8a5 c1b1 d7b6 e2c1 b6c4 d1g1 a5d8 f2f3 e7d6
>isready
<readyok
>go wtime 1328256 btime 1239809 winc 16000 binc 16000
<info depth 3 seldepth 16 time 11091 nodes 6719 score cp 23 hashfull 1000 nps 0 pv g2g4 d6f4 e3f4 d8d6 f4f5 a7a5 g4g5 f6h5 f5f6 a5a4 f6g7 a4a3 b2b3 c4e3 c2d2 a8a7
<info depth 3 seldepth 16 time 13975 nodes 6975 score cp 23 hashfull 1000 nps 18 pv g2g4 d6f4 e3f4 d8d6 f4f5 a7a5 g4g5 f6h5 f5f6 a5a4 f6g7 a4a3 b2b3 c4e3 c2d2 c8f5
<info string c2a4  (262 ) N:       1 (+ 0) (P:  0.64%) (Q: -0.90750) (U: 0.90263) (Q+U: -0.00487) (V:  -.----)
<info string h1h3  (183 ) N:       2 (+ 0) (P:  0.57%) (Q: -0.55255) (U: 0.54340) (Q+U: -0.00915) (V:  -.----)
<info string c2b3  (258 ) N:       3 (+ 0) (P:  0.75%) (Q: -0.69535) (U: 0.52919) (Q+U: -0.16616) (V:  -.----)
--- cut ---
<info string c3d1  (469 ) N:     220 (+ 0) (P:  7.46%) (Q: -0.04061) (U: 0.09580) (Q+U:  0.05519) (V:  -.----)
[...]

PR #175, opened 2018-07-14
Garbage collector spinlock

RedDenver:


mooskagh:
As I've written here: https://github.com/LeelaChessZero/lc0/issues/173#issuecomment-405048552, it's not a correct place to optimize.

PR #176, opened 2018-07-14
Change Edge::p_ from float to uint16_t

dubslow:
This saves 2 actual bytes, or 4 aligned bytes after the Move savings, resulting in roughly 40% savings in total tree size:
35\*8 + 64 = 344  bytes per node --> 35\*4 + 64 = 204 bytes per node

```
# with new policy

|   _ | |
|_ |_ |_| built Jul 14 2018
Move size 2 for estimated Edge size 4 vs actual Edge size 4 Node size 64 for total estimated bytes/Node of 204
go nodes 5
Creating backend [blas]...
BLAS vendor: OpenBlas.
OpenBlas [NO_LAPACKE DYNAMIC_ARCH NO_AFFINITY Sandybridge].
OpenBlas found 8 Sandybridge core(s).
OpenBLAS using 1 core(s) for this backend.
BLAS max batch size is 256.
info depth 2 seldepth 2 time 33 nodes 2 score cp 6 hashfull 0 nps 60 pv d2d4 d7d5
info depth 2 seldepth 3 time 60 nodes 4 score cp 5 hashfull 0 nps 66 pv d2d4 d7d5 c2c4
info depth 2 seldepth 3 time 74 nodes 5 score cp 5 hashfull 0 nps 67 pv d2d4 d7d5 c2c4
info string g2g4  (378 ) N:       0 (+ 0) (P:  0.36%) (Q:  0.01514) (U: 0.00861) (Q+U:  0.02375) (V:  -.----) 
info string f2f3  (346 ) N:       0 (+ 0) (P:  0.51%) (Q:  0.01514) (U: 0.01230) (Q+U:  0.02745) (V:  -.----) 
info string f2f4  (351 ) N:       0 (+ 0) (P:  0.56%) (Q:  0.01514) (U: 0.01333) (Q+U:  0.02847) (V:  -.----) 
info string g1h3  (161 ) N:       0 (+ 0) (P:  0.56%) (Q:  0.01514) (U: 0.01337) (Q+U:  0.02851) (V:  -.----) 
info string h2h4  (403 ) N:       0 (+ 0) (P:  0.63%) (Q:  0.01514) (U: 0.01512) (Q+U:  0.03027) (V:  -.----) 
info string b1a3  (34  ) N:       0 (+ 0) (P:  0.66%) (Q:  0.01514) (U: 0.01593) (Q+U:  0.03107) (V:  -.----) 
info string a2a4  (207 ) N:       0 (+ 0) (P:  1.11%) (Q:  0.01514) (U: 0.02670) (Q+U:  0.04184) (V:  -.----) 
info string b2b4  (234 ) N:       0 (+ 0) (P:  1.15%) (Q:  0.01514) (U: 0.02761) (Q+U:  0.04276) (V:  -.----) 
info string d2d3  (288 ) N:       0 (+ 0) (P:  1.28%) (Q:  0.01514) (U: 0.03076) (Q+U:  0.04591) (V:  -.----) 
info string g2g3  (374 ) N:       0 (+ 0) (P:  1.42%) (Q:  0.01514) (U: 0.03413) (Q+U:  0.04928) (V:  -.----) 
info string c2c3  (259 ) N:       0 (+ 0) (P:  1.55%) (Q:  0.01514) (U: 0.03721) (Q+U:  0.05235) (V:  -.----) 
info string a2a3  (204 ) N:       0 (+ 0) (P:  1.82%) (Q:  0.01514) (U: 0.04373) (Q+U:  0.05887) (V:  -.----) 
info string b2b3  (230 ) N:       0 (+ 0) (P:  1.96%) (Q:  0.01514) (U: 0.04699) (Q+U:  0.06213) (V:  -.----) 
info string b1c3  (36  ) N:       0 (+ 0) (P:  2.17%) (Q:  0.01514) (U: 0.05211) (Q+U:  0.06726) (V:  -.----) 
info string h2h3  (400 ) N:       0 (+ 0) (P:  3.26%) (Q:  0.01514) (U: 0.07826) (Q+U:  0.09340) (V:  -.----) 
info string e2e3  (317 ) N:       0 (+ 0) (P:  7.25%) (Q:  0.01514) (U: 0.17399) (Q+U:  0.18913) (V:  -.----) 
info string c2c4  (264 ) N:       0 (+ 0) (P:  8.88%) (Q:  0.01514) (U: 0.21317) (Q+U:  0.22832) (V:  -.----) 
info string e2e4  (322 ) N:       0 (+ 0) (P: 10.20%) (Q:  0.01514) (U: 0.24485) (Q+U:  0.26000) (V:  -.----) 
info string g1f3  (159 ) N:       1 (+ 0) (P: 19.11%) (Q:  0.02653) (U: 0.22934) (Q+U:  0.25587) (V:  0.0265) 
info string d2d4  (293 ) N:       3 (+ 0) (P: 35.52%) (Q:  0.01257) (U: 0.21311) (Q+U:  0.22568) (V:  0.0147) 
bestmove d2d4
quit

# with old policy
       _
|   _ | |
|_ |_ |_| built Jul 10 2018
go nodes 5
Creating backend [blas]...
BLAS vendor: OpenBlas.
OpenBlas [NO_LAPACKE DYNAMIC_ARCH NO_AFFINITY Sandybridge].
OpenBlas found 8 Sandybridge core(s).
OpenBLAS using 1 core(s) for this backend.
BLAS max batch size is 256.
info depth 1 seldepth 1 time 22 nodes 2 score cp 6 hashfull 0 nps 90 pv d2d4 d7d5
info depth 1 seldepth 2 time 43 nodes 4 score cp 5 hashfull 0 nps 93 pv d2d4 d7d5 c2c4
info depth 1 seldepth 2 time 52 nodes 5 score cp 5 hashfull 0 nps 96 pv d2d4 d7d5 c2c4
info string g2g4  (378 ) N:       0 (+ 0) (P:  0.36%) (Q:  0.01514) (U: 0.00864) (Q+U:  0.02378) (V:  -.----) 
info string f2f3  (346 ) N:       0 (+ 0) (P:  0.51%) (Q:  0.01514) (U: 0.01234) (Q+U:  0.02748) (V:  -.----) 
info string f2f4  (351 ) N:       0 (+ 0) (P:  0.56%) (Q:  0.01514) (U: 0.01337) (Q+U:  0.02852) (V:  -.----) 
info string g1h3  (161 ) N:       0 (+ 0) (P:  0.56%) (Q:  0.01514) (U: 0.01340) (Q+U:  0.02854) (V:  -.----) 
info string h2h4  (403 ) N:       0 (+ 0) (P:  0.63%) (Q:  0.01514) (U: 0.01516) (Q+U:  0.03030) (V:  -.----) 
info string b1a3  (34  ) N:       0 (+ 0) (P:  0.67%) (Q:  0.01514) (U: 0.01596) (Q+U:  0.03111) (V:  -.----) 
info string a2a4  (207 ) N:       0 (+ 0) (P:  1.11%) (Q:  0.01514) (U: 0.02673) (Q+U:  0.04188) (V:  -.----) 
info string b2b4  (234 ) N:       0 (+ 0) (P:  1.15%) (Q:  0.01514) (U: 0.02765) (Q+U:  0.04280) (V:  -.----) 
info string d2d3  (288 ) N:       0 (+ 0) (P:  1.28%) (Q:  0.01514) (U: 0.03080) (Q+U:  0.04594) (V:  -.----) 
info string g2g3  (374 ) N:       0 (+ 0) (P:  1.42%) (Q:  0.01514) (U: 0.03415) (Q+U:  0.04929) (V:  -.----) 
info string c2c3  (259 ) N:       0 (+ 0) (P:  1.55%) (Q:  0.01514) (U: 0.03727) (Q+U:  0.05241) (V:  -.----) 
info string a2a3  (204 ) N:       0 (+ 0) (P:  1.82%) (Q:  0.01514) (U: 0.04378) (Q+U:  0.05892) (V:  -.----) 
info string b2b3  (230 ) N:       0 (+ 0) (P:  1.96%) (Q:  0.01514) (U: 0.04699) (Q+U:  0.06213) (V:  -.----) 
info string b1c3  (36  ) N:       0 (+ 0) (P:  2.17%) (Q:  0.01514) (U: 0.05214) (Q+U:  0.06728) (V:  -.----) 
info string h2h3  (400 ) N:       0 (+ 0) (P:  3.26%) (Q:  0.01514) (U: 0.07829) (Q+U:  0.09344) (V:  -.----) 
info string e2e3  (317 ) N:       0 (+ 0) (P:  7.25%) (Q:  0.01514) (U: 0.17403) (Q+U:  0.18918) (V:  -.----) 
info string c2c4  (264 ) N:       0 (+ 0) (P:  8.88%) (Q:  0.01514) (U: 0.21322) (Q+U:  0.22836) (V:  -.----) 
info string e2e4  (322 ) N:       0 (+ 0) (P: 10.20%) (Q:  0.01514) (U: 0.24490) (Q+U:  0.26005) (V:  -.----) 
info string g1f3  (159 ) N:       1 (+ 0) (P: 19.11%) (Q:  0.02653) (U: 0.22935) (Q+U:  0.25588) (V:  0.0265) 
info string d2d4  (293 ) N:       3 (+ 0) (P: 35.52%) (Q:  0.01257) (U: 0.21312) (Q+U:  0.22569) (V:  0.0147) 
bestmove d2d4

```

Only one move has a different rounded decimal digit, all the rest are identical. Remains to be tested in gameplay. Should probably be tested with actual 16bit weights as well.

### DIFF
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -115,7 +115,7 @@ Move Edge::GetMove(bool as_opponent) const {
 
 std::string Edge::DebugString() const {
   std::ostringstream oss;
-  oss << "Move: " << move_.as_string() << " P:" << p_;
+  oss << "Move: " << move_.as_string() << " P: " << GetP();
   return oss.str();
 }
 
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -113,29 +113,36 @@ Move Edge::GetMove(bool as_opponent) const {
   return m;
 }
 
-// We use a 16 bit significand for storage savings in memory-critical Edges.
-// Non-zero priors are not allowed, as we use them multiplicatively to
-// calculate U. We have 16 bits, so 2^16 unique values available (we don't
-// reserve a value for "uninitialized"); we want them to be evenly
-// balanced around 1/2 (so that converting floats to ints preserves normality),
-// so we have to adjust the conversion slightly. For example, with 2 bits:
-// 0       0.25       0.5       0.75       1
-// |---------|---------|---------|---------|
-// A    B    C    D    E    F    G    H    I
-// In order to evenly balance the 4 available values, we need to actually put
-// them at 0.125, 0.375, 0.625, and 0.875, i.e. B, D, F, and H.
-// For converting floats to ints: just multiply by 2^(n=2) and truncate, which
-// maps [A,C) -> 0, [C,E) -> 1, [E,G) -> 2, [G,I) -> 3.
-// To convert these integers back to floats, divide by 2^(n=2), mapping 0 -> A,
-// 1 -> C, 2 -> E, 3 -> G. Recover the *balanced around 1/2* floats by adding
-// 0.125, i.e. 1/2^(n+1).
-float Edge::GetP() const {
-  return   static_cast<float>(p_) / static_cast<float>(1 << 16)
-         + static_cast<float>(1) / static_cast<float>(1 << 17);
+// We use strictly 16 bits to store policy values in memory-critical Edges.
+// We manage this by storing 16 bits taken directly from the float.
+// Floats are 1 sign bit, 8 bits of exponent, and 23 significand bits. The
+// exponent is stored a bit funny, in such a way that its most significant bit
+// is always off in the interval [0, 2).
+// See https://www.h-schmidt.net/FloatConverter/IEEE754.html for details.
+// Note that that means we don't actually get 2^16 unique values in [0,1],
+// rather some spread into (1,2), but (although it's possible) it's more
+// complicated to make use of the extra values. At any rate, it's true that the
+// available values are disproportionately clustered (and thus more precise)
+// towards 0, which is how NN policy behaves anyways (since policy always adds
+// up to 1). We also implement round-to-nearest by shifting the significand up
+// by half the precision that is ultimately kept (addition overflow into the 7
+// exponent bits is possible, but also works correctly).
+
+// Toss the sign and MSB of the exponent, take the next 16 bits, that leaves
+// the 14 trailing LSB of the significand to be lost as well. Round-to-nearest
+// by adding a 1 in that 14th bit before truncating.
+void Edge::SetP(float p) {
+  assert(0.0f <= p && p < 2.0f);
+  uint32_t tmp;
+  std::memcpy(&tmp, &p, sizeof(float));
+  p_ = static_cast<uint16_t>((tmp + (1 << 13)) >> 14);
 }
 
-void Edge::SetP(float val) {
-  p_ = static_cast<uint16_t>(val * (1 << 16));
+float Edge::GetP() const {
+  uint32_t tmp = static_cast<uint32_t>(p_) << 14;
+  float tmpf;
+  std::memcpy(&tmpf, &tmp, sizeof(uint32_t));
+  return tmpf;
 }
 
 std::string Edge::DebugString() const {
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -115,7 +115,7 @@ Move Edge::GetMove(bool as_opponent) const {
 
 std::string Edge::DebugString() const {
   std::ostringstream oss;
-  oss << "Move: " << move_.as_string() << " P: " << GetP();
+  oss << "Move: " << move_.as_string() << " p_: " << p_ << " GetP: " << GetP();
   return oss.str();
 }
 
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -113,36 +113,41 @@ Move Edge::GetMove(bool as_opponent) const {
   return m;
 }
 
-// We use strictly 16 bits to store policy values in memory-critical Edges.
-// We manage this by storing 16 bits taken directly from the float.
-// Floats are 1 sign bit, 8 bits of exponent, and 23 significand bits. The
-// exponent is stored a bit funny, in such a way that its most significant bit
-// is always off in the interval [0, 2). Furthermore, the next two bits are only
-// off for microscopic floats near 0; we ignore those and round up to the
-// smallest possible 5-bit-exp float above 0 (i.e. assume those 2 bits are set).
+// Policy priors (P) are stored in a compressed 16-bit format.
 //
+// Source values are 32-bit floats:
+// * bit 31 is sign (zero means positive)
+// * bit 30 is sign of exponent (zero means nonpositive)
+// * bits 29..23 are value bits of exponent
+// * bits 22..0 are significand bits (plus a "virtual" always-on bit: s ∈ [1,2))
+// The number is then sign * 2^exponent * significand, usually.
 // See https://www.h-schmidt.net/FloatConverter/IEEE754.html for details.
 //
-// Note that that means we don't actually get 2^16 unique values in [0,1],
-// rather some spread into (1,2), but (although it's possible) it's more
-// complicated to make use of the extra values. At any rate, it's true that the
-// available values are disproportionately clustered (and thus more precise)
-// towards 0, which is how NN policy behaves anyways (since policy always adds
-// up to 1). We also implement round-to-nearest by shifting the significand up
-// by half the precision that is ultimately kept (addition overflow into the 7
-// exponent bits is possible, but also works correctly).
+// In compressed 16-bit value we store bits 27..12:
+// * bit 31 is always off as values are always >= 0
+// * bit 30 is always off as values are always < 2
+// * bits 29..28 are only off for values < 4.6566e-10, assume they are always on
+// * bits 11..0 are for higher precision, they are dropped leaving only 11 bits
+//     of precision
+//
+// When converting to compressed format, bit 11 is added to in order to make it
+// a rounding rather than truncation.
+//
+// Out of 65556 possible values, 2047 are outside of [0,1] interval (they are in
+// interval (1,2)). This is fine because the values in [0,1] are skewed towards
+// 0, which is also exactly how the components of policy tend to behave (since
+// they add up to 1).
 
-// Toss the sign and 3 highest bits of the exponent (assuming them to be 011),
-// take the next 16 bits, that leaves the 12 trailing LSB of the significand to
-// be lost as well. Round-to-nearest by adding a 1 in that 12th bit before
-// truncating. The minimum number that can be represented is 4.656613E-10.
+// If the two assumed-on exponent bits (3<<28) are in fact off, the input is
+// rounded up to the smallest value with them on. We verify this by subtracting
+// the two bits from the input and checking for a negative result. This is
+// combined with the round-to-nearest addition (1<<11).
 void Edge::SetP(float p) {
+  constexpr int32_t roundings = (1 << 11) - (3 << 28);
   assert(0.0f <= p && p <= 1.0f);
   int32_t tmp;
   std::memcpy(&tmp, &p, sizeof(float));
-  // If those 2 exponent bits (3<<28) aren't set, round up by setting tmp = 0.
-  // This is combined with the round-to-nearest addition (1<<11) into one op.
-  tmp -= 0x2FFFF800; // == (3 << 28) - (1 << 11)
+  tmp += roundings;
   p_ = (tmp < 0) ? 0 : static_cast<uint16_t>(tmp >> 12);
 }
 
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -139,12 +139,13 @@ Move Edge::GetMove(bool as_opponent) const {
 // they add up to 1).
 
 // If the two assumed-on exponent bits (3<<28) are in fact off, the input is
-// rounded up to the smallest value with them on. We verify this by subtracting
-// the two bits from the input and checking for a negative result. This is
-// combined with the round-to-nearest addition (1<<11).
+// rounded up to the smallest value with them on. We accomplish this by
+// subtracting the two bits from the input and checking for a negative result
+// (the subtraction works despite crossing from exponent to significand). This
+// is combined with the round-to-nearest addition (1<<11) into one op.
 void Edge::SetP(float p) {
-  constexpr int32_t roundings = (1 << 11) - (3 << 28);
   assert(0.0f <= p && p <= 1.0f);
+  constexpr int32_t roundings = (1 << 11) - (3 << 28);
   int32_t tmp;
   std::memcpy(&tmp, &p, sizeof(float));
   tmp += roundings;
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -113,6 +113,31 @@ Move Edge::GetMove(bool as_opponent) const {
   return m;
 }
 
+// We use a 16 bit significand for storage savings in memory-critical Edges.
+// Non-zero priors are not allowed, as we use them multiplicatively to
+// calculate U. We have 16 bits, so 2^16 unique values available (we don't
+// reserve a value for "uninitialized"); we want them to be evenly
+// balanced around 1/2 (so that converting floats to ints preserves normality),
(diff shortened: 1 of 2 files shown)


src/mcts/node.cc
```diff
@@ -105,7 +105,7 @@ Move Edge::GetMove(bool as_opponent) const {
 
 std::string Edge::DebugString() const {
   std::ostringstream oss;
-  oss << "Move: " << move_.as_string() << " P:" << p_;
+  oss << "Move: " << move_.as_string() << " P: " << GetP();
```
mooskagh: Maybe output both int and float? It's only used for debug, may be useful to see both.

src/mcts/node.h
```diff
@@ -69,12 +70,14 @@ class Edge {
...
-
-  // Sets move probability.
-  void SetP(float val) { p_ = val; }
+  // Returns or sets value of Move policy prior returned from the neural net
+  // (but can be changed by adding Dirichlet noise). We use a 16 bit significand
+  // for storage savings in memory-critical Edges. Priors are floored at
+  // 1/0xFFFF, since we use priors multiplicatively to calculate U.
+  float GetP() const { return   static_cast<float>(p_)
+                              / static_cast<float>(0xFFFF); }
+  void SetP(float val) { p_ = std::max(static_cast<uint16_t>(1),
```
mooskagh: That way p_ can never be 0, kind of a waste of capacity.

What we would want is rounding up:
(example with if we had only 2 bits to store rather than 16)
```
0       0.25       0.5       0.75       1
|---------|---------|---------|---------|
A    B    C     D   E   F   G   H   I   J
```
ABC would be encoded as 0 (0.25 when converted back)
ED as 1  (0.5)
FG as 2 (0.75)
HIJ as 3 (1.0)

For that:
GetP should return ` (p_ + 1) / static_cast<float>(0x4)` (or 0x10000 in case of 16 bit).

SetP should round up. It's more compicated, but it's called much rarer, so probably something like:
```cpp
  auto x = ceil(val * 0x10000);
  p_ = x ? (x-1) : 0;
```

mooskagh: Hm... Or probably it's better those 4 bits to mean:
0.125,  0.375, 0.625, 0.875..
and round do the closest.
That way overall error will be less. (and also expected sum of probabilities will remain 1.0 which is good)

dubslow: I thought the conclusion on Discord was that it should not be possible for policy to be 0, because that would mess up the computation of `U` and thus significantly alter `Search` behavior.

mooskagh: Yes, with rounding up it's never 0.

dubslow: Ah, the other reason I had a value off limits for policy: how do we distinguish uninitialized-policy Edges from initialized-policy ones? Though I guess the code had never cared about that before (except for one case in PrefetchIntoCache), everywhere just assumes that policy is non-zero, perhaps we should change that?

Videodr0me: There is one thing to keep in mind here. Ideally, original order should be preserved. So if for example two moves are rounded to the same policy value., that would be suboptimal. So if a > b before rounding it should also be a > b after rounding. There is some Literature on this, and some even claim that order is more important than actual value. I would not go that far, but its something to keep in mind. There are algos which ensure that rounding does not change (strict) order, but i think thats would be clutter the codebase even more. 

Also optimization vs. code clarity is an issue, and instead of casting here and there and working with a number of different datatypes, I would suggest to for example just use FP16 throughout the code.

mooskagh: Another idea, given that smaller values much more frequent and need better precision:

```cpp
void SetP(float p) { p = static_cast<uint16_t>(reinterpret_cast<uint32_t>(p) >> 13); }
float GetP() { return reinterpret_cast<float>(static_cast<uint32_t>(p_ << 13)); }
```

The whole bit is wasted to denote value 1.0f, but at least conversion is fast and resulting possible value distribution is really good!

On this site:
https://www.h-schmidt.net/FloatConverter/IEEE754.html

Those are those bits:
image
Green arrow is a wasted bit, only set for values between 1.00 and 1.999999999999.
We can throw it away and replace with one more bit of mantissa (by more bit magic or by multiplying/dividing value by 2).

dubslow:
Latest commit untested for now

src/mcts/node.h
borg323: Instead of chopping the low bits, you can add 0x2000 to sort-off round to nearest. 
```
void SetP(float p) { p = static_cast<uint16_t>((reinterpret_cast<uint32_t>(p) + 0x2000) >> 13); }
float GetP() { return reinterpret_cast<float>(static_cast<uint32_t>(p_ << 13)); }
```

mooskagh: It's a good idea with adding 0x2000, but I think it's better to do that in GetP() rather than in SetP().
```cpp
void SetP(float p) { p = static_cast<uint16_t>(reinterpret_cast<uint32_t>(p) >> 13); }
float GetP() { return reinterpret_cast<float>(static_cast<uint32_t>((p_ << 13) + 0x2000)); }
```

borg323: I think doing it in SetP is more accurate because it is a very good approximation of round to nearest. The only difference is when the 14 lowest bits are exactly equal to 0x2000, where real round to nearest will round up or down depending on the next higher bit.

mooskagh: Suppose we have two bit mantissa, then those numbers:
```
0       0.25       0.5       0.75       1  (× 2ⁿ)
|---------|---------|---------|---------|
   A   B     C    D    E   F    G     H
```
With adding 0x2000 in GetP() will become
```
0       0.25       0.5       0.75       1  (× 2ⁿ)
|---------|---------|---------|---------|
     A         C         E         G
     B         D         F         H
```
And with adding 0x2000 in SetP() they will be:
```
0       0.25       0.5       0.75       1  (× 2ⁿ)
|---------|---------|---------|---------|
A         C         E         G
          B         D         F         H
```
Hm... Do you think that the trick with overflowing mantissa into the exponent (H) will always work?

borg323: Yes, overflowing the mantissa in the exponent works in ieee standard representation, with the exception of inf which is never the case here. I always assumed this is by design to simplify rounding hardware.

dubslow:
Well the current commit works as far as Exa can tell, so I guess I'd be interested in seeing how the fancier way compares. (Feel free to push to my repo/this branch if you beat me to implementing it.)

dubslow:
I'm all for simplicity, especially when simplicity means less branching.

dubslow:
Okay, this latest commit passed a sanity check checking the policy from the start position. Still probably worth a long haul test by Exa.

borg323:
I have a working draft of the idea to trade exponent range for saving more mantissa bits: https://github.com/borg323/lc0/commit/d24f0e299ebfe254faa0f0d84f33326a1e4d007e
This is just to show the idea, the full version shouldn't rely on the assert to check the range but have a check and second path to handle it, e.g. by setting p_ to zero which will be recovered as 1.0842022E-19 by GetP().

dubslow:
> MTGOStark: @Dubslow PR176: 
>
>-6 +- 22
> compared to v16

>I think it's reasonably safe to conclude there's not really any performance impact. Can keep testing if you want.

This PR seems to be ready for merging.

src/mcts/node.cc
```diff
@@ -113,9 +113,50 @@ Move Edge::GetMove(bool as_opponent) const {
   return m;
 }
 
+// We use strictly 16 bits to store policy values in memory-critical Edges.
```
mooskagh:
```cpp

// Priors (P) are stored in compressed 16-bit format.

// Source values are 32-bit floats:
// * bit 31 is sign (zero means positive)
// * bit 30 is sign of exponent (zero means negative)
// * bits 29..23 are value bits of exponent
// * bits 22..0 are mantissa bits
// (see https://www.h-schmidt.net/FloatConverter/IEEE754.html for details).

// In compressed 16-bit value we store bits 27..12:
// * bit 31 is always 0 as values are always >= 0
// * bit 30 is always 0 as values are always < 2
// * bits 29..28 are only 0 for values < 4.65e-10, assume they are always 1
// * bits 11..0 are for higher precision, they are droppes leaving only 11 bits
//              of precision.

// When converting to compressed format, bit 11 is added to make it
// a rounding rather than truncation.

// Out of 65556 possible values, 2047 are outside of [0;1] interval (they are in
// interval (1;2)).
```

src/mcts/node.cc
```diff
@@ -113,9 +113,50 @@ Move Edge::GetMove(bool as_opponent) const {
...
+// take the next 16 bits, that leaves the 12 trailing LSB of the significand to
+// be lost as well. Round-to-nearest by adding a 1 in that 12th bit before
+// truncating. The minimum number that can be represented is 4.656613E-10.
+void Edge::SetP(float p) {
+  assert(0.0f <= p && p <= 1.0f);
+  int32_t tmp;
+  std::memcpy(&tmp, &p, sizeof(float));
+  // If those 2 exponent bits (3<<28) aren't set, round up by setting tmp = 0.
+  // This is combined with the round-to-nearest addition (1<<11) into one op.
+  tmp -= 0x2FFFF800; // == (3 << 28) - (1 << 11)
```
mooskagh: If `(3 << 28) - (1 << 11)` is more readable, maybe write it directly as an expression rather than comment?

src/mcts/node.h
```diff
@@ -97,8 +96,8 @@ class Edge {
   Move move_;
 
   // Probability that this move will be made. From policy head of the neural
-  // network.
-  float p_ = 0.0;
+  // network. Stored with a 16 bit significand in the form of a uint16_t.
```
mooskagh: Stored in 16-bit compressed format (5 bits of exponent and 11 bits of mantissa).

src/mcts/node.cc
dubslow: Excellent comment!

src/mcts/node.cc
dubslow: Done, but still not entirely sure about the proposed solution. Please re-review. I want to emphasize that they're conceptually different, even though they can be implemented as one unified cpu instruction.

mooskagh: Not really "Done", but ok. :)

mooskagh: Ah, actually done! It just showed previous version.

src/mcts/node.cc
```diff
@@ -113,9 +113,55 @@ Move Edge::GetMove(bool as_opponent) const {
...
+// If the two assumed-on exponent bits (3<<28) are in fact off, the input is
+// rounded up to the smallest value with them on. We verify this by subtracting
+// the two bits from the input and checking for a negative result. This is
+// combined with the round-to-nearest addition (1<<11).
+void Edge::SetP(float p) {
+  constexpr int32_t roundings = (1 << 11) - (3 << 28);
+  assert(0.0f <= p && p <= 1.0f);
+  int32_t tmp;
+  std::memcpy(&tmp, &p, sizeof(float));
+  tmp += roundings;
```
mooskagh: it was `-=` prior to change, is `+=` needed instead?

dubslow: Notice I flipped the order of the two constants as well, it's equivalent either way.

src/mcts/node.cc
```diff
@@ -139,12 +139,13 @@ Move Edge::GetMove(bool as_opponent) const {
...
-// the two bits from the input and checking for a negative result. This is
-// combined with the round-to-nearest addition (1<<11).
+// rounded up to the smallest value with them on. We accomplish this by
+// subtracting the two bits from the input and checking for a negative result
+// (the subtraction works despite crossing from exponent to significand). This
+// is combined with the round-to-nearest addition (1<<11) into one op.
 void Edge::SetP(float p) {
-  constexpr int32_t roundings = (1 << 11) - (3 << 28);
   assert(0.0f <= p && p <= 1.0f);
+  constexpr int32_t roundings = (1 << 11) - (3 << 28);
```
borg323: This is fine

PR #177, opened 2018-07-15
Fix castling comparison to not crash NodeTree

dubslow:
Broken by the reduce Move size PR, which incorrectly changed how castles are compared. Previously, they weren't compared, `Move::operator==` only compared `from`, `to`, and promotion. We revert now to that behavior, since that's the information specified by UCI `position moves ...` long algebraic notation.

### DIFF
--- a/src/chess/bitboard.h
+++ b/src/chess/bitboard.h
@@ -217,7 +217,12 @@ class Move {
   // 0 .. 1857, to use in neural networks.
   uint16_t as_nn_index() const;
 
-  bool operator==(const Move& other) const { return data_ == other.data_; }
+  // We ignore the castling bit, because UCI's `position moves ...` commands
+  // specify squares and promotions, but NOT whether or not a move is castling.
+  // NodeTree::MakeMove and all Move::Move constructors are thus so ignorant.
+  bool operator==(const Move& other) const {
+    return (data_ | kCastleMask) == (other.data_ | kCastleMask);
+  }
 
   bool operator!=(const Move& other) const { return !operator==(other); }
   operator bool() const { return data_ != 0; }


dubslow:
The crash on master can be reproduced by attempting to reuse any tree with a castle in its history. Example:

```
position startpos moves b2b3 b7b6 c1b2 c8b7 c2c4 g8f6 g1f3 e7e6 e2e3 f8e7 f1e2 e8g8
position startpos moves b2b3 b7b6 c1b2 c8b7 c2c4 g8f6 g1f3 e7e6 e2e3 f8e7 f1e2 e8g8 d1c1
```

The second command will cause an assert to fail (you need asserts activated, not optimized away).

src/chess/bitboard.h
```diff
@@ -251,9 +256,9 @@ class Move {
   // bit 15 whether move is a castling
 
   enum Masks : uint16_t {
-    kToMask = 0b0000000000111111,
-    kFromMask = 0b0000111111000000,
-    kPromoMask = 0b0111000000000000,
+    kToMask     = 0b0000000000111111,
```
Tilps: Why are you putting random formatting changes in a bugfix?

src/chess/bitboard.h
```diff
@@ -217,7 +217,12 @@ class Move {
   // 0 .. 1857, to use in neural networks.
   uint16_t as_nn_index() const;
 
-  bool operator==(const Move& other) const { return data_ == other.data_; }
+  // We ignore the castling bit, because UCI's `position moves ...` commands
+  // specify squares and promotions, but NOT whether or not a move is castling.
+  // NodeTree::MakeMove and all Move::Move constructors are thus so ignorant.
+  bool operator==(const Move& other) const {
+    return (data_ | kCastleMask) == (other.data_ | kCastleMask);
```
mooskagh: `& ~kCastleMask` would look cleaner to me.

PR #181, opened 2018-07-16
First draft of CONTRIBUTING.md

dubslow:
The only thing that's different from status quo is that I added that rebase merges may be specifically requested, although squash merges are the default and the preference. I added this because I have in mind on particular PR that's mostly just a cleanup of a bunch of minor/inconsequential things to make the codebase prettier, none of which deserve a PR but which do deserve separate commits for each.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -58,6 +58,8 @@ The internal code dependency structure looks like this:
 
 Pull Requests are squahsed when merged. This means all commits in the branch will be squashed into one commit applied onto master, so branches and their PRs should stick to *one* topic only. If you think changes deserve separate commits, make separate PRs for each commit.
 
+This also means it's not possible to reuse one branch for multiple PRs; new PRs must either use entirely new branches, or else you could use `git reset --hard` on the current branch.
+
 
 ### Command line/UCI options
 


dubslow:
I've removed the section about rebase merges, as requested.

dubslow:
@mooskagh I would very much appreciate your specific approval for this one, it is your codebase after all

CONTRIBUTING.md
```diff
@@ -0,0 +1,25 @@
+# Contributing to lc0
+
+There are several project guidelines we prefer that contributors stick to.
+
+This codebase uses Semantic Versioning. This codebase uses Semantic Versioning. A release is the final commit for that version number, and all subsequent commits are development for the next version. Updating the version (i.e. making a release) should be done only with the automated tool for it, `scripts/bumpversion.py`. Any commit that changes the version must consist solely of that version change; any commit which changes anything but the version cannot also change the version.
+
+`master` is the default branch, and the active development branch. `release` will always point to the latest tag. `master` (when it differs from `release`, i.e. almost always) will have a version which includes a postfix attached to the patch section of the version. `release` does not have a postfix attached to the version.
+
```
mooskagh: """
### Allowed features

The Lc0 is still in early stages of development, and not yet reached the point where we are ready to add small tweaks to add few points of a rating. Large code changes still happen, and having lots of small optimizations when one has to keep slows process down.

Therefore, as a rule, search algorithm tweaks that give gain of less than ~30 Elo points are discouraged at this point.
(this limit will gradually be lowered as Lc0 code matures, eventually to 0.0 Elo).

#### Adding new command line flags/ uci parameters

Only add new parameters if users can significantly (>30 Elo) benefit by tweaking it. We don't want to make every single constant configurable (or rather, users don't want to see hundreds of parameters which don't do anything).

Try to minimize number of parameters that your feature introduces. If your feature introduces several parameters, every individual parameter should be significant (i.e. tweaking it with others fixes will give >30 Elo).

#### Adding feature for testing

It is fine to temporary commit a feature of unknown Elo gain to that people could test it. It's also fine to expose many parameters for the feature initially so that people could tune them. However, if the tweak didn't prove to be significant, it should be removed after few weeks.
Same for parameters: it's fine to have them initially for tuning, but if gain of tweaking particular parameter is minuscule, that parameter should be removed.

"""

mooskagh: Also about command line flags:

"""
Every parameter consists of command line flag and description.
Description is used as:
- Help text when `lc0` is run with `--help` parameter
- Uci option name (for parameters used in `uci` mode).

Therefore description should:
- Not end with a period (UCI options are not ended with a dot, and dot is added automatically when outputting help)
- Be short (otherwise UCI parameter won't fit in chess UI)
- Be different from command line flag (otherwise it looks silly in `--help`). In should explain command line flag instead
- It should be clear what it means as UCI parameter.

"""

CONTRIBUTING.md
```diff
@@ -0,0 +1,89 @@
+# Contributing to lc0
+
+These are the guidelines and standards followed by this codebase.
+
+The language is C++, specifically C++14. The compiler used must support C++14, and the entire C++14 standard featureset may be used. As such, manual `new` and `delete` memory mangement is disallowed; use the standard library tools for managing memory (such as `unique_ptr`, `shared_ptr` etc.). When compiler support is more widespread, the project may upgrade to the C++17 standard in the future.
```
mooskagh: Second sentence doesn't add any new information on top of the first. :)
`new` and `delete` are not disallowed, they are strongly discouraged, but sometimes necessary.

CONTRIBUTING.md
```diff
@@ -0,0 +1,89 @@
...
+
+Style is of course the first guideline on every new contributor's mind :)
+
+This codebase largely complies with the [Google C++ style guide](https://google.github.io/styleguide/cppguide.html). The maintainers recommend the use of [Clang's auto formatter](https://clang.llvm.org/docs/ClangFormatStyleOptions.html).
+
+Notable exceptions:
+ 1. C++ exceptions are allowed (in fact, only `lczero::Exception`, defined in `utils/exception.h`, is allowed)
+ 2. We use `#pragma once` instead of header guards.
+ 3. Default function parameters are sometimes allowed, not only for constructors and assignment operators.
+ 4. Rvalue reference function params are sometimes allowed.
```
mooskagh: Oops, I don't know what I was thinking of, but `, not only for constructors and assignment operators` actually was related to rvalue reference parameters (item 4), not default function arguments.

CONTRIBUTING.md
```diff
@@ -0,0 +1,89 @@
...
+ * Code in `src/chess` only depends on `src/utils`
+
+ * Code in `src/neural` only depends on `src/utils` and `src/chess`
+
+ * Code in `src/mcts` only depends on `src/utils`, `src/chess` and `src/neural`
+
+
+### Git history
+
+Pull Requests are squahsed when merged. This means all commits in the branch will be squashed into one commit applied onto master, so PRs (and branches) should stick to *one* topic only. If you think changes deserve separate commits, make separate PRs for each commit. (This also means it's not possible to reuse one branch for multiple PRs; new PRs must either use entirely new branches, or else you could use `git reset --hard` on the current branch.)
```
mooskagh: `(This also means it's not possible to reuse...` was related to squashing, but now that comment is too far from mentioning of squashing, so it's not clear.
I suggest to replace `This` with `Squashing`, and remove parentheses around that sentence.

CONTRIBUTING.md
dubslow: I've heard of plenty of projects that use subsets of features from various standards, enough so that "this is C++14" may be considered ambiguous; but I don't have that much experience with C++, so I'll defer to you on this matter.

CONTRIBUTING.md
dubslow: I had it in parentheses because that sentence isn't really related to our codebase standards, it's only a practical tip for people who understand git very little (which admittedly once included me).

PR #182, opened 2018-07-16
Add submodule update to build instructions

gsobala:
Updated instructions for Ubuntu and Macos, submodule update required for protobuf


dubslow:
@gsobala Perhaps you could add a comment at the top of the README about how users should `git checkout release` before building. That ought to reset the hanging status check.

PR #189, opened 2018-07-17
Appveyor: do not skip commits to fix clash with GitHub status expectations

borg323:
This will make Appveyor build all commits, even trivial ones that only change text files. There is a misalignment with GitHub that expects a status report, while Appveyor doesn't process the commits at all and thus no status is reported. The other two commits in this series save some time by building in parallel and caching the files than meson downloads.


dubslow:
Oh the irony :P

PR #190, opened 2018-07-18
discover gzipped files

Error323:
The DiscoveryWeightsFile() function for protobuf compressed files didn't work. This PR makes sure it works again by checking the file header of a gzipped file. Unfortunately gzopen() doesn't distinguish between opening a regular file or a gzipped file.

Furthermore, binary protobuf does not contain some kind of header but goes straight into serialized data. I don't know of a better way to check for these type of files.

### DIFF
--- a/src/neural/loader.cc
+++ b/src/neural/loader.cc
@@ -101,7 +101,8 @@ void DenormConvBlock(const pblczero::Weights_ConvBlock& conv,
 FloatVectors LoadFloatsFromPbFile(const std::string& buffer) {
   auto net = pblczero::Net();
   FloatVectors vecs;
-  net.ParseFromString(buffer);
+  if (!net.ParseFromString(buffer))
+    throw Exception("Invalid weight file: parse error.");
 
   if (net.magic() != kWeightMagic)
     throw Exception("Invalid weight file: bad header.");
@@ -192,7 +193,7 @@ Weights LoadWeightsFromFile(const std::string& filename) {
   PopulateLastIntoVector(&vecs, &result.ip_pol_w);
   PopulateConvBlockWeights(&vecs, &result.policy);
 
-  // Version, Input + all the residual should be left.
+  // Input + all the residual should be left.
   if ((vecs.size() - 4) % 8 != 0)
     throw Exception("Invalid weight file: parse error.");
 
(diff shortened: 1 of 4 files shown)


src/neural/loader.cc
```diff
@@ -33,6 +33,8 @@
 
 namespace lczero {
 
+static const std::uint32_t kWeightMagic = 0x1c0;
```
mooskagh: C++ way is to put it into anonymous namespace (two lines below).
Then `static` is not really needed.

src/neural/loader.cc
```diff
@@ -100,6 +102,9 @@ FloatVectors LoadFloatsFromPbFile(const std::string& buffer) {
   FloatVectors vecs;
   net.ParseFromString(buffer);
```
mooskagh: `if (!net.ParseFromString(buffer)) throw Exception("Cannot parse weights file.")`

src/neural/loader.cc
```diff
@@ -100,6 +102,9 @@ FloatVectors LoadFloatsFromPbFile(const std::string& buffer) {
...
 
+  if (net.magic() != kWeightMagic)
+    throw Exception("Invalid weight file");
```
mooskagh: Maybe period at the end of a sentence? :)
Also maybe a bit more verbose: `Invalid weight file: bad header.`

src/neural/loader.cc
```diff
@@ -219,8 +224,8 @@ std::string DiscoveryWeightsFile() {
 
   std::sort(time_and_filename.rbegin(), time_and_filename.rend());
 
-  // Open all candidates, from newest to oldest, possibly gzipped, and try to
-  // read version for it. If version is 2, return it.
+  // Open all candidates, from newest to oldest, possibly gzipped, and try to 
+  // read version for it. If version is 2 or if the file is gzipped, return it.
```
mooskagh: "or if file is gzipped" is no longer true

src/neural/loader.cc
```diff
@@ -235,7 +242,15 @@ std::string DiscoveryWeightsFile() {
     int val = 0;
     data >> val;
     if (!data.fail() && val == 2) {
-      std::cerr << "Found network file: " << candidate.second << std::endl;
+      std::cerr << "Found txt network file: " << candidate.second << std::endl;
+      return candidate.second;
+    }
+
+    // first byte of the protobuf stream is 0x0d for fixed32, so we ignore it as
```
dubslow: Capital `F`

PR #196, opened 2018-07-20
Versioning2

Error323:
We got a file `version.inc` and a script `bumpversion.py` that can manipulate it. This should define a sane basis for our versioning needs.

### DIFF
--- a/scripts/bumpversion.py
+++ b/scripts/bumpversion.py
@@ -1,48 +1,80 @@
 #!/usr/bin/env python3
 
 import argparse
-import subprocess
+import tempfile
+import textwrap
+import os
+from string import Template
+from subprocess import call
 
-def extract(git_version):
-    version_array = git_version.split('.')
-    patch_array = version_array[2].split('-')
-    major = int(version_array[0][1:])
-    minor = int(version_array[1])
-    patch = int(patch_array[0])
-    postfix = ""
+VERSION_FILE = os.path.join(os.path.dirname(os.path.realpath(__file__)), "../src/version.inc")
+VERSION_CONTENT = """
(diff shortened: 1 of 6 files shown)


Error323:
The idea is now to flesh out bumpversion.py together such that it will suit our needs.

src/version.h
```diff
@@ -2,20 +2,7 @@
 
 // Versioning is performed according to the standard at <https://semver.org/>
 // Our postfix consists of the 8 character git-sha1 id iff the commit is not
```
Tilps: This sentence looks obsolete, consider removing.

scripts/bumpversion.py
```diff
@@ -1,48 +1,88 @@
...
+    with tempfile.NamedTemporaryFile(suffix='.md') as f:
+        f.write(init_msg.encode('utf-8'))
+        f.flush()
+        call([EDITOR, f.name])
+        f.seek(0)
+        return f.read().decode('utf-8')
+
+
+def update(major, minor, patch):
+    # this is a workflow suggestion:
```
Tilps: Maybe drop the workflow suggestion for now?

scripts/bumpversion.py
```diff
@@ -1,48 +1,88 @@
...
+    call(["git", "commit", "-am", msg])
+
+
 def tag(major, minor, patch):
     version = "v{}.{}.{}".format(major, minor, patch)
-    subprocess.getoutput("git tag -m {} {}".format(version, version))
+    call(["git", "-m", version, version])
+
+
+def editor(init_msg):
```
Tilps: This seems like deadcode - I can see its potential use in future, but it doesn't look terribly complicated so maybe just drop for now?

PR #205, opened 2018-07-21
Release script.

mooskagh:
New minor release, and cherrypick part.
New patch release and documentation will be in a separate commit.

### DIFF
--- a/scripts/deploy.md
+++ b/scripts/deploy.md
@@ -4,7 +4,7 @@ To release a new version, `deploy.py` is used.
 
 ### Prerequisites
 
-`depoy.py` should be run from the repo, directly cloned from
+`deploy.py` should be run from the repo, directly cloned from
 `LeelaChessZero/lc0` (not fork), and upstreams from `master` and `release`
 branches should be configured to point to that repo.
 
@@ -25,22 +25,22 @@ Every pipeline consists of series of steps.
 
 To start pipeline, run it like this:
 ```bash
-$ ./depoy.py new_patch_release
+$ ./deploy.py new_patch_release
 ```
 
 To just get a list of steps in a pipeline, do that:
 ```bash
-$ ./depoy.py --list new_patch_release
+$ ./deploy.py --list new_patch_release
 ```
 
 To start pipeline from particular step (5 in this example), do that:
 ```bash
-$ ./depoy.py --start 5 new_patch_release
+$ ./deploy.py --start 5 new_patch_release
 ```
 
 To make pipeline ask confirmation before executing every step, run like this:
 ```bash
-$ ./depoy.py --steps new_patch_release
+$ ./deploy.py --steps new_patch_release
 ```
 
 The state of the pipeline is stored in file `/tmp/<pipeline-name>_state.json`, 
@@ -97,7 +97,7 @@ During patch release, only patch version is updated. Examples:
  one pipeline run is supported.
 
 ```bash
-$ ./depoy.py cherrypick
+$ ./deploy.py cherrypick
 ```
 The cherrypick pipeline commits cherrypicks, but does not push them (as
 patch release pipeline will do that).
--- a/scripts/deploy.md
+++ b/scripts/deploy.md
@@ -64,8 +64,8 @@ with `dev` suffix added.
 
 Examples:
 
-- `v0.13.1 -> v0.14-rc1`, master branch becomes `v0.15-dev`
-- `v0.13.1 -> v1.0-rc1`, master branch becomes `v1.1-dev`
+- `v0.13.1 -> v0.14.0-rc1`, master branch becomes `v0.15.0-dev`
+- `v0.13.1 -> v1.0.0-rc1`, master branch becomes `v1.1-dev`
 
 To do that, just run:
 ```bash
@@ -84,8 +84,9 @@ During patch release, contents of `master` is NOT merged into `release`.
 Instead, you run `cherrypick` pipeline to pick individual changes to merge into.
 
 During patch release, only patch version is updated. Examples:
-- `v0.13-rc1` -> `v0.13-rc2`
-- `v0.13-rc1` -> `v0.13.0`
+- `v0.13.1-rc1` -> `v0.13.1-rc2`
+- `v0.13.0-rc1` -> `v0.13.0`
+- `v0.13.4` -> `v0.13.5-rc1`
 - `v0.13.4` -> `v0.13.5`
 
 
@@ -152,8 +153,8 @@ Found current version v1.3.1
 
 [ 8/23] Select new major/minor version....
 Current version is v1.3.1. What will be the new one?
-  1. 1.4-rc1
-  2. 2.0-rc1
+  1. 1.4.0-rc1
+  2. 2.0.0-rc1
 >>>>>>> 1
 
 [ 9/23] Merge master into release...
@@ -178,8 +179,8 @@ $ git add /home/crem/dev/test/changelog.txt
 
 [16/23] Write version into version.inc, commit and tag....
 $ git add /home/crem/dev/test/src/version.inc
-$ git commit -m "Change version.inc to v1.5-dev."
-[release 70e12c6] Change version.inc to v1.5-dev.
+$ git commit -m "Change version.inc to v1.5.0-dev."
+[release 70e12c6] Change version.inc to v1.5.0-dev.
  2 files changed, 8 insertions(+), 3 deletions(-)
 
 [17/23] Checkout release...
@@ -203,10 +204,10 @@ Your branch is ahead of 'origin/release' by 1 commit.
 
 [20/23] Write version into version.inc, commit and tag....
 $ git add /home/crem/dev/test/src/version.inc
-$ git commit -m "Change version.inc to v1.4-rc1."
-[release 2b515ea] Change version.inc to v1.4-rc1.
+$ git commit -m "Change version.inc to v1.4.0-rc1."
+[release 2b515ea] Change version.inc to v1.4.0-rc1.
  1 file changed, 2 insertions(+), 2 deletions(-)
-$ git tag -a v1.4-rc1 -m "Adding tag v1.4-rc1"
+$ git tag -a v1.4.0-rc1 -m "Adding tag v1.4.0-rc1"
 
 [21/23] Checkout master...
 $ git checkout master
@@ -233,7 +234,7 @@ Counting objects: 100% (1/1), done.
 Writing objects: 100% (1/1), 171 bytes | 171.00 KiB/s, done.
 Total 1 (delta 0), reused 0 (delta 0)
 To bitbucket.org:mooskagh/test.git
- * [new tag]         v1.4-rc1 -> v1.4-rc1
+ * [new tag]         v1.4.0-rc1 -> v1.4.0-rc1
 
 The pipeline has finished.
 ```
@@ -328,11 +329,11 @@ Switched to branch 'release'
 Your branch is up to date with 'origin/release'.
 
 [ 7/18] Read current version from version.inc file...
-Found current version v1.4-rc1
+Found current version v1.4.0-rc1
 
 [ 8/18] Select new patch version....
-Current version is v1.4-rc1. What will be the new one?
-  1. 1.4-rc2
+Current version is v1.4.0-rc1. What will be the new one?
+  1. 1.4.0-rc2
   2. 1.4.0
 >>>>>>> 2
 
--- a/scripts/deploy.md
+++ b/scripts/deploy.md
@@ -1,0 +1,388 @@
+# Releasing a new version
+
+To release a new version, `deploy.py` is used.
+
+### Prerequisites
+
+`depoy.py` should be run from the repo, directly cloned from
+`LeelaChessZero/lc0` (not fork), and upstreams from `master` and `release`
+branches should be configured to point to that repo.
+
+It's recommended to have a separate repository cloned which you will use
+exclusively for releases, and won't do other development on that repo.
+
+To push version changes, scripts needs write access to this repo.
+
+
+### General script information
+
+There are three pipelines implemented in this script:
+- `new_minor_release`
+- `cherrypick`
+- `new_patch_release`
+
+Every pipeline consists of series of steps.
+
+To start pipeline, run it like this:
+```bash
+$ ./depoy.py new_patch_release
+```
+
+To just get a list of steps in a pipeline, do that:
+```bash
+$ ./depoy.py --list new_patch_release
+```
+
+To start pipeline from particular step (5 in this example), do that:
+```bash
+$ ./depoy.py --start 5 new_patch_release
+```
+
+To make pipeline ask confirmation before executing every step, run like this:
+```bash
+$ ./depoy.py --steps new_patch_release
+```
+
+The state of the pipeline is stored in file `/tmp/<pipeline-name>_state.json`, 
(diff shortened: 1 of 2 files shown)


dubslow:
Why is this separate from `bumpversion.py`? They handle the same functionality, they should be a common script

mooskagh:
Yes it will replace bumpversion.py.

mooskagh:
Ready for review!

mooskagh:
Resulting python turned out to be pretty messy, feel free to ask to clean it up!

scripts/deploy.py
```diff
@@ -0,0 +1,417 @@
...
+############################################################################
+
+
+def FormatVersion(version_dict):
+    res = '%d.%d' % (version_dict['MAJOR'], version_dict['MINOR'])
+    if version_dict['POSTFIX']:
+        res += '-%s' % version_dict['POSTFIX']
+    else:
+        res += '.%d' % version_dict['PATCH']
+    return res
```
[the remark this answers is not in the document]
mooskagh: semver format allows 1.23-blah as an alias of 1.23.0-blah
And we don't really use semver as we don't really have API.

But anyway, changed to always show patch number. And any patch may have rc now.
(after 1.23.5, we may choose if we want 1.23.6 or 1.23.6-rc1)

scripts/deploy.md
```diff
@@ -0,0 +1,388 @@
+# Releasing a new version
+
+To release a new version, `deploy.py` is used.
+
+### Prerequisites
+
+`depoy.py` should be run from the repo, directly cloned from
```
dubslow: Typo here :)

scripts/deploy.md
```diff
@@ -0,0 +1,388 @@
...
+When issuing new minor/major release, entire contents of `master` branch is
+merged into `release` branch, new minor version change happens, and `-rc1`
+suffix is added.
+
+The version in `master` branch is after that updated to a next minor version,
+with `dev` suffix added.
+
+Examples:
+
+- `v0.13.1 -> v0.14-rc1`, master branch becomes `v0.15-dev`
```
dubslow: Without having yet reviewed the python, postfixes must come after the patch version. `0.14-rc1` isn't a valid version, but `0.14.0-rc1` is. Ditto `0.15-dev` no bueno, needs must be `0.15.0-dev`

scripts/deploy.md
```diff
@@ -0,0 +1,388 @@
...
+$ git tag -a v1.4-rc1 -m "Adding tag v1.4-rc1"
+
+[21/23] Checkout master...
+$ git checkout master
+Switched to branch 'master'
+Your branch is ahead of 'origin/master' by 6 commits.
+  (use "git push" to publish your local commits)
+
+[22/23] Push release changes to github...
+$ git push --all origin
```
dubslow: How does this work?

As our repo is currently designed, no commits may be pushed to master without an approving review, i.e. all pushes must have a PR with input from another human.

scripts/deploy.md
dubslow: Without having yet reviewed the python, postfixes must come after the patch version. `0.14-rc1` isn't a valid version, but `0.14.0-rc1` is. Ditto `0.15-dev` no bueno, needs must be `0.15.0-dev`

mooskagh: Done.
I've edited the example run manually though instead of running the script.

scripts/deploy.md
mooskagh: I'm pretty sure all of those branch protections are only in github UI, and through command line it's all unrestricted, at least to repo admins.

dubslow: That is not true (as mentioned on Discord, but worth duplicating here for those not on Discord). The branch protections apply to all means of pushing, GH UI or git CLI.

scripts/deploy.py
```diff
@@ -0,0 +1,497 @@
...
+    ctx['new-version'] = SelectionPrompt(
+        "Current version is v%s. What will be the new one?" % FormatVersion(v),
+        [(FormatVersion(x), x) for x in variants])
+    return True
+
+
+def ReadCurrentVerionFromFile(ctx):
+    """Read current version from version.inc file"""
+    version = {}
+    macro_re = re.compile(r'#define LC0_VERSION_(\S+) (.*)')
```
oscardssmith: Shouldn't the regex be compiled outside the function?

dubslow: *could* be placed outside this function, at global level as _hidden_variable. not a big deal either way. honestly, since it's only used a handful of times and not interspersed with other regexes, you could remove it entirely with no performance hit (re module automatically caches recently compiled regexes)

scripts/deploy.py
```diff
@@ -0,0 +1,497 @@
...
+        self.list_only = False  # Only list steps, do not execute.
+        self.step_by_step = False  # Ask confirmation before every step.
+        self.context = {}
+        self.cmd_name = 'unknown'
+
+    def AddStep(self, func):
+        self.steps.append(func)
+
+    def StateFileName(self):
+        return os.path.join('/tmp', '%s_state.json' % self.cmd_name)
```
dubslow: Shouldn't we generate a properly unique tmp file name via the stdlib module for this?

dubslow: Or if not, at least add a lc0 specific portion to the name somewhere

scripts/deploy.py
```diff
@@ -0,0 +1,497 @@
...
+                if self.step_by_step:
+                    if not click.confirm('Should I run it?'):
+                        raise click.Abort
+                if task_f(self.context):
+                    self.context['idx'] += 1
+                    continue
+            except click.Abort:
+                raise
+            except Jump as jmp:
+                self.context['idx'] += jmp.whereto
```
dubslow: `whereto` should be the absolute target, not a relative offset. If it is a relative offset (as the `+=` indicates), then it should be renamed to `offset` or similar

scripts/deploy.py
```diff
@@ -0,0 +1,497 @@
...
+    p.AddStep(CompareFileSha(CHANGELOG_PATH, require_equal=False))
+    p.AddStep(RunCmdStep('git add %s' % CHANGELOG_PATH))
+
+
+############################################################################
+# Pipeline machinery
+############################################################################
+
+
+def RunJump(where):
```
dubslow: See below also, but "steps *forward*" indicates this should be named `offset` or similar, not `where`, which implies an absolute target, not a relative one

scripts/deploy.py
```diff
@@ -0,0 +1,497 @@
...
+    def f(ctx):
+        raise Jump(where)
+
+    if where > 0:
+        f.__doc__ = "Go %d steps forward" % where
```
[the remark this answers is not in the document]
mooskagh: It's allowed in the style guide: https://github.com/google/styleguide/blob/gh-pages/pyguide.md#310-strings

But I've replaced it in the entire file except one place where % does look more readable.

scripts/deploy.py
mooskagh: Done.

scripts/deploy.py
```diff
@@ -0,0 +1,497 @@
...
+        f.__doc__ = "Go %d steps backwards" % -where
+    return f
+
+
+def FormatVersion(version_dict):
+    res = '%d.%d.%d' % (
+        version_dict['MAJOR'],
+        version_dict['MINOR'],
+        version_dict['PATCH'],
+    )
```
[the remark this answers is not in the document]
mooskagh: Done.

scripts/deploy.py
```diff
@@ -0,0 +1,497 @@
...
+        for i, n in enumerate(choices):
+            click.secho("%3d. %s" % (i + 1, n[0]), fg='yellow')
+        r = click.prompt('', prompt_suffix='>>>>>>> ')
+        try:
+            r = int(r) - 1
+            if zero_choice and r == -1:
+                return None
+            if 0 <= r < len(choices):
+                return choices[r][1]
+        except:
```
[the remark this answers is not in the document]
mooskagh: Done.

scripts/deploy.py
```diff
@@ -0,0 +1,497 @@
...
+    """Cherry-pick a commit"""
+    return RunCmdStep('git cherry-pick %s' % ctx['cherrypick-commit'])(ctx)
+
+
+def WriteVersionToFile(version):
+    with open(VERSION_PATH, "w") as f:
+        f.write('// This file is automatically generated by deploy.py. '
+                'Do not edit this file.\n')
+        for k, v in version.items():
+            f.write('#define LC0_VERSION_%s %s\n' % (k, json.dumps(v)))
```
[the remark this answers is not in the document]
mooskagh: Added comment.
But in general it's because it will handle both strings correctly (including ones with special character etc, like `"asd\"sdf\n"`) and numbers.
`repr()` will use single quotes instead.

scripts/deploy.py
mooskagh: It should not be unique, it's (only) intention is to be reused in the case of crash.
Added lc0-specific portion.

scripts/deploy.py
mooskagh: Done.

scripts/deploy.md
mooskagh: Yes indeed, I've checked and it doesn't allow me to do that. Will need to find a solution.

dubslow:
I don't see any actual new code pushed?

scripts/deploy.py
```diff
@@ -0,0 +1,497 @@
...
+def new_minor_release(ctx):
+    p = ctx.obj['pipeline']
+
+    CheckOwnIntegrityAndFetch(p)
+    p.AddStep(RunCmdStep('git checkout release', doc="Checkout release"))
+    p.AddStep(ReadCurrentVerionFromFile)
+    p.AddStep(GetNextVersion)
+    p.AddStep(RunCmdStep('git merge master', doc="Merge master into release"))
+    UpdateChangeLog(p)
+    p.AddStep(WriteAndCommitVersion('dev-version', add_tag=False))
```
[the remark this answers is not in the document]
dubslow: I have no idea what the purpose of that convoluted sequence is. I had thought we never, ever, merge anything from release to master. All merges are one direction only, or so I had thought. When we want to release whatever the current state of master is, simply:

1) increment the version.inc in master from `vX.Y.Z-dev` to `vX.Y.Z`

2) merge master into release

3) increment version.inc in master to `vX.Y+1.0-dev`.

No collisions in version.inc, no confusing wrong way merges, and as far as I can tell it has the same end result as what you wrote.

scripts/deploy.py
```diff
@@ -0,0 +1,497 @@
...
+    p.AddStep(RunCmdStep('git checkout master', doc="Checkout release"))
+    p.AddStep(RunCmdStep('git merge release', doc="Merge release into master"))
+    p.AddStep(RunCmdStep('git checkout release', doc="Checkout release"))
+    p.AddStep(WriteAndCommitVersion('new-version', add_tag=True))
+    p.AddStep(RunCmdStep('git checkout master', doc="Checkout master"))
+    p.AddStep(
+        RunCmdStep(
+            'git push --all origin', doc="Push release changes to github"))
+    p.AddStep(RunCmdStep('git push --tags origin', doc="Push tags to github"))
```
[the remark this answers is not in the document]
dubslow: That's.... complicated, isn't it? What's the point of an isolated repo for making releases?

As far as what repo is cloned, that doesn't matter; all developer clones will eventually have to have both upstream (to pull upstream changes) and the fork (to push local changes) as added remotes, so which one you use to make the clone at first is really quite irrelevant, they make the same end result.

scripts/deploy.py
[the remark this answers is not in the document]
dubslow: Well the zen of python is that there should only be one way to do things, and string formatting is certainly one of the minor blemishes where they failed to hold to that ideal. If you're writing in a codebase that is not Python 2 compatible, such as this one, there's no reason really to use the outdated % formatting, and perhaps the Google style guide will reflect that in the future.

At any rate, f-strings (with `format()` as a supplement) are certainly the way of the future, but alas, it will be a few years before they can be assumed present.

scripts/deploy.py
```diff
@@ -110,54 +124,50 @@ def UpdateChangeLog(p):
     p.AddStep(RunStoreFileSha(CHANGELOG_PATH))
     p.AddStep(
         RunCmdStep(
-            '%s %s' % (os.getenv('EDITOR', 'vim'), CHANGELOG_PATH),
-            doc="Edit changelog"))
+            '{0} {1}'.format(os.getenv('EDITOR', 'vim'), CHANGELOG_PATH),
+            doc='Edit changelog'))
     p.AddStep(CompareFileSha(CHANGELOG_PATH, require_equal=False))
-    p.AddStep(RunCmdStep('git add %s' % CHANGELOG_PATH))
+    p.AddStep(RunCmdStep(f'git add {CHANGELOG_PATH}'))
```
dubslow: If we're going to use `f` strings, then we need to be sure all release devs have 3.6 available, and if so, then all possible formatting should be done with this (e.g. two lines above). Otherwise, use `format()` exclusively.

scripts/deploy.py
```diff
@@ -171,28 +181,30 @@ def f(ctx):
...
 def WriteVersionToFile(version):
-    with open(VERSION_PATH, "w") as f:
+    with open(VERSION_PATH, 'w') as f:
         f.write('// This file is automatically generated by deploy.py. '
                 'Do not edit this file.\n')
         for k, v in version.items():
-            f.write('#define LC0_VERSION_%s %s\n' % (k, json.dumps(v)))
+            # Value may be string or number. JSON formatting of those data
+            # types is conveniently the same as in C++, so outputting JSON.
+            f.write(f'#define LC0_VERSION_{k} {json.dumps(v)}\n')
```
dubslow: I guess the part that had confused me before is that the string should already be suitably interpreted/converted/whatever when it's placed into the `version` dict. For doing string to string concatenation, there should be no interpretation/converted/whatever.

scripts/deploy.py
```diff
@@ -40,22 +40,36 @@ def cli(ctx, start, list, steps):
...
+    # 2. Change version.inc in release branch to new "dev"
+    # 3. Merge release to master (so that master has new "dev" too)
+    # 4. Change version.inc in release branch to new release version.
+
+    # Then do the push, so all that merging back and forth is kind of atomic
+    # from github point of view).
+
+    # Without that (e.g. modifying version.inc independently rather than having
+    # a sequence), all merges will have a collision in version.inc and that's
+    # painful to resolve.
```
dubslow: Er, see my other comment that I wrote 30 minutes ago as a reply to previous comments, but perhaps I misunderstood you. Perhaps by "collision" you actually mean "git conflict"? At any rate, I still don't quite see how doing this the simple way would generate merge conflicts. Suffice it to say, I really think doing any release->master merges, ever in the history of ever, is a bad idea.

scripts/deploy.py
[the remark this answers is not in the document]
mooskagh: Then when you want to release a patch, what is the sequence?

scripts/deploy.py
mooskagh: I'm pretty sure both I and Error have python 3.7 available.
I use f'' where it's more readable and format (or I'd actually prefer `%`) where it's not.

scripts/deploy.py
dubslow: Oh, I suppose that having patch versions on release separately from master might generate conflicts. Even so, I'd rather build this script to handle conflicts than do any sort of backwards merges.

PR #206, opened 2018-07-22
Don't show patch for -dev and -rc releases.

mooskagh:
Also rename dev version to be 0.17-dev rather than 0.16.1-dev


src/version.cc
```diff
@@ -1,12 +1,17 @@
...
-           std::to_string(patch);
-  if (postfix.empty()) return v;
-  return v + "-" + postfix;
+std::string GetVersionStr(int major, int minor, int patch,
+                          const std::string& postfix) {
+  auto v = std::to_string(major) + "." + std::to_string(minor);
+  if (postfix.empty()) {
+    return v + "." + std::to_string(patch);
+  } else {
+    assert(patch == 0);
```
Tilps: I think it would be fine to potentially have an rc for a patch.

mooskagh: Ok, then closing this.

Issue #209, opened 2018-07-22
Lc0 doesn't populate repetition planes


mooskagh:
Example:
```
position fen rnbqkbnr/pp2pppp/2pp4/8/1P6/5N2/P1PPPPPP/RNBQKB1R w KQkq - 0 1 moves c2c4 d8b6 d1b3 a7a5 b4b5 g8f6 b1c3 e7e5 a2a3 c6b5 e2e3 e5e4 f3g5 h7h6 g5e4 f8e7 e4f6 e7f6 d2d4 a5a4 b3b5 b6b5 c3b5 e8g8 b5d6 f6e7 d6c8 f8c8 a1b1 c8c7 c4c5 b8d7 f1b5 g7g6 e1e2 h6h5 b5d7 c7d7 b1b4 f7f5 c1d2 a8a7 h1b1 e7d8 b4b5 g8g7 h2h3 g7f6 e2d3 g6g5 d3c4 f6e6 d2e1 g5g4 e1d2 g4h3 g2h3 e6f7 d2e1 f7f8 f2f4 d7h7 c4d3 h7d7 d3c4

go nodes 0
```

Populated planes:
https://gist.github.com/mooskagh/327e2380056ed41ee917fa5738ca20fd

For comparison, lczero populated planes for the same position:
https://gist.github.com/mooskagh/140e5d59cefc2340f8ce9ba454e11b3b

DanielUranga:
how you get the populated planes?

mooskagh:
You mean how to I dump them?

```cpp
for (auto& x : result) std::cerr << std::bitset<64>(x.bits) << ' ' << x.value << std::endl;
```

mooskagh:
That game on lichess: https://lichess.org/qJ5Dhirq#65

DanielUranga:
I think the issue is here: https://github.com/LeelaChessZero/lc0/blob/2321011913d8a7914d8177e82ceaf34fbe2d6ee8/src/chess/position.cc#L102
the loop should start from `int idx = positions_.size() - 1`, if not is not considering the last position.

Tilps:
'last' is the last position - if you make that loop positions_.size() - 1(i.e. last) it will immediately terminate because last == last and thus return 1 because its repetition count is 0.
(That being said I've not worked out the problem yet.)

DanielUranga:
[...]

PR #210, opened 2018-07-22
Issue209

DanielUranga:
Fixes https://github.com/LeelaChessZero/lc0/issues/209

### DIFF
--- a/src/chess/position_test.cc
+++ b/src/chess/position_test.cc
@@ -1,0 +1,46 @@
+/*
+  This file is part of Leela Chess Zero.
+  Copyright (C) 2018 The LCZero Authors
+
+  Leela Chess is free software: you can redistribute it and/or modify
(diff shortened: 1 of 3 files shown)


src/chess/position_test.cc
```diff
@@ -0,0 +1,47 @@
...
+*/
+
+#include <gtest/gtest.h>
+
+#include <iostream>
+#include "src/chess/position.h"
+
+namespace lczero {
+
+TEST(PositionHistory, ComputeLastMoveRepetitions) {
```
Tilps: This test should probably be called ComputeLastMoveRepetitionsForPawnPushWithoutEnpassant (or similar)

And ideally we should have a ComputeLastMoveRepetitionsforPawnPushWithLegalEnpassant to cover both sides.

PR #211, opened 2018-07-23
After finding a checkmate, never bother searching for alternatives to mate

dubslow:
Was just wondering about this in relation to Leela not finding a mate in 5 posted on Discord.

Turns out that after searching a checkmating move, all the other (non-mate) alternatives still get searched like normal, which in turn "pollutes" the Q value in the nodes above the tree. This is of course exactly what we want in most cases, PUCT is supposed to be least upper confidence bound after all, but when we find a checkmate, our confidence is infinite, the game result is without uncertainty -- so never bother with alternatives.

This will go a long way towards Leela giving quicker mates. When the game is already totally won, all Q values in the search tree will already be north of 0.9, so finding mates currently doesn't do much, because averaging in a 1.0 into a whole bunch of other 0.95 and 0.98 or whatever doesn't actually change much of anything higher up the search tree. This will guarantee that such higher-up values converge monotonically to 1, rather than asymptotically only over several tens or hundreds of thousands of nodes. (And to be clear, even without this patch, parents-to-mate will *eventually* converge to having Q=1.0, but only after literally thousands or millions of nodes as the sqrt(N) term in the normal U decreases. Instead with this patch, converge immediately.)

Demonstrations of these qualitative descriptions incoming.

### DIFF
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -160,28 +160,16 @@ Edge* Node::GetEdgeToNode(const Node* node) const {
 
 std::string Node::DebugString() const {
   std::ostringstream oss;
-  oss << " BitBools:" << bit_bools_ << " This:" << this << " Parent:" << parent_
+  oss << " Term:" << is_terminal_ << " This:" << this << " Parent:" << parent_
       << " Index:" << index_ << " Child:" << child_.get()
       << " Sibling:" << sibling_.get() << " Q:" << q_ << " N:" << n_
       << " N_:" << n_in_flight_ << " Edges:" << edges_.size();
   return oss.str();
 }
 
-bool Node::IsTerminal() const {
-  return bit_bools_ & kTerminalMask;
-}
-
 void Node::MakeTerminal(GameResult result) {
-  bit_bools_ |= kTerminalMask;
+  is_terminal_ = true;
   q_ = (result == GameResult::DRAW) ? 0.0f : 1.0f;
-}
-
-bool Node::IsCertain() const {
-  return bit_bools_ & kCertainMask;
-}
-
-void Node::MakeCertain() {
-  bit_bools_ |= kCertainMask;
 }
 
 bool Node::TryStartScoreUpdate() {
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -160,16 +160,28 @@ Edge* Node::GetEdgeToNode(const Node* node) const {
 
 std::string Node::DebugString() const {
   std::ostringstream oss;
-  oss << " Term:" << is_terminal_ << " This:" << this << " Parent:" << parent_
+  oss << " BitBools:" << bit_bools_ << " This:" << this << " Parent:" << parent_
       << " Index:" << index_ << " Child:" << child_.get()
       << " Sibling:" << sibling_.get() << " Q:" << q_ << " N:" << n_
       << " N_:" << n_in_flight_ << " Edges:" << edges_.size();
   return oss.str();
 }
 
+bool Node::IsTerminal() const {
+  return bit_bools_ & kTerminalMask;
+}
+
 void Node::MakeTerminal(GameResult result) {
-  is_terminal_ = true;
+  bit_bools_ |= kTerminalMask;
   q_ = (result == GameResult::DRAW) ? 0.0f : 1.0f;
+}
+
+bool Node::IsCertain() const {
+  return bit_bools_ & kCertainMask;
+}
+
+void Node::MakeCertain() {
+  bit_bools_ |= kCertainMask;
 }
 
 bool Node::TryStartScoreUpdate() {
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -657,6 +657,13 @@ void SearchWorker::ExtendNode(Node* node) {
     // Could be a checkmate or a stalemate
     if (board.IsUnderCheck()) {
       node->MakeTerminal(GameResult::WHITE_WON);
+      // Set this move and all its siblings as certain for search purposes.
+      // TODO: probably should take a mutex here? Maybe move this into MakeTerminal?
+      Node* parent = node->GetParent();
+      for (auto& edge : parent->Edges()) {
+        Node* sibling = edge.GetOrSpawnNode(parent);
+        sibling->MakeCertain();
+      }
     } else {
       node->MakeTerminal(GameResult::DRAW);
     }
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -591,7 +591,7 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend() {
     // n_in_flight_ is incremented. If the method returns false, then there is
     // a search collision, and this node is already being expanded.
     if (!node->TryStartScoreUpdate()) return {node, true};
-    // Unexamined leaf node. We've hit the end of this playout.
+    // Either terminal or unexamined leaf node -- the end of this playout.
     if (!node->HasChildren()) return {node, false};
     // If we fall through, then n_in_flight_ has been incremented but this
     // playout remains incomplete; we must go deeper.
@@ -627,6 +627,11 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend() {
         ++possible_moves;
       }
       float Q = child.GetQ(parent_q);
+      if (Q == 1.0f && child.IsTerminal()) {
+        // If we find a checkmate, then the confidence is infinite, so ignore U.
+        best_edge = child;
+        break;
+      }
       const float score = child.GetU(puct_mult) + Q;
       if (score > best) {
         best = score;
@@ -657,13 +662,6 @@ void SearchWorker::ExtendNode(Node* node) {
     // Could be a checkmate or a stalemate
     if (board.IsUnderCheck()) {
       node->MakeTerminal(GameResult::WHITE_WON);
-      // Set this move and all its siblings as certain for search purposes.
-      // TODO: probably should take a mutex here? Maybe move this into MakeTerminal?
-      Node* parent = node->GetParent();
-      for (auto& edge : parent->Edges()) {
-        Node* sibling = edge.GetOrSpawnNode(parent);
-        sibling->MakeCertain();
-      }
     } else {
       node->MakeTerminal(GameResult::DRAW);
     }
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -238,6 +238,8 @@ void Search::SendMovesStats() const {
     }
     oss << ") ";
 
+    oss << "(T: " << edge.IsTerminal() << ") ";
+
     info.comment = oss.str();
     info_callback_(info);
   }
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -56,6 +56,7 @@ const char* Search::kCacheHistoryLengthStr =
 const char* Search::kPolicySoftmaxTempStr = "Policy softmax temperature";
 const char* Search::kAllowedNodeCollisionsStr =
     "Allowed node collisions, per batch";
+const char* Search::kStickyCheckmateStr = "Ignore alternatives to checkmate";
 
 namespace {
 const int kSmartPruningToleranceNodes = 100;
@@ -86,6 +87,7 @@ void Search::PopulateUciParams(OptionsParser* options) {
                             "policy-softmax-temp") = 1.0f;
   options->Add<IntOption>(kAllowedNodeCollisionsStr, 0, 1024,
                           "allowed-node-collisions") = 0;
+  options->Add<BoolOption>(kStickyCheckmateStr, "sticky-checkmate") = false;
 }
 
 Search::Search(const NodeTree& tree, Network* network,
@@ -112,7 +114,8 @@ Search::Search(const NodeTree& tree, Network* network,
       kFpuReduction(options.Get<float>(kFpuReductionStr)),
       kCacheHistoryLength(options.Get<int>(kCacheHistoryLengthStr)),
       kPolicySoftmaxTemp(options.Get<float>(kPolicySoftmaxTempStr)),
-      kAllowedNodeCollisions(options.Get<int>(kAllowedNodeCollisionsStr)) {}
+      kAllowedNodeCollisions(options.Get<int>(kAllowedNodeCollisionsStr)),
+      kStickyCheckmate(options.Get<bool>(kStickyCheckmateStr)) {}
 
 namespace {
 void ApplyDirichletNoise(Node* node, float eps, double alpha) {
@@ -629,7 +632,7 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend() {
         ++possible_moves;
       }
       float Q = child.GetQ(parent_q);
-      if (Q == 1.0f && child.IsTerminal()) {
+      if (search_->kStickyCheckmate && Q == 1.0f && child.IsTerminal()) {
         // If we find a checkmate, then the confidence is infinite, so ignore U.
         best_edge = child;
         break;
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -283,7 +283,16 @@ class EdgeAndNode {
   // Returns U = numerator * p / N.
   // Passed numerator is expected to be equal to (cpuct * sqrt(N[parent])).
   float GetU(float numerator) const {
-    return numerator * GetP() / (1 + GetNStarted());
+    if (IsTerminal() && GetQ(0.0f) == 1.0f) {
+      // Never bother with alternatives to checkmate. In theory, U represents
+      // the "uncertainty that this move is actually better than the leading
+      // move", so when we have a certain winner, we should set all sibling U
+      // to 0, meaning MCTS is guided by value alone. That's complicated to do,
+      // however, so instead we create the same behavior with this:
+      return std::numeric_limits<float>::infinity();
+    } else {
+      return numerator * GetP() / (1 + GetNStarted());
+    }
   }
 
   std::string DebugString() const;
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -154,20 +154,14 @@ class Node {
   // for terminal nodes.
   float GetQ() const { return q_; }
 
+  // Returns whether the node is known to be draw/lose/win.
+  bool IsTerminal() const { return is_terminal_; }
   uint16_t GetFullDepth() const { return full_depth_; }
   uint16_t GetMaxDepth() const { return max_depth_; }
+  uint16_t GetNumEdges() const { return edges_.size(); }
 
-  // Returns whether the node is known to be draw/lose/win.
-  bool IsTerminal() const;
   // Makes the node terminal and sets it's score.
   void MakeTerminal(GameResult result);
-
-  // In MCTS+PUCT, search is guided by Q+U; U is a sort of "uncertainty that
-  // this move might be better than the leading move"; when a checkmate is
-  // available, we can then say that it and its siblings are certain (the
-  // siblings being certainly worse than mate).
-  bool IsCertain() const;
-  void MakeCertain();
 
   // If this node is not in the process of being expanded by another thread
   // (which can happen only if n==0 and n-in-flight==1), mark the node as
@@ -193,7 +187,6 @@ class Node {
   V3TrainingData GetV3TrainingData(GameResult result,
                                    const PositionHistory& history) const;
 
-  uint16_t GetNumEdges() const { return edges_.size(); }
   // Returns range for iterating over edges.
   ConstIterator Edges() const;
   Iterator Edges();
@@ -236,13 +229,8 @@ class Node {
   uint16_t max_depth_ = 0;
   // Complete depth all subnodes of this node were fully searched.
   uint16_t full_depth_ = 0;
-
-  // Stores miscellaneous booleans: IsTerminal, and IsCertain.
-  uint8_t bit_bools_ = 0;
-  enum Masks_ : uint8_t {
-    kTerminalMask = 0x1,
-    kCertainMask = 0x2,
-  };
+  // Does this node end game (with a winning of either sides or draw).
+  bool is_terminal_ = false;
 
   // Pointer to a parent node. nullptr for the root.
   Node* parent_ = nullptr;
@@ -285,24 +273,17 @@ class EdgeAndNode {
   int GetNStarted() const { return node_ ? node_->GetNStarted() : 0; }
   uint32_t GetNInFlight() const { return node_ ? node_->GetNInFlight() : 0; }
 
-  // Whether the node is known to be terminal or certain.
+  // Whether the node is known to be terminal.
   bool IsTerminal() const { return node_ ? node_->IsTerminal() : false; }
-  bool IsCertain() const  { return node_ ? node_->IsCertain()  : false; }
 
   // Edge related getters.
   float GetP() const { return edge_->GetP(); }
   Move GetMove(bool flip = false) const { return edge_->GetMove(flip); }
 
-  // Returns U, which in a sense measures the relative uncertainty that this
-  // move might be better than the leading move.
-  // U = numerator * p / N, where the passed numerator is expected to be equal
-  // to (cpuct * sqrt(N[parent])).
+  // Returns U = numerator * p / N.
+  // Passed numerator is expected to be equal to (cpuct * sqrt(N[parent])).
   float GetU(float numerator) const {
-    if (IsCertain()) {
-      return 0.0f;
-    } else {
-      return numerator * GetP() / (1 + GetNStarted());
-    }
+    return numerator * GetP() / (1 + GetNStarted());
   }
 
   std::string DebugString() const;
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -154,14 +154,20 @@ class Node {
   // for terminal nodes.
   float GetQ() const { return q_; }
 
-  // Returns whether the node is known to be draw/lose/win.
-  bool IsTerminal() const { return is_terminal_; }
   uint16_t GetFullDepth() const { return full_depth_; }
   uint16_t GetMaxDepth() const { return max_depth_; }
-  uint16_t GetNumEdges() const { return edges_.size(); }
 
+  // Returns whether the node is known to be draw/lose/win.
+  bool IsTerminal() const;
   // Makes the node terminal and sets it's score.
   void MakeTerminal(GameResult result);
+
+  // In MCTS+PUCT, search is guided by Q+U; U is a sort of "uncertainty that
+  // this move might be better than the leading move"; when a checkmate is
+  // available, we can then say that it and its siblings are certain (the
+  // siblings being certainly worse than mate).
+  bool IsCertain() const;
+  void MakeCertain();
 
   // If this node is not in the process of being expanded by another thread
   // (which can happen only if n==0 and n-in-flight==1), mark the node as
@@ -187,6 +193,7 @@ class Node {
   V3TrainingData GetV3TrainingData(GameResult result,
                                    const PositionHistory& history) const;
 
+  uint16_t GetNumEdges() const { return edges_.size(); }
   // Returns range for iterating over edges.
   ConstIterator Edges() const;
   Iterator Edges();
@@ -229,8 +236,13 @@ class Node {
   uint16_t max_depth_ = 0;
   // Complete depth all subnodes of this node were fully searched.
   uint16_t full_depth_ = 0;
-  // Does this node end game (with a winning of either sides or draw).
-  bool is_terminal_ = false;
+
+  // Stores miscellaneous booleans: IsTerminal, and IsCertain.
+  uint8_t bit_bools_ = 0;
+  enum Masks_ : uint8_t {
+    kTerminalMask = 0x1,
+    kCertainMask = 0x2,
+  };
 
   // Pointer to a parent node. nullptr for the root.
   Node* parent_ = nullptr;
@@ -273,23 +285,21 @@ class EdgeAndNode {
   int GetNStarted() const { return node_ ? node_->GetNStarted() : 0; }
   uint32_t GetNInFlight() const { return node_ ? node_->GetNInFlight() : 0; }
 
-  // Whether the node is known to be terminal.
+  // Whether the node is known to be terminal or certain.
   bool IsTerminal() const { return node_ ? node_->IsTerminal() : false; }
+  bool IsCertain() const  { return node_ ? node_->IsCertain()  : false; }
 
   // Edge related getters.
   float GetP() const { return edge_->GetP(); }
   Move GetMove(bool flip = false) const { return edge_->GetMove(flip); }
 
-  // Returns U = numerator * p / N.
-  // Passed numerator is expected to be equal to (cpuct * sqrt(N[parent])).
+  // Returns U, which in a sense measures the relative uncertainty that this
+  // move might be better than the leading move.
+  // U = numerator * p / N, where the passed numerator is expected to be equal
+  // to (cpuct * sqrt(N[parent])).
   float GetU(float numerator) const {
-    if (IsTerminal() && GetQ(0.0f) == 1.0f) {
-      // Never bother with alternatives to checkmate. In theory, U represents
-      // the "uncertainty that this move is actually better than the leading
-      // move", so when we have a certain winner, we should set all sibling U
-      // to 0, meaning MCTS is guided by value alone. That's complicated to do,
-      // however, so instead we create the same behavior with this:
-      return std::numeric_limits<float>::infinity();
+    if (IsCertain()) {
+      return 0.0f;
     } else {
       return numerator * GetP() / (1 + GetNStarted());
     }
--- a/src/mcts/search.h
+++ b/src/mcts/search.h
@@ -102,6 +102,7 @@ class Search {
   static const char* kCacheHistoryLengthStr;
   static const char* kPolicySoftmaxTempStr;
   static const char* kAllowedNodeCollisionsStr;
+  static const char* kStickyCheckmateStr;
 
  private:
   // Returns the best move, maybe with temperature (according to the settings).
@@ -174,6 +175,7 @@ class Search {
   const bool kCacheHistoryLength;
   const float kPolicySoftmaxTemp;
   const int kAllowedNodeCollisions;
+  const bool kStickyCheckmate;
 
   friend class SearchWorker;
 };


jjoshua2:
This looks good

dubslow:
Discordite OldWolf2, who originally reported that position wondering why Leela couldn't find a mate in 5/4 (mate in 5 from the FEN, Leela got the first move right but then couldn't find the followup mate in 4) reports that without this patch, it takes 1.38M nodes to find the mate in 4; with this patch, a mere 23K, a better than 10x improvement (and probably with better nps to boot).

dubslow:
Note that this patch only really makes an immediate impact in positions which are already clearly won, though it might have a minor impact on other tactical positions which aren't otherwise already won. I don't expect a big improvement in overall Elo.

DanielUranga:
Normal a/b engines usually substract the current search tree depth from the mate value, so the shortest mates are favoured. Does it make sense and is possible to apply that here?

dubslow:
No reason to. This could be considered a tweak to the interpretation of PUCT, more specifically `U`, leaving the rest of the MCTS intact. Adding special handling of mates to deliberately favor shorter ones would mean not following MCTS, and perhaps more importantly, is totally unnecessary: it's always been true that the search ought to favor shorter mates, because as Q gets averaged, longer mates will have more opportunities for Leela to throw/the opponent to make unexpected moves, so the Q will be lower for longer mates than for shorter mates anyways. The only thing this patch changes is the variability in that noise. Instead of taking millions or billions of nodes for that noise to average out towards favoring shorter mates, it will now do so in thousands of nodes.

Tilps:
Making eval depend on depth would invalidate tree reuse.  

I suspect this patch probably overlaps some work which VideoDr0me has done.  Its very similar to a patch I wrote back in the day for lczero as well, just more focused on checkmate moves.

dubslow:
Yes it has some minimal overlap, but this is only a slight reinterpretation of what `U` should be in the puct formula. Most of VideoDr0me's work involves adding other features/tweaks to MCTS itself, which is left intact here.

Most importantly, we don't "cheat" by going pre-searching for mates either in the current ply or in the next ply -- the network still has to produce a playout which searches the mate node before the U-reset takes place.

Tilps:
My patch (which i'm saying is kind of equivalent) instead increased Q by a value which was larger than anything U could calculate.

dubslow:
Yes I had considered that as well :) But I like this way better, since it can be described as a "more accurate" implementation of `U`, which is supposed to be upper confidence bound. Anything involving setting Q outside of [-1,1] would be a hack that (strictly speaking) violates MCTS.

In fact framing it in terms of `U`, one might make the argument that not doing this was a bug in our PUCT implementation, failing to set infinite upper confidence on totally certain moves...

dubslow:
It's worth noting that this change would probably significantly affect training.

Tilps:
I think that is an odd interpretation, as we become more confident through repeated visits, the U value decreases, but here you are proposing that 'infinite confidence' has an infinite U.  As for my case of increasing Q, its also an invalid interpretation, I chose it because it represented the idea that 'definitely mate now' has more value than 'maybe mate later'.

dubslow:
Yeah you're right, that is actually totally backwards. My bad. I guess it's just whatever is needed to make it ignore pointless alternatives.

dubslow:
I guess MCTS chooses a node to explore based on value and "uncertainty", in other words, more uncertain nodes get a higher exploration. But then what I'm saying here is for checkmate, further exploration is unnecessary -- so to effect that, we "tell the MCTS that the uncertainty is infinite", so that it never checks other nodes. I guess it is still a hack.

mooskagh:
For me it would be much cleaner to do in `PickNodeToExtend()` (rather than trying to trick PickNodeToExtend by providing hacky values from outside). In the edge selection loop, if you see a checkmate, just pick it as best and break.

dubslow:
How would you do that? Much of the point of this PR is that the checkmate doesn't "stick" until *after* it's been found "naturally" by a previous search. We return U=inf only if the node is already terminal, which is only done in ExtendNode, which comes after PickNodeToExtend. Changing PickNodeToExtend to examine the chess-contents of a given node (as opposed to mcts-contents) would be even more of a hack IMO. So what exactly did you have in mind for that?

dubslow:
Tilps also pointed out to me on Discord that this probably isn't suitable for training as is: although it's correct, so to speak, the potential problem is that if we train the policy against only one checkmate move when multiple are available, then we may accidentally blind the network to certain kinds of checkmate moves. This is only a problem at the root node though, for policy training. For any child nodes, this technique is fine.

mooskagh:
I cannot comment on training data difference (although I doubt it would matter),

but for the `PickNodeToExtend`, I suggested moving condition check `if (q==1.0f and is_terminal)` from inside of `GetU` to a place where that `GetU` is called.  (if would make sense to have a helper function `bool Node::IsCheckmate() const` which would do that `q==1.0f and is_terminal` check).

dubslow:
About the training part, wouldn't that be covered by Dirichlet noise? That ought to ensure that it's reasonably random which of multiple valid checkmate moves gets all the visits

dubslow:
Meh, I guess I still don't like that, because we would need to add that code *everywhere* that GetU is called, 4 times in SendMovesStats, once in PickNodeToExtend, and once in PrefetchIntoCache... and if we're putting it everywhere that GetU is called, why not put it straight into GetU?

dubslow:
I should think `SendMovesStats` should reflect the search that actually happened. Who cares what the "real" U is if that "real" U wasn't used to guide the search? I'd be fine with adding a separate is_terminal output though... (and in such cases I'd rather print no U at all than a value which is totally irrelevant to search)

And `PrefetchIntoCache` should never even dream about trying to prefetch an already-determined node. I guess I should add an error check for it to ignore terminal nodes and their siblings.

Tilps:
On the comment about Dirichlet noise, I don't think it will be sufficient give a first visit to anything other than an existing high policy node - Noise helps avoid all the visits going to the same place policy wants them to go.

killerducky:
@dubslow did you test this in matches to see if it gains Elo? @Tilps did a similar thing and I think he found it lost Elo.

Tilps:
It was pretty close to in the noise - I think I had variants which looked positive, but not much.  I hear VideoDr0me did better by doing something a bit more complicated.

AlexGreason:
While this should be a strict improvement in theory, in practice it would be a good idea to run some games to verify that it's at least not a regression to check if there's some error in implementation or unconsidered issue.

dubslow:
I'm streaming a cpu-speed check. I disagree that it's not a U thing, per the current comment in the source, but I'll give a whack at implementing it the other way. (And no finite U or Q would suffice -- that would leave an arbitrarily large N where visits would start being spread out again.)

And as mentioned earlier in the comments, this is not suspected to gain any immediate elo, engine-to-engine compared, only to convert wins substantially faster. In the long term, I think in training this will improve network strength, but in engine-to-engine comparisons, with identical networks, I don't expect it gains nearly any Elo.

dubslow:
`"Q+U" cannot be larger than "1 + Cpuct", so any Q or U > Cpuct + 1 would ensure the move will be chosen with any N.`

This is not true, U is unbounded: https://pastebin.com/raw/PcLi6eYh

mooskagh:
Indeed, growth with sqrt(total N).

dubslow:
crem, would the implementation sketched by the most recent commit be acceptable?

It is the theory-correct version which the first commit was a hacked approximation of.

Videodr0me:
@dubslow As you seem highly motivated to finally add certainty propagation to the main branch, I would propose we work together, with your expertise on the evolved codebas, I believe we can get it done in a day or two. I would have one requirement, that is to make it optional, for one it makes self-play testing easy, as its a quick way to expose potential bugs and because its not clear at all how this will affect training (which needs some thought and extensive testing). I would gladly walk you through my code, pointing out subtle things i initally missed and learned along the way. 

As for the new proposed change, I am a little afraid to say it, but its neither theoretically sound nor does it do anything different from the first proposal or what crem and i proposed. You are essentially setting U to 0 for the certain win node (which woud be theoretically correct, but is not done by DM for a reason, i'll come back to that later) and also set all siblings to certain making their U zero, too. Ultimately you decide the node to be picked to expand on  C only which naturally is highest for the win, reaching the same outcome as the other proposals just in a more convoluted way. But setting the siblings U to 0 is not theoretically correct at all as their U is what it is, and is not affected by a siblings. The only connection you make is by the argument I mentioned in my last post "If we are at that node we would only play the winning move", which falls (iin case of certain winning moves only slightly) short in an avering backpropagation scheme in ignoring the signal inherent in possibly visiting these siblings for earlier branch decisions. From an coding style perspective it also is awkward to have a method that calculates U be overloaded by what are essentially search decisions.

The only really sound way is to propagate the certainty up in the tree and not sideways. But again if you like I can walk you through the Winands paper and my Code so we can get it done properly.

dubslow:
For the purposes of this PR, I'm not interested in certainty *propagation*, only in the certainty for the sibling moves of the terminal move (and not their parent or grandparents).

>"If we are at that node we would only play the winning move", which fails (iin case of certain winning moves only slightly) short in an averaging backpropagation scheme, bnecause it ignores the signal inherent in possibly visiting these siblings for earlier branch decisions.

I'm sorry but I simply do not agree with this statement.

I guess having been outvoted 2-1, I'll change this to what crem originally suggested. I still don't agree that it is a feature of search, but I'm more interested in the effect than in the underlying theory.

(And how do you know that DM didn't do it this way, and if so, why not?)

Videodr0me:
Well, well this change will not get 90% of the benefits. Its 0 ELo vs. 10 Elo. Also I would advocate to only do it at root, because there its totally safe (no more signal for grandparents) and also gets leela to play an instant mate instead of teasing the opponent.

To clarify my statement about the signal, It just boils down to if some information can be yielded by also visiting the siblings. If in line A along a path with many nodes (grandparents, grandgrandparents etc) the  visited endnode has one terminal winning child and  all other childs are loosing, it makes sense (and MCTS does this) is to choose another line B early where lets say all childs of the Endnode are winning. We are basically in a deterministic sampling situation, and there is uncertainty in all the grandparents and grandgrandparents along the line. If the "winningness" of the final node correlates with the "goodness" of the hole line then changes like yours that only always visit the winning child ignore this info. But as I said for winning nodes its probably a non-issue and you get +/- 0 Elo or an insignificant loss. But as you can safely do it at root without any issues (not even theoretical) why not just do it there. 

You can see the effect of this signal most clearly in avoiding terminal loosing nodes - much the same of your argument can be made here: In your language the Q+UCB of the certain loss should be -1, and all siblings should be preferred. The thing is if you do that you loose measurable elo. The reason is that even an "unforced" loss is a signal that the hole line is bad, and its better to find out early.

For terminal draws a similar argument holds, here also the UCB+Q should be 0 at all time. And heres the reason that DM did not do that (they never mention checking any siblings like you do in any of their papers) lets say there are three moves at root, one first visitied leading to a terminal draw, and two moves with an initial Q from the NN of -0.02. Now the UCB would now with a high probability (ceterus paribus of policy) visit the -0.02 moves, until their ucb pushes them to lets say -0.01.... now if we stop search here... these moves got more visits than the initial draw we only visited once. So it would make Leela more optimistic and even sometimes not take terminal draws. Now is this theoretical behaviour a bad thing? No, only testing can show if its good or bad. And at least from my testing it seems that doing this (setting the U to 0 for terminal draws) is a measurable elo loss.

dubslow:
Changed to crem's suggested implementation, with short circuiting in search and extra field in SendMovesStats.

```|   _ | |
|_ |_ |_| v0.16.1-dev built Jul 24 2018
position fen k2r1b2/P3qp2/7r/1Q6/1P6/6P1/4PPp1/R1R3K1 w - - 0 28 moves b5d5 h6c6 c1c6 d8d5 c6c8 a8b7 a7a8q b7b6
go nodes 10000
Creating backend [blas]...
BLAS vendor: OpenBlas.
OpenBlas [NO_LAPACKE DYNAMIC_ARCH NO_AFFINITY Sandybridge].
OpenBlas found 8 Sandybridge core(s).
OpenBLAS using 1 core(s) for this backend.
BLAS max batch size is 256.
info depth 1001 seldepth 2 time 116 nodes 2 score cp 12800 hashfull 0 nps 17 pv a8c6
info depth 1001 seldepth 2 time 384 nodes 10000 score cp 12800 hashfull 0 nps 26041 pv a8c6
info string a1f1  (4   ) N:       0 (+ 0) (P:  0.10%) (Q:  0.99998) (U: 0.11986) (Q+U:  1.11983) (V:  -.----) (T: 0) 
info string a8a3  (1602) N:       0 (+ 0) (P:  0.15%) (Q:  0.99998) (U: 0.18027) (Q+U:  1.18024) (V:  -.----) (T: 0) 
info string a8a5  (1606) N:       0 (+ 0) (P:  0.18%) (Q:  0.99998) (U: 0.21496) (Q+U:  1.21494) (V:  -.----) (T: 0) 
info string a8a4  (1604) N:       0 (+ 0) (P:  0.19%) (Q:  0.99998) (U: 0.22403) (Q+U:  1.22401) (V:  -.----) (T: 0) 
info string a1e1  (3   ) N:       0 (+ 0) (P:  0.21%) (Q:  0.99998) (U: 0.25110) (Q+U:  1.25107) (V:  -.----) (T: 0) 
info string a1d1  (2   ) N:       0 (+ 0) (P:  0.22%) (Q:  0.99998) (U: 0.26399) (Q+U:  1.26397) (V:  -.----) (T: 0) 
info string a8a2  (1600) N:       0 (+ 0) (P:  0.22%) (Q:  0.99998) (U: 0.26763) (Q+U:  1.26761) (V:  -.----) (T: 0) 
info string g1h2  (157 ) N:       0 (+ 0) (P:  0.23%) (Q:  0.99998) (U: 0.28005) (Q+U:  1.28003) (V:  -.----) (T: 0) 
info string f2f4  (351 ) N:       0 (+ 0) (P:  0.25%) (Q:  0.99998) (U: 0.29687) (Q+U:  1.29685) (V:  -.----) (T: 0) 
info string c8c7  (1660) N:       0 (+ 0) (P:  0.25%) (Q:  0.99998) (U: 0.30212) (Q+U:  1.30210) (V:  -.----) (T: 0) 
info string a1a7  (19  ) N:       0 (+ 0) (P:  0.25%) (Q:  0.99998) (U: 0.30375) (Q+U:  1.30373) (V:  -.----) (T: 0) 
info string a8a7  (1611) N:       0 (+ 0) (P:  0.27%) (Q:  0.99998) (U: 0.32669) (Q+U:  1.32667) (V:  -.----) (T: 0) 
info string a1b1  (0   ) N:       0 (+ 0) (P:  0.30%) (Q:  0.99998) (U: 0.36568) (Q+U:  1.36565) (V:  -.----) (T: 0) 
info string f2f3  (346 ) N:       0 (+ 0) (P:  0.31%) (Q:  0.99998) (U: 0.36726) (Q+U:  1.36724) (V:  -.----) (T: 0) 
info string a8b7  (1612) N:       0 (+ 0) (P:  0.32%) (Q:  0.99998) (U: 0.38659) (Q+U:  1.38656) (V:  -.----) (T: 0) 
info string c8c1  (1645) N:       0 (+ 0) (P:  0.33%) (Q:  0.99998) (U: 0.39404) (Q+U:  1.39401) (V:  -.----) (T: 0) 
info string c8c3  (1647) N:       0 (+ 0) (P:  0.33%) (Q:  0.99998) (U: 0.39703) (Q+U:  1.39700) (V:  -.----) (T: 0) 
info string a1a5  (15  ) N:       0 (+ 0) (P:  0.34%) (Q:  0.99998) (U: 0.40200) (Q+U:  1.40197) (V:  -.----) (T: 0) 
info string g1g2  (156 ) N:       0 (+ 0) (P:  0.34%) (Q:  0.99998) (U: 0.40262) (Q+U:  1.40259) (V:  -.----) (T: 0) 
info string e2e3  (317 ) N:       0 (+ 0) (P:  0.34%) (Q:  0.99998) (U: 0.40852) (Q+U:  1.40849) (V:  -.----) (T: 0) 
info string e2e4  (322 ) N:       0 (+ 0) (P:  0.34%) (Q:  0.99998) (U: 0.41113) (Q+U:  1.41110) (V:  -.----) (T: 0) 
info string a1c1  (1   ) N:       0 (+ 0) (P:  0.35%) (Q:  0.99998) (U: 0.41719) (Q+U:  1.41717) (V:  -.----) (T: 0) 
info string b4b5  (695 ) N:       0 (+ 0) (P:  0.35%) (Q:  0.99998) (U: 0.42594) (Q+U:  1.42592) (V:  -.----) (T: 0) 
info string c8c5  (1651) N:       0 (+ 0) (P:  0.36%) (Q:  0.99998) (U: 0.42971) (Q+U:  1.42969) (V:  -.----) (T: 0) 
info string c8d8  (1665) N:       0 (+ 0) (P:  0.36%) (Q:  0.99998) (U: 0.43146) (Q+U:  1.43143) (V:  -.----) (T: 0) 
info string c8c4  (1649) N:       0 (+ 0) (P:  0.37%) (Q:  0.99998) (U: 0.44522) (Q+U:  1.44519) (V:  -.----) (T: 0) 
info string c8e8  (1666) N:       0 (+ 0) (P:  0.40%) (Q:  0.99998) (U: 0.47545) (Q+U:  1.47543) (V:  -.----) (T: 0) 
info string g3g4  (615 ) N:       0 (+ 0) (P:  0.40%) (Q:  0.99998) (U: 0.48252) (Q+U:  1.48249) (V:  -.----) (T: 0) 
info string c8f8  (1667) N:       0 (+ 0) (P:  0.44%) (Q:  0.99998) (U: 0.52319) (Q+U:  1.52316) (V:  -.----) (T: 0) 
info string c8c2  (1646) N:       0 (+ 0) (P:  0.45%) (Q:  0.99998) (U: 0.53746) (Q+U:  1.53743) (V:  -.----) (T: 0) 
info string a1a3  (10  ) N:       0 (+ 0) (P:  0.45%) (Q:  0.99998) (U: 0.53976) (Q+U:  1.53973) (V:  -.----) (T: 0) 
info string a1a4  (13  ) N:       0 (+ 0) (P:  0.45%) (Q:  0.99998) (U: 0.54556) (Q+U:  1.54554) (V:  -.----) (T: 0) 
info string a1a2  (7   ) N:       0 (+ 0) (P:  0.49%) (Q:  0.99998) (U: 0.58206) (Q+U:  1.58204) (V:  -.----) (T: 0) 
info string a8b8  (1614) N:       0 (+ 0) (P:  0.85%) (Q:  0.99998) (U: 1.02295) (Q+U:  2.02293) (V:  -.----) (T: 0) 
info string a8d5  (1607) N:       0 (+ 0) (P:  1.00%) (Q:  0.99998) (U: 1.19645) (Q+U:  2.19643) (V:  -.----) (T: 0) 
info string c8b8  (1664) N:       0 (+ 0) (P:  1.36%) (Q:  0.99998) (U: 1.63489) (Q+U:  2.63487) (V:  -.----) (T: 0) 
info string a8a6  (1608) N:       0 (+ 0) (P:  6.10%) (Q:  0.99998) (U: 7.31839) (Q+U:  8.31836) (V:  -.----) (T: 0) 
info string a1a6  (17  ) N:       0 (+ 0) (P: 12.85%) (Q:  0.99998) (U: 15.42125) (Q+U: 16.42123) (V:  -.----) (T: 0) 
info string c8c6  (1655) N:       0 (+ 0) (P: 21.52%) (Q:  0.99998) (U: 25.82792) (Q+U: 26.82790) (V:  -.----) (T: 0) 
info string a8c6  (1610) N:    9999 (+ 0) (P: 45.98%) (Q:  1.00000) (U: 0.00552) (Q+U:  1.00552) (V:  1.0000) (T: 1) 
bestmove a8c6```

The problem here is that actually a8a6 is also a terminal node, but never gets expanded, so it shows (T: 0). This is the same problem as the training problem that Tilps mentioned.

**Comment by Videodr0me** [comment]
@dubslow 
I think you still miss one crucial point in mcts, all backpropagated Q yield information about the whole line. If you suddenly limit this by always visiting a winning child, or by always avoiding a loosing child (which only can occur if you propagate the win of one of the grandchilds as a loss to that child), you loose some information. It does not matter if it will or will not have been played. How large this signal is, is another matter and can be tested. So there are no assumptions or axioms, its a plain fact that information is lost. Only, the value of that information can be disputed.

You claim it does not matter, while my testing revealed that it does matter greatly for avoiding loosing nodes and also for always visiting winning childs. Granted in the latter case its almost equal with only a slight loss within the margins of error, but if your theory were true it would be a plus - and that i do not see in the data.  And the term "signal" is not my invention, its exactly what Winands describes in their paper.

I think, that all this is not that important, if you want to err on the cautios side you only do it at root, if you want a little larger effect in playing mates more quickly you can do it everywhere in the tree but might loose a fractional amount of elo. In both cases it should be done in the picknodetoextend loop.

**Comment by dubslow** [comment]
> I think you still miss one crucial point in mcts, all backpropagated Q yield information about the whole line. 

I agree. I also hold that non-mate-siblings-to-terminal-nodes aren't "part of the whole line". The line ends at the terminal node.

> always visiting a winning child

Given my previous sentence, this does not lose information.

> always avoiding a loosing child (which only can occur if you propagate the win of one of the grandchilds as a loss to that child)

That is outside the scope of this PR. It neither implements this, nor was it ever intended to, and if you thought I intended to, then my mistake for not communicating that.

> It does not matter if it will or will not have been played.

On the contrary, this is the fundamental point of importance. MCTS is designed to find good moves in the face of uncertain replies by opponents. The entire premise upon which MCTS is designed, is that we can't predict how opponents will play. Therefore, *for nodes where the opponent never gets a chance to play*, MCTS is irrelevant.

I do not claim to fully understand what you implemented or what you tested, but I stand by the above sentences. This PR does nothing more than eliminate irrelevant cruft from the game tree.

**Comment by Videodr0me** [comment]
> The entire premise upon which MCTS is designed, is that we can't predict how opponents will play. Therefore, for nodes where the opponent never gets a chance to play, MCTS is irrelevant. In other words, this PR only changes behavior that is outside the scope of MCTS.

This is plain wrong on so many levels. If we really couldn't predict how opponents will play, then MCTS  would not work at all. Obvioulsy it does, so we can somewhat predict opponents moves better than chance would allow - this is what the whole policy head is for, trying to predict moves, which in turn is improved by MCTS.  

>Therefore, for nodes where the opponent never gets a chance to play, MCTS is irrelevant. 
This is plain wrong. With the change only the terminal win is visited and only 1.0 are backpropagated. Without the change sometimes other siblings are visited and other values are backpropagated. As backpropagated Qs are used for further to decide which node to expand next. The opposite is true visits of these nodes change behaviour of MCTS and are relevant.

You failed to even adress the training target issues, so i would strongly advise to first make it optional and not use it in training before thouroug testing.

**Comment by dubslow** [comment]
> With the change only the terminal win is visited and only 1.0s are backpropagated. 

This is only true assuming the network has good policy. *Only after the terminal node is itself found* will it overwrite its siblings.

> Without the change sometimes other siblings are visited and other values are backpropagated. 

They should not be (after having found a mate), these other siblings, these other values, are irrelevant to the game, assuming that the neural network is trying to win. I can't imagine any circumstance under which we aren't trying to win (and that doesn't include things like training a network to only draw, that's merely a redefinition of what "favorable terminal node" means, the same concept applies).

> As backpropagated Qs are used for further to decide which node to expand next, visits of these nodes change behaviour of MCTS and are relevant.

They change the behavior in our code, for now, but are irrelevant to the abstract MCTS algorithm.

> You failed to even adress the training target issues, so i would strongly advise to first make it optional and not use it in training before thorough testing.

This is correct, this PR is currently not in merge-able form. I would like a further review from @mooskagh and @Tilps before deciding how to proceed as regards to training (either disable this or better track all available terminal nodes).

**Comment by Videodr0me** [comment]
> They should not be (after having found a mate), these other siblings, these other values, are irrelevant to the game, assuming that the neural network is trying to win. 

This is wrong, as we are not yet at that node where we can play the terminal win. I am again pointing to my initial example. Which alternative A or B would you trust more? A that leads deep down in the tree over many nodes to a node which has only one terminal winining child, and a lot of Q=-0.99 (near loosing) childs or B which over many nodes leads to a node whose children are all terminal wins? I would rather trust B and not A, even though if i would finally reach those pre-terminal nodes it would not matter - but there is no gurantee that we reach them, we can only take previous visits as an indicator for the whole path. You can only argue that the signal is so weak that it hardly matters - and thats what i get from testing: It matters but only very very slightly, but enough so to outweigh the positive aspects (i am not saying that there are no positives of this change). So maybe thats what we argue about, the size of the effect, you say it does not matter at all elo wise while i was surprised to find in testing that instead of a very slight gain i found it to be at best equalish if not a minute loss.

**Comment by dubslow** [comment]
> This is wrong, as we are not yet at that node where we can play the terminal win.

What? Yes we are! That's the only node that's ever been in question in this PR, is exactly the node where we can play the winning move!

**Comment by dubslow** [comment]
Unless are you referring to something like the position in the original comments?

`position fen k2r1b2/P3qp2/7r/1Q6/1P6/6P1/4PPp1/R1R3K1 w - - 0 28 moves b5d5 h6c6`

The "best" move for the winning side here -- in the sense of fewest moves to forced mate -- is c1c6, with mate following at most 4 moves later. MainLeela, under current master, takes millions of nodes to "find" this. With this PR, it takes a couple tens of thousands of nodes to find the shortest win, because it evaluates the final position, where one move gives mate regardless of the opponent, as won, since it is in fact won. It does *not* ignore the alternatives to any move before the final mate, as I've emphasized over and over again.

In fact I would daresay that the current implementation is bugged, and quite possibly a reason that Leela is so poor at tactical mates, because those are the kinds of mates you specifically try to avoid with your example-that-I-don't-fully-understand. If you find yourself in a position where one move gives mate, but any other mate loses the game, that's currently evaluated as a bad position -- even though it's a won position!

**Comment by dubslow** [comment]
Here's a clearer description of my point.

Consider this position: https://lichess.org/h22QmY2u#44 (given by fen r4r1k/ppqb1B1p/4p3/4p1Q1/3P1n2/2P4R/P4PPP/2R3K1 w - - 4 23)

How do you as a player evaluate this position? Hint: after finding the only mating move (Qf6#), the answer is "the evaluation of this position is 1". Nothing else. There is no possible alternative. After you find the mate, this position is *totally won*. Nothing you say about MCTS or heuristics or anything else matters -- *this position is won*. Stockfish says it's won, any human who can spot the mate move will say it's won, because it is totally, deterministically, without uncertainty, infinite confidence won. After finding the mate, MCTS suddenly goes out the picture, it is irrelevant. That position on the board is a win for white, no matter how you slice it.

Therefore, I say it's a bug that Leela doesn't report Q=1 when examining this position after finding the mate.
Example output. The average value printed at the end is the weighted-by-visits average of all the searched move Qs, as is normally done and propagated up the tree. This is for current master. https://pastebin.com/raw/G4kct4Fh

Note how after 1000 nodes, after finding the mate, Leela still only gives a 0.924 evaluation for a totally won position. After 10000 nodes, Leela only gives a 0.988. That's ridiculous, and matters a *lot* at the fringes of deeper searches far from possible tactics, where near-leaf nodes only get dozens of playouts.

Meanwhile, for the fixed version from this PR:

https://pastebin.com/raw/3KCeTqx1

0.923 -> 0.996,  0.988 -> 0.9996. Arguably, according to the strictest interpretation, I should remove the pre-mate-found Q inputs to the average so that the parent node sees strictly Q=1 for this position, but that's *way* more difficult and pointless code, when merely restricting further playouts does the same job just fine. And the examples listed in earlier comments demonstrate how in less cliff-edge circumstances, current master can actually *lose* Q-value by adding more playouts to a *strictly won mate-in-right-now* position. 

So how  would you evaluate this position? I very much hope you evaluate it as Q=1.

(To be clear: this PR does *not* propagate this `Q=1` any further up the tree. Any possible "route through the tree" that leads to this position will be weighted-averaged according to all the alternative destinations *besides* this position. But *the position itself has Q=1*, without doubt, without room for interpretation of MCTS or propagation or uncertainty or whatever.)

**Comment by killerducky** [comment]
Sorry I didn't follow all this discussion. But is it true the Videodr0me and Tilps both have results that say something like this did not gain Elo? @dubslow do you have any different results? Can you run some tests?

**Comment by dubslow** [comment]
In selfplay, this will gain approximately zero elo, in an engine-vs-engine comparison. You can watch my twitch stream for a comparison.

What this *will* do is, when a position is already won, find and play moves towards quick checkmates, in other words, curb the "playing with its food behavior" that leela sometimes does. I suspect it *might* also be useful in tactics against AB engines, but that is uncertain as well. (potato_chess on Twitch is testing right now).

**Comment by dubslow** [comment]
Anyways, I just pushed a hypothetical implementation of equalizing the policy for multiple checkmate moves in training. Please review, Tilps.

**Comment by Videodr0me** [comment]
@killerducky There is no dispute over the finding that it does not gain any elo. I think my testing and Dubslows asre in agreement here. I did many runs back when i tested that and never got anything outside the margins of error, and even if i did the next run would point in the opposite direction. I had slightly more negative runs than positives, though. 

@dubslow I had the exact same notion about this, as you do when I started looking at this. Only later when i dug deeper i came to understand that this reasoning of "moves that would not have been played" should not get visited is not the whole story. And to answer one of your comment: No its not bugged! Its just one of the many degrees of freedom MCTS offers.

This is a discussion very much worth having, because it will become much more relevant if and when i merge my code in to main (though it will remain optional, because of the unsettled training issues). And its great that Dubslow is willing to go into the details here.

Let me first try a new approach to explain some aspects which maybe i neglected to put some emphasis on :

1) There are advantages from always choosing a terminal win, not only because of human aesthetics (playing a mate quicker without "dancing" aroung), but also it should gain Elo, because it might stumble on a low policy mate more quickly.  The example positions Dubslow show go in this direction. Yet, elo gain is neutral.

2) The disadvantage is that some signal is lost. I think by now we are on the same page that with current implementation (whether you call it bugged or not) is affects search and could potentially loose its ability to discriminate between lines A or B. Dubslow thinks this could not matter at all, I think that his argument is game theoretically flawed and there is a tradeoff. And this tradeoff is exactly the reason that the gains from (1) are nullified or possibly even slightly outweighed.

3) In other Games, the tradeoff might be different. There it might gain a lot or with other games might loose a lot more. This tradeoff is also dependend on our policy predictor, if it would be completely blind to mates it would surely help. And yet elo gain is neutral. I tested mainly with ID350, and policy did not get worse, that this would have changed.

4) Humans tend to judge play on their aesthetic subjetive notions, which bias us to favour the "mating combinaton". Where as Leela is oblivous to that, and just needs to win, even if that drags out the game. Again if this would be meaningful in an objective way, elo would suffer, it does not.

5) The whole discussion, is forshadowing the same discusiion for avoiding a loosing node (which currently does not happen, but will if we propagate certainty). And there the exact same argument should hold, as for winning nodes, because who would ever "play a move that leads to a certain loss".  And here the data is completely unambigous, its a bad idea to avoid visiting certain loosing (by setting U to 0, whch in this case has even some theoretical merit). I know that we are not doing this with the proposed change, but thinking about this, reveals that the same mechanism is at play.

6) Whatever you think, it should be an option and not standard behaviour until its affect on training is tested.

File: src/mcts/node.cc [hunk 1]
```diff
@@ -237,20 +237,40 @@ uint64_t ReverseBitsInBytes(uint64_t v) {
...
   result.version = 3;
 
   // Populate probabilities.
-  float total_n = static_cast<float>(
-      GetN() - 1);  // First visit was expansion of it inself.
   std::memset(result.probabilities, 0, sizeof(result.probabilities));
-  for (const auto& child : Edges()) {
-    result.probabilities[child.edge()->GetMove().as_nn_index()] =
-        child.GetN() / total_n;
+  if (checkmate) {
```
mooskagh: Why is that change needed?
If something less trivial than 5 lines change is needed to get nearly zero Elo improvement, we should not push that. That complicates the code with no benefit.

Videodr0me:
@mooskagh I think this change wants to make it more compatible with training in giving all terminal wins equal visits, One problem might be that it equalizes vistis for terminals but does not produce visits for Q=+0.99 childs. And yes its a 0 elo change (within the margins of error) - but i can fully understand Dubslow for pursuing this. Because if you follow his line of thought (and if you would have asked me 4 month ago, i would probably have argued similarily) it should never loose elo, so even while we do not measure a gain, thats only because the margins of error are very large for such a change and many many games would be needed to push it into positive territory. Thats why he is defending it rigerously. And if he would be correct about this, it would be a ever so slightly gain with no downside. My understanding of mcts and the experiments i did over time convinced me that things are not that simple, and that the results are not only due to the margin of error but also because there is a tradeoff here. Basically thats what we argue about.

As for the change: I would propose leaving the simple version in as an option. That option should be disabled in training - and thus no need for the mentioned adjustment. Or we leave it for the time being and revisit it with a more complete mcts-solver implementation.

src/mcts/node.cc
dubslow: This is exactly what Tilps was worried about earlier. The extensive comment describes why.

dubslow: And I wrote a comment a few up that I believe this patch could lead to training strength gains, even if in a same-weight comparison it doesn't add elo, it does affect training, and in a positive way IMO.

dubslow:
You continue to insist that moves that are irrelevant to the outcome of the game are meaningful, and that you held this same view yourself once, but have offered zero data for the other conclusion -- only highly unconvincing philosophical arguments. What is the result of these experiments? Numbers, code. Is it in your repository? I kind of assume so, but haven't checked recently. (For instance, this quote `The disadvantage is that some signal is lost. I think by now we are on the same page that with current implementation (whether you call it bugged or not) is affects search and could potentially loose its ability to discriminate between lines A or B.` -- I have repeatedly stated that this PR does not change the calculation of *lines* leading to mate. Lines A or B of length more than one ply are still fully searched as before. No signal is lost.)

I have long commented -- you ought to remember me mentioning this to you on Discord a month or two ago -- that your data necessarily will always be complete, because you conduct your experimental code modifications *only with nets that have not been trained on the modified code*. That is a severe shortcoming in actually being able to judge the merit of any of your tests. Even if they don't show Elo in same weight, engine-to-engine comparison, there still could be (probably would be?) noticeable Elo gain after applying the code changes to the training data and watching how the network changes. Of course it's not really practical for you to test that yourself, but my point is that none of your results to date are anywhere near definitive.

Videodr0me:
> and that you held this same view yourself once, but have offered zero data for the other conclusion -- only highly unconvincing philosophical arguments. What is the result of these experiments? Numbers, code. Is it in your repository? 

https://github.com/Videodr0me/leela-chess-experimental/blob/3ed4fad87df06234e9d0b4cc286e41b68d694c0c/src/mcts/search.cc#L739

This is from a june 2 commit. Nice thing that github does not forget anything. Thats the exact same change we are discussing here. And the only philosophical statement in this thread is yours about the "abstract MCTS algorithm":

> They change the behavior in our code, for now, but are irrelevant to the abstract MCTS algorithm.

Your own experiments confirm mine and yielded the +-0 elo gain as you stated here and on discord. I believe others tested this too with the same result. Everybody can easily test this in self play. 

> Even if they don't show Elo in same weight, engine-to-engine comparison, there still could be (probably would be?) noticeable Elo gain after applying the code changes to the training data and watching how the network changes. 

This is true. As its a trade-off there is a theoretical possibilty that the freed capacity of finding mate patterns actually helps the NN and offsets the lost signal. I do not have any opinion on whether that might actually reflect in elo in a meaningful way. 

I think we are at a point where its moot trying to convince each other, you believe current MCTS is bugged because it does not force visits to terminal wins. I say its a deegree of freedom you have in applying MCTS that has its pros and cons.

dubslow:
I've stashed the training commit in [a separate branch](https://github.com/dubslow/lc0/tree/terminalsearch-training) for later reconsideration. I still believe it will yield a noticeably stronger net.

In the meantime, I added an option to control this feature, default off for now.

src/mcts/search.cc
```diff
@@ -86,6 +87,7 @@ void Search::PopulateUciParams(OptionsParser* options) {
                             "policy-softmax-temp") = 1.0f;
   options->Add<IntOption>(kAllowedNodeCollisionsStr, 0, 1024,
                           "allowed-node-collisions") = 0;
+  options->Add<BoolOption>(kIgnoreNonmateStr, "ignore-nonmate") = false;
```
mooskagh: "Ignore nonmate = false" is a triple negative!
Maybe come up with some "shiny" parameter name?
Like.. "sticky-checkmate" or at least maybe "visit-mate-alternatives" (=true by default)

dubslow: I was definitely not so sure about the name :) Either of those are fine by me, I guess I lean towards the shorter one by default

dubslow:
Current results of strength test I'm running:

```
Score of lc0 t9 9155 checkmate vs lc0 t9 9155 master: 121 - 132 - 461 [0.492]
Elo difference: -5.35 +/- 15.16

714 of 2000 games finished.
```

Inconclusive.

I asked potato_chess to test it, and it came up about 7 games short of the master-gauntlet, but then again he's redoing the same net-and-master gauntlet with longer TC, which is supposed to favor Leela,   and that's currently on pace to fall short of the original gauntlet by several games as well, so that one also strikes me as inconclusive. Much further testing is required. What is certain is that the strength change, whether loss or gain, isn't of great magnitude (probably less than 5 elo either way), as we all expected.

I still expect that if/when this is implemented in *training*, it will result in a relatively stronger net compared to current. All these tests on networks trained without this change aren't very meaningful as to the how the strength would improve (or regress) with the change in training.

With approval, I'm merging now with default off. Maybe this will make it easier to engage others' help on Discord to get the thousands of games needed to truly determine how safe it is to activate this by default.

Videodr0me:
@dubslow Just to be fair you should mention that this was not your original argument at all, your original argument was that this change corrects a bug in MCTS, and thus should be strictly dominant to the old way. The addtional testing now points more in a negative direction (even if ever so slight and inconclusive) - which again supports my view that there are tradeoffs involved. You now seem to accept this tradeoff and seem to hope that when training on the change, it might edge out a small plus. This is also fair, and might possibly work, But it could equally get worse. For example if other +0.99 NN eval childs are no longer visited, and this skews the distribution in a way that leela generalizes winning moves in slightly different positions less well. Anyhow, i am looking forward to the training results and i sincerely hope its a plus then, because the same tradeoffs are in play for a certainty propagation up the tree - only there the good already outweighs the bad. But an additonal boost is always welcome. 

So please let me know as soon as you have results.

dubslow:
`your original argument was that this change corrects a bug in MCTS`

I maintain this is true. MCTS converges in the infinite limit to perfect play -- and as I mentioned in my post, it is *currently* true that given sufficiently many visits, Q converges to 1 for all checkmate-producing moves, and in particular, the fraction of visits received by non-checkmate moves converges to 0. That is to say, MCTS converges to eliminating all information whatsoever from non-checkmate moves; in other words, these noncheckmate moves offer zero information. All this PR does is speed up the convergence.

That the data hints at a -5 elo loss is nothing more than noise so far. And, even if it is a true strength loss (which I still doubt, more data needed), that does *not* mean that the other nodes actually do generate information, for mathematically they do not (again, MCTS is proven to converge to perfect play, and alternatives to mate converge to 0% of the visit share). All it means is that we are using a search technique that the network wasn't trained for. (In other words, although all my statements about converged infinite-playout MCTS are true, in reality the Leela NN is only a finite approximation, and that approximation may or may not gain 3-5 elo by using mathematically irrelevant information as a heuristic, and using this technique where not so trained may mess up those approximation heuristics, which again, I'm still not convinced of that either. That still doesn't mean that this simplification is incorrect, it just means we haven't run the proper tests yet.)

My results will be ready once I've convinced someone on Discord to generate the data :) should be easier after next release, when option is available by default.

Videodr0me:
@dubslow You stick to your guns. Fair enough! I am convinced that you are wrong. With infinite visits MCTS with or without this change will converge. So this case is irrelevant. Sometimes the change helps to find a good move faster, and sometimes it does not (this latter part you do not share). I also get that you strongly believe that we are not seeing the elo gains according to your view, because of the margin of error.  Would you agree that if i showed you one example of a tree where with a fixed search of lets say between 5 and 10 visits, the changed selection rule would select an inferior move earlier in the tree, that then your whole argument is falsified?
Because then it would boid down to a tradeoff (what i am saying all along) that there are x positions where it helps but also y positions where it might hurt. In your world there should be no such cases. Agreed?

About training: This is actually a good thing if this whole discussion will lead to trying this in training at some point, because the tradeoff might shift and that would be even better for full certainty propagation. But again it might not. Buts its a worthwhile experiment nonetheless. Also, even if it gains some elo, I would not expect too much, at most 3-5 Elo over 10.000 games (unfortunately we need this many). And it might easily go the other way.

mooskagh:
This is pretty small change, so if within a few weeks there will be minimal proven benefits (like 5 Elo), it's fine turn it on by default and remove the flag.

However if there will be no benefit, it's best to revert the change to simplify the code.

dubslow:
`With infinite visits MCTS with or without this change will converge. So this case is irrelevant.` On the contrary, it tells us that your argument of "the non-mate alternatives provide valuable information" is wrong, at least in theory if not necessarily in practice.

And I'm not sure there are elo gains, at least not without also changing training; what there ought to be is shortened, more direct (and more aesthetic) end games.

`Would you agree that if i showed you one example of a tree where with a fixed search of lets say between 5 and 10 visits, the changed selection rule would lead to MCTS play an inferior move earlier in the tree, that then your whole argument is falsified?`

No, and see my previous comment why: just because the network may have taught itself to accidentally use non-relevant information in its heuristics doesn't have anything to do with the mathematical validity of the argument. That said, if you do have a case as you describe, I'd *love* to see it. I strongly doubt such a scenario exists, in theory or practice.

`Because then it would boil down to a tradeoff (what i am saying all along) that there are x positions where it helps but also y positions where it hurts. In your world there should be no cases that hurt. Agreed?` Again, it comes down to theory vs practice. Mathematically speaking, there is no tradeoff, and no such positions y. In practice, in finite MCTS with a NN which has been trained incorrectly, the NN might rely on information which is theoretically irrelevant to make a better move, so such positions y *might* exist, or might not. Again, they would be eliminated when this change is implemented in training.

As for reverting, this patch is not expected to gain much or any elo -- what it does do is find shorter, more direct, and more aesthetic mating patterns, especially as there are more and more pieces on the board. I'll try and provide some data that bears this out.

Videodr0me:
@Dubslow Sorry, but you are contradicting yourself in every two half sentences. How can a method that according to your argument should be in theory strictly dominant -  no matter what NN you use - because remember you claim MCTS is bugged -  ever be worse in practice? This is plain illogical. Either it is game theoretically strictely dominant regardless of policy and eval or it is not. Otherwise there is a tradeoff, exactly the tradeoff you dispute..

> And I'm not sure there are elo gains, at least not without also changing training; what there ought to be is shortened, more direct (and more aesthetic) end games.

Really, are we talking (human) aesthetics now?

> Mathematically speaking, there is no tradeoff, and no such positions y. 

And now watch how you contradict yourself:

> In practice, in finite MCTS with a NN which has been trained incorrectly, the NN might rely on information which is theoretically irrelevant to make a better move, so such position y might exist

What is it now? A mathematical impossibility or something that might happen? For the theoretical properties of MCTS the NN is irrelevant it could be random. Either you conceed that it is a mathematical possibility then we can talk how often such positions occur and the tradeoff involved or you still claim they cannot possibly exits. Lets quote you on this:

> In fact framing it in terms of U, one might make the argument that not doing this was a bug in our PUCT implementation, failing to set infinite upper confidence on totally certain moves...

or

> but the terminal node's siblings, do not matter. The siblings of the move before mate matter; the siblings of the mate itself are totally irrelevant to the rest of the tree.

or

>  Therefore, for nodes where the opponent never gets a chance to play, MCTS is irrelevant. In other words, this PR only changes behavior that is outside the scope of MCTS.

or 

> They should not be (after having found a mate), these other siblings, these other values, are irrelevant to the game, assuming that the neural network is trying to win

or

> They change the behavior in our code, for now, but are irrelevant to the abstract MCTS algorithm.

or

> Therefore, I say it's a bug that Leela doesn't report Q=1 when examining this position after finding the mate. 

or

>  I have repeatedly stated that this PR does not change the calculation of lines leading to mate. Lines A or B of length more than one ply are still fully searched as before. No signal is lost.)

or

>  That is to say, MCTS converges to eliminating all information whatsoever from non-checkmate moves; in other words, these noncheckmate moves offer zero information. All this PR does is speed up the convergence

You can't have it both ways. If it is as "abstract", "theoretical or "mathematical" as you claim then these positions cannot exits not even as an constructed example.

> In fact I would daresay that the current implementation is bugged, and quite possibly a reason that Leela is so poor at tactical mates, because those are the kinds of mates you specifically try to avoid with your example-that-I-don't-fully-understand

Thats why I insist that you agree, to the notion that if such an example exits (i maybe falsely assumed that this is obvious from my outline examples already given) you conceed that as falsification of your original argument. Because i want to spare us calculating  a 20 node tree and 15 visits If you then just wave your hands and say "this proves nothing".  But i can also fully understand if you do not want to take this further.

dubslow:
I guess I haven't made clear the differences between theory and reality. There are several potential scenarios to consider.

	NN trained with alts to mate  |	NN trained without alts to mate  |	Infinite MCTS = Perfect Play (with nonperfect NN)
--- | --- | ---
Engine searches alts to mate	 | Elo=x, mating sequence length=y	 | Elo=x+e-ε, n unknown (maybe dozens?), ε putative, mating sequence length=y'<y | 	Elo=infinite, nodes-to-converge-to-shortest-mate=n
Engine ignores alts to mate | 	Elo:x+ε (putative), mating sequence length=y'<y | 	Elo=x+e (maybe dozens?) mating sequence length=y''<y'<y | 	Elo=infinite, nodes-to-converge-to-shortest-mate=n'<n

I'm sure you won't find it hard to fit my various statements into this nice and easy table. In theory, alts to mate offer nothing. *In practice*, the NN might accidentally be trained to use it in some heuristic manner, much like the history planes. They're totally unnecessary to play chess, but since the network is trained with history input, it cannot play without history input. Same here: because the network was trained with alternatives to mate, tweaking the search in this way might possibly mess up the net's heuristics. I really don't see how though, and you have not offered a satisfying explanation as to how this might occur from a game theory perspective, as opposed to from a net-trained-with-bad-heuristics-like-history perspective, which is independent of the certainty of the game theory. And I don't believe you can find a low nodecount position where this change worsens play. If you do, by all means present it and we can see how far it changes my mind. But I strongly doubt that such exists.

Videodr0me:
> I'm sure you won't find it hard to fit my various statements into this nice and easy table. In theory, alts to mate offer nothing.
This is incorrect. The fallacy here is twofold. For one we are not in a beauty contest. It does not matter for strength (measured in elo) if you mate quicker or longer, as long as you mate.  Rewards are not discounted in chess, a win is a win. So in theory your table adds nothing, because mating faster in itself adds nothing (using your language). 

Also and that is the second fallacy, always choosing winning childs does NOT gurantee you find (a) the fastest and (b) not even that you always find a  faster mates then not always choosing winning childs.  This is a piure theoretical fact,, because as the change affects backpropageted Qs the tree is searched differently a shorter mate might be lurking in a now unvisited side branches much earlier along the path.

In practice the change on average might very well shorten mating sequencs pleasing human aestethics, especially in the pet examples that motivated this change in the first place. But will it win more often? Here the data is piling up showing that it is not, its even pointing very slightly in the opposite direction.  And that inspite of a number of positions of type x where it should increase elo. How does that compute in your world of a "mathematical" impossibility?

> In practice, the NN might accidentally be trained to use it in some heuristic manner, much like the history planes. They're totally unnecessary to play chess, but since the network is trained with history input, it cannot play without history input. 

This is again your main contradiction. A purely dominant strategy would even win more games if both players played with the same random NN. That holds even if you consider the current performance of the change as an artifact of training, it just would not matter, as relatively speaking in your world the change should always outperform the pre-change behaviour. Remember this started out as an argument over MCTS, that you sidetracked towards learning to somehow avoid facing the reality of the incoming data. 

And please let me not get into the history planes, they are a relatively cheap way to increase strength in relation to NN size. You are now disputing this, too?  Do you sincerely believe there are no tradeoffs involveld with histroy planes? Or do you even think history planes are a bug, thats only somewhat beneficial, because the NN mislearned something?
 
Even in theory there is something to be said for history planes, but I will not go into details here, because its a deeply theoretical argument with little or no practical bearing at the moment (It involves opponent modelling and best play in heteregenous tournaments that include imperfect opponents among other perfect opponents).

>  Same here: because the network was trained with alternatives to mate, tweaking the search in this way might possibly mess up the net's heuristics.
Or it could help it to generalize better as it is  learning against a target that does not  excessively favour terminal wins over +0.99 positions. It really is a tradeoff here, too.

> really  don't see how though, and you have not offered a satisfying explanation as to how this might occur from a game theory perspective, as opposed to from a net-trained-with-bad-heuristics-like-history perspective, which is independent of the certainty of the game theory.

This seems like progress, so you agree that these examples exits, even if only because of  some mislearned behaviour? Or do you still believe they do not exist not even in theory, which implies that they also can't exist in practice? 

This after all is a simple two player zero-sum game -and to me it seems you somehow mix up game theoretical arguments with arguments that root in a believe about trade-offs (these i would call practical).

If you conceed that these example exist, no matter why they exits then we could  move one step ahead and leave game theory and turn to the practical aspects of the trade-off involved. But i will only put in more effort to make the finer points of MCTS clear to you (e.g. by taking the time to construct such an example, or further continueing this discussion), if we share the same game theoretic understanding, that such an example would falsify your theoretical arguement. This does not imply  that practically the tradeoff can't be shifted into positive elo territory by learning, but you would have acknowledge that such a tradeoff exists.

dubslow:
We are talking past each other. This conversation is not productive at this point, this is a total waste of time on both our parts.

`they are a relatively cheap way to increase strength in relation to NN size. You are now disputing this, too? ` I have always disputed this for the duration of my involvement in this project. There is zero data to suggest that history planes improve play in chess. (Supposedly there exists some such data in Go, but Go has wildly different locality properties than chess, which is supposedly what history helps with.)

Videodr0me:
Agreed, lets not continue the discussion at this point. Still thanks for a heated, but always civil discussion, even if we could not convince each other. Kudos @dubslow.

mooskagh:
Actually, indeed that PR started with all that `U` tweaks, so when it changed to something simpler I felt revealed and approved that PR.

But in fact it's still questionable whether it's useful at all, and especially whether it worth a command line flag/uci parameter. It's not rare that people ask whey would they want to change one parameter or another, and for this, I don't know what to say.

For this feature to stay, we need to have some measurable difference.
Elo difference has to be tested.
If it's not because of Elo but for some other reasons, this has to be checked too, and not just be a mental conclusion. Like, if it wins won positions quicker, take a sample of positions where eval is +3, and gather stats of "moves till checkmate".

If it's purely for aestetics, I think it worth asking other people's opinions on discord, whether they think it's useful. Again, generate set of positions and show difference how version with and without the flag plays from it.

If there will be no confirmed benefit, this PR should be reverted.
If there will be confirmed benefit, the flag should be removed, making this feature "always on".

dkappe:
Some stats from playing out randomly selected endgame epd’s, played both ways, against sf9tb @ 0.25 sec per move.

```
250 12 man ccrl
250 12 man kingbase
100 12 man semi random (sr)
100 6 man ccrl
100 6 man kingbase
50 6 man sr
50 5 man sr
50 4 man sr
50 3 man sr
```
That’s 1000 epd’s, played both ways, is 2000 endgames played out to terminal.

Here are the elo stats without and with sticky mate.

```
Elo difference: 72.61 +/- 23.19
Elo difference: 66.10 +/- 23.22
Elo difference: 10.43 +/- 47.42
Elo difference: 41.89 +/- 31.92
Elo difference: 45.42 +/- 34.81
Elo difference: 24.36 +/- 63.53
Elo difference: 3.47 +/- 59.46
Elo difference: 3.47 +/- 51.72
Elo difference: 0.00 +/- 57.41
```

now with

```
Elo difference: 61.08 +/- 24.17
Elo difference: 76.98 +/- 23.64
Elo difference: 10.43 +/- 47.67
Elo difference: 36.62 +/- 35.66
Elo difference: 47.19 +/- 37.99
Elo difference: 10.43 +/- 61.08
Elo difference: 10.43 +/- 61.08
Elo difference: 0.00 +/- 54.85
Elo difference: 0.00 +/- 60.66
```

The average ply length for white and black victories against sf9tb is:

Without:

```
wavg: 39.8740458015267, wcount: 262
bavg: 38.8, bcount: 230
```

and with

```
wavg: 39.4734982332156, wcount: 283
bavg: 40.64453125, bcount: 256
```

dkappe:
I will run the same epd’s next, instead of a random sampling next.

dubslow:
At 0.25 s/move, how many nodes was that per move?

PR #212, opened 2018-07-23
Additional permission for linking with cuDNN/CUDA, closes #184

### LINKED ISSUE #184: Add permission under GPL v3 section 7 for linking with cuDNN/CUDA.
In order to be able to distribute a windows cuDNN/CUDA lc0 binary together with the required dlls we need to comply with both GPL and nVidia license terms. In short, GPL requires no proprietary dlls distributed with lc0, while nVidia requires the dlls distributed only with lc0 and not independently. 

There is a way around this by adding an additional permission as allowed by GPL v3 in section 7. An example for such an additional permission for CUDA is found [here](https://gitweb.gentoo.org/repo/gentoo.git/plain/licenses/GPL-3+-with-cuda-exception). 

This can be done for all files currently in lc0 with the agreement of all contributors, except for the leela/lczero derived ones used for BLAS/OpenCL, so that lc0 can be distributed, with the cudnn backend only, together with the required dlls.

The suggested text is the following
```
Additional permission under GNU GPL version 3 section 7

If you modify this Program, or any covered work, by linking or
combining it with NVIDIA Corporation's libraries from the
NVIDIA CUDA Toolkit and the the NVIDIA CUDA Deep Neural
Network library (or a modified version of those libraries),
containing parts covered by the terms of the respective license
agreement, the licensors of this Program grant you additional
permission to convey the resulting work.
```

borg323:
Now that everybody's agreement was received in #184 (with the exception of @ron-wolf who has just a single character comment fix), we can go on with adding the permission to all source files (except the blas and opencl related ones). I also added a license section to the README.


README.md
```diff
@@ -105,3 +105,33 @@ Or.
...
+
+You should have received a copy of the GNU General Public License
+along with Leela Chess.  If not, see <http://www.gnu.org/licenses/>.
+
+### Additional permission under GNU GPL version 3 section 7
+
+_The source files of Lc0 with the exception of the BLAS and OpenCL
+backends (namely `network_blas.cc`, `network_opencl.cc` and all files
+in the `blas` and `opencl` directories have the following additional
+permission as allowed under GNU GPL version 3 section 7:_
```
mooskagh: I think we should isolate those files, at least put them into a separate directory, with it's own COPYING and README (which i.e. reminds that it's not allowed to copy-paste the code from them into other sources, and is discouraged to edit files of different licenses in one PR).

Probably we can do that later, but we should not forget.

borg323:
The latest version moves network_blas.cc in the blas directory and network_opencl.cc in the opencl directory. There is now a simple README.md file in both directories saying the license is GPL and emphasizing that no additional permission applies. I don't see the need for additional COPYING files, the one in the root is unmodified.

PR #213, opened 2018-07-24
Node::GetVisitedPolicy uses cached value to avoid expensive recalcula…

dje-dev:
Severe slowdown in nodes per second (nps) has been commonly seen in high-end multi-GPU systems (starting notably after a few seconds to a few minutes). Analysis under a debugger shows the CPU threads become the bottleneck, with only one thread making progress at any given time as it traverses and updates the increasingly deep large tree (and other threads waiting on the single tree mutex held by the active thread). Aggregate GPU utilization often drops from near 100% to 25% or less.

As GPUs become increasingly faster, (but CPUs much less so) the ultimate scalability of LeelaChessZero will be increasingly by gated by the efficiency of the search code. Thus incremental improvements in this area would seem to be important. 

Analysis under a debugger revealed considerable CPU time being spent in the method Node::GetVisitedPolicy, which does a lot of pointer chasing as it accumulates the visited policy statistic. This method executes approximately 100,000,000 times for each 5,000,000 nodes visited. 

This PR proposes an enhancement whereby the Node::GetVisitedPolicy method now just returns a cached value. This new cached value is stored in a new member variable at Node (at a cost of 4 bytes). The cached value is easily kept up to date (needing a potential update in only one place, at Node::FinalizeScoreUpdate).

Analysis from the starting position using a 3-GPU (Titan V) system showed a aggregate nps improvement of 13% when run for 60 seconds (73.6 to 83.3 knps), and about 15% (51.6 to 59.4 knps) when run for 15 minutes.

**SAMPLE TEST COMMAND LINE** 
lc0 -t 8 -w d:\weights\big.txt --no-smart-pruning --nncache=20000000 --minibatch-size=512 --backend=multiplexing "--backend-opts=(backend=cudnn-fp16,gpu=0),(backend=cudnn-fp16,gpu=1),(backend=cudnn-fp16,gpu=2)"

### DIFF
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -184,8 +184,9 @@ void Node::FinalizeScoreUpdate(float v) {
   // Recompute Q.
   q_ += (v - q_) / (n_ + 1);
   // If first visit, update parent's sum of policies visited at least once.
-  if (n_ == 0 && parent_ != nullptr)
+  if (n_ == 0 && parent_ != nullptr) {
     parent_->visited_policy_ += parent_->edges_[index_].GetP();
+  }
   // Increment N.
   ++n_;
   // Decrement virtual loss.
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -149,13 +149,8 @@ void Node::CreateEdges(const MoveList& moves) {
 Node::ConstIterator Node::Edges() const { return {edges_, &child_}; }
 Node::Iterator Node::Edges() { return {edges_, &child_}; }
 
-float Node::GetVisitedPolicy() const {
-  float res = 0.0f;
-  for (const auto* node = child_.get(); node; node = node->sibling_.get()) {
-    if (node->n_ > 0) res += edges_[node->index_].GetP();
-  }
-  return res;
-}
+float Node::GetVisitedPolicy() const { return visited_policy_;  }
+
 
 Edge* Node::GetEdgeToNode(const Node* node) const {
   assert(node->parent_ == this);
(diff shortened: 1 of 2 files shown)


dubslow:
+4 bytes is not a small thing to ask, though at least we have Edge-Node separation now. Did you measure memory usage before and after?

Also, this implies that setting fpu reduction to zero would see pretty much the same speed gain, can you confirm that?

dje-dev:
Setting fpu-reduction to zero yields a very different result and I'm afraid it doesn't help to test potential modification. In the 60 second test the average nps is much higher at 99k nps. However the tree shape is very different (depth 4/28 instead of 6/39). I this less deep tree, the work needed to be done by CPU is not (yet) the bottleneck. 

Regarding memory, granted 4 bytes would be unfortunate to lose, but seemed bearable because:
* as noted, Edge-Node separation means there are many fewer Nodes
* memory consumption is only likely to become critical on a high-end GPU system, but that's exactly the situation where this speedup will be important to help keep making reasonable progress in long analyses
* with time, GPUs speed and DRAM capacity will improve fairly rapidly but not CPUs, which would make this PR increasingly beneficial

dubslow:
(gpu memory has nothing to do with it, this is entirely cpu memory)

dje-dev:
I just meant that you are unlikely to run out of RAM without a GPU because you’ll never be able to generate enough nodes....

src/mcts/node.cc
```diff
@@ -188,6 +183,9 @@ void Node::CancelScoreUpdate() { --n_in_flight_; }
 void Node::FinalizeScoreUpdate(float v) {
   // Recompute Q.
   q_ += (v - q_) / (n_ + 1);
+  // If first visit, update parent's sum of policies visited at least once.
+  if (n_ == 0 && parent_ != nullptr)
+    parent_->visited_policy_ += parent_->edges_[index_].GetP();
```
mooskagh: As statement is on a separate line, it's better to have `{..}` around it.

dubslow:
Note that pending the shrink-Edge-policy PR, it might also be worthwhile to reduce this to 16 bits from 32 as well. It would depend on current size and alignment issues.

PR #218, opened 2018-07-28
Out of order eval for terminal nodes and cache hits.

mooskagh:
Currently if the first node in a batch turns out to be a cache hit, it's back-propagated immediately without gathering full batch and without querying NN.

This PR extends that (behind `--out-of-order-eval` flag) to:
- Cache hit at any index in the batch
- Also to terminal nodes

When during the batch collection such node is encountered, it's processed out of order. On the next iteration, batch gathering continues from where previous attempt stopped.

Gives +8% nps.
(Having `--cache-history-length=0` with that change gives +12% nps on top of that, may be worth to reconsider using it, maybe it will translate to Elo gain this time)

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -57,11 +57,8 @@ const char* Search::kCacheHistoryLengthStr =
 const char* Search::kPolicySoftmaxTempStr = "Policy softmax temperature";
 const char* Search::kAllowedNodeCollisionsStr =
     "Allowed node collisions, per batch";
-<<<<<<< HEAD
 const char* Search::kOutOfOrderEvalStr = "Out-of-order cache backpropagation";
-=======
 const char* Search::kStickyCheckmateStr = "Ignore alternatives to checkmate";
->>>>>>> upstream/master
 
 namespace {
 const int kSmartPruningToleranceNodes = 100;
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -548,8 +548,8 @@ void SearchWorker::GatherMinibatch() {
   int number_out_of_order = 0;
 
   // Gather nodes to process in the current batch.
-  // If we had too many (kMiniBatchSize) nodes out of order, also exit so that
-  // search has a chance to exit search.
+  // If we had too many (kMiniBatchSize) nodes out of order, also interrupt the
+  // iteration so that search can exit.
   // TODO(crem) change that to checking search_->stop_ when bestmove reporting
   // is in a separate thread.
   while (minibatch_size < search_->kMiniBatchSize &&
@@ -575,10 +575,10 @@ void SearchWorker::GatherMinibatch() {
       // Node was never visited, extend it.
       ExtendNode(node);
 
-      // Only send non-terminal nodes to neural network
+      // Only send non-terminal nodes to a neural network.
       if (!node->IsTerminal()) {
-        minibatch_.back().nn_queried = true;
-        minibatch_.back().is_cache_hit = AddNodeToComputation(node, true);
+        picked_node.nn_queried = true;
+        picked_node.is_cache_hit = AddNodeToComputation(node, true);
       }
     }
 
@@ -586,16 +586,15 @@ void SearchWorker::GatherMinibatch() {
     // doesn't require NN eval (i.e. it's a cache hit or terminal node), do
     // out of order eval for it.
     if (search_->kOutOfOrderEval) {
-      if (node->IsTerminal() || minibatch_.back().is_cache_hit) {
+      if (node->IsTerminal() || picked_node.is_cache_hit) {
         // Perform out of order eval for the last entry in minibatch_.
-        FetchSingleNodeResult(&minibatch_.back(),
-                              computation_->GetBatchSize() - 1);
-        DoBackupUpdateSingleNode(minibatch_.back());
+        FetchSingleNodeResult(&picked_node, computation_->GetBatchSize() - 1);
+        DoBackupUpdateSingleNode(picked_node);
 
         // Remove last entry in minibatch_, as it has just been
         // processed.
         // If NN eval was already processed out of order, remove it.
-        if (minibatch_.back().nn_queried) computation_->PopCacheHit();
+        if (picked_node.nn_queried) computation_->PopCacheHit();
         minibatch_.pop_back();
         --minibatch_size;
       }
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -586,8 +586,7 @@ void SearchWorker::GatherMinibatch() {
     // doesn't require NN eval (i.e. it's a cache hit or terminal node), do
     // out of order eval for it.
     if (search_->kOutOfOrderEval) {
-      const bool is_terminal = node->IsTerminal();
-      if (is_terminal || minibatch_.back().is_cache_hit) {
+      if (node->IsTerminal() || minibatch_.back().is_cache_hit) {
         // Perform out of order eval for the last entry in minibatch_.
         FetchSingleNodeResult(&minibatch_.back(),
                               computation_->GetBatchSize() - 1);
@@ -596,7 +595,7 @@ void SearchWorker::GatherMinibatch() {
         // Remove last entry in minibatch_, as it has just been
         // processed.
         // If NN eval was already processed out of order, remove it.
-        if (minibatch_.back().nn_queried) computation_->PopInput();
+        if (minibatch_.back().nn_queried) computation_->PopCacheHit();
         minibatch_.pop_back();
         --minibatch_size;
       }
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -531,44 +531,41 @@ bool SearchWorker::IsSearchActive() const {
 // ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 void SearchWorker::InitializeIteration(
     std::unique_ptr<NetworkComputation> computation) {
-  // If last iteration was out of order eval, continue gathering batch.
-  if (out_of_order_eval_) {
-    // If NN eval was already processed on the last iteration, remove it.
-    if (nodes_to_process_.back().nn_queried) computation_->PopInput();
-    // Was already processed on the last iteration.
-    nodes_to_process_.pop_back();
-    out_of_order_eval_ = false;
-    return;
-  }
-
-  // It was not an out-of-order eval, batch is fresh.
   computation_ = std::make_unique<CachingComputation>(std::move(computation),
                                                       search_->cache_);
-  nodes_to_process_.clear();
-  collisions_found_ = 0;
+  minibatch_.clear();
 }
 
 // 2. Gather minibatch.
 // ~~~~~~~~~~~~~~~~~~~~
 void SearchWorker::GatherMinibatch() {
-  int nodes_found = nodes_to_process_.size() - collisions_found_;
+  // Total number of nodes to process.
+  int minibatch_size = 0;
+  int collisions_found = 0;
+  // Number of nodes processed out of order.
+  int number_out_of_order = 0;
 
   // Gather nodes to process in the current batch.
-  while (nodes_found < search_->kMiniBatchSize) {
+  // If we had too many (kMiniBatchSize) nodes out of order, also exit so that
+  // search has a chance to exit search.
+  // TODO(crem) change that to checking search_->stop_ when bestmove reporting
+  // is in a separate thread.
+  while (minibatch_size < search_->kMiniBatchSize &&
+         number_out_of_order < search_->kMiniBatchSize) {
     // If there's something to process without touching slow neural net, do it.
-    if (nodes_found > 0 && computation_->GetCacheMisses() == 0) return;
+    if (minibatch_size > 0 && computation_->GetCacheMisses() == 0) return;
     // Pick next node to extend.
-    nodes_to_process_.emplace_back(PickNodeToExtend());
-    auto& picked_node = nodes_to_process_.back();
+    minibatch_.emplace_back(PickNodeToExtend());
+    auto& picked_node = minibatch_.back();
     auto* node = picked_node.node;
 
     // There was a collision. If limit has been reached, return, otherwise
     // just start search of another node.
     if (picked_node.is_collision) {
-      if (++collisions_found_ > search_->kAllowedNodeCollisions) return;
+      if (++collisions_found > search_->kAllowedNodeCollisions) return;
       continue;
     }
-    ++nodes_found;
+    ++minibatch_size;
 
     // If node is already known as terminal (win/loss/draw according to rules
     // of the game), it means that we already visited this node before.
@@ -578,9 +575,8 @@ void SearchWorker::GatherMinibatch() {
 
       // Only send non-terminal nodes to neural network
       if (!node->IsTerminal()) {
-        nodes_to_process_.back().nn_queried = true;
-        nodes_to_process_.back().is_cache_hit =
-            AddNodeToComputation(node, true);
+        minibatch_.back().nn_queried = true;
+        minibatch_.back().is_cache_hit = AddNodeToComputation(node, true);
       }
     }
 
@@ -589,10 +585,18 @@ void SearchWorker::GatherMinibatch() {
     // out of order eval for it.
     if (search_->kOutOfOrderEval) {
       const bool is_terminal = node->IsTerminal();
-      if (is_terminal ||
-          (!is_terminal && nodes_to_process_.back().is_cache_hit)) {
-        out_of_order_eval_ = true;
-        return;
+      if (is_terminal || minibatch_.back().is_cache_hit) {
+        // Perform out of order eval for the last entry in minibatch_.
+        FetchSingleNodeResult(&minibatch_.back(),
+                              computation_->GetBatchSize() - 1);
+        DoBackupUpdateSingleNode(minibatch_.back());
+
+        // Remove last entry in minibatch_, as it has just been
+        // processed.
+        // If NN eval was already processed out of order, remove it.
+        if (minibatch_.back().nn_queried) computation_->PopInput();
+        minibatch_.pop_back();
+        --minibatch_size;
       }
     }
   }
@@ -764,8 +768,6 @@ bool SearchWorker::AddNodeToComputation(Node* node, bool add_if_cached) {
 // 3. Prefetch into cache.
 // ~~~~~~~~~~~~~~~~~~~~~~~
 void SearchWorker::MaybePrefetchIntoCache() {
-  // If out of order eval is going on, don't prefetch anything into cache.
-  if (out_of_order_eval_) return;
   // TODO(mooskagh) Remove prefetch into cache if node collisions work well.
   // If there are requests to NN, but the batch is not full, try to prefetch
   // nodes which are likely useful in future.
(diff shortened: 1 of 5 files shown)


src/neural/cache.h
```diff
@@ -73,6 +73,8 @@ class CachingComputation {
   float GetQVal(int sample) const;
   // Returns P value @move_id of @sample.
   float GetPVal(int sample, int move_id) const;
+  // Pops last input from the computation. Only allowed for inputs from cache.
```
dubslow: When I first reviewed, "inputs from cache" slightly confused me. I understand better now having read the whole thing, but probably still better to tweak this comment to say "Only allowed for inputs which were cached" or "inputs which were cache hits" or something like that, the "inputs *from* cache" just confused me ever so slightly.

src/mcts/search.h
```diff
@@ -236,17 +238,26 @@ class SearchWorker {
...
   void ExtendNode(Node* node);
-  bool AddNodeToComputation(Node* node, bool add_if_cached = true);
+  bool AddNodeToComputation(Node* node, bool add_if_cached);
   int PrefetchIntoCache(Node* node, int budget);
+  void FetchSingleNodeResult(NodeToProcess* node_to_process,
+                             int idx_in_computation);
+  void DoBackupUpdateSingleNode(const NodeToProcess& node_to_process);
 
   Search* const search_;
+  // List of nodes to process.
```
dubslow: This comment is totally redundant with the declaration :) I think better to say "This is the minibatch" or something, emphasize that it is the main purpose of the class is this member

src/mcts/search.h
```diff
@@ -236,17 +238,26 @@ class SearchWorker {
     Node* node;
     bool is_collision = false;
     bool nn_queried = false;
+    bool is_cache_hit = false;
```
dubslow: With 3 bools in a row, do modern compilers optimize these to fit all into one byte? Or do alignment issues mean that such an optimization wouldn't help anyways?

src/mcts/search.cc
```diff
@@ -518,16 +521,27 @@ bool SearchWorker::IsSearchActive() const {
 // ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 void SearchWorker::InitializeIteration(
     std::unique_ptr<NetworkComputation> computation) {
-  nodes_to_process_.clear();
+  // If last iteration was out of order eval, continue gathering batch.
```
dubslow: Maybe slightly more detailed would be "If last iteration was {interrupted,paused} by OOO eval, resume gathering the [same][old] batch"

src/mcts/search.cc
```diff
@@ -541,22 +555,35 @@ void SearchWorker::GatherMinibatch() {
...
+      }
+    }
+
+    // If out of order eval is enabled and the node to compute we added last
+    // doesn't require NN eval (i.e. it's a cache hit or terminal node), do
+    // out of order eval for it.
+    if (search_->kOutOfOrderEval) {
+      const bool is_terminal = node->IsTerminal();
+      if (is_terminal ||
+          (!is_terminal && nodes_to_process_.back().is_cache_hit)) {
```
dubslow: Isn't the `!is_terminal` here redundant with the first part of the condition?

IOW, couldn't this just be `if (is_terminal || nodes_to_process_.back().is_cache_hit) {`?

src/mcts/search.cc
```diff
@@ -818,48 +847,62 @@ int SearchWorker::PrefetchIntoCache(Node* node, int budget) {
 // 4. Run NN computation.
 // ~~~~~~~~~~~~~~~~~~~~~~
 void SearchWorker::RunNNComputation() {
-  // This function is so small as to be silly, but its parent function is
-  // conceptually cleaner for it.
-  if (computation_->GetBatchSize() != 0) computation_->ComputeBlocking();
+  // Do not do NN eval if there's nothing to compute, or if it's an out-of-order
+  // eval, don't compute.
```
dubslow: Surely this comment can be simplified :)

src/mcts/search.cc
```diff
@@ -818,48 +847,62 @@ int SearchWorker::PrefetchIntoCache(Node* node, int budget) {
...
+    // they require any further processing after value retrieval.
+    node_to_process->v = node->GetQ();
+    return;
+  }
+  // For NN results, we need to populate policy as well as value.
+  // First the value...
+  node_to_process->v = -computation_->GetQVal(idx_in_computation);
+  // ...and secondly, the policy data.
+  float total = 0.0;
+  for (auto edge : node->Edges()) {
```
dubslow: It occurs to me now, for cached NN results, we recompute the policy normalization, and if non-unit, we recompute the softmax temperature each time as well. Why isn't this done directly in the CachingComputation and stored as part of the cache, so we don't re-do these policy shenanigans for every cache hit?

(This comment is totally ancillary to the rest of the review)

src/mcts/search.cc
```diff
@@ -868,47 +911,59 @@ void SearchWorker::FetchMinibatchResults() {
...
+    return;
+  }
+
+  for (const NodeToProcess& node_to_process : nodes_to_process_) {
+    DoBackupUpdateSingleNode(node_to_process);
+  }
+}
+
+void SearchWorker::DoBackupUpdateSingleNode(
+    const NodeToProcess& node_to_process) REQUIRES(search_->nodes_mutex_) {
```
dubslow: Breaking the line here seems awkward to me, why not 

```
void Class::Function(arg)
                  REQUIRES(foo) {
```
?

(Again, ancillary. If clang did this automatically, then I guess ignore this comment.)

src/mcts/search.cc
```diff
@@ -541,22 +555,35 @@ void SearchWorker::GatherMinibatch() {
...
+
+    // If out of order eval is enabled and the node to compute we added last
+    // doesn't require NN eval (i.e. it's a cache hit or terminal node), do
+    // out of order eval for it.
+    if (search_->kOutOfOrderEval) {
+      const bool is_terminal = node->IsTerminal();
+      if (is_terminal ||
+          (!is_terminal && nodes_to_process_.back().is_cache_hit)) {
+        out_of_order_eval_ = true;
+        return;
```
dubslow: Okay, only serious question about the implementation: if an `out_of_order_eval_` *only* ever processes *one* node out of order at a time, why bother having this member variable and faking out the rest of the functions? Why not just be rid of the member variable and instead directly call `FetchSingleNodeResults` and `DoBackupUpdateSingleNode` (or whatever they're named) right here instead? Kinda seems to me like the same argument about the sticky checkmate change and messing with U in some way rather than just implementing the change where the change needs to occur. For such a simple optimization, we don't need to fake out the whole rest of the iteration.

src/mcts/search.h
dubslow: Hell, maybe we should just straight up rename it to `minibatch_`?

sergiovieri:
Hi, can anyone verify this. I tried the `--out-of-order-eval` flag and ran the engine through uci. The engine stops around 10 moves into the match. Tried this several times, but it yields the same result. Turning off this flag immediately fixed this problem and able to run the game till the end.

src/mcts/search.h
SunnyWar: I tried it by adding "/d1reportSingleClassLayoutNode" as a compiler option. It result in a ton of information on the output but at the top is this: 

: added class lczero::Node	size(72):
2>	+---
2> 0	| EdgeList edges_
2>16	| index_
2>  	| <alignment member> (size=2)
2>20	| q_
2>24	| n_
2>28	| n_in_flight_
2>  	| <alignment member> (size=2)
2>32	| visited_policy_
2>36	| max_depth_
2>38	| full_depth_
2>40	| is_terminal_
2>  	| <alignment member> (size=7)
2>48	| parent_
2>56	| ?$unique_ptr@VNode@lczero@@U?$default_delete@VNode@lczero@@@std@@ child_
2>64	| ?$unique_ptr@VNode@lczero@@U?$default_delete@VNode@lczero@@@std@@ sibling_
2>	+---

dubslow: wow, that bool is a killer for alignment purposes. and it's a bit odd that it doesn't combine `index_` and `n_in_flight_` ? Does it preserve the order in which they're declared?

SunnyWar: Note also by adding "#pragma pack(1)" above the class definition the size is reduced. This is not always advisable thought as mis-alignment can have other performance problems.

class lczero::Node	size(61):
1>	+---
1> 0	| EdgeList edges_
1>16	| index_
1>18	| q_
1>22	| n_
1>26	| n_in_flight_
1>28	| visited_policy_
1>32	| max_depth_
1>34	| full_depth_
1>36	| is_terminal_
1>37	| parent_
1>45	| ?$unique_ptr@VNode@lczero@@U?$default_delete@VNode@lczero@@@std@@ child_
1>53	| ?$unique_ptr@VNode@lczero@@U?$default_delete@VNode@lczero@@@std@@ sibling_
1>	+---

SunnyWar: @dubslow I lurking and trying to observe what is important and what I can contribute, if anything. I gather that the size of Node is important. You can reduce easily by simple moving the declaration of n_in_flight to right before index. Node is now smaller.

1>class lczero::Node	size(64):
1>	+---
1> 0	| EdgeList edges_
1>16	| index_
1>18	| n_in_flight_
1>20	| q_
1>24	| n_
1>28	| visited_policy_
1>32	| max_depth_
1>34	| full_depth_
1>36	| is_terminal_
1>  	| <alignment member> (size=3)
1>40	| parent_
1>48	| ?$unique_ptr@VNode@lczero@@U?$default_delete@VNode@lczero@@@std@@ child_
1>56	| ?$unique_ptr@VNode@lczero@@U?$default_delete@VNode@lczero@@@std@@ sibling_

dubslow: Indeed :) size of Node is important, though less important than it once was, since we have Edge-Node separation now. Each node reported by the search spawns one `Node`, but also all the relevant child `Edges` for that node, so if the search tree has n nodes reported, then on average the search tree takes `n * (sizeof(Node) + 35 * sizeof(Edge))`, where 35 is the average chess branching factor. So the size of Edges dominates (see PR176), but at the same time we could probably get a several-percent improvement here with the alignment. Indeed rearranging the 2 byte ints helps, but I think if we're going to rearrange for optimization, we may as well go whole-hog and put all members in order of size, i.e. pointers first. That way it's easier to modify in the future. I'll make another PR shortly. Many thanks for the pointer to the correct compiler flag, and of course printing the output here to spike my curiosity!

oscardssmith: Given this disparity, are we sure edges don't have any similar problems?

SunnyWar: @dubslow I arranged things as you suggested and it comes out nicely:

1>class lczero::Node	size(64):
1>	+---
1> 0	| EdgeList edges_
1>16	| ?$unique_ptr@VNode@lczero@@U?$default_delete@VNode@lczero@@@std@@ child_
1>24	| ?$unique_ptr@VNode@lczero@@U?$default_delete@VNode@lczero@@@std@@ sibling_
1>32	| parent_
1>40	| q_
1>44	| n_
1>48	| visited_policy_
1>52	| index_
1>54	| n_in_flight_
1>56	| max_depth_
1>58	| full_depth_
1>60	| is_terminal_
1>  	| <alignment member> (size=3)

dubslow: Yes. They are currently 6 bytes theoretically, 8 in practice on my g++; PR176 as mentioned will reduce that to 4 and 4 respectively. Edges have only two members.

dubslow: Turns out the EdgeList itself has 6 bytes of padding. I guess that would be next on the list.

SunnyWar: @oscardssmith This is the result for the Edge class

1>class lczero::Edge	size(8):
1>	+---
1> 0	| Move move_
1>  	| <alignment member> (size=2)
1> 4	| p_
1>	+---

Moving p_ to be first it's a tiny bit better:

1>class lczero::Edge	size(8):
1>	+---
1> 0	| p_
1> 4	| Move move_
1>  	| <alignment member> (size=2)
1>	+---

I doubt making this change would have any measurable impact.

src/mcts/search.cc
```diff
@@ -518,16 +521,27 @@ bool SearchWorker::IsSearchActive() const {
...
+  nodes_to_process_.clear();
+  collisions_found_ = 0;
 }
 
 // 2. Gather minibatch.
 // ~~~~~~~~~~~~~~~~~~~~
 void SearchWorker::GatherMinibatch() {
-  int nodes_found = 0;
-  int collisions_found = 0;
+  int nodes_found = nodes_to_process_.size() - collisions_found_;
```
Videodr0me: Is this correct? collisions can also be due to collisions with other threads batches, if we subtract those from our batch, our batch might get smaller and smaller. Maybe a reason for the decreasing nps some observed

mooskagh:
I'll redo this PR as @dubslow suggested (do backpropagation right in "picknodetoextend"), which will indeed simplify things. (A bit of care should be done to exit from that loop though.)
Until then, there's no need to review!

I said that on discord on Saturday(?) already (and planned to fix shortly after that), but for now I'm not doing anything, so repeating here.

dubslow:
Well worth repeating, I totally missed the Discord message!

src/mcts/search.cc
mooskagh: No longer relevant as out-of-order eval is now inplace.

src/mcts/search.cc
mooskagh: Indeed, fixed.

src/mcts/search.cc
mooskagh: Good idea indeed, changed. Simplified things a lot.

src/mcts/search.cc
mooskagh: Done. :)

src/mcts/search.cc
mooskagh: Caching computation doesn't always know the list of legal moves. In prefetch-into-cache, it stores values for all pseudo-legal moves instead.
It's a good optimization to keep in mind, especially when we get rid of history, but it's not that easy to implement nicely.

src/mcts/search.cc
mooskagh: That's clang. I usually write everything in one line and then just run clang-format. :)

src/mcts/search.h
mooskagh: Three bools in a row won't share the same byte (C++ requires that it should be possible to take address of each).
However, i can write:
```cpp
    bool is_collision:1 = false;
    bool nn_queried:1 = false;
    bool is_cache_hit:1 = false;
```
and then they will indeed reside in one byte.

But it probably will slow down things.  NodeToProcess is not memory critical structure at all, there are only minibatch_size * num_threads of those in total at any moment.
So even if we'll ever be able to optimize code to allow 64 threads and minibatch of 2000 nodes in each (unrealistic), it will still be <2MB in total.

src/mcts/search.h
mooskagh: Done renaming. After renaming, the comment makes sense. :)

src/neural/cache.h
mooskagh: Done.

src/mcts/search.h
dubslow: We happy engineers who live in the days when 2MB can be considered trivial :)

mooskagh: Well if we have 64 threads and 2000 nodes in each, and every thread does say 4 NN queries per second, that's 500000nps, which is memory growth of ~100MB per second.
Comparing to that, static 2MB is fine.

Current reality is more like 2 threads of 256 nodes, which is more like 5KB.

src/mcts/search.cc
```diff
@@ -551,19 +565,39 @@ void SearchWorker::GatherMinibatch() {
...
 
-    // Only send non-terminal nodes to neural network
-    if (!node->IsTerminal()) {
-      nodes_to_process_.back().nn_queried = true;
-      AddNodeToComputation(node);
+    // If out of order eval is enabled and the node to compute we added last
+    // doesn't require NN eval (i.e. it's a cache hit or terminal node), do
+    // out of order eval for it.
+    if (search_->kOutOfOrderEval) {
+      const bool is_terminal = node->IsTerminal();
```
dubslow: Not sure this line is necessary?

src/neural/cache.h
```diff
@@ -73,6 +73,9 @@ class CachingComputation {
   float GetQVal(int sample) const;
   // Returns P value @move_id of @sample.
   float GetPVal(int sample, int move_id) const;
+  // Pops last input from the computation. Only allowed for inputs which were
+  // cached.
+  void PopInput();
```
dubslow: Actually, reading the `SearchWorker` code, perhaps this is better renamed as `PopCacheHit`? Which is far more semantic than `PopInput`

src/mcts/search.cc

src/neural/cache.h
mooskagh: Done.

src/mcts/search.cc
```diff
@@ -556,18 +567,38 @@ void SearchWorker::GatherMinibatch() {
...
     // If node is already known as terminal (win/loss/draw according to rules
     // of the game), it means that we already visited this node before.
-    if (node->IsTerminal()) continue;
+    if (!node->IsTerminal()) {
+      // Node was never visited, extend it.
+      ExtendNode(node);
 
-    // Node was never visited, extend it.
-    ExtendNode(node);
+      // Only send non-terminal nodes to neural network
```
dubslow: period!

src/mcts/search.cc
```diff
@@ -530,24 +533,32 @@ bool SearchWorker::IsSearchActive() const {
...
+  // Total number of nodes to process.
+  int minibatch_size = 0;
   int collisions_found = 0;
+  // Number of nodes processed out of order.
+  int number_out_of_order = 0;
 
   // Gather nodes to process in the current batch.
-  while (nodes_found < search_->kMiniBatchSize) {
+  // If we had too many (kMiniBatchSize) nodes out of order, also exit so that
+  // search has a chance to exit search.
```
dubslow: I think that there's a redundant word in this comment?

src/mcts/search.cc
```diff
@@ -556,18 +567,38 @@ void SearchWorker::GatherMinibatch() {
...
+    if (!node->IsTerminal()) {
+      // Node was never visited, extend it.
+      ExtendNode(node);
 
-    // Node was never visited, extend it.
-    ExtendNode(node);
+      // Only send non-terminal nodes to neural network
+      if (!node->IsTerminal()) {
+        minibatch_.back().nn_queried = true;
+        minibatch_.back().is_cache_hit = AddNodeToComputation(node, true);
```
dubslow: You already have the `auto& picked_node = minibatch_.back()` local variable, may as well use it here

src/mcts/search.cc
```diff
@@ -556,18 +567,38 @@ void SearchWorker::GatherMinibatch() {
...
+        FetchSingleNodeResult(&minibatch_.back(),
+                              computation_->GetBatchSize() - 1);
+        DoBackupUpdateSingleNode(minibatch_.back());
+
+        // Remove last entry in minibatch_, as it has just been
+        // processed.
+        // If NN eval was already processed out of order, remove it.
+        if (minibatch_.back().nn_queried) computation_->PopCacheHit();
+        minibatch_.pop_back();
+        --minibatch_size;
```
dubslow: Ditto this block as well about the local variable!

src/mcts/search.cc
```diff
@@ -835,104 +866,96 @@ int SearchWorker::PrefetchIntoCache(Node* node, int budget) {
 
 // 4. Run NN computation.
 // ~~~~~~~~~~~~~~~~~~~~~~
-void SearchWorker::RunNNComputation() {
-  // This function is so small as to be silly, but its parent function is
-  // conceptually cleaner for it.
-  if (computation_->GetBatchSize() != 0) computation_->ComputeBlocking();
-}
+void SearchWorker::RunNNComputation() { computation_->ComputeBlocking(); }
```
dubslow: Is it, uh, worthwhile to remove this particular abstraction? It isn't really much of one lol. Totally up to you

src/mcts/search.cc
```diff
@@ -835,104 +866,96 @@ int SearchWorker::PrefetchIntoCache(Node* node, int budget) {
...
   }
 }
 
 // 6. Propagate the new nodes' information to all their parents in the tree.
 // ~~~~~~~~~~~~~~
 void SearchWorker::DoBackupUpdate() {
-  // Initialize playout and depth counters.
-  uint16_t max_depth_batch = 0;
-  uint32_t cum_depth_batch = 0;
-  uint16_t playouts_batch = 0;
```
dubslow: This PR now has the side effect of changing the depth updates from per batch to per playout, but in the short term at least that's not a problem

src/mcts/search.cc

src/mcts/search.cc
mooskagh: Yeah it was weirdly stated, probably that way is better.

src/mcts/search.cc

src/mcts/search.cc

src/mcts/search.cc
mooskagh: It all was done in a first place for support within-one-thread multiple games (for training games generation), and this function was the main reason to do that. That feature was never finished, but we'll need the same for tree detaching.

It will be:

```
  for all searches:
    Do steps 1..3
  for all searches:
    Do step 4.   (with network_st_batch, only last one will actually block)
  for all searches:
    Do steps 5..7
```

So step 4 is important enough to have a trivial (for now) abstration function.

src/mcts/search.cc
mooskagh: I'm not convinced that doing batch updates is faster than direct updates, as we hold exclusive mutex at that time anyway.

Videodr0me: Yes currently its a wash. I was hoping for a mutex reorg to gain some traction, but maybe there are more pressing things?!

Issue #193, opened 2018-07-19
Try minimax eval


mooskagh:
(that's an idea for much later)

I've tried to replace `Q` from being subtree average to be simple minimax and ran a few games. Although minimax variant was mostly losing, it played quite decently, so it's an idea worth exploring.

The problem with "subtree Q" is that if there is a long tactical losing line, Lc0 doesn't drop the move immediately but it needs many visits for Q to slowly drop, and then many visits for N of other edge to exceed N of "bad node".

The problem with minimax is opposite: when some tactics is found, low score may propagate quite far up, and this node is temporary out of consideration.

Possible solution would be to to have a mix: for `theta` between 0 and 1,
`Q = theta*minimax_Q  + (1-theta)*subtree_avg_Q`

Hacky patch of minimax `Q` if someone wants to play around.
```diff
diff --git a/src/mcts/node.cc b/src/mcts/node.cc
index adde34a..8ba0b73 100644
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -178,10 +178,13 @@ void Node::CancelScoreUpdate() { --n_in_flight_; }
 
 void Node::FinalizeScoreUpdate(float v, float gamma, float beta) {
   // Recompute Q.
-  if (gamma == 1.0f && beta == 1.0f) {
-    q_ += (v - q_) / (n_ + 1);
+  if (n_ == 0) {
+    q_ = v;
   } else {
-    q_ += (v - q_) / (std::pow(static_cast<float>(n_), gamma) * beta + 1);
+    q_ = 1.0f;
+    for (Node* child : ChildNodes()) {
+      q_ = std::min(q_, -child->q_);
+    }
   }
   // Increment N.
   ++n_;
```

Videodr0me:
I tried this and similar schemes some months back and could not make them work, it always lost a substantial amount of elo. But any new datapoint on this is welcome. One idea i did not follow through yet is to use alpha beta bounds not for backpropagating value but just to prune branches essentially penalizing policy for visiting out of bound branches..... so many ideas, so little time....

DanielUranga:
Also something else to take into consideration: when there are no more child moves it could be that we checkmated the opposing king, or that we drowned it, so te score could be `0.0f` in that case, right?.

DanielUranga:
IMHO there is no need to do that, AB engines have really low branching factors and they still use the minimax value propagation.

DanielUranga:
[...]

PR #221, opened 2018-07-31
[Not for merge] Issue 193: Try using a MinMax component on the MCTS search

DanielUranga:
Trying https://github.com/LeelaChessZero/lc0/issues/193
Making this PR partly to get the Appveyor binaries hehe.

### DIFF
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -169,7 +169,7 @@ std::string Node::DebugString() const {
 
 void Node::MakeTerminal(GameResult result) {
   is_terminal_ = true;
-  q_ = (result == GameResult::DRAW) ? 0.0f : 1.0f;
+  minmax_q_ = q_ = (result == GameResult::DRAW) ? 0.0f : 1.0f;
 }
 
 bool Node::TryStartScoreUpdate() {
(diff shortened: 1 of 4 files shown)


src/mcts/node.cc
```diff
@@ -181,8 +181,17 @@ bool Node::TryStartScoreUpdate() {
...
 
 void Node::FinalizeScoreUpdate(float v) {
-  // Recompute Q.
+  // Recompute MCTS Q.
   q_ += (v - q_) / (n_ + 1);
+  // Recompute MinMax Q.
+  if (n_ == 0) {
+    minmax_q_ = v;
+  } else {
+    for (Node* child : ChildNodes()) {
```
DanielUranga: I noticed that I'm missing the `minmax_q_ = 1.0f` initialization here, but also it would be better to use `min_element` from C++ std.

DanielUranga:
Initial results (10 games match):
```Lc0 - Lc0-minmax : 3,5/10 1-4-5 (=0==000==1?)  35%  -108
-----------------Lc0-minmax-----------------
Lc0-minmax - Lc0 : 6,5/10 4-1-5 (=1==111==0?)  65%  +108```
Sample size is really small, but at least is not undoubtedly weaker.

dubslow:
`but at least is not undoubtedly weaker` you can't even conclude that. At most you can conclude it's not 300 elo weaker.

DanielUranga:
Superceded by https://github.com/LeelaChessZero/lc0/pull/243

Issue #86, opened 2018-06-15
Depth tracking needs re-design and re-implementation


dubslow:
The current `depth` output is totally useless for people and protocol alike, since it never goes above 2, ever.

PR #74 would address that particular issue, replacing the current depth with something that's approximately right, and at least somewhat useful, but it isn't necessarily ideal.

So, even if/when 74 is merged, we need a better design for depth. Doing a logarithmic approximation with calculated branching factor is considered suboptimal; perhaps better is a linear average of the depth of all nodes. There are other choices as well.

And, even aside from design choices and issues, the current implementations, either of them, suffer the drawback of using storage inside the `Node` structure, which is a precious resource per issue #13. No matter which design is chosen, it should not require any storage inside each `Node`. This shouldn't be too hard; it should be relatively straightforward to maintain a variable or two at the top level `Search` , which is updated during node eval backpropagation. This would be effective with the caveat that it is stateful: tree reuse would require complete recursive calculation of the top level variables for the subtree in question. (One potential workaround is to have `Search` do a different depth calculation for each potential first and second ply move; then, when the tree is pruned for these two moves, we already have the requisite calculation done, and save the full recursive recalculation only for more extreme (i.e. rarer) tree pruning scenarios.)

mooskagh:
I do have some ideas to share, but let's address issues that are blocking for lc0 launch for training first.

Videodr0me:
Just a couple of thoughts for whenever this gets any priority (obviously not now):
1) I would propose  moving depth calculation to picknode instead of backpropagation. Mainly because in backpropagation we never know during traversel of the tree upwards what depth we are really in until we reach root. This makes the info not really usable for any (future) backpropagaton schemes (e.g. depth depending formulas). While in PickNode we traverse top down so depth at any level has actual meaning and can be used (in the future) for deciding which node to expand (depth dependend PUCT).

2) Full depth in my opinion is not needed at all. It does not convey any useful info (even in long searches  over hours hardly ever reaches 4 - 5) and it does have excessive overhead as in the backpropagation traversal on every levels all nodes are checked. Also in contrast to a/b engines full depth does not gurantee that a tactic within those moves is really found (it might take a while for visits to catch up, even if Q is already properly backpropagated)

3) Max depth is useful and should be used for what traditional engines call selective depth (as it is currently the case)

[...]

PR #222, opened 2018-07-31
Change "full" depth to average depth

dubslow:
Since there remain major disagreements about the performance of recalculating reused-tree depths outside of the Search code, this patch instead leaves the updated depth inside the Node structure, as currently exists in master. Even so, it's still technically less cpu cycles, although probably trivially so. It is also +2 bytes in Node, but it seems that these +2 bytes are already wasted by alignment, so at least on g++/x86 compiles, this is memory neutral. With this patch: `Move size 2 for estimated Edge size 6 vs actual Edge size 8 Node size 72 for total estimated bytes/Node of 352` and on master: `Move size 2 for estimated Edge size 6 vs actual Edge size 8 Node size 72 for total estimated bytes/Node of 352`. I'm pretty sure it didn't used to be this way until the introduction of the cached TotalVisitedPolicy thing, I guess that extra float is causing the alignment problem. (And yes, I double-triple checked I was actually using master when running the comparison.)

So it seems that this patch is strictly better than status quo, being identical memory and trivially less cpu, and also having the benefit that "full" depth is replaced with something much more [useful](https://github.com/LeelaChessZero/lc0/pull/124#issue-198461389) yet which also satisfies my "one move analysis, one depth" [criterion](https://github.com/LeelaChessZero/lc0/issues/86#issuecomment-408179482). (In particular, full depth is bugged with having sometimes depth > 1000 output, but most importantly, it always shows the ~same value for nearly all searches of any sort, including trees that are order of magnitudes different in size, or trees which are the same size but wildly different shape. Average depth fixes all these problems.)

Although I haven't had time to investigate further about the performance of my old PR #160, which I still provisionally believe to be superior pending further investigation (several bytes fewer memory, less total cpu time, less cpu time in search code), I'm willing to close that one if this one is approved.

### DIFF
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -175,10 +175,10 @@ class Node {
   void FinalizeScoreUpdate(float v);
 
   // Depth tracking and retrieval functions.
-  void UpdateMaxDepth(const uint16_t& depth) {
+  void UpdateMaxDepth(const uint16_t depth) {
     if (depth > max_depth_) max_depth_ = depth;
   }
-  void UpdateAverageDepth(const uint16_t& depth) {
+  void UpdateAverageDepth(const uint16_t depth) {
     // Denominator should be total+1, but we exclude the root from the average.
     average_depth_ += (depth - average_depth_) / n_;
   }
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -27,6 +27,7 @@
 
 #pragma once
 
+#include <cmath>
 #include <iostream>
 #include <memory>
 #include <mutex>
@@ -177,9 +178,15 @@ class Node {
   void UpdateMaxDepth(const uint16_t& depth) {
     if (depth > max_depth_) max_depth_ = depth;
   }
+  void UpdateAverageDepth(const uint16_t& depth) {
+    // Denominator should be total+1, but we exclude the root from the average.
+    average_depth_ += (depth - average_depth_) / n_;
+  }
+
   uint16_t GetMaxDepth() const { return max_depth_; }
-  void UpdateAverageDepth(const uint16_t& depth) { cumulative_depth_ += depth; }
-  uint16_t GetAverageDepth() const;
+  uint16_t GetAverageDepth() const {
+    return static_cast<uint16_t>(std::round(average_depth_));
+  }
 
   V3TrainingData GetV3TrainingData(GameResult result,
                                    const PositionHistory& history) const;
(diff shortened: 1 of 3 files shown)


dubslow:
Actually, SunnyWar just provided an analysis of `Node`'s layout and alignment, and the biggest offender is that the `bool is_terminal_` is costing 8 of those 72 bytes, though I think this PR changes it to the bool costing 4 bytes and `uint32_t cumulative_depth_` costing 6 bytes, though there's other alignment penalties that can probably be optimized as well. (We could even shrink `visited_policy_` to 2 bytes like in the policy-shrink PR if that would help.)

SunnyWar:
@dubslow I've been poking around at some of the other objects that appear to be used a lot.  This one seems unusually large and I see several places where it is copied by value.

1>class lczero::MoveExecution	size(64):
1>	+---
1> 0	| Move move
1>  	| <alignment member> (size=6)
1> 8	| ChessBoard board
1>56	| reset_50_moves
1>  	| <alignment member> (size=7)
1>	+---

Rearranging saves a bit of space:

1>class lczero::MoveExecution	size(56):
1>	+---
1> 0	| ChessBoard board
1>48	| Move move
1>50	| reset_50_moves
1>  	| <alignment member> (size=5)
1>	+---

src/mcts/node.h
```diff
@@ -177,16 +174,25 @@ class Node {
   // * N-in-flight (-=1)
   void FinalizeScoreUpdate(float v);
 
-  // Updates max depth, if new depth is larger.
-  void UpdateMaxDepth(int depth);
+  // Depth tracking and retrieval functions.
+  void UpdateMaxDepth(const uint16_t& depth) {
```
SunnyWar: Why pass by reference? As is it compiles to void UpdateAverageDepth(unsigned short const & __ptr64)
versus void UpdateAverageDepth(const uint16_t depth) which compile to void UpdateAverageDepth(unsigned short)

src/mcts/node.h
```diff
@@ -177,16 +174,25 @@ class Node {
...
+    if (depth > max_depth_) max_depth_ = depth;
+  }
+  void UpdateAverageDepth(const uint16_t& depth) {
+    // Denominator should be total+1, but we exclude the root from the average.
+    average_depth_ += (depth - average_depth_) / n_;
+  }
```
SunnyWar: Why pass by reference? As is it compiles to void UpdateAverageDepth(unsigned short const & __ptr64)
versus void UpdateAverageDepth(const uint16_t depth) which compile to void UpdateAverageDepth(unsigned short)

dubslow:
In support of this:

depth_poll

mooskagh:
Closing for now.
But after tree detaching is implemented, there will be no depth reporting (initially) and we'll have a new round of depth discussion with fresh strength. :)

PR #223, opened 2018-07-31
Node rearrangement to improve alignment and save padding memory

dubslow:
Per the discussion with @SunnyWar in the sidetracked out of order PR. Starting this PR now mostly so that we have a place to continue the discussion without polluting the other PR.

I made this commit off of the depth PR, since as demonstrated those +2 bytes aren't actually costing anything given current alignment. I can change that as necessary. 

This does reduce 72 bytes to 64 without changing any members, only reordering them. That's a bit over 2% relative memory savings, admittedly not much; after the Edge shrink is merged, it will be closer to 4% relative savings, and the potential future savings mentioned below could make it more than 5% savings.


src/mcts/node.cc
```diff
@@ -193,22 +193,10 @@ void Node::FinalizeScoreUpdate(float v) {
...
-  if (full_depth_ > *depth) return false;
-  for (Node* child : ChildNodes()) {
-    if (*depth > child->full_depth_) *depth = child->full_depth_;
-  }
-  if (*depth >= full_depth_) {
-    full_depth_ = ++*depth;
-    return true;
-  }
-  return false;
+uint16_t Node::GetAverageDepth() const {
```
mooskagh: How does this change belong to member reordering?

src/mcts/node.h
```diff
@@ -209,35 +208,49 @@ class Node {
   std::string DebugString() const;
 
  private:
-  // List of edges.
+  // Turns out the alignment and padding of the node was costing more than 8
```
mooskagh: Those "turns out"-like comments are really too long for someone to read them and update when they are wrong. I don't think that any of padding-related comments are needed.
Maybe just something like: `// This class has lots of instances, take care about padding when adding new members` but even that is too obvious.

src/mcts/node.h
```diff
@@ -209,35 +208,49 @@ class Node {
...
+  // bytes on modern compilers. Simply rearranging the members leads to
+  // significant savings on padding. And if rearranging members for alignment,
+  // may as well arrange them in order of size, to ease future changes.
+
+  // First, the 16-byte EdgeList, which is itself 10 real bytes and 6 pad bytes.
+  // TODO: shrink the padding on this somehow?
+  // Maybe even merge it into this class??
```
mooskagh: My initial plan was to make EdgeList packed, and take care that it will be with proper alignment in Node, with uint16 right after it.
Then I decided that it's too hacky. I plan to replace pointer to index in EdgePool anyway, will reduce it from 8 to 4 bytes.

Out of EdgeList edgesize is how I initially implemented it for alignment reasons, didn't like it.

src/mcts/node.h
oscardssmith: given how dramatic an effect it can have, I'd keep an arguably too obvious comment. honestly I'm kind of surprised that compilers don't do this themselves

src/mcts/node.h
oscardssmith: 2 questions
1 do I understand correctly that edge list is just a linked list?
2. if so, could you store 4 values per node to fix alignment and save memory?

SunnyWar: @oscardssmith The edge list is a pointer to a vector. I've looked it over and I don't see how to make it smaller. Though it does look like there are few ways to make faster.

src/mcts/node.cc
mooskagh: Ping.

dubslow: I'll redo this PR shortly

dubslow:
Rebased the relevant commit on current master.

Master currently has 53 real bytes and 7 padding bytes, for 64 aligned bytes; however the padding is scattered throughout the Node such adding any single new 4 byte member would increase usage to 72 aligned bytes, and adding 2 byte members could only be done where such padding already exists.

Instead, this commit rearranges all the padding to the end, such that removing even 1 real byte would save 8 aligned bytes, or we could add a 4 byte member *and* a 2 byte member and stay at the current aligned byte total. So even though this doesn't save memory instantly, it will lead to savings after future changes compared to without this commit.

src/mcts/node.h
```diff
@@ -206,33 +206,42 @@ class Node {
   std::string DebugString() const;
 
  private:
-  // List of edges.
+  // Turns out the alignment and padding of the Node was costing more than 8
```
mooskagh:
```cpp
// To minimize number of padding bytes and avoid having unnecessary padding when
// new fields are added, we store fields sorted by size: first large structures, then 8-byte fields, then 4-byte, 
// 2-byte and 1-byte.
```

src/mcts/node.h
```diff
@@ -206,33 +206,42 @@ class Node {
...
+  // significant savings on padding. As such, instead of being organized
+  // conceptually, the members are merely ordered from largest to smallest.
+
+  // TODO: shrink the padding on this somehow? It takes 16 bytes even though
+  // only 10 are real! Maybe even merge it into this class??
   EdgeList edges_;
-  // Index of this node is parent's edge list.
-  uint16_t index_;
+
+  // Pointer to a parent node. nullptr for the root.
```
mooskagh:
```cpp
// 8-byte fields
```
(I think you had that initially)

src/mcts/node.h
```diff
@@ -206,33 +206,42 @@ class Node {
...
+  Node* parent_ = nullptr;
+  // Pointer to a first child. nullptr for a leaf node.
+  std::unique_ptr<Node> child_;
+  // Pointer to a next sibling. nullptr if there are no further siblings.
+  std::unique_ptr<Node> sibling_;
+
   // Average value (from value head of neural network) of all visited nodes in
```
mooskagh: `// 4-byte fields`

src/mcts/node.h
```diff
@@ -206,33 +206,42 @@ class Node {
...
   // of the player who "just" moved to reach this position, rather than from the
   // perspective of the player-to-move for the position.
   float q_ = 0.0f;
+  // Sum of policy priors which have had at least one playout.
+  float visited_policy_ = 0.0f;
   // How many completed visits this node had.
   uint32_t n_ = 0;
-  // (aka virtual loss). How many threads currently process this node (started
+
+  // Index of this node is parent's edge list.
```
mooskagh: `// 2-byte fields`

src/mcts/node.h
```diff
@@ -206,33 +206,42 @@ class Node {
...
+  // (AKA virtual loss.) How many threads currently process this node (started
   // but not finished). This value is added to n during selection which node
   // to pick in MCTS, and also when selecting the best move.
   uint16_t n_in_flight_ = 0;
-  // Sum of policy priors which have had at least one playout.
-  float visited_policy_ = 0.0f;
 
-  // Does this node end game (with a winning of either sides or draw).
+  // Whether or not this node end game (with a winning of either sides or draw).
   bool is_terminal_ = false;
```
mooskagh: `// 1-byte fields.`

src/mcts/node.h
```diff
@@ -206,33 +206,42 @@ class Node {
...
 
-  // Pointer to a parent node. nullptr for the root.
-  Node* parent_ = nullptr;
-  // Pointer to a first child. nullptr for a leaf node.
-  std::unique_ptr<Node> child_;
-  // Pointer to a next sibling. nullptr if there are no further siblings.
-  std::unique_ptr<Node> sibling_;
 
   // TODO(mooskagh) Unfriend NodeTree.
   friend class NodeTree;
```
mooskagh: I think it would also be useful to add static assert just to watch total structure size and avoid surprises with other compilers, something like:
```cpp
// This assert is here just to track structure size. Feel free to change when changing Node object.
static_assert(sizeof(Node) == 52, "Unexpected size of Node");
```
right after Node structure.

src/mcts/node.h
dubslow: Done, with slightly tweaked wording, rereview

src/mcts/node.h
dubslow: I did, but it seemed slightly unnecessary in combination with other comments from you about unnecessary comments. I guess it doesn't hurt to have them

src/mcts/node.h
dubslow: Done

src/mcts/node.h
dubslow: done

src/mcts/node.h
dubslow: done

src/mcts/node.h
dubslow: done

PR #224, opened 2018-08-02
ispc support

ra1u:
It turns out that ispc tool https://ispc.github.io/ enables few percent faster evaluation in tool.  This patch is collaboration of @frpays, @borg323 and other contributors who were involved with testing and ideas. 

Ispc tool if available on system is used to generate object files with implementation of WinogradConvolution3::TransformIn and WinogradConvolution3::TransformOut and is relevant for users running blas on CPU. 

Current default is that build generates object files appropriate for native cpu employing instruction set of that cpu. When building binary that will run on other cpu it is advisable to disable meson flag ispc_native_only. This allows running on any instruction set. This was disabled by default, because such linking does not work on every PC. Both Linux and Windows are supported.


borg323:
I think this is a perfectly safe since in my tests the code produced identical results with the default Winograd convolution code. The ispc code is only used if the ispc compiler is installed but this is trivial since only a single executable needs to be in the path.

PR #225, opened 2018-08-02
cudnn message

borg323:
I'm not sure whether this is necessary, or even a good idea, but @mooskagh asked for it.

### DIFF
--- a/src/neural/network_cudnn.cu
+++ b/src/neural/network_cudnn.cu
@@ -997,6 +997,10 @@ class CudnnNetwork : public Network {
       }
     }
 
+    printf(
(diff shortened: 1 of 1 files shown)


src/neural/network_cudnn.cu
```diff
@@ -997,6 +997,10 @@ class CudnnNetwork : public Network {
       }
     }
 
+    printf(
```
mooskagh: Should go to stderr if it's even a good idea. But probably it's fine without it.
So let's drop it! Sorry for making you make this PR, it's probably good to go without the message.

PR #227, opened 2018-08-03
[WIP] Adding UCI multipv

kylemiller3:
Functions on my Arena Chess but some of the output is not 100% correct yet. Probably obsoletes some variables like best_move_edge_ as well since you can just get that from the list object now. I could use some guidance on how to integrate further.

### DIFF
--- a/src/chess/callbacks.h
+++ b/src/chess/callbacks.h
@@ -71,6 +71,8 @@ struct ThinkingInfo {
   optional<int> score;
   // Best line found. Moves are from perspective of white player.
   std::vector<Move> pv;
+  // Multi pv index
+  int pv_idx = -1;
   // Freeform comment.
   std::string comment;
 
--- a/src/chess/callbacks.h
+++ b/src/chess/callbacks.h
@@ -72,7 +72,7 @@ struct ThinkingInfo {
   // Best line found. Moves are from perspective of white player.
   std::vector<Move> pv;
   // Multi pv index
-  int pv_idx = -1;
+  int multipv_idx = -1;
   // Freeform comment.
   std::string comment;
 
--- a/src/chess/callbacks.h
+++ b/src/chess/callbacks.h
@@ -71,7 +71,7 @@ struct ThinkingInfo {
   optional<int> score;
   // Best line found. Moves are from perspective of white player.
   std::vector<Move> pv;
-  // Multi pv index
+  // Multi pv index for UCI.
   int multipv_idx = -1;
   // Freeform comment.
   std::string comment;
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -162,7 +162,7 @@ class Node {
   uint16_t GetNumEdges() const { return edges_.size(); }
 
   // Calculates score in centipawns.
-  float GetCpScore() const {
+  int GetCpScore() const {
     return 290.680623072 * tan(1.548090806 * q_);
   }
 
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -27,10 +27,10 @@
 
 #pragma once
 
+#include <cmath>
 #include <iostream>
 #include <memory>
 #include <mutex>
-#include <cmath>
 #include "chess/board.h"
 #include "chess/callbacks.h"
 #include "chess/position.h"
@@ -161,6 +161,11 @@ class Node {
   uint16_t GetMaxDepth() const { return max_depth_; }
   uint16_t GetNumEdges() const { return edges_.size(); }
 
+  // Calculates score in centipawns.
+  float GetCpScore() const {
+    return 290.680623072 * tan(1.548090806 * q_);
+  }
+
   // Makes the node terminal and sets it's score.
   void MakeTerminal(GameResult result);
 
@@ -285,11 +290,6 @@ class EdgeAndNode {
   // Passed numerator is expected to be equal to (cpuct * sqrt(N[parent])).
   float GetU(float numerator) const {
     return numerator * GetP() / (1 + GetNStarted());
-  }
-
-  // Calculates score in centipawns
-  static float GetCpScore(const EdgeAndNode& edge) {
-    return 290.680623072 * tan(1.548090806 * edge.GetQ(0));
   }
 
   std::string DebugString() const;
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -30,6 +30,7 @@
 #include <iostream>
 #include <memory>
 #include <mutex>
+#include <cmath>
 #include "chess/board.h"
 #include "chess/callbacks.h"
 #include "chess/position.h"
@@ -284,6 +285,11 @@ class EdgeAndNode {
   // Passed numerator is expected to be equal to (cpuct * sqrt(N[parent])).
   float GetU(float numerator) const {
     return numerator * GetP() / (1 + GetNStarted());
+  }
+
+  // Calculates score in centipawns
+  static float GetCpScore(const EdgeAndNode& edge) {
+    return 290.680623072 * tan(1.548090806 * edge.GetQ(0));
   }
 
   std::string DebugString() const;
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -57,6 +57,7 @@ const char* Search::kPolicySoftmaxTempStr = "Policy softmax temperature";
 const char* Search::kAllowedNodeCollisionsStr =
     "Allowed node collisions, per batch";
 const char* Search::kStickyCheckmateStr = "Ignore alternatives to checkmate";
+const char* Search::kMultiPvStr = "MultiPV";
 
 namespace {
 const int kSmartPruningToleranceNodes = 100;
@@ -89,6 +90,7 @@ void Search::PopulateUciParams(OptionsParser* options) {
   options->Add<IntOption>(kAllowedNodeCollisionsStr, 0, 1024,
                           "allowed-node-collisions") = 0;
   options->Add<BoolOption>(kStickyCheckmateStr, "sticky-checkmate") = false;
+  options->Add<IntOption>(kMultiPvStr, 1, 100, "multipv") = 1;
 }
 
 Search::Search(const NodeTree& tree, Network* network,
@@ -116,7 +118,8 @@ Search::Search(const NodeTree& tree, Network* network,
       kCacheHistoryLength(options.Get<int>(kCacheHistoryLengthStr)),
       kPolicySoftmaxTemp(options.Get<float>(kPolicySoftmaxTempStr)),
       kAllowedNodeCollisions(options.Get<int>(kAllowedNodeCollisionsStr)),
-      kStickyCheckmate(options.Get<bool>(kStickyCheckmateStr)) {}
+      kStickyCheckmate(options.Get<bool>(kStickyCheckmateStr)),
+      kMultiPv(options.Get<int>(kMultiPvStr)) {}
 
 namespace {
 void ApplyDirichletNoise(Node* node, float eps, double alpha) {
@@ -140,7 +143,7 @@ void ApplyDirichletNoise(Node* node, float eps, double alpha) {
 }  // namespace
 
 void Search::SendUciInfo() REQUIRES(nodes_mutex_) {
-  if (!best_move_edge_) return;
+  if (!best_move_edge_ || asc_move_edges_.size() == 0) return;
   last_outputted_best_move_edge_ = best_move_edge_.edge();
   uci_info_.depth = root_node_->GetFullDepth();
   uci_info_.seldepth = root_node_->GetMaxDepth();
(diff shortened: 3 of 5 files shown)


src/chess/callbacks.h
```diff
@@ -71,6 +71,8 @@ struct ThinkingInfo {
   optional<int> score;
   // Best line found. Moves are from perspective of white player.
   std::vector<Move> pv;
+  // Multi pv index
+  int pv_idx = -1;
```
dubslow: Probably better to straight up name this `multipv_idx` or `multipv_index`

src/mcts/search.h
```diff
@@ -152,6 +159,7 @@ class Search {
   const int64_t initial_visits_;
 
   mutable SharedMutex nodes_mutex_;
+  std::list<EdgeAndNode> asc_move_edges_ GUARDED_BY(nodes_mutex_);
```
dubslow: Not sure this needs to be a member variable? Can't it just be a local variable in like the two places it's needed?

src/mcts/search.h
```diff
@@ -103,10 +104,16 @@ class Search {
...
   static const char* kStickyCheckmateStr;
+  static const char* kMultiPvStr;
 
  private:
   // Returns the best move, maybe with temperature (according to the settings).
   std::pair<Move, Move> GetBestMoveInternal() const;
+  
+  // Returns a list of all children with, with or without temperature in descending order of visits.
+  std::list<EdgeAndNode> GetAscChildrenNoTemperature(Node* parent) const;
+  std::list<EdgeAndNode> GetAscChildrenWithTemperature(Node* parent,
```
dubslow: Kind of unnecessary (not to mention meaningless) function, per Discord discussion

src/mcts/search.h
```diff
@@ -30,6 +30,7 @@
 #include <functional>
 #include <shared_mutex>
 #include <thread>
+#include <list>
```
dubslow: list -> vector, per Discord discussion

src/mcts/search.cc
```diff
@@ -915,8 +938,10 @@ void SearchWorker::DoBackupUpdate() {
       // Best move.
       if (n->GetParent() == search_->root_node_ &&
           search_->best_move_edge_.GetN() <= n->GetN()) {
+        search_->asc_move_edges_ =
+            search_->GetAscChildrenNoTemperature(search_->root_node_);
```
dubslow: Don't see a reason we need to always keep track of this, see other comments

src/mcts/search.cc
```diff
@@ -381,35 +392,46 @@ std::pair<Move, Move> Search::GetBestMoveInternal() const
...
 
-// Returns a child with most visits.
 EdgeAndNode Search::GetBestChildNoTemperature(Node* parent) const {
-  EdgeAndNode best_edge;
+  return GetAscChildrenNoTemperature(parent).back();
+}
+
+EdgeAndNode Search::GetBestChildWithTemperature(Node* parent,
+                                                   float temperature) const {
+  return GetAscChildrenWithTemperature(parent, temperature).back();
```
dubslow: It's simpler code to do this, but is it worth the cpu time of fully sorting the list instead of just finding the single best one every time? This function is called once for every playout, it might be worth the optimization to keep them both separate

src/mcts/search.cc
```diff
@@ -293,6 +303,7 @@ void Search::MaybeTriggerStop() {
     best_move_callback_({best_move_.first, best_move_.second});
     responded_bestmove_ = true;
     best_move_edge_ = EdgeAndNode();
+    asc_move_edges_ = std::list<EdgeAndNode> { best_move_edge_ };
```
dubslow: again, not necessary to always be tracking this

src/mcts/search.cc
```diff
@@ -150,17 +153,24 @@ void Search::SendUciInfo() REQUIRES(nodes_mutex_) {
...
+  uci_info_.comment.clear();
 
-  bool flip = played_history_.IsBlackToMove();
-  for (auto iter = best_move_edge_; iter;
-       iter = GetBestChildNoTemperature(iter.node()), flip = !flip) {
-    uci_info_.pv.push_back(iter.GetMove(flip));
-    if (!iter.node()) break;  // Last edge was dangling, cannot continue.
+  int pv_idx = 0;
+  for (auto edge_ptr = asc_move_edges_.rbegin(); 
+         edge_ptr != asc_move_edges_.rend(); ++edge_ptr) {
```
dubslow: Couldn't this loop just be written as `for (const auto& edge : asc_move_edges) {}` ?

src/mcts/search.cc
```diff
@@ -150,17 +153,24 @@ void Search::SendUciInfo() REQUIRES(nodes_mutex_) {
...
+    bool flip = played_history_.IsBlackToMove();
+    uci_info_.score = 290.680623072 * tan(1.548090806 * edge_ptr->GetQ(0));
```
dubslow: I think this PR should take the opportunity to factor out this formula into a `Node::GetCPScore()` function, instead of repeating it two or three times in the uci parts of this file. Kinda old miniwart in the codebase, may as well fix it here with the rest of this PR

src/mcts/search.cc
```diff
@@ -150,17 +153,24 @@ void Search::SendUciInfo() REQUIRES(nodes_mutex_) {
...
 
-  bool flip = played_history_.IsBlackToMove();
-  for (auto iter = best_move_edge_; iter;
-       iter = GetBestChildNoTemperature(iter.node()), flip = !flip) {
-    uci_info_.pv.push_back(iter.GetMove(flip));
-    if (!iter.node()) break;  // Last edge was dangling, cannot continue.
+  int pv_idx = 0;
+  for (auto edge_ptr = asc_move_edges_.rbegin(); 
+         edge_ptr != asc_move_edges_.rend(); ++edge_ptr) {
+    bool flip = played_history_.IsBlackToMove();
```
dubslow: Per previous Discord discussion, maybe we could take the opportunity with this PR to rename this variable more descriptively, like `is_black` or `is_mirrored` or something

src/mcts/search.cc
```diff
@@ -150,17 +153,24 @@ void Search::SendUciInfo() REQUIRES(nodes_mutex_) {
...
+    uci_info_.score = 290.680623072 * tan(1.548090806 * edge_ptr->GetQ(0));
+    uci_info_.pv.clear();
+    for (auto iter = *edge_ptr; iter;
```
dubslow: No harm in making this `const auto&` I think

src/mcts/search.h
kylemiller3: I really wasn't sure if I was going to cause a bug with multi-threading when I made this decision but I see that SendUciInfo requires the nodes_mutex_ so on second thought it shouldn't be modified in that function.

src/mcts/search.cc
kylemiller3: Would have written it like if I didn't but one of the lists was naturally ascending so had to use reverse iterators.

kylemiller3:
I think I got all the requested changes in this commit. I'm not sure what other UCI information should be updated in each multipv loop iteration.

src/mcts/search.cc
```diff
@@ -381,8 +392,36 @@ std::pair<Move, Move> Search::GetBestMoveInternal() const
   return {best_node.GetMove(played_history_.IsBlackToMove()), ponder_move};
 }
 
+// all children in descending order of visits
```
dubslow: Capitalized and period

mooskagh:
I didn't look into that at all, but when it will be ready and not [WIP], I'd surely like to take a look before it's merged.

src/mcts/node.h
```diff
@@ -286,6 +287,11 @@ class EdgeAndNode {
     return numerator * GetP() / (1 + GetNStarted());
   }
 
+  // Calculates score in centipawns
+  static float GetCpScore(const EdgeAndNode& edge) {
```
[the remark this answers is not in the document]
kylemiller3: That's fine. I use static methods when I don't modify the state of the object or need to override its behavior. Since this was a pure calculation I thought that was appropriate. I did make a mistake on my own terms here I would have just passed in the Q score as the parameter since there is no need for EdgeAndNode object at all.

src/mcts/search.cc
```diff
@@ -150,17 +153,25 @@ void Search::SendUciInfo() REQUIRES(nodes_mutex_) {
...
-    uci_info_.pv.push_back(iter.GetMove(flip));
-    if (!iter.node()) break;  // Last edge was dangling, cannot continue.
+  // Handle UCI multipv. Loop until we are out of moves or kMultiPv
+  std::vector<EdgeAndNode> sorted_edges = GetDescChildrenNoTemperature(root_node_);
+  for (int pv_idx = 0;
+       pv_idx < std::min(kMultiPv, static_cast<int>(sorted_edges.size()));
+       ++pv_idx) {
+    EdgeAndNode edge = sorted_edges[pv_idx];
+    bool is_black = played_history_.IsBlackToMove();
+    uci_info_.score = EdgeAndNode::GetCpScore(edge);
```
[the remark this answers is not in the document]
kylemiller3: I believe I was doing this (just with a static method) and I don't believe the node method is always non-null because it was crashing on me. I decided to go with EdgeAndNode for that reason.

kylemiller3:
@mooskagh It can be used as-is if you just want the top lines and evaluations. Not all of the information is accurate such as depth and nodes per line. I haven't looked into that or how it works but perhaps @dubslow could give me some more guidance on updating those fields or I can update them myself if I look at how those work when I get around to it.

src/mcts/node.h
```diff
@@ -30,6 +30,7 @@
 #include <iostream>
 #include <memory>
 #include <mutex>
+#include <cmath>
```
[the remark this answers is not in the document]
mooskagh: Fyi this will be done automatically if you run clang-format

src/mcts/search.cc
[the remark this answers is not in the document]
kylemiller3: Without checking for nulls I get a crash like this 

```
2018-08-07 17:30:30.397<--1:Creating backend [cudnn]...
2018-08-07 17:30:31.042<--1:readyok
2018-08-07 17:30:31.042<--1:readyok
2018-08-07 17:30:31.042<--1:readyok
2018-08-07 17:30:31.080<--1:info depth 2 seldepth 2 time 19 nodes 3 score cp 18 hashfull 0 nps 157 pv c2c4 e7e6 multipv 1
2018-08-07 17:30:31.088<--1:info depth 2 seldepth 2 time 19 nodes 3 score cp 10 hashfull 0 nps 157 pv e2e3 g7g6 multipv 2
2018-08-07 17:30:31.112<--1:BytesRead > 0
2018-08-07 17:30:31.113<--1:-- LiesThread.Execute terminated --
```
With edge.node()->GetCpScore(); sometimes

kylemiller3:
I think that's all the changes requested.

src/mcts/node.h
```diff
@@ -160,6 +161,11 @@ class Node {
   uint16_t GetMaxDepth() const { return max_depth_; }
   uint16_t GetNumEdges() const { return edges_.size(); }
 
+  // Calculates score in centipawns.
+  float GetCpScore() const {
```
dubslow: Should just return an `int` I think, to match the `ThinkingInfo` type, which is typed as such because the precision is specifically capped at .01 pawns

src/mcts/search.cc
```diff
@@ -150,17 +153,25 @@ void Search::SendUciInfo() REQUIRES(nodes_mutex_) {
...
+    EdgeAndNode edge = sorted_edges[pv_idx];
+    bool is_mirrored = played_history_.IsBlackToMove();
+    if (edge.HasNode()) uci_info_.score = edge.node()->GetCpScore();
+    uci_info_.pv.clear();
+    for (auto iter = edge; iter;
+         iter = GetBestChildNoTemperature(iter.node()), is_mirrored = !is_mirrored) {
+      uci_info_.pv.push_back(iter.GetMove(is_mirrored));
+      if (!iter.node()) break;  // Last edge was dangling, cannot continue.
+    }
+    uci_info_.multipv_idx = pv_idx + 1;
```
dubslow: We should only set this if indeed `kMultiPv` > 1

src/mcts/node.h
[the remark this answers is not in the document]
kylemiller3: Thanks

kylemiller3:
I noticed you cannot dynamically change the multipv mode variable while the engine is running. It needs to be updated somehow in a method or the option should cause the engine to restart.

dubslow:
One may change the UCI setting at any time.

This looks good to me if we don't include include specific-to-variation depth or node counts. Doing those will be a bit harder to implement.

kylemiller3:
I'm not able to change it and have my UCI engine show it properly with this implementation without stopping and restarting the analysis.

I'd be fine with implementing those but I don't really know where to start. This is fine for now though.

kylemiller3:
Ah nevermind it seems to take the multipv setting fine on the next move.

mooskagh:
Hi @kylemiller3, are you still active on this?
I think we want multipv to be in for v0.19.

mooskagh:
Implemented independently in #375

kylemiller3:
@mooskagh Do you need me to do anything if you implemented it in #375 ?

Issue #19, opened 2018-06-02
lc0: Support configuration files


mooskagh:
_From @mooskagh on May 27, 2018 6:11_

Add configuration file as yet another way to change `lc0` parameters (in addition to command line flags and UCI parameters).

It's convenient to e.g. setup backend/hardware configuration only once, or some people like --verbose-move-stats always enabled.
Also some GUIs have no easy way to either supply command line flags or uci params. That would be helpful for them too.

_Copied from original issue: glinscott/leela-chess#669_

mooskagh:
Implemented in https://github.com/LeelaChessZero/lc0/pull/231

PR #231, opened 2018-08-03
Implement Configuration Files

cwbriscoe:
This PR will hopefully satisfy the https://github.com/LeelaChessZero/lc0/issues/19 help request.

The configuration file layout in the PR simply uses the same command syntax that you would use on the command line.  You also have the option to put each flag on a different line if you choose.  An "#" symbol at the beginning of the line is treated like a comment.  For example:

```
#this line is a comment
       #this line is also a comment
-t 1
--minibatch-size=128
#--minibatch-size=256
--max-prefetch=0
```
Will be the same as a command line that looks like:

`-t 1 --minibatch-size=128 --max-prefetch=0`

Parameters on the command line override parameters in the config file.  For instance:

`./lc0 -t 2`

This will cause two threads to be used instead of the one specified in the config file.

The default configuration file is currently set to .config.  This can be changed by using the new -c/--config command line option.

This has not been tested on Windows or Mac.

### DIFF
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -254,7 +254,7 @@ EngineLoop::EngineLoop()
 }
 
 void EngineLoop::RunLoop() {
-  if (!options_.ProcessAllFlags()) return;
+  if (!ConfigFile::Init(&options_) || !options_.ProcessAllFlags()) return;
   UciLoop::RunLoop();
 }
 
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -33,6 +33,7 @@
 #include "mcts/search.h"
 #include "neural/factory.h"
 #include "neural/loader.h"
+#include "utils/configfile.h"
 
 namespace lczero {
 namespace {
@@ -96,6 +97,8 @@ void EngineController::PopulateOptions(OptionsParser* options) {
                             "time-curve-right-width") = 74.0f;
 
   Search::PopulateUciParams(options);
+
+  ConfigFile::PopulateOptions(options);
 
   auto defaults = options->GetMutableDefaultsOptions();
 
--- a/src/selfplay/loop.cc
+++ b/src/selfplay/loop.cc
@@ -27,6 +27,7 @@
 
 #include "selfplay/loop.h"
 #include "selfplay/tournament.h"
+#include "utils/configfile.h"
 
 namespace lczero {
 
@@ -45,7 +46,7 @@ void SelfPlayLoop::RunLoop() {
   options_.Add<BoolOption>(kInteractive, "interactive") = false;
   SelfPlayTournament::PopulateOptions(&options_);
 
-  if (!options_.ProcessAllFlags()) return;
+  if (!ConfigFile::Init(&options_) || !options_.ProcessAllFlags()) return;
   if (options_.GetOptionsDict().Get<bool>(kInteractive)) {
     UciLoop::RunLoop();
   } else {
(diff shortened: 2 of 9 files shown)


src/selfplay/loop.cc
```diff
@@ -45,7 +46,7 @@ void SelfPlayLoop::RunLoop() {
   options_.Add<BoolOption>(kInteractive, "interactive") = false;
   SelfPlayTournament::PopulateOptions(&options_);
 
-  if (!options_.ProcessAllFlags()) return;
+  if (!ConfigFile::Init(&options_) || !options_.ProcessAllFlags()) return;
```
mooskagh: It's hard to say how it should work in selfplay vs uci mode.
It surely should not be the same config (if a person did some tuning for uci mode, and then started training, it should not use the same params), but maybe it makes sense to allow some config (e.g. for gpu config).

For now I think it's safer not to read any configs at all in selfplay mode.

Or... Maybe make default config name uci.config, selfplay.config, etc. RegisterMode and ConsumeCommand can be refactored to store mode, but probably that's better to do later. So, no config for selfplay mode for now.

src/engine.cc
```diff
@@ -97,6 +98,8 @@ void EngineController::PopulateOptions(OptionsParser* options) {
 
   Search::PopulateUciParams(options);
```
mooskagh: This place in file is too sparse, maybe remove that empty line?

src/utils/configfile.cc
```diff
@@ -0,0 +1,95 @@
...
+
+  std::cerr << "Found configuration file: " << filename << std::endl;
+
+  for(std::string line; getline( input, line );) {
+    // Remove all leading and trailing whitespace.
+    line = Trim(line);
+    // Ignore comments.
+    if (line.substr(0, 1) == "#") continue;
+
+    args += line + " ";
```
mooskagh: Why not append directly to `vector<string>` but instead build a string and then split it?

mooskagh:
As asked in Discord:

crem Yesterday at 9:49 AM

@cwbriscoe  in config, why not just have one parameter per line? Currently if there is some parameter with space inside, it will be split into two
Like..  `--weights=C:\Program Files\weights.txt`.   And `--weights="C:\Program Files\weights.txt"` won't work either because it's usually shell who removes the quotes, and because it will be still split.

cwbriscoe:
Crem - I am Midboss on discord so that is probably why I missed it.  I didn't think about windows and having spaces in filenames.  I could do as you suggested but how should I handle the short command options such as "-t 0".  Just force the long form or do special handling for that case?

mooskagh:
Forcing long options is fine, I think.
You can even check that every option starts with `--` and throw error if not.

cwbriscoe:
I changed to only allow one argument per line and to only allow long argument names.  "--" is enforced for now.  It didn't seem that there was a consensus on allowing arguments without "--".  I would be happy to add that later if desired.  Making the "=" optional is a bit trickier because you run into the same issue with splitting the line that I had before with Windows file paths.

dubslow:
I still think that using the `--` prefix is highly nonstandard and that there's no reason to use it.

But, as long as all distributions of lc0 (binary or source or...) include a default/example config file, it will probably be fine in practice.

mooskagh:
I think we need to package leela with a sample configuration file.
All lines could be commented there but it will be easy to see the format and uncomment.

PR #230, opened 2018-08-03
renamed smart-pruning-aggresiveness option to smart-pruning

borg323:
As @mooskagh asked. Only the option is renamed, the variable names are left as they were.

PR #234, opened 2018-08-04
Pruning rename

dubslow:
Alternative to #230 , separating time pruning from node pruning.

Things I'm not sure about: the name "node pruning", perhaps "visit pruning" would be better?; the description string for this option is in need of help too

Things I'm more sure about but willing to change: node pruning default off.

This would also "fix" the behavior of not searching forced moves, which might mess with resign.

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -49,7 +49,8 @@ const char* Search::kTemperatureStr = "Initial temperature";
 const char* Search::kTempDecayMovesStr = "Moves with temperature decay";
 const char* Search::kNoiseStr = "Add Dirichlet noise at root node";
 const char* Search::kVerboseStatsStr = "Display verbose move stats";
-const char* Search::kAggressiveTimePruningStr = "Aggressive smart pruning threshold";
+const char* Search::kTimePruningStr = "Time pruning aggression";
+const char* Search::kNodePruningStr = "Node pruning";
 const char* Search::kFpuReductionStr = "First Play Urgency Reduction";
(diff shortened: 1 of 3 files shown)


oscardssmith:
I like visit pruning

jjoshua2:
I thought crem specifically asked for only one parameter since we don't need a ton of parameter that don't help elo?

mooskagh:
Why would we need two separate parameters?
In general it's better to keep as few parameters as possible.

dubslow:
Separate parameters because they're kinda separate options -- it would be better, in my opinion, to have node pruning default off even while time management remains default on pruned. Honestly I'd be fine with disabling node pruning altogether, and just have time pruning. I don't hold these opinions especially strongly tho

mooskagh:
There is really no distinction between "time pruning" and node pruning.
They both do the same: if remaining nodes is not enough for particular move to win, this move is not considered. "Time" just does additional step of estimating "remaining nodes" rather than knowing it exactly.

The "aggressiveness" parameter is not to adjust to imperfections to that time estimation, it's something different and makes sense for nodes too:
If
MoveA   has 10000 visits
MoveB..MoveZ   have  10 visits each
Remaining moves is 15000, then it's possible that some of MoveB..MoveZ will overtake MoveA, but unlikely. Because of that it makes sense to risk and stop the search early.

dubslow:
Suffice it to say I don't agree.

Having said that, I don't think it's that big a deal either way.

PR #243, opened 2018-08-10
Minimax take2

DanielUranga:
This is a clean up of what worked best in https://github.com/LeelaChessZero/lc0/pull/221
The minmax component is weighted using the amount of visits for best child (the one with best Q) divided by the amount of visits of parent node.
Test results
- at fixed 200 nodes:
```
Score of lc0-minmax.exe vs lc0-main.exe: 70 - 34 - 96  [0.590] 200
Elo difference: 63.23 +/- 34.83
```
- at fixed 400 nodes:
```
Score of lc0-minmax.exe vs lc0-main.exe: 63 - 49 - 88  [0.535] 200
Elo difference: 24.36 +/- 36.13
```
- at fixed 800 nodes:
```
Score of lc0-minmax.exe vs lc0-main.exe: 70 - 33 - 97  [0.593] 200
Elo difference: 65.02 +/- 34.65
```
All test results available at https://jenkins.pirho.me/, also PGNs available there.
Will post results with higher node counts once I have them.


DanielUranga:
1000 nodes:
```
Score of lc0-minmax.exe vs lc0-main.exe: 52 - 29 - 119  [0.557] 200
Elo difference: 40.13 +/- 30.58
```

DanielUranga:
I think the main issue is that this adds 4 bytes to the Node structure.

mooskagh:
Adding 4 bytes for something that gives 40 Elo is worthwhile from my point of view.

dubslow:
The Elo gain seems to diminish as the search depth increases though. Are there any results from, say, 100K nodes/move?

DanielUranga:
Haven't tested at higher nodes/move than 10K, results
```
Score of lc0-minmax.exe vs lc0-main.exe: 14 - 9 - 77  [0.525] 100
Elo difference: 17.39 +/- 32.61
Finished match
```
We are checking the optimal `cpuct` value before going to higher node count. (OT: it looks like the current default `cpuct` is outdated).

DanielUranga:
Also I noticed something with this approach: let's say your best child node has Q=0.8 with N=80, and the current parent "MCTS" Q is 0.7 with N=100, the final minimax node value for the node will be: `0.8*0.8 + 0.7*0.2 = 0.78`.
Let's say that a new best child node is found with Q=0.9, but it just has one visit (N=1), minimax Q will be: `0.9*0.01+0.7*0.99 = 0.702`, which is a lower Q than before after finding a new best move.

So empirically this works at low node counts, but it has it's flaws. My main interest with this is to show that the search function can be improved and we could gain a decent amount of ELO here, and get input and ideas from others to achieve that.

thepok:
If it works for low node counts, it should be great in Training game generation?!?

Eddh:
I think I've seen some ideas around this multiple times. Most interesting one to me would be to consider late visits as more representative of a node value than early visits. Since they are more likely to be the best available option. so if a node at 1000 visit gets a new visit that is very good, then even this single visit should be weighted higher than the previous visits. I've never seen an implementation of this idea though.

borg323:
Another thing that needs to be considered is whether it works correctly with smart pruning.

DanielUranga:
I'm not familiar with smart pruning, I thought it was time management related only?

DanielUranga:
At 20k nodes, cpuct=1.8 for minimax:
```
Score of lc0-minmax2.exe vs lc0-main.exe: 25 - 15 - 110  [0.533] 150
Elo difference: 23.20 +/- 28.62
```

Videodr0me:
I am still unsure whether you are not just testing optimal puct vs. non-optimal puct. If this change is highly sensitive to cpuct (as it seems), a valid test would be a clopped cpuct for lc0-main vs. a clopped cpuct for the mini-max version. I might have missed this, and you are already doing this. Generally, cpuct is a moving target with (clop) optimal cpucts ranging from 1.6 to 3.2 for different NNs.

The second caveat I have is the low number of games (not so much the visits). Testing with 100 game matches basically has very large error margins. The multiplexing backend and lc0 selfplay mode usually allows one to rack up large number of games much faster e.g.:

```
lc0 selfplay --parallelism=8 --no-share-trees --backend=multiplexing "--backend-opts=cudnn(threads=2)" --games=1000 --visits=1000 --temperature=1 --tempdecay-moves=10 player1: --sticky-checkmate player2: --no-sticky-checkmate
```

Videodr0me:
@Eddh The idea that later visits convey more recent and thus more relevant information has been tested in lc0. That was the idea behind the backpropagate-beta and gamma parameters. Unfortunately they yielded only gains upto 1000 visit searches, longer searchers yielded no gains and even moved the elo gain into negative territory. I suspect there is something to be gained from this idea, and have some theories as to what additional factors might be at play, but did not come around to explore it more deeply. Some data is here: [https://github.com/Videodr0me/leela-chess-experimental](url).

Of course this could change with the refined minimax version of this commit so it might be an idea to retest this on the minimax component, q or even on both q and minimax adaptively. So much to test, so little time.

DanielUranga:
@Videodr0me the changes were tested with thousands of games at low nodes per move. (See the spreadsheet Adrian posted). We tested 200, 400 (at various cpuct), at 1000 and 2000 nodes per move and minimax always performed better than default outside of bayeselo error bounds.
  At higher node count this seems to converge to regular MCTS, we only see improvement when we tune `cpuct`, so it might just that we found a better `cpuct` for the nets we tested.
  So the question is how to follow this, options are:
1. Don't merge, keep investigating different approaches.
2. Merge as a disabled by default uci-option, possibly something like "alternative-q-backpropagation" (true/false). This way more people will have easy access to it, might get interested, fiddle with the config and report results back.

Maybe for a future super-non-A0-like test net we could try training with the option enabled, since at 800 nodes it's record against regular is really good: `Score of lc0-minmax.exe vs lc0-main.exe: 201 - 123 - 324  [0.560] 648`, but this is just a guess for the distant future.

DanielUranga:
Or I can make it uci-optional and we will see, right?

dubslow:
I'm fine with merging this in principle, but it will mark a significant divergence from the DeepMind algorithm. Such should be mentioned and disseminated.

In practice, it either needs to be default off in UCI, or else we need to get the training server to implement maximum version enforcement as well.

DanielUranga:
I like those ideas, @AdrianAcala has setup a really good testing framework using Jenkins that we might use to test ideas, maybe some contributors could donate their resources to be used as Jenkins slaves so we can test changes that are independant from the NN faster.
About this PR will make the uci option changes tonight/tomorrow and commit here.

DanielUranga:
In case someone with powerful GPU wants to test:
```
lc0 selfplay --verbose-move-stats --parallelism=8 --no-share-trees --backend=multiplexing "--backend-opts=cudnn(threads=2)" --games=1000 --visits=1000 --temperature=1 --tempdecay-moves=10 player1: --alternative-q-update-algorithm player2: --no-alternative-q-update-algorithm
```

Videodr0me:
Result at standard cpuct: 
```
tournamentstatus final win 99 53 lose 88 122 draw 313 325
```
This translates to W 152/L 210/D 638 and to -20 elo against standard lc0. This was expected, so i am rerunning with cpuct = 1.775 for minimax vs. cpuct 1.6 for standard lc0.

P.S.: Soon the tournemant status will look pretty and already include all elo calculations

DanielUranga:
How is that possible? This was tested and winning after several thousand games at several cpuct, it even has the record scores on potato_chess twitch channel. Which node count was used for the test? Do we know that the option is being used? Because using cutechess-cli, I was no able to see gains enabling it through command line ("args" cutechess parameter), but only when using the uci option ("initstr" cutechess parameter).

DanielUranga:
Yes, I constructed that command, but didn't test with that but with cutechess-cli using 2 moves random openings. Was easily able to tell if this was on/off due to winrate, this version w/l/d at 800 nodes is about 2X/X/2X vs normal.

Videodr0me:
This:
```
.\lc0-minimaxq selfplay --verbose-move-stats --parallelism=8 --no-share-trees --backend=multiplexing "--backend-opts=cudnn(threads=2)" --games=1000 --visits=1000 --temperature=1 --tempdecaymoves=10 player1: --alternative-q-update-algorithm --cpuct=1.775 player2: --no-alternative-q-update-algorithm --cpuct=1.6``` 
yieldeld this:
```
tourmentstatus final win 98 81 lose 112 131 draw 290 288
```
This translates to W 179/L 243/D 578 and -22 Elo vs. lc0 standard with cpuct 1.6. Maybe something went wrong with the option, so that its not properly disabled? Also, i am using an older net. I can spare another couple of hours on that machine, so if you have a preferred net and/or settings I could try another run.

**Comment by DanielUranga** [comment]
```
./cutechess-cli \
-engine name=ID{ID1} cmd=/content/lc0/build/lc0 arg="--weights=/content/networks/{network_hash_1}" \
nodes=800 \
stderr=leela1.err \
-engine name=ID{ID1}-minimax cmd=/content/lc0/build/lc0 arg="--weights=/content/networks/{network_hash_1}" initstr="setoption name Use alternative experimental Q update algorithm value true\n" \
nodes=800 \
stderr=leela2.err \
-each tc=inf proto=uci -recover -games 2 -rounds {rounds} -repeat -pgnout output_pgn.pgn -openings file=../2moves_LT_1000.pgn order=random
```
That is what I used on Google Collab, based on a Collab provided by scs-ben.

On a side note: we should improve the selfplay mode so it prints which values are being used for the different options, also to refuse to start if an unrecognized option is passed, and a new option to only print games results.

**Comment by DanielUranga** [comment]
This command:
```
!rm -f cutechess-cli/output_pgn.pgn && rm -f cutechess-cli/leela*.err # remove old file if there is one
!cd cutechess-cli && QT_LOGGING_RULES="*.debug=true" ./cutechess-cli \
-engine name=ID{ID1}-minimax cmd=/content/lc0/build/lc0 arg="--weights=/content/networks/{network_hash_1}" \
arg="--alternative-q-update-algorithm" \
nodes=200 \
stderr=leela1.err \
-engine name=ID{ID1} cmd=/content/lc0/build/lc0 arg="--weights=/content/networks/{network_hash_1}" \
nodes=200 \
stderr=leela2.err \
-each tc=inf proto=uci -recover -games 2 -rounds {rounds} -repeat -pgnout output_pgn.pgn -openings file=../gaviota-starters.pgn order=random
```
Yielded this result:
```
Score of ID10745-minimax vs ID10745: 177 - 141 - 282  [0.530] 600
Elo difference: 20.87 +/- 20.24
Finished match
```
Using net 10745.

**Comment by DanielUranga** [comment]
@Videodr0me Maybe temperature is far worse for minimax Q so it was getting bad positions from the opening?

**Comment by DanielUranga** [comment]
Just finished another 400 games match at 400 nodes per move; command:
```
!rm -f cutechess-cli/output_pgn.pgn && rm -f cutechess-cli/leela*.err # remove old file if there is one
!cd cutechess-cli && QT_LOGGING_RULES="*.debug=true" ./cutechess-cli \
-engine name=ID{ID1}-minmax cmd=/content/lc0-minmax/build/lc0 arg="--weights=/content/networks/{network_hash_1}" \
arg="--alternative-q-update-algorithm" \
nodes={playouts_per_move_1} \
stderr=leela2.err \
-engine name=ID{ID1} cmd=/content/lc0-minmax/build/lc0 arg="--weights=/content/networks/{network_hash_1}" \
nodes={playouts_per_move_1} \
stderr=leela1.err \
-each tc=inf proto=uci -recover -games 2 -rounds {rounds} -resign movecount=4 score=650 -draw movenumber=40 movecount=5 score=9 -repeat -pgnout output_pgn.pgn -openings file=../2moves_LT_1000.pgn order=random
```
result:
```
Score of ID10745-minmax vs ID10745: 125 - 87 - 188  [0.547] 400
Elo difference: 33.11 +/- 24.81
Finished match
```
Going to try to understand why `./lc0 selfplay` is producing different results.

Videodr0me:
Please try with 1000 visits, better 2000. I got all sort of results with <= 800 visits. For example the moving average backprop worked great below 800 visits, but turned out not helping at all or even turning negative at higher visit searches.

killerducky:
Getting some timeouts with this PR:
http://ec2-34-217-73-2.us-west-2.compute.amazonaws.com:8000/viewTest/4/

mooskagh:
Is that ready for review?
How promising is that Elo-wise?
That will be non-trivial to marry with tree detaching, so I think it would make sense that they are useful separately and then merge.

killerducky:
Test running here. http://ec2-34-217-73-2.us-west-2.compute.amazonaws.com:8000/viewTest/12/
There are still a few timeouts, I'd like to rerun later with the new watchdog thread, I think that will fix it.

But this is a pretty major change so I guess we will need more tests. Also there are some other competing ideas out there.

mooskagh:
Does this still look promising?

DanielUranga:
As promising as this: https://docs.google.com/spreadsheets/d/1CL6iEnxzTu1lgLAoHDnlu_gRj4fswwqtOcIAR_z2khg/edit?usp=sharing

The most interesting thing is that this shows that maybe is possible to come with something better than the averaging that is used for the backpropagation step.

Long term imho the ideal would be to be able to decouple the search code so it can be switched between different versions: maybe try to make a lockless one from scratch, the "VOI" Oscard Smith proposes, maybe even some sort of A/B or B*, etc. And be able to choose from them as we choose backends.

DanielUranga:
Daniel Shawul does something similar in Scorpio: http://talkchess.com/forum3/viewtopic.php?f=7&t=70069&start=10#p791950

mooskagh:
Does this still look promising in 2020? :)

DanielUranga:
Will just close this PR.

PR #244, opened 2018-08-10
Appveyor update cudnn to 7.2.1

borg323:
The new version of cudnn was released a few days ago. This updates the appveyor script to build against it.


ankan-ban:
none of the features/bug fixes seem to affect any of our cuda/cudnn code, but it's always better in general to use latest version of the library in case they optimized some kernels/algorithms. So the change looks good!

dubslow:
I was gonna say the opposite actually, better to give it more than a few days to ensure it's bug free. The risk of requiring a long bug discovery process quite outweighs the "need 2% extra performance right now". Performance is secondary to stability.

cn4750:
Honestly, until these kinds of things are in the hands of the masses, we don't have the testing resources to detect these kinds of things for such a safe-seeming change.

borg323:
Anyway, this will only be user visible after a release, where the packaged zip will contain the new dlls. Which looks like is still a few days away.

PR #245, opened 2018-08-10
add minimal temperature and tempdecay offset

kostuek:
Example:
temperature = 5
tempmin = 2
tempdecaymoves =10
tempdecayoffset = 4

first 4 moves will be at temp 5, moves 5 -> 10 linear decay to temp 2, temp 2 from move 10 to end

### DIFF
--- a/src/mcts/params.cc
+++ b/src/mcts/params.cc
@@ -61,9 +61,16 @@ const OptionId SearchParams::kTemperatureId{
     "while making the move."};
 const OptionId SearchParams::kTempDecayMovesId{
     "tempdecay-moves", "TempDecayMoves",
-    "Reduce temperature for every move from the game start to this number of "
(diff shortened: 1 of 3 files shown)


mooskagh:
What is the reason of this change? Having 4 parameters related to temperature is kind of excessive..
(but I'm fine with removing old way and replacing with better one)

kostuek:
when played around with training an own net, I found the "old way" a little bit unflexible, for example I wanted to try a high constant temp for the first 5 or 10 moves when generating games, then decay it down but not to zero but say 1 or 0.5 and keep it there until the end, to still keep some variety in the endgame. but I agree to the amount of options being too high, just didn't really know, what to do about it. maybe this is more usefull for some "experimental" branch, if there were one?

Tilps:
This PR is very stale now, but I think I'd like something like it for v20 to test out more options for training.

kostuek:
I'll try to change it.

Issue #14, opened 2018-06-02
Make parameter parsers validate input


mooskagh:
_From @mooskagh on April 29, 2018 20:6_

(pretty straightforward, so if someone wants an easy task to start contributing, that's a good candidate)

Currently while command line / uci parameters have restrictions (min/max for numeric, list of choices for choice params), they are not checked. Validate them.

_Copied from original issue: glinscott/leela-chess#470_

Synthetica9:
This should be closed, right?

PR #246, opened 2018-08-11
Validate command line parameter constraints

cwbriscoe:
https://github.com/LeelaChessZero/lc0/issues/14

This PR validates that the values entered for parameters on the command line are valid based on their allowed values.  The following validations take place based on the Option type:
IntOption:  the value is between min and max or is equal to the min or max.
FloatOption:  the value is between min and max or is equal to the min or max.
ChoiceOption:  the value is equal to one of the choices.
BoolOption:  the value is either "true" or "false".
StringOption:  no validation.

If the validation fails, then lc0 will exit with a message describing why the value was invalid.

It seemed that the code intended to allow "on" and "off" for BoolOption.  I did not see any current options that use that.  I took those values out as valid.  I can add back if desired.

I also added a few test cases in new file optionsparser_test.cc.


cwbriscoe:
Ready for review.

src/utils/optionsparser.cc
```diff
@@ -328,8 +325,14 @@ IntOption::ValueType IntOption::GetVal(const OptionsDict& dict) const {
   return dict.Get<ValueType>(GetName());
 }
 
-void IntOption::SetVal(OptionsDict* dict, const ValueType& val) const {
+bool IntOption::SetVal(OptionsDict* dict, const ValueType& val) const {
+  if (val < min_ || val > max_) {
+    std::cerr << "Flag '--" << GetLongFlag() << "' must be between "
```
mooskagh: Instead of outputting it to stderr and reporting the error up (where it is used?), it's better to throw an Exception.
In this case it will also e.g. show error through uci protocol when bad parameters are attempted to be set through uci option command.

cwbriscoe: I am a c++ exception noob.  I throw an exception now.  Should I be catching it somewhere or just let it crash and burn?  It crashes and burns right now!

cwbriscoe:
Ready for review again.  I am not sure I did the exceptions correctly because I am not sure if you want me to catch them somewhere or just let it crash.  I haven't used exceptions much at all in the past...

src/utils/optionsparser.cc
```diff
@@ -50,12 +50,14 @@ std::vector<std::string> OptionsParser::ListOptionsUci() const {
   return result;
 }
 
-void OptionsParser::SetOption(const std::string& name, const std::string& value,
+bool OptionsParser::SetOption(const std::string& name, const std::string& value,
                               const std::string& context) {
   auto option = FindOptionByName(name);
   if (option) {
     option->SetValue(value, GetMutableOptions(context));
+    return true;
```
mooskagh: Maybe also just throw Exception rather than return `false`? :)

src/utils/optionsparser.cc
```diff
@@ -329,6 +338,12 @@ IntOption::ValueType IntOption::GetVal(const OptionsDict& dict) const {
 }
 
 void IntOption::SetVal(OptionsDict* dict, const ValueType& val) const {
+  if (val < min_ || val > max_) {
+    std::stringstream buf;
```
mooskagh: `ostringstream`?

src/utils/optionsparser.h
```diff
@@ -116,11 +116,14 @@ class OptionsParser {
   void AddContext(const std::string&);
 
  private:
+  // Print help to std::cerr
```
mooskagh: `Prints` (it's 3rd person, describing what function does).
Also, period in the end of the sentence. :)

PR #247, opened 2018-08-11
ispc  fix linking error

ra1u:
Adding `--pic` flag (position independent code) solves linking error with object code generated from ispc. This also allows building binary with multiple architectures, that was for this error disabled with meson flag  `ispc_native_only`

Error that should be fixed:
`../winograd_ispc.o: relocation R_X86_64_32S against .rodata.cst32 can not be used when making a PIE object; recompile with -fPIC`

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -199,7 +199,7 @@ if get_option('build_backends')
     endif
 
     ispc = find_program('ispc', required: false)
-
+    ispc_extra_args = []
     if has_blas and get_option('ispc') and ispc.found()
       ispc_native_only = get_option('ispc_native_only')
       if host_machine.system() == 'windows'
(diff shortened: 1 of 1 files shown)


meson.build
```diff
@@ -210,7 +210,8 @@ if get_option('build_backends')
 			    '@BASENAME@_avx2.obj' ]
 	endif
       else
-        outputnames = [ '@BASENAME@.o']
+	ispc_extra_args += ['--pic']
```
mooskagh: It seems that this part of the meson build file has weird indentation. Would it be possible for you to fix it so that it's clear what matches what?

borg323:
Does this work if we pass `-Ddefault_library=static` to meson?

meson.build
ra1u: There were tabs and spaces in mix for ispc part of meson code. Last patch for this PR a443c44 should fix this.

ra1u:
@borg323 I dont know it that it is relevant for object files. Meson flag `-Ddefault_library=static` does not seem to change list of required libraries in resulting  `lc0` binary when comparing default build trough output of `ldd`.

Issue #48, opened 2018-06-07
Tablebase support


dubslow:
Perhaps the biggest feature remaining that lczero has that lc0 doesn't. Syzygy in particular

jjoshua2:
So we need a strategy. To use the syzygy  code I need a lot of stockfish
code like its material hash in each position which inguess is a board
object here. The alternative is to use fathom dll which no one has
experience with I'm guessing. Although I think ethereal or laser author
just implemented it so we could look at diffs and maybe get some help there.

Tilps:
If I can find the time, I'm happy to port the syzygy code to use lc0 move gen rather than stockfish. It will take several hours of dev time though - which is not trivial to come by :P

oscardssmith:
One possible way to implement this without concerns about nonzero contamination is to make it so that when training, we propagate the win/draw/loss values through search the same way we do with checkmates so we don't train on games poisoned by stupid tactical mistakes, but don't use tables for policy so that the NN still learns how to play endgames. This would strengthen the training games, and prevent a large portion of incorrect training data, but should allow users the option of whether to play games with or without the table.

jjoshua2:
I think now with test30 this is working in server too, so not really much left to implement anywhere...

PR #250, opened 2018-08-12
Port syzygy to lc0 movegen.

Tilps:
This time starting from the cfish version - combined with code I wrote for the previous version with intent to have a final product which is compatible with GPL with exception.

There are a couple of places I'm sure this has performance issues - and in all its very ugly - but it compiles.  I even tested that it could successfully load the 6-piece tablebases for both wdl and wdl+dtz without crashing.

No probing has been tested at all yet.

Has an identical surface API to the previous attempt, so attempts at integrating it into search should be able to be reused.

### DIFF
--- a/src/syzygy/syzygy.cc
+++ b/src/syzygy/syzygy.cc
@@ -1327,6 +1327,10 @@ class SyzygyTablebaseImpl {
   std::vector<TbHashEntry> tbHash_;
 };
 
+SyzygyTablebase::SyzygyTablebase() : max_cardinality_(0) {}
+
+SyzygyTablebase::~SyzygyTablebase() = default;
+ 
 void SyzygyTablebase::init(const std::string& paths) {
   paths_ = paths;
   impl_.reset(new SyzygyTablebaseImpl(paths_));
--- a/src/syzygy/syzygy.cc
+++ b/src/syzygy/syzygy.cc
@@ -1,0 +1,1557 @@
+/*
+ * Original from cfish tbprobe.c
+ * Copyright (c) 2013-2018 Ronald de Man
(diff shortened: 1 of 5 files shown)


src/syzygy/syzygy.cc
```diff
@@ -0,0 +1,1656 @@
...
+
+  bool test_tb(const char* str, const char* suffix) {
+    return !name_for_tb(str, suffix).empty();
+  }
+
+  void* map_tb(const char* name, const char* suffix, map_t* mapping) {
+    std::string fname = name_for_tb(name, suffix);
+    void* base_address;
+#ifndef _WIN32
+    stat statbuf;
```
cwbriscoe: This line needs to be `struct stat statbuf; `to compile on the linux build.

Tilps: Yeah, overzealous cleanup of extraneous use of the word struct.  My bad, fixed.

Tilps:
This is a step towards #48 being addressed. (This comment added to link them.)

src/syzygy/syzygy.cc
```diff
@@ -0,0 +1,1667 @@
...
+
+    return true;
+  }
+
+  int probe_table(const ChessBoard& pos, int s, int* success, const int type) {
+    // Obtain the position's material-signature key
+    Key key = calc_key_from_position(pos);
+
+    // Test for KvK
+    if (type == WDL && (pos.ours() + pos.theirs()).count() == 2) return 0;
```
borg323: I think this would be better as 
```
if (type == WDL && pos.ours() == pos.our_king() && pos.theirs() == pos.their_king()) return 0;
```

dubslow:
Merged in e00707b

PR #251, opened 2018-08-12
Efficient and informative depth computation.

Videodr0me:
This computes depth more efficiently and replaces full_depth with the more informative average depth:

1. both max_depth and full_depth are removed from the node structure freeing space
2. the checking of all siblings for full_depth computation is no longer needed
3. full_depth rarely got larger than 5 even on fast machines and long searches
4. average depth replaces full_depth as a more informative measure
5. max_depth reflects search depth not tree depth
6. max_depth is consistent with length of pv displayed (if pv stems from the max depth line)
7. max_depth and cum_depth (for averaging) are computed in a two step process, first per batch and only after the mutexed search->max_depth and search->cum_depth are updated. This is in preparation of a mutex reorg.
8. For discussion: average depth is true average depth and thus can also decrease (informing about leela currently exploring a wider tree). This is rare, my own testing with chessbase gui and cutechess as well as Dubslows testing showed no problems with this. It could easily be modified to be monotonically increasing.

### DIFF
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -194,23 +194,6 @@ void Node::FinalizeScoreUpdate(float v) {
   --n_in_flight_;
 }
 
-void Node::UpdateMaxDepth(int depth) {
-  if (depth > max_depth_) max_depth_ = depth;
-}
-
-bool Node::UpdateFullDepth(uint16_t* depth) {
-  // TODO(crem) If this function won't be needed, consider also killing
-  //            ChildNodes/NodeRange/Nodes_Iterator.
-  if (full_depth_ > *depth) return false;
-  for (Node* child : ChildNodes()) {
-    if (*depth > child->full_depth_) *depth = child->full_depth_;
-  }
-  if (*depth >= full_depth_) {
-    full_depth_ = ++*depth;
-    return true;
-  }
-  return false;
-}
 
 Node::NodeRange Node::ChildNodes() const { return child_.get(); }
 
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -156,8 +156,6 @@ class Node {
 
   // Returns whether the node is known to be draw/lose/win.
   bool IsTerminal() const { return is_terminal_; }
-  uint16_t GetFullDepth() const { return full_depth_; }
-  uint16_t GetMaxDepth() const { return max_depth_; }
   uint16_t GetNumEdges() const { return edges_.size(); }
 
   // Makes the node terminal and sets it's score.
@@ -225,10 +223,6 @@ class Node {
   // Sum of policy priors which have had at least one playout.
   float visited_policy_ = 0.0f;
 
-  // Maximum depth any subnodes of this node were looked at.
-  uint16_t max_depth_ = 0;
-  // Complete depth all subnodes of this node were fully searched.
-  uint16_t full_depth_ = 0;
   // Does this node end game (with a winning of either sides or draw).
   bool is_terminal_ = false;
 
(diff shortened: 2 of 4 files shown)


dubslow:
I don't approve of this for [this](https://github.com/LeelaChessZero/lc0/issues/86#issuecomment-408179482) reason. On TCEC this would engender all sorts of horrifying misconceptions: on one move, when Lc0 makes a new tree with 1M nodes, gets an average depth of (say) 20 and max depth 45, which is well and good; then the next move is one of those 600ms smart pruned moves, will result in a average depth of possibly like 30 and a max depth of 30 as well. So with a 600ms move, average depth will increase by several ply and max depth will decrease by several ply, in a 600ms move. That will cause mass confusion as to how exactly the depth is calculated, and will lead to falsehoods galore being spread as truth because no one knows any better.

Videodr0me:
I am aware of your understanding of what depth should be, which is kind of "unique". SF and other engines have no reproducability of depth due to multi threadedness and using hash info. Selective depth of SF also fluctuates from move to move, especially on hash hits.  "Normal" engines depth output is not clearly defined either, in particular when considering the extensive pruning and extension regimes. Average depth is exactly what it says it is, and thus if only one line is searched, max depth and average depth are naturally identical. The main advantages of this proposal over your previous ones are:
- no full tree traversal
- no wasted space in nodes
- depth is real search depth and not tree depth
- a vote on discord preferred this solution
This is essentially a cosmetic issue having no effect on playing strength and therefore should not lead to needless overcomplication and wasted resources (time or space).

dubslow:
Please do not make references to votes on Discord, especially when I made one that showed that this option was *not* preferred. And the "search" depth is far less important than the move-analysis depth. I cannot speak to other engines, except to say that their analysis is wiped fresh and restarted for each subsequent move, which differs quite dramatically from our tree reuse.

Videodr0me:
Hash is not wiped between moves - some engines do, most don't. I do not share your view on "move-analysis" depth (whatever thats supposed to be) and I think not outputting anything at all would be less informative. However you interpret depth, average depth, max depth, tree depth or "move-analysis depth", calling any of these "lies" is unprofessional and besides the point. All of these are computed from actual search data and the way they are computed is completely transparent and can be verified by looking at the source. 

I played a bit of chess in the day and used engine analysis regularily for opening preparation and game analysis. In my view the only thing which might warrant a discussion is that average depth, both in this and your proposal, can decrease occasionally, which will be of a much bigger surprise to anybody than the cases you find objectionable.

dubslow:
You approve this and merge without so much as a by your leave? Really?

dubslow:
Hash is not anywhere near the same thing as preserving the tree -- the shape of the search may be quite different even as the hash is recycled between moves. This is *not* so for Leela, the old search is reused in exactly the same way as it was made, a sort of checkpoint from which to continue that old search, so the old depth information must be included as well.

These numbers are generated from search data, but as I explained in my comment, the data from one `Search` instantiation does *not* describe the entirety of the move analysis generated by Leela in the case of tree reuse, and presenting depth information from only the most recent `Search`, not from the `NodeTree`, is misleading to the point of being simply wrong, and so it's not a stretch to call it a lie. It is perhaps linguistically hyperbolized, but it is also accurate.

cn4750:
@Videodr0me, I am not sure it will be a large surprise to people if average depth decreases occasionally. Houdini 3 was one of the first popular engines to have decreasing depths in search due to its use of Accelerated Principal Variation Search, a.k.a. smart fail-high. I think most of the popular GUIs fixed any bugs related to decreasing depths as a result of this.

Videodr0me:
@cn4750 Thats good to hear. In my limited testing i also found no issues with chessbase gui and cutechess.

PR #253, opened 2018-08-13
History plane initialization corrects misevals

dje-dev:
In recent weeks/months, it has been noted that the recent testnet versions (10xxx) perform poorly compared to the mainnet when judged on the basis of test suites (such as STS) or when playing match games at shorter time controls. This finding has seemed puzzling, given the dominance of these nets by seemingly all other measures.  To give just one example, the mainnet 532 scores about 10,800 points in STS compared to only 8,670 points for testnet 10664 (both at 1000 visits per position).

It turns out this is related to the fact that tests positions are provided to Leela without history (as noted by ObiWanBenoni and possibly others on Discord several weeks ago) This causes nonsensical evaluations by Leela (increasingly so with later versions of the network which are more sensitized to history planes). 

Consider for example running LC0 as follows:
```
lc0.exe --verbose-move-stats --backend=cudnn-fp16 --minibatch-size=1 --max-prefetch=0 --weights=D:\weights\testserver.lczero.org\weights_10664.txt.gz 
```
and providing this fairly balanced position for analysis at low visit counts:
```
position fen r1bq1rk1/pp2bppp/2np1n2/3Np3/4P3/3B1N2/PPP2PPP/R1BQ1RK1 b - - 7 9
go nodes 0
```
This yield Q=-0.95166 (i.e. almost certainly a lost position). If instead we provided some made-up shuffling history:
```  
position fen r1bq1rk1/pp2bppp/2np1n2/3Np3/4P3/3B1N2/PPP2PPP/R1BQ1RK1 b - - 7 9 moves d8d7 d1d2 d7d8 d2d1
go nodes 0
```
then a more reasonable Q=-0.05435 is returned. Furthermore, the policy vector completely changes to something more reasonable.

The implications of the above include:
- test suite benchmark results are seriously flawed
- matchplay results at low to medium visits may yield dramatically inferior play by Leela, depending on the method by which the driving program sends the positions to the engine (if a single explicit FEN of the current position is provided rather than a sequence of moves starting from the startpos)
- use of Leela interactively for study of ad hoc positions is compromised

A simple workaround seems possible, and is provided in this PR for review and discussion. When a single FEN is provided to PositionHistory::Reset, we first fill history planes with positions which are obtained by applying a null move to the current position. (Interestingly, simply adding the current position multiple times fails to yield similar improvement).

The improvement is dramatic at low to medium visit counts. For example, the STS suite score improves: 
  - from 8,670 to 12,267 at  1,000 visits 
  - from 9,550 to 12,965 at 10,000 visits 

Hopefully many other Leela benchmark tests could benefit from rerun if this PR is incorporated.

### DIFF
--- a/src/chess/position.cc
+++ b/src/chess/position.cc
@@ -91,7 +91,17 @@ GameResult PositionHistory::ComputeGameResult() const {
 
 void PositionHistory::Reset(const ChessBoard& board, int no_capture_ply,
                             int game_ply) {
+  // No move history is available to initialize history planes.
+  // However presence of empty history planes yields nonsensical evaluations.
+  // As a workaround all history planes are filled with the current position.
   positions_.clear();
+  Position currentPos = Position(board, no_capture_ply, game_ply);
+  Move nullMove = Move(0);
+  const int kMoveHistory = 8;
+  for (int idx = 0; idx < kMoveHistory - 1; idx++) {
+    positions_.emplace_back(Position(currentPos, nullMove).GetBoard(),
+                            no_capture_ply, game_ply);
+  }
   positions_.emplace_back(board, no_capture_ply, game_ply);
 }
 


dubslow:
The only possible way I would approve of this, personally, is if it was behind a default-off option.

(As it currently stands, this would also change how leela plays the startpos and first few moves after start, potentially causing chaos in the training pipeline.)

dje-dev:
For sure, at a minimum issues of synchronization with existing nets has to be considered, and I’ll provide more data on that in the coming days. My hope and very preliminary data points suggest Leela may play better chess in early moves with this change. If so, default on arguably not dangerous. Intriguingly the first move preferred switches from c4 to e4, which is also the move it converges to after deep search.

I conjecture that Leela seeing uninitialized input data (perhaps sitting at a wildly incongruous point away from all other valid chess positions) in training of early moves might impede the training process.

dubslow:
That's an indictment of the history planes; this treats the symptom, not the cause, sweeping the problem under the rug rather than eliminating it. That's why I'm not really in favor of this fix

oscardssmith:
I also agree. The history pane really should not be needed in general as the only case history matters is for 3 fold repetition. If we just replaced it with a boolean if the position is a repeat, and an int 50 move rule, we would be able to correct for this, and also get a significant speed bonus as well.

dje-dev:
I don’t see history planes as a problem, but rather that almost all of Leela training data came from inputs where the history planes were filled with legal positions. When no history is available, substituting a repeated position seemed to me most logical, setting the inputs to a relevant point not lying discontinuously far away in the distribution.

oscardssmith:
The problem I see is that it seems kind of like we were feeding in weather forecast. Sure it shouldn't hurt anything, but if that became necessary to make good predictions, something clearly has gone wrong.

AlexGreason:
My opinion: this should definitely be an option, since it's necessary for sensible analysis from FENs or test suites, but I agree on leaving it off for training (either by having it default off in general, or default on and add a parameter disabling it to the training run parameters)

borg323:
If en passant capture is possible, the previous move can be reconstructed, and this may be important for evaluation of the current position. At a minimum this case should be handled. See also the previous effort in <https://github.com/glinscott/leela-chess/pull/633>.

mooskagh:
1. It's a nice option to have, but it should be be an option (default off, until proved to be useful) rather than just a change.
2. It's better do have this logic in EncodePositionForNN() function rather than in play history. Adding it into history is kind of hacky and can interact with other things (like threefold repetition detection).

borg323:
Also there are some test failures. Appveyor output:
```
[==========] Running 2 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 2 tests from PositionHistory
[ RUN      ] PositionHistory.ComputeLastMoveRepetitionsWithoutLegalEnPassant
unknown file: error: SEH exception with code 0xc0000005 thrown in the test body.
[  FAILED  ] PositionHistory.ComputeLastMoveRepetitionsWithoutLegalEnPassant (0 ms)
[ RUN      ] PositionHistory.ComputeLastMoveRepetitionsWithLegalEnPassant
unknown file: error: SEH exception with code 0xc0000005 thrown in the test body.
[  FAILED  ] PositionHistory.ComputeLastMoveRepetitionsWithLegalEnPassant (0 ms)
[----------] 2 tests from PositionHistory (1 ms total)
[----------] Global test environment tear-down
[==========] 2 tests from 1 test case ran. (1 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 2 tests, listed below:
[  FAILED  ] PositionHistory.ComputeLastMoveRepetitionsWithoutLegalEnPassant
[  FAILED  ] PositionHistory.ComputeLastMoveRepetitionsWithLegalEnPassant
 2 FAILED TESTS
```

roy7:
Forgive me if I'm clueless here, but this is only for FEN being input into lc0 right, so how would it affect training? Training games will start from the normal game start and play to the end with no "position fen" command being used.

We discussed this sort of approach a lot on Discord earlier today, and my only concern about this sort of easy fix (@mardak did it via adding "moves a1a1 a1a1" to positions) is that loading a FEN that is from the first few moves of a game will have a history adding where in reality the history boards are blank at game start. But people normally won't be loading a FEN for a move 2 or move 3 board state, I'd imagine. Unless a GUI using a book feeds the move to LC0 as a "position fen"?

I personally dislike history plains and would like to see 3-fold handled in a different way. A suggestion I made on discord for consideration is 3 plains of inputs (0,1) where it is set to 1 for any valid move space where a repetition results in first, second, or third repetition. That would allow the NN to learn about forcing or avoiding 3-fold ties when it is losing or winning. But loading in a FEN with no history is no longer a problem since it is what the training examples normally look like (most moves result in zero repetitions.)

roy7:
The problem with this being a flag, or the inputs not being changed to address history in some other way that is FEN-compatible, is the average user at home will connect Leela to their GUI and input a FEN to study and get wrong/bad results with no idea why.

Mardak:
As mentioned in https://github.com/LeelaChessZero/lc0/pull/253#issuecomment-415845789, one can avoid the "position fen …" issue by doing "position fen … moves a1a1 a1a1" instead, so if this is more of a "UI" issue than back-end encoding issue, I've put a front-end fix in #303.

dubslow:
@roy7 `position startpos` is an alias for fen, it amounts to the same thing as far as neural network inputs go. In fact we already have evidence that the startpos suffers some problems because it doesn't get any history input.

Mardak:
I did notice in testing #303 where there were different moves with injected history or not:
```
position startpos
info string g2g3  (374 ) N:       0 (+ 0) (P:  6.59%) (Q:  0.03243) (U: 0.22422) (Q+U:  0.25666) (V:  -.----) 
info string c2c4  (264 ) N:       0 (+ 0) (P:  7.74%) (Q:  0.03243) (U: 0.26303) (Q+U:  0.29546) (V:  -.----) 
info string e2e4  (322 ) N:       0 (+ 0) (P:  9.63%) (Q:  0.03243) (U: 0.32757) (Q+U:  0.36000) (V:  -.----) 
info string d2d4  (293 ) N:       0 (+ 0) (P: 12.29%) (Q:  0.03243) (U: 0.41774) (Q+U:  0.45017) (V:  -.----) 
info string g1f3  (159 ) N:       0 (+ 0) (P: 16.08%) (Q:  0.03243) (U: 0.54681) (Q+U:  0.57925) (V:  -.----) 

position startpos moves a1a1 a1a1
info string b2b3  (230 ) N:       0 (+ 0) (P:  5.37%) (Q: -0.62694) (U: 0.18272) (Q+U: -0.44422) (V:  -.----) 
info string c2c4  (264 ) N:       0 (+ 0) (P:  6.93%) (Q: -0.62694) (U: 0.23553) (Q+U: -0.39141) (V:  -.----) 
info string g1f3  (159 ) N:       0 (+ 0) (P:  7.35%) (Q: -0.62694) (U: 0.24975) (Q+U: -0.37719) (V:  -.----) 
info string d2d4  (293 ) N:       0 (+ 0) (P: 12.27%) (Q: -0.62694) (U: 0.41711) (Q+U: -0.20983) (V:  -.----) 
info string e2e4  (322 ) N:       0 (+ 0) (P: 17.31%) (Q: -0.62694) (U: 0.58853) (Q+U: -0.03842) (V:  -.----) 
```
Both value and policy are different although the value without dummy moves seems more reasonable.

dje-dev:
Thanks to all for the multiple comments and suggestions. Based on feedback, I will shortly open a separate PR with a reworked implementation which incorporates the suggestions to move this change into encoder.cc (crem) and also to handle the special case of en passant (borg)

PR #257, opened 2018-08-14
WIP: Track time saved by smart pruning and use it as soon as possible.

roy7:
~~When testing this use --slowmover 1. I didn't mess with the default or remove it, but that should be obsolete using the pruned "bonus" time in this way instead.~~

slowmover default changed to 1 in the PR itself.


dubslow:
I'm pretty sure this is pointless, since the time saved is left in the normal clock, and the time per move is calculated from the normal clock, i.e. already includes the previously-saved time.

roy7:
Yes currently the saved time is added to clock and spread over the rest of the game. But that means if we get more tree reuse than expected, we end game with lots of time left over and that was making people upset. Also it isn't possible to predict from game to game how much tree reuse will happen leading to smart pruning. This way we don't need to know and you can still aim for an optimal time curve without trying to guess using the slowmover variable.

roy7:
This seg faults in testing because search_ isn't set until after I'm trying to access the variable I added. Going to relocate it to _tree or inside the engine controller.

roy7:
Got it working now. Moved the variable out to be a global for easy testing. If it works out well it can be cleaned up later and made a non-global.

roy7:
Running a 100 game series to test it, 45s+0.5 timing.

> ./cutechess-cli -engine conf=leela -engine conf=leela-test arg="--slowmover=1.0" -each arg="--threads=2" arg="--weights=/home/roy/chess/id10520" tc=45+.5 timemargin=500 proto=uci -repeat -recover -games 2 -rounds 50 -pgnout test1.pgn  -openings file=Noomen_2-move_Testsuite.pgn order=random

```
[{
    "name": "leela-test",
    "command": "lc0",
    "protocol": "uci",
    "workingDirectory": "/home/roy/chess/lc0/build/release"
  },
  {
    "name": "leela",
    "command": "lc0",
    "protocol": "uci",
    "workingDirectory": "/home/roy/chess/lc0-original/build/release"
}]
```

Using these settings based on feedback from Discord. Need to get to bed shortly but I'll report initial results leela vs leela-test: 10 - 15 - 46  [0.465]. A lot more testing needed of course, and longer time controls.

roy7:
j33 from Discord noted that the bonus time is applied directly to next move but is also accounted for in the total time remaining for the rest of the match. An experiment for another evening would be to subtract the bonus time from the total time remaining when calculating the time curve targets for each move.

roy7:
francis from Discord shared a couple graphs he created from the first 80 games.

Time per move:
https://cdn.discordapp.com/attachments/425419999096733706/479145318311526459/avg_mt_lc0_clock_time.png

Clock remaining per move:
https://cdn.discordapp.com/attachments/425419999096733706/479146004810039297/avg_lc0_clock_time.png

First 80 matches PGN file: http://termbin.com/iklt

fli:
We try to emulate this statically with _Scale thinking time_>1 so this is a good shot at making that dynamic and removing _Scale thinking time_ as a parameter. Probably will eventually need to put some condition to reset bonus time if a new game is started. (I'm francis from discord)

roy7:
Did some basic stderr logging to be sure things were working as I expected, and that showed me bonus time was being stored many times per move (as we moved out of the tree it stored once per level). So I moved the storage logic to Search::MaybeTriggerStop and it only fires if smart pruning kicked in.

roy7:
TODO: Need to reset bonus time to -1 someplace in case game is reset or a new game is started with existing client.

roy7:
My 300+5 test is impressively slow. About 9 hours in and I'm only on game 24.

PGN for first 23 games: http://termbin.com/ljxwz

Test version is losing 2 - 1 - 20  [0.522] so far, sample size too small to matter atm.

roy7:
It has been suggested I try slowmover 1.95 which was the default before the agressive smart pruning was added. Before smart pruning it apparently just did a fixed amount per move? I thought it was more fancy than that, copying SF or some sort of known 'optimal' curve from chess programming research.

The current graph is clearly much too flat but that's a result of slowmover=1, not my real change. I was trying to avoid the "scale based of smart pruning" purpose of slowmover...

fli:
I ran some tests last night playing this patch and prepatch against a gauntlet of handicapped SF/Komodo/Ethereal with TCEC adjudication rules. TC was 9.9+.155 w/ .1s move overhead to try and emulate TCEC conditions at a lower TC.
Scale thinking time = 1.00, aggressive smart pruning = 0.0, + this patch scored **-124 +/- 24.**
Scale thinking time = 1.95, aggressive smart pruning = 1.0 scored **-109 +/- 13**.
avg_lc0_clock_time
avg_mt_lc0_clock_time

So this patch with current settings is roughly 15 ELO worse, but it is interesting that their scores are so close with such different time curves. Maybe retuning the curves will result in further ELO increase since I think @jjoshua2 tuned them with scale thinking time on.

roy7:
Stopping 300+5 test after 55 games because it's too slow.

Score of leela vs leela-test: 4 - 4 - 47  [0.500] 55
Elo difference: 0.00 +/- 35.09

Next I want to try same slowmover for both, and try removing the bonus time from the global time left.

roy7:
Ok with slowmover=1 I used @fli's spreadsheet and came up with a new curve to emulate LC0's default behavior. I suspect the first few moves could move faster since she should be very confident about her openings, at leas the first few moves, right? But for now I'm trying to stack time on move 1 to match LC0's own behavior. Curve settings could be re-tuned later.

Also, bonus time is no longer counted in total_moves_time to make that more correct.

100 match series is under way:

Score of leela vs leela-test: 1 - 2 - 8  [0.455] 11

First 11 games: http://termbin.com/zinu

Some graphs on the first 7 games so you can see my new time curves are much closer to LC0 default behavior.

clock time / move

move time / move

roy7:
100 games using a time curve similar to LC0: http://termbin.com/189w

Score of leela vs leela-test: 11 - 24 - 65 [0.435] 100
Elo difference: -45.42 +- 40.11

roy7:
Side note, I believe the way the time curve calculation works now in default LC0, if a lot of time is saved by smart pruning then the extra time won't really get used up properly since the time curve assumes all prior time has been using the same curve since game start. I know slowmover 2.66 was an estimated way to try to shift time forward based on the 'average' smart prune time savings, but it might cause us to use time too fast if we don't prune much or not fast enough if we prune a ton.

I'm not sure on the best way to solve that in default LC0 as-is, but if my approach here did end up eventually adopted that problem goes away. We'd be using a specific time curve for all normal game time that should mostly look right, with saved time used as soon as possible above that curve. It shouldn't shift the curve around in weird ways or we end up at game end move 60 with 15 minutes on the clock somehow.

The curve needs tuned and experimented with probably, I feel like maybe we can move faster on the first few moves of a game if we don't use an external opening book, but right now I'm emulating LC0's default of max time on move 1 and then down from there for rest of game (the slowmover 2.66 sort of strongly overrides the raw time curve shape).

roy7:
As mentioned by jjosh on Discord:

> we do nps = nodes(time- 200ms)
> so at 200ms nps is infinite assuming we have gotten a node yet. that's why on tcec we've never seen a pruning before 600ms because it takes a bit for it to come down to the real nps

With my testing at +.5 time controls, there isn't time for smart pruning to really do anything in the end game since once we run out of time, we'll spend about 500ms on every move. It might be interesting to re-test after all of this with +1 time controls, which would allow end-game smart pruning...

jkormu:
Ran tournament last night with lc0+pr257 net 10729 vs lc0 net 10729 vs sf dev from last winter.
Time control 30s +0.5s, 2 move opening book.

This was pr257 with subtract bonus time from total time -correction. Slow mover set as 1.0 for new tm leela. Other settings as default.  

`
   1 sf9+                                128      36     184   67.7%   48.4%
   2 lc0_new_tm                    -32       34     184   45.4%   54.9%
   3 lc0_old_tm                      -93      32      184   37.0%   57.6%
Finished match
`

Will build latest master and rerun.

jkormu:
I used default curve settings. Started new tournament with 45+1.0 time control.

roy7:
My extra games so far:

Score of leela vs leela-test: 24 - 22 - 118  [0.506] 164

So about a wash so far, leaning towards default.

@jkormu Please note that's fine but the with slowmover=1 you will have steady time over a long period of time, instead of lots of time early going down (like the original graphs at start of this thread). If you change the curve with the commands I provided so it looks more similar to lc0's current time usage, you'll get the later graphs I posted. I believe using time earlier is stronger than spreading it evenly since by turn 30 the game might be decided. But removing --slowmover needs big changes to the time curve itself to still be similar.

Thanks for your extra test runs.

roy7:
Side note/TODO, I don't believe current time management code in LC0 accounts for opening book. So with some book moves you'll be further down the intended time curve than otherwise. I'm not sure if that's something to look at or doesn't matter.

oscardssmith:
Since we've figured out that the problems in TCEC were caused by thermal throttle, is this change still needed?

jjoshua2:
This is from dkappe's 12 man endings. A lot of the endings are very drawish or immediately resigning so take elo with a grain of solt, but its definitely more on the gaining than the losing side as expected if syzygy is not inverted and getting any hits...
```Score of lc0tb vs lc0: 433 - 399 - 584 [0.512]
Elo difference: 8.34 +/- 13.86```

**Comment by jkormu** [comment]
@roy7 Wanted to test without any curve changes as based on the first test it seemed like it is Elo win without any parameter changes other than removing slow mover. For me it is a very convincing if there are straight Elo gains without tuning vs tuned values. We can only go better with tuning the curve for the 
 new approach.

Here are the results from my overnight tournament with 45.0 + 1.0.   
```
   1 sf9+                           62      50      90   58.9%   51.1%
   2 lc0_pr257                     -12      45      88   48.3%   62.5%
   3 lc0                           -51      45      90   42.8%   61.1%
```

**Comment by roy7** [comment]
@jkormu Cool thank you!

If your tourney is all wrapped up do you want to run one with the time curve commands for lc0_pr257 and see what happens? Or you could even add it separately as a lc0_pr257_curve?

I'm hopeful the +1 time is long enough there is some smart pruning benefits happening in late game. But really the biggest win should be early game by getting that extra time when we need it, as soon as possible in the game.

**Comment by roy7** [comment]
Running another small 30+1 test of my own last night:

```
Rank Name         Elo    +    - games score oppo. draws
   1 stockfish     74   31   30   148   68%   -37   47%
   2 leela-test   -29   29   29   150   43%    14   57%
   3 leela        -44   29   29   150   39%    21   55%
```

**Comment by killerducky** [comment]
@roy7 did you know MTGOStark has been testing v0.16 against the version in TCEC with new TM settings, and early results show v0.16 is ~40 Elo better? This means you will probably need to beat v0.16, not the version you're testing against.

**Comment by roy7** [comment]
I've been using /master. Is something newer than /master available?

**Comment by killerducky** [comment]
We spoke on Discord, but for the record:
* /master is the latest
* v0.16 did not have Aggressive smart prune (ASP), so it's like ASP=1
* A later release added it with ASP=0.68
* MTGOStark's tests show ASP=0.68 is 40 Elo worse than ASP=1.

Also note https://github.com/LeelaChessZero/lc0/pull/260 (kCacheHistoryLength change from 1 to 7) appears to have lost Elo also. More tests TBD.

**Comment by roy7** [comment]
In general I wouldn't be a fan of ASP, I think if it can change let it change and keep thinking earlier in the game. But I was trying to test my change without changing anything else, although removing slowmover of course requires time curve changes to be anything similar to LC0's current default time settings.

Perhaps I should run a fresh test with ASP set to 1 for both copies of Leela?

**Comment by roy7** [comment]
Updated results posted above. Starting a fresh test with ASP set to 1 for both.

**Comment by jkormu** [comment]
I might be able to run some test on Sunday.

**Comment by roy7** [comment]
With AGP set to 1 for both bots, they seem to play about identically so far.

```
Rank Name         Elo    +    - games score oppo. draws
   1 stockfish     82   38   36   101   70%   -41   46%
   2 leela        -40   35   35   102   40%    19   55%
   3 leela-test   -42   35   35   103   40%    20   54%
```

**Comment by dubslow** [comment]
I'm not surprised, this is really no different from just upping the slowmover slightly.

**Comment by roy7** [comment]
I set slowmover to 1 and adjusted curve defaults to look similar to the prior lc0 that had slowmover 2.66. Obviously the curve might be better in some totally different shape when doing smart pruning and immediate use of time. But at least this way the defaults are what I'm sort of intending/testing with no command lines needed by testers.

**Comment by roy7** [comment]
Since the time allocated to a move can be higher than the time left in the game, I'm wondering if when I calculate how much bonus to store I should use min() of the allocated time and the hard cap time (real clock left). So we don't count or roll forward bonus time that could never have been spent? Don't know. Maybe it's fine as it. The thing is the curve is based on the total time for the game, which is based on increment * movestogo. So the game's total time is way higher than just our time on clock, and the curve is based on total game time not clock time. That's  how sometimes curve target time for a move can be higher than our actual time remaining.

**Comment by roy7** [comment]
Updated to latest /master.

**Comment by jjoshua2** [comment]
merely testing that time pruning is still helpful. 68 default vs 1.0 (disabled aggressive)
```
Score of lc0 10810 roy vs lc0 10810 1.0 asp: 49 - 31 - 402 [0.519]
Elo difference: 12.98 +/- 12.59
```

**Comment by roy7** [comment]
To clarify you were testing aggressive smart pruning option. Normal time pruning is still helpful even if the aggressive test failed. :)

**Comment by roy7** [comment]
Since the time for a move is calculated off the curve shape and the estimated time left for rest of the game based on movestogo and the increment, the time limit for a move can be more than the actual clock itself. It's safe to ignore it I guess since we'll always move before actual clock expires, but I wonder of storage of the bonus should account for this somehow.

**Comment by roy7** [comment]
This will be even more valuable now that LC0 does pondering. If it could get some more testing/love? :)

**Comment by fli** [comment]
Because the first move doesn't have bonus time yet, what do you think about multiplying the time spent on the first move (I guess the move where `bonus_time = -1`) by some factor? I think this adjustment will kind of prime the bonus time to its steady state value and the first few moves won't, on average, have less time spent on them.

I think the best factor to multiply by is the average % pruned by move aversion, which would have to be guessed. E.g if, on average, moves are pruned to 70% of their allocated time, then the factor to multiply the first move by is `1/.7`. This value would obviously depend on the `move aversion` parameter

**Comment by jjoshua2** [comment]
Theoretically first move isn't long enough and should depend on futile
aversion like you say. Probably doenst matter much. Something like double
time is probably about right scales by futile aversion?

**Comment by fli** [comment]
Yeah, that seems right to me

**Comment by roy7** [comment]
Do you think it would be common for the smart pruning to kick in frequently on the very first move the game? Are we assuming there was no book and Leela has such a sharp preference on the opening move, or that we have an opening book the continuation move is so obvious other moves won't be searched?

**Comment by jjoshua2** [comment]
It's no more or less likely than any other move. So it won't use the time
it is allocated. Try it yourself. Go move time 120s uses like 60s

**Comment by roy7** [comment]
As it should if most of the search is all going into a single move.

When i was trying to see if there's some sort of optimal time curve to base things off of, I did find a plot of time used by human Grandmasters. They do course use the majority of their time early in the game but not on the very first few moves the game. I imagine that's the human version of an opening book, that for any major opening you recognize you know the preferred lines you like to play.

**Comment by jjoshua2** [comment]
Its OK to not use the whole time curve on the first move if starting from
start position, but if it's on playchess when you have a 18 move opening
book, then definitely not, because you get dropped out on a key move most
of the time. With a tournament like TCEC conditions are probably in the
middle...

**Comment by fli** [comment]
The adjustment should apply to the first move out of book though (which might be pretty late). I think not having this first move adjustment might've been why in our early tests the time curve looked so different.

**Comment by jkormu** [comment]
Match results from last night with TC 45+1 and no ponder:

```  
   1 sf9                            13      42     108   51.9%   59.3%
   2 lc0_pr257                      10      37     108   51.4%   67.6%
   3 lc0                           -23      35     108   46.8%   71.3%
Finished match
```

**Comment by jjoshua2** [comment]
This pr also has advantage in reducing noise in all other tests. Since time
use is more consistent.

**Comment by jkormu** [comment]
I am trying to test how this works together with ponder but no luck so far.
First night I messed up settings and ponder was off. Last night on 45+1 tc Leelas (release and pr) ran out of time in every other game. Investigating the reason now. Top candidates:
1. SF ponder steals all the cpu resources. I might have falsely assumed that only 1 thread is used by default.
2. 1 sec increment is not enough for Leela with ponder on.

Anyone aware of issue with Leela timing out when ponder on and low increments?

**Comment by jkormu** [comment]
Release Leela seems to be unsuitable for pondering tests as she keeps losing on time with ponder on even with 3 second increments. I will switch to master as #309 might help.

**Comment by roy7** [comment]
That's very weird/concerning. Hopefully #309 fixes it.

**Comment by roy7** [comment]
@fli was kind enough to give me some new graphs. I ran 10 games of 300+5 with pondering vs stockfish. The PGNs: http://termbin.com/3ld3 .

Pondering caused no damage to my time curve on average, we used all time by move 50 as intended:

avg_lc0_clock_time

Here's the same information (move time should say clock time) for all 10 games as individual lines. The jagged edges moving right are instant moves (clock didn't go down), following by bigger drops (bonus time being used up):

avg_mt_lc0_clock_time all 10

Here's the example amount of time used per move for one of the games. Big spikes are use of time saved from prior moves. With no smart pruning it'd just be a simple smooth curve of time used per move as game progresses.

avg_mt_lc0_clock_time

**Comment by roy7** [comment]
This should be compatible with https://github.com/LeelaChessZero/lc0/pull/271 to replace the current time curve code with his more advanced version. If anyone tries using both at once please report back. It'd be cool to see if the clock graph ends up matching his targets on average.

**Comment by jkormu** [comment]
Finally some results with ponder. tc 45+1.0 and 500ms move overhead. Will gather more games to bring the error bars down a bit.


```
Score of lc0_ponder_pr vs sf9_ponder: 12 - 18 - 40  [0.457] 70
Elo difference: -29.85 +/- 53.57
Finished match

Score of lc0_ponder vs sf9_ponder: 18 - 16 - 36  [0.514] 70
Elo difference: 9.93 +/- 57.15
Finished match
```

remaining time

time used

**Comment by roy7** [comment]
It's weird to me the time graphs are so similar with or without the PR, and apparently for this test the PR makes things a lot weaker in ELO?

**Comment by jkormu** [comment]
Yes, I agree it is strangely similar. I have double checked I am using right Leelas and ponder is indeed on.

Hard to draw conclusions about the ELO. The test gives only some indication it might be worse with ponder.

**Comment by jjoshua2** [comment]
Using more time is not obviously better when ponder is on, like it is
without ponder. Moving really fast is actually good if it doesn't hurt your
move quality because you spend most of your thinking time in their turn.
That's why these elo tests are important.

**Comment by roy7** [comment]
Perhaps for the time curves to diverge more requires more time on the clock, I don't know. If you look at the time usage graphs in the chess.com tournament, you'll see the effect of pondering is making the curve bulge out a lot to the right. Reflecting all those early fast moves where the time is spread to the rest of the game instead of used quickly. It also means we have plenty of time left around move 50, where we generally aim for no time left at move 50.

**Comment by roy7** [comment]
Perhaps for the time curves to diverge more requires more time on the clock, I don't know. If you look at the time usage graphs in the chess.com tournament, you'll see the effect of pondering is making the curve bulge out a lot to the right. Reflecting all those early fast moves where the time is spread to the rest of the game instead of used quickly. It also means we have plenty of time left around move 50, where we generally aim for no time left at move 50.

**Comment by jkormu** [comment]
I don't understand the remaining time graph I am getting. @roy7 your 300+5 graph looks more sensible. I would expect this pr to follow more closely the no-ponder time usage as shown in your test. I can't find mistake in my test. I will next check the time usage on bit longer tc and eventually with your 300+5. Did SF have ponder on in your test? 

The results don't  look good for this pr when ponder on with 45+1 tc.
```
Score of lc0_pr_ponder vs sf9_ponder: 10 - 23 - 37  [0.407] 70
Elo difference: -65.28 +/- 56.18

Score of lc0_ponder vs sf9_ponder: 10 - 16 - 44  [0.457] 70
Elo difference: -29.85 +/- 49.75
```

**Comment by roy7** [comment]
I did use ponder for both. No reason sf needs pondering. I limited it to 4 threads so it wouldn't mess anything up (my Ryzen 1700 has 16 threads available). My test was ran with:

`[
  {
    "name": "leela-test",
    "command": "lc0",
    "protocol": "uci",
    "workingDirectory": "/home/roy/chess/lc0/build/release"
  },
  {
    "name": "leela",
    "command": "lc0",
    "protocol": "uci",
    "workingDirectory": "/home/roy/chess/lc0-original/build/release"
  },
  {
    "name": "stockfish",
    "command": "stockfish-9-64",
    "protocol": "uci",
    "workingDirectory": "/home/roy/chess/stockfish-9-linux/Linux",
    "options": [
      {
        "alias" : "",
        "default" : 1,
        "max" : 512,
        "min" : 1,
        "name" : "Threads",
        "type" : "spin",
        "value" : 4
      }
    ]
  }
]
`

And

`./cutechess-cli -engine conf=stockfish -engine conf=leela-test arg="--ponder" arg="--threads=2" arg="--weights=/home/roy/chess/11252.w" -each tc=300+2 ponder timemargin=500 proto=uci -repeat -recover -games 2 -rounds 5 -pgnout 500+2ponder_francis.pgn -openings file=Noomen_2-move_Testsuite.pgn order=random -resign movecount=4 score=650 -draw movenumber=40 movecount=5 score=9`

**Comment by roy7** [comment]
@jkormu If you plot just one game of data does your time remaining graph look like mine, where there are jumps to the right any time pondering gives us an instant move?

**Comment by roy7** [comment]
If you use pondering without my PR, the total time left curve should be bloating towards the right like seen after CCCC games:

image

My time curve targets out of time at move 50 and it hit that in all of my tests and in the average graph pretty well. But the CCC example above without PR had time up through move 75 or so. Leela was black.

**Comment by roy7** [comment]
For reference, here is time used by human GMs. They have a memorized 'opening book' of sorts, so time kicks up after they 'leave the book'.

image

**Comment by jkormu** [comment]
I see same curvature with your pr and vanilla lc0, no bloating in 45+1.

1 game sample of your pr:
remaining time

1 game sample of vanilla lc0:
remaining time

**Comment by roy7** [comment]
It doesn't appear your vanilla LC0 is using pondering then. There's never a horizontal step in the time graph. Ie: Pondering never once got enough visits on the right expected move for Leela to move instantly (or at least much faster than normal).

Like at move 11 when SF uses 10 seconds, LC0's graph barely has a blip?

**Comment by jkormu** [comment]
You are right. Sample from wrong pgn file. Generated new sample for vanilla lc0 and verified 100% gpu utilization all time, so ponder is definitely on now. (Ignore 10 last moves as my script does not handle increments correctly when no game reached 100 moves.) 
remaining time

**Comment by roy7** [comment]
Ok that looks more like we'd expect. When SF has a big thinking move, we slide sideways from pondering savings. Without PR, the time isn't used asap, so we have no huge drops of our own to offset that. We run out of time around move 70.

With PR we should see some large drops when bonus time is burned up, and game should run out of time around move 50.

**Comment by roy7** [comment]
Is it possible your earlier graphs also used wrong PGN file? With and without PR should not be so similar and running out of time around move 70. With PR should be more steep and run out of time by 50.

**Comment by jkormu** [comment]
My previous build of pr257 had something wrong. Merged pr257 to master and rebuilt and now thing look more sensible (tc 45+1). 


```
Score of lc0_pr257_ponder vs sf9_ponder: 13 - 9 - 52  [0.527] 74
Elo difference: 18.80 +/- 43.24

Score of lc0_ponder vs sf9_ponder: 16 - 14 - 44  [0.514] 74
Elo difference: 9.39 +/- 50.68
```

Average remaining time:

remaining time

One sample games lc0_pr257:
remaining time_pr


One sample game lc0:
remaining time

**Comment by roy7** [comment]
Now that's more like it! ;)

**Comment by jkormu** [comment]
Ordo ratings for no ponder tournament vs time handicapped thread compensated sf:

```
   # PLAYER       :  RATING  ERROR  POINTS  PLAYED   (%)
   1 sf9          :    26.0   18.3   175.5     316    56
   2 lc0_pr257    :     3.2   18.6   158.5     313    51
   3 lc0          :   -29.2   18.4   137.0     313    44
```

I think this pr should be merged for the following reasons:
1. Seems like elo gain without ponder on short time control
2. Seems like there is no elo loss with ponder on short time control
3. Removes slowmover parameter, making it easier to tune.
4. We should let Leela think longer than normally when she has hard time deciding between moves. This pr does this in a limited way:
During moves that are easy to decide we build up bonus time thanks to smart pruning. When a hard move is encountered, Leela uses most of the bonus time.
5. Allows extensions where 4. is addressed properly. For example, use some portion of bonus time on next move and save rest for hard moves.

Thanks for the pr @roy7. I think I am done with the testing for now.

roy7:
Thank you @jkormu!

mooskagh:
I think it makes sense to make it a command line switch for one release to be able to compare and switch back and forth. After one month (our release cycle), we'll remove the switch and keep either "new" or "old" way a default.

roy7:
It's possible the bonus time value needs reset someplace(s) in case a game is reset in the middle or the engine is asked about multiple separate games one after another. I don't know the code well enough for that situation, but I hope the review or some other contributor might.

mooskagh:
Indeed, it should be a member of `EngineController` rather than a global variable.

roy7:
At first I tried to make it a member of one of the classes but couldn't find one that would work since I needed the information in two different locations. So eventually I gave up and used the global variable to at least test the idea. :)

mooskagh:
Implemented as suggested in https://github.com/LeelaChessZero/lc0/pull/358.

roy7:
I'm closing this one since @mooskagh has re-implemented it in a better way in https://github.com/LeelaChessZero/lc0/pull/358. His version handles game resets and also provides a command line option.

PR #260, opened 2018-08-15
Cache history length is int. Turned out it has always been 1.

mooskagh:


killerducky:
This bug is related to calculating the hash to use for the nncache hash. Instead of using all 8 history planes, it would use only 2 (current board, and board from previous ply). This means we will get more nncache hits and higher NPS performance (good), but at the cost of less accurate network results (bad).

This bug has been in the code since at least May 9, 2018. So it's unlikely to explain recent TCEC problems.

This is a tradeoff that has to be measured to know what is best. Long ago on lczero.exe Tilps measured that using all 8 history planes is better than using only 1: https://github.com/glinscott/leela-chess/pull/546. But that was long ago and there were many bug fixes and other changes since then.

killerducky:
@mooskagh with this patch we run 25% slower from start position (2410NPS vs 3191NPS). Measuring the Elo difference will take longer. Instead I'm thinking about looking into setting up Andrew Grant's Openbench. Will discuss that in Discord and maybe make an issue for it.

```
./lc0 -w ~/lcnetworks/t10_10520
go nodes 100000

0a205fee6ec2245a23f1345cade63160af18581f - History length is int. (#260)
info depth 8 seldepth 27 time 34136 nodes 82546 score cp 26 hashfull 400 nps 2418 pv e2e4 e7e6 d2d4 d7d5 e4d5 e6d5 g1f3 f8d6 f1d3 g8e7 e1g1 e8g8 c1g5 b8c6 g5h4 c6b4 d3e2 c8f5
info depth 8 seldepth 27 time 34118 nodes 81925 score cp 26 hashfull 397 nps 2401 pv e2e4 e7e6 d2d4 d7d5 e4d5 e6d5 g1f3 f8d6 f1d3 g8e7 e1g1 e8g8 c1g5 c8g4 g5h4 d8d7 h4g3 e7f5 g3d6


35f2483520aee29ae28132d6e930e075c913d895 - Previous version
info depth 8 seldepth 28 time 26107 nodes 83523 score cp 26 hashfull 288 nps 3199 pv e2e4 e7e6 d2d4 d7d5 e4d5 e6d5 g1f3 f8d6 f1d3 g8e7 e1g1 e8g8 c1g5 b8c6 g5h4 c6b4 d3e2 c8f5
info depth 8 seldepth 28 time 26241 nodes 83523 score cp 26 hashfull 288 nps 3182 pv e2e4 e7e6 d2d4 d7d5 e4d5 e6d5 g1f3 f8d6 f1d3 g8e7 e1g1 e8g8 c1g5 b8c6 g5h4 c6b4 d3e2 c8f5
```

PR #262, opened 2018-08-17
Fix UCI output to consistently use playouts

dubslow:


mooskagh:
I think showing reused nodes is more useful than resetting it from 0, and this time it costs us nothing to do that.
So, I anticipate another round of discussions. Although preferred solution would be just to close this PR. :)

mooskagh:
Closing for now.

PR #267, opened 2018-08-18
Add hidden options for controlling Dirichlet noise Alpha and Epsilon.

Mardak:
In preparation for experiments adjusting some magic numbers from A0-chess (α = 0.3, ε = 0.25), we can add options for the server to control these values. ~NB: with `--noise` replaced by `--noise-epsilon`, existing commands that use the old option will fail: `Unknown command line flag: --noise.`~

```
./lc0 --show-hidden --help
  -n,  --[no-]noise
               Add Dirichlet noise to root node prior probabilities. This allows the engine to
               discover new ideas during training by exploring moves which are known to be bad.
               Not normally used during play.
               [UCI: DirichletNoise  DEFAULT: false]

       --noise-epsilon=0.00..1.00
               Amount of Dirichlet noise to combine with root priors. This allows the engine to
               discover new ideas during training by exploring moves which are known to be bad.
               Not normally used during play.
               [UCI: DirichletNoiseEpsilon  DEFAULT: 0.00  MIN: 0.00  MAX: 1.00]

       --noise-alpha=0.00..10000000.00
               Alpha of Dirichlet noise to control the sharpness of move probabilities. Larger
               values result in flatter / more evenly distributed values.
               [UCI: DirichletNoiseAlpha  DEFAULT: 0.30  MIN: 0.00  MAX: 10000000.00]
```

The 1e7 for alpha is to allow for flat noise, e.g.,
```
./lc0 --verbose-move-stats --noise-epsilon=1 --noise-alpha=10000000
go nodes 1
info string a2a3  (204 ) N:       0 (+ 0) (P:  5.00%) (Q:  0.03735) (U: 0.17010) (Q+U:  0.20745) (V:  -.----) (T: 0) 
info string h2h4  (403 ) N:       0 (+ 0) (P:  5.00%) (Q:  0.03735) (U: 0.17010) (Q+U:  0.20745) (V:  -.----) (T: 0) 
info string a2a4  (207 ) N:       0 (+ 0) (P:  5.00%) (Q:  0.03735) (U: 0.17014) (Q+U:  0.20748) (V:  -.----) (T: 0) 
```

### DIFF
--- a/src/mcts/params.cc
+++ b/src/mcts/params.cc
@@ -87,6 +87,12 @@ const OptionId SearchParams::kTemperatureVisitOffsetId{
     "Adjusts visits by this value when picking a move with a temperature. If a "
     "negative offset reduces visits for a particular move below zero, that "
     "move is not picked. If no moves can be picked, no temperature is used."};
+const OptionId SearchParams::kNoiseId{
+    "noise", "DirichletNoise",
+    "Add Dirichlet noise to root node prior probabilities. This allows the "
+    "engine to discover new ideas during training by exploring moves which are "
+    "known to be bad. Not normally used during play.",
+    'n'};
 const OptionId SearchParams::kNoiseEpsilonId{
     "noise-epsilon", "DirichletNoiseEpsilon",
     "Amount of Dirichlet noise to combine with root priors. This allows the "
@@ -205,6 +211,7 @@ void SearchParams::Populate(OptionsParser* options) {
   options->Add<FloatOption>(kTemperatureWinpctCutoffId, 0.0f, 100.0f) = 100.0f;
   options->Add<FloatOption>(kTemperatureVisitOffsetId, -1000.0f, 1000.0f) =
       0.0f;
+  options->Add<BoolOption>(kNoiseId) = false;
   options->Add<FloatOption>(kNoiseEpsilonId, 0.0f, 1.0f) = 0.0f;
   options->Add<FloatOption>(kNoiseAlphaId, 0.0f, 10000000.0f) = 0.3f;
   options->Add<BoolOption>(kVerboseStatsId) = false;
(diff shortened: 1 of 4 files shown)


mooskagh:
Is it planned to remove those parameters later after the experiments?
If not, it makes sense to remove --noise parameter (and check for --noise-epsilon = 0 instead). Less params is better. :)

Mardak:
Nod. There's some duplicate noise configurations possible, although relying on --noise-epsilon=0 (or default value 0) for noise off and --noise-epsilon=0.25 to turn on forces people to provide the correct 0.25 value to turn on noise. Although unclear if people are manually turning on --noise outside of automatic self-play.

Also, removing --noise and adding --noise-epsilon would require the server to change at the same time as clients that don't understand --noise-epsilon would reject, and newer clients similarly would reject --noise.

Mardak:
Updated PR addressing comments to remove --noise/-n and only have --noise-epsilon (default 0) and --noise-alpha instead.

```
       --noise-epsilon=0..1           Add Dirichlet noise at root node (weighting).
                                      (default: 0  min: 0  max: 1)
       --noise-alpha=0..1e+07         Dirichlet noise alpha (scaling).
                                      (default: 0.3  min: 0  max: 1e+07)
```

Looks like self-play doesn't have noise set by the server anyway as it locally turned on noise, so the earlier concerns of needing a simultaneous server update aren't needed.

I suppose alternatively, the options could be `--noise=0.25` instead of `--noise-epsilon=0.25` ?

Tilps:
yes we aren't sending the noise flag from server so this change wouldn't break training in transition.

Mardak:
Rebased with the new `OptionId` approach. I saw there was some discussion of using noise for matches. Might be useful use an epsilon / alpha different from the default 0.25 and 0.3.

Mardak:
Rebased fixing minor conflicts from #771 and #750.

mooskagh:
Is this parameter planned to be used?..
It's only useful for training, right? Are we still planning to adjust those? Need input from @Tilps.

Mardak:
There was some discussion of adding more variety to matches, which could be done with different noise. Not sure if it's less of an issue now with larger temp visit offsets used for match variety.

Tilps:
I'm not sure if we're planning on adjusting these until someone produces some evidence that our current settings are wrong for some reason - but without the settings maybe no one will do the testing to work that out.

mooskagh:
Maybe make them hidden options (we hide some similar options)? So that people don't ask what those Uci params mean.

Mardak:
Marked the options as hidden. Probably should merge this as part of a major version as backwards compatibility can be broken for those expecting `--noise` or `-n` command line options or `DirichletNoise` uci option.

Tilps:
Both selfplay and matches uses selfplay mode - and both use noise enabled now - so we could just stop sending the noise flag in parameters since its the default - then it would be safe to submit this.

jhorthos:
I would like to use tunable noise parameters in training - hidden is fine with me.

kiudee:
It turns out that for our experiment with the new equilibrium-based U formula, we need the possibility to set the parameter(s) for the Dirichlet noise.

Is there anything preventing this from being merged? It rebases cleanly onto the current master.

See https://github.com/LeelaChessZero/lc0/issues/924#issuecomment-521997069 for the rationale and #918 for the U formula.

mooskagh:
I have no objections, but I think we should remove `--noise=true` from match params for all runs on the server before releasing this.

Tilps:
or just put the noise flag back until this has been included in a version which has been made mandatory.

Mardak:
Rebased and put back noise flag and treating as 0.25 epsilon.

PR #268, opened 2018-08-18
Add --lagbuffer to support bots avoiding network timeouts

roy7:
we don't want to calculate this * movestogo and mess up time management. But setting move_overhead to like 1000 or more will do that, so it's a poor way to make the bot move early to avoid network timeouts. 

--lagbuffer is the same term Leela Zero uses. It subtracts this from the max ceiling time allowed to think.


fli:
This patch is good for local play too. Below is an image of time lost to overhead/lag just on my local machine from lc0 to cutechess.

screen shot 2018-08-19 at 4 31 23 pm

To completely avoid timeouts I would need to set my move overhead to the minimum value (100ms), but it only reaches that minimum on extremely rare occasions and the time management code will plan as if that move overhead happens on every move. So it would be better for me to set move overhead to the average of the time lost (17ms), then a parameter such as _lagbuffer_ to the worst case (100ms) so that all time is used effectively.

src/engine.cc
```diff
@@ -131,14 +133,15 @@ SearchLimits EngineController::PopulateSearchLimits(int ply, bool is_black,
...
+  int64_t lagbuffer = options_.Get<int>(kLagBufferStr);
   float time_curve_peak = options_.Get<float>(kTimeCurvePeak);
   float time_curve_left_width = options_.Get<float>(kTimeCurveLeftWidth);
   float time_curve_right_width = options_.Get<float>(kTimeCurveRightWidth);
 
   // Total time till control including increments.
   auto total_moves_time =
       std::max(int64_t{0},
-               time + increment * (movestogo - 1) - move_overhead * movestogo);
+               time - lagbuffer + increment * (movestogo - 1) - move_overhead * movestogo);
```
fli: If it is working as a "worst case" buffer then I don't think you need this change in this line.

src/engine.cc
```diff
@@ -162,7 +165,7 @@ SearchLimits EngineController::PopulateSearchLimits(int ply, bool is_black,
   // Make sure we don't exceed current time limit with what we calculated.
   limits.time_ms = std::max(
       int64_t{0},
-      std::min(static_cast<int64_t>(this_move_time), time - move_overhead));
+      std::min(static_cast<int64_t>(this_move_time), time - lagbuffer - move_overhead));
```
fli: Again if it is working as a "worst case" buffer then I think this line just needs to be `time - max(lagbuffer, move_overhead)`

roy7:
Funny thing is on ultrabullet I still sometimes time out even if I hard code in like a 2500 ms buffer. Not sure why though. Maybe it's just my connection to lichess is so laggy I'm losing time on every move even if bot moves instantly, which isn't LC0's fault of course. ;)

src/engine.cc
roy7: Good call using max() instead.

roy7:
@AndyGrant said the total_moves_time being calculated by subtracting move_overhead * movestogo is in error and the multiplication shouldn't be there. He suggests this fix https://github.com/LeelaChessZero/lc0/compare/master...AndyGrant:patch-1. If that is done, then this PR can be closed since move_overhead would identically to lagbuffer.

mooskagh:
#306 is pushed, so closing this.

PR #270, opened 2018-08-19
Make cache-history-length 1 by default in UCI mode.

mooskagh:


src/engine.cc
```diff
@@ -111,6 +111,10 @@ void EngineController::PopulateOptions(OptionsParser* options) {
   defaults->Set<float>(Search::kCpuctStr, 3.4f);         // CPUCT = 3.4
   defaults->Set<float>(Search::kPolicySoftmaxTempStr, 2.2f);  // Psoftmax = 2.2
   defaults->Set<int>(Search::kAllowedNodeCollisionsStr, 32);  // Node collisions
+  // Cache key has a history of 1 ply back. That's to be compatible with old
+  // bug. Also tests show that for now 1 has better strength than 7.
```
borg323: Maybe rephrase to "**likely** has better strength than 7."

dubslow:
I had thought MTGOStark had several hundred games showing that 7 was better.

xghostxx:
Just a question: if the current Leela has been trained on cache-history-length=1, then wouldn't it naturally be stronger at that value than at 7? Wouldn't a proper test be training a net with this value only ever set to 7 vs a net with its value set to 1 over the same set of training games and then pitting the two nets against each other?

almaudoh:
> Just a question: if the current Leela has been trained on cache-history-length=1, then wouldn't it naturally be stronger at that value than at 7? Wouldn't a proper test be training a net with this value only ever set to 7 vs a net with its value set to 1 over the same set of training games and then pitting the two nets against each other?

My thoughts exactly. Even though the result above:

> `Score of lc0_hist0 vs lc0_hist1: 40 - 36 - 124  [0.510] 200
> Elo difference: 6.95 +/- 29.71`

seems to imply no history is better than 1-ply history, even for nets trained with 1-ply history. That would be interesting.

mooskagh:
Could you also test out-of-order-eval vs no-out-of-order-eval, both with `history-cache-length=0`?

borg323:
I did the out-of-order-eval vs no-out-of-order-eval with history-cache-length=0 test:
```
Score of lc0_hist0_out_of_order vs lc0_hist0_no_out_of_order: 52 - 41 - 107  [0.527] 200
Elo difference: 19.13 +/- 32.88
```

dubslow:
Please copy paste that result to 276 as well! Also, I'd be very much interested in that test continuing!

PR #271, opened 2018-08-19
Replace time curve logic with movestogo approximator.

fli:
This replaces the previous time curve/move importance logic with a function which guesses how many moves left in the game are worth planning time management for, and then splits time evenly among them.

I wanted to make this change as the previous logic had trouble planning for games with increment. Games with a decent amount of increment had large discontinuous drops in the time curve (see images below), because the previous logic would plan to use a lot of increment that hadn't been added to the clock yet.

You can also make a copy of [this spreadsheet](https://docs.google.com/spreadsheets/d/1viJP-gffkpRGVEQFqTNSsaXVwQ8cmxD303cfNeqnMTE/edit#gid=0) and play with some values to compare the old time curve with this proposed change.

*TCEC time settings. Move time on left axis, clock time on right axis*
screen shot 2018-08-19 at 4 01 49 pm

_20s+10s increment_
screen shot 2018-08-19 at 4 23 47 pm

The function which computes the weight at each ply is the [Survival function of the log-logistic distribution](https://en.wikipedia.org/wiki/Log-logistic_distribution) which I found fit very well to some analysis I did of when Leela's games tend to end.
depdatavsindepdata1_modelplotlrp_11345_1534646361_9010134
_Regression of ply vs P(game still alive at ply)_

I'm still in the process of tuning the parameters of the curve.


src/engine.cc
```diff
@@ -133,33 +136,49 @@ SearchLimits EngineController::PopulateSearchLimits(int ply, bool is_black,
...
+  // Compute first weight separately as it will be used to normalise.
+  float this_move_prob = ComputePlyWeight(ply, time_curve_midpoint,
+                                          time_curve_scale, time_curve_shape);
+
+  // Integrate over a large range of plies to approximate integrating to
+  // infinity.
+  for (int i = 1; i < ply + 300; ++i) {
+    guessed_movestogo += ComputePlyWeight(ply + 2 * i, time_curve_midpoint,
+                                          time_curve_scale, time_curve_shape);
+  }
```
dubslow: Surely you can compute the analytic integral, or a good approximation to it, and then compute that directly instead of this tiresome loop?

fli:
I've removed the shape parameter by making sure that the function is 1 at ply=0. Also removed the normalise logic as I don't think that was correct.

src/engine.cc
fli: Yes, possibly. I'll investigate.

src/engine.cc
```diff
@@ -90,12 +89,10 @@ void EngineController::PopulateOptions(OptionsParser* options) {
   options->Add<StringOption>(kNnBackendOptionsStr, "backend-opts");
   options->Add<FloatOption>(kSlowMoverStr, 0.0f, 100.0f, "slowmover") = 2.66f;
   options->Add<IntOption>(kMoveOverheadStr, 0, 10000, "move-overhead") = 100;
-  options->Add<FloatOption>(kTimeCurvePeak, -1000.0f, 1000.0f,
-                            "time-curve-peak") = 26.2f;
-  options->Add<FloatOption>(kTimeCurveLeftWidth, 0.0f, 1000.0f,
-                            "time-curve-left-width") = 82.0f;
-  options->Add<FloatOption>(kTimeCurveRightWidth, 0.0f, 1000.0f,
-                            "time-curve-right-width") = 74.0f;
+  options->Add<FloatOption>(kTimeCurveMidpoint, 0.0f, 200.0f,
```
fli: TODO: set minimum to 1 to avoid dividing by 0

fli:
I updated this with the current best tuned values. There is a closed solution for summing to infinity if we keep the parameter values constant (or at least their ratio).

fli:
I realised yesterday that the normalising logic in my first commit was correct (but other parts weren't) so I've fixed that, rebased, and am retuning.

The loop should essentially be calculating:


Where `midpoint > 0`, `scale > 0`, `move >= 0 and is the current move the game is up to`.
So if anyone has an elegant solution to computing that, I'm all ears!

oscardssmith:
What's x? The thing you are summing over?

fli:
Sorry, should be x=move under that summation symbol.

oscardssmith:
Some quick playing with desmos shows that this perfectly fits ax^b+c, but I'm not sure how to solve for a, b, and c. I'll keep you updated.

roy7:
A reference if it has any value for you is [TIME MANAGEMENT PROCEDURE IN COMPUTER CHESS](http://facta.junis.ni.ac.rs/acar/acar200901/acar2009-07.pdf).

fli:
Thanks @roy7 I think I have skimmed that paper before. I guess our strategy is quite similar in that we estimate the remaining moves, and spend less time on moves with only 1 option. I would like to eventually investigate what is pretty much the inverse of smart pruning, where if two moves are close then extend thinking time (what is said on the bottom of pg 78 of that paper)

fli:
I've simplified the function a little bit and changed the parameter names to `Time curve midpoint ply` and `Time curve steepness`.

I've also updated the default values to those picked by a 26165 game CLOP tune at 12.42s+0.069s against multiple engines, with no adjudication, no futile-aversion, scale thinking time = 1, and with PR #309. My goal was to determine a baseline time curve without any dynamic TM features from which dynamic TM features could then be added. For higher time controls and with tablebases or adjudication the optimal values will probably shift leftwards some, but I think this is a good conservative default.

I'll keep tuning a bit longer and update the default values again when I'm satisfied with it. I'll do some tests after that, but this PR is ready for others to test if anyone wishes to do so.

This is the current time curve vs the new one at CCCC time controls
image

fli:
Some test results against master and time handicapped stockfish, w/ default settings and no adjudication:

18s+.1s, 95% chance pr is better than master after 150 games
```
   # PLAYER        :  RATING  ERROR  POINTS  PLAYED   (%)  CFS(%)
   1 stockfish     :    50.5   36.2    60.5     100    60      93
   2 lc0-pr        :     2.5   36.5    50.5     100    50      95
   3 lc0-master    :   -52.9   37.1    39.0     100    39     ---
```

30s+.5s, 87% chance pr is better than master after 372 games
```
   # PLAYER        :  RATING  ERROR  POINTS  PLAYED   (%)  CFS(%)
   1 stockfish     :    13.3   23.2   131.0     248    53      66
   2 lc0-pr        :     4.7   22.7   126.5     248    51      87
   3 lc0-master    :   -18.1   22.6   114.5     248    46     ---
```

So at these time controls it is an ELO gain, at longer time controls it may be a wash. But the biggest plus about this patch is more stable time curve behaviour at different time controls with different sized increments.

oscardssmith:
Does this converge on long tc to the current approach? If not, how do we know it's not worse?

jjoshua2:
yah, doubling TC losing ~20 elo is not a good trend. but a longer TC test would show if it is converging to zero at longer TC or actually negative. What I have done is test 20 games at 30 min + 10s TCEC time controls, and gotten all draws, and then I look at the curve average and see if I like it better. If so, I conclude it is good even at long TC. But an intermediate test or two is also necessary I think, like 60s+0.5s and 3'+1".

fli:
Had to use my compute for some other things recently. I'll merge latest and should be able to pick this test up again in the next day or two.

fli:
Results of 180+1 test.
```
Head to head statistics:
1) lc0-pr               9.3 :    272 (+37,=207,-28),  51.7 %

   vs.                       :  games (  +,   =,  -),   (%) :    Diff,    SD, CFS (%)
   lc0-master                :     68 (  5,  59,  4),  50.7 :    +1.0,  12.4,   53.3
   lc0-pr-dynamic            :     68 (  5,  59,  4),  50.7 :    +4.1,  12.6,   62.9
   lc0-master-dynamic        :     68 (  8,  49, 11),  47.8 :    +5.2,  12.3,   66.3
   stockfish                 :     68 ( 19,  40,  9),  57.4 :   +36.3,  12.7,   99.8

2) lc0-master           8.3 :    272 (+26,=228,-18),  51.5 %

   vs.                       :  games (  +,   =,  -),   (%) :    Diff,    SD, CFS (%)
   lc0-pr                    :     68 (  4,  59,  5),  49.3 :    -1.0,  12.4,   46.7
   lc0-pr-dynamic            :     68 (  7,  57,  4),  52.2 :    +3.1,  12.6,   59.7
   lc0-master-dynamic        :     68 (  3,  61,  4),  49.3 :    +4.1,  12.4,   63.1
   stockfish                 :     68 ( 12,  51,  5),  55.1 :   +35.3,  12.8,   99.7

3) lc0-pr-dynamic       5.2 :    272 (+29,=219,-24),  50.9 %

   vs.                       :  games (  +,   =,  -),   (%) :    Diff,    SD, CFS (%)
   lc0-pr                    :     68 (  4,  59,  5),  49.3 :    -4.1,  12.6,   37.1
   lc0-master                :     68 (  4,  57,  7),  47.8 :    -3.1,  12.6,   40.3
   lc0-master-dynamic        :     68 (  3,  61,  4),  49.3 :    +1.0,  12.5,   53.3
   stockfish                 :     68 ( 18,  42,  8),  57.4 :   +32.2,  12.7,   99.4

4) lc0-master-dynamic   4.2 :    272 (+26,=224,-22),  50.7 %

   vs.                       :  games (  +,   =,  -),   (%) :    Diff,    SD, CFS (%)
   lc0-pr                    :     68 ( 11,  49,  8),  52.2 :    -5.2,  12.3,   33.7
   lc0-master                :     68 (  4,  61,  3),  50.7 :    -4.1,  12.4,   36.9
   lc0-pr-dynamic            :     68 (  4,  61,  3),  50.7 :    -1.0,  12.5,   46.7
   stockfish                 :     68 (  7,  53,  8),  49.3 :   +31.1,  12.4,   99.4

5) stockfish          -27.0 :    272 (+30,=186,-56),  45.2 %

   vs.                       :  games (  +,   =,  -),   (%) :    Diff,    SD, CFS (%)
   lc0-pr                    :     68 (  9,  40, 19),  42.6 :   -36.3,  12.7,    0.2
   lc0-master                :     68 (  5,  51, 12),  44.9 :   -35.3,  12.8,    0.3
   lc0-pr-dynamic            :     68 (  8,  42, 18),  42.6 :   -32.2,  12.7,    0.6
   lc0-master-dynamic        :     68 (  8,  53,  7),  50.7 :   -31.1,  12.4,    0.6
```

I wasn't sure whether to test with the new dynamic slowmover or not, so I tested with and without. Looks like all versions are roughly equal (probably need 10-100x more games to be able to say with certainty which is better), although the PR versions seem to have scored better against SF (though maybe too few games played).

I think this shows that this PR is at least on par and likely better at STC compared to the original time curve. Last thing I want to test is parameters chosen based on analysis of when Leela games are decided by as opposed to these parameters which were chosen based on STC clopping. If parameters based on analysis are comparable, I'd rather choose them because they would probably be less overtuned and make more conceptual sense.

If this PR is taken, time management could be simplified by removing slowmover, making the steepness parameter static, and using the `Time curve midpoint ply` as the only parameter adjusting the time curve. It would be a very intuitive way to adjust the time curve, i.e. "When do I want to start spending half of the max allocated time per move?"

roy7:
If you are testing this PR vs master without this PR but using immediate-time=1, you also need arg="--slowmover=1.0" arg="--time-curve-peak=1" arg="--time-curve-left-width=60" arg="--time-curve-right-width=60". None of those defaults were changed in the PR that was merged. At least if you want to replicate my original results.

Of course with your PR, the time curve settings all get deleted thankfully.

jjoshua2:
@fli Is it possible for you to resolve conflicts? I'd like to test this again with all our new changes like use time immediate and fixed puct scaling, using time right spots can actually change moves now!

fli:
Resolved conflicts. Haven't tested thoroughly but seems to work. I'll have that time script for you later on.

Here's the spreadsheet to see what it should be doing: https://docs.google.com/spreadsheets/d/1viJP-gffkpRGVEQFqTNSsaXVwQ8cmxD303cfNeqnMTE/edit#gid=0

zz4032:
I tuned the two parameters myself just to make sure that I get similar results.
Tuning in selfplay and all other lc0 settings on default.

Tuning results:
```
TimeMidpointPly = 101 (default: 100)
TimeSteepness = 9.10 (default: 7)
```
TimeSteepness is a bit higher than default.

@fli Is there a sheet where I can observe how the new time curve changes? The Google sheet diagram only changes when I modify old time curve parameters, not the new curve parameters.
Btw in line 62 in src/engine.cc it should be:
`      "time-midpoint-halfmove", "TimeMidpointPly", `

Time control: 19.68s/game + 0.082s/move (same as in tuning)
Increment is 1/240th of main time and I get ~6000 nodes on the first moves .

Test match vs. lc0:
```
   # PLAYER           :  RATING  ERROR   GAMES  LOS(%)  DRAWS(%)
   1 lc0-PR271-tuned  :     0.0   ----    1000    73.4      53.7
   2 lc0              :    -4.4   13.8    1000     ---      53.7
```
6-men-TBs WDL+DTZ (and some 7-men WDL) were used in search and for 6-men adjudication. Unbalanced book to avoid draws was used (+ replaying with reversed colors of course).

The idea of removing Slowmover and making TimeSteepness a constant sounds great.

fli:
@zz4032 Thanks. You may have to switch tabs back and forth to make it update.

I don't expect it to be a big win against the current time stuff, but conceptually it is simpler, it works more predictably for different time controls and doesn't have the discontinuous drops. I think it's a better baseline than the current code.

Slowmover will probably need to be greater than 1 to get best performance because ImmediateTimeUse is 0.6 and not 1.

roy7:
I'm still a fan of immediate time use 1.0. :)

zz4032:
What could be a plan to reduce time parameters with PR271? I don't have complete understanding for all of them.
1) fix: Slowmover=1, TimeSteepness=9 (or 8 or 7?)
2) tune: TimeMidpointPly, ImmediateTimeUse, SmartPruningFactor
3) remove/hide as parameters: Slowmover, TimeSteepness, ImmediateTimeUse?

For ponder=on TimeMidpointPly would need to be retuned again (replacing Slowmover).

Test matches vs. current version at 1) short TC, 2) long TC, 3) short TC + ponder=on ?

jjoshua2:
I think a good end goal is to have only one uci param for time control, that is easy to understand, like slowmover for SF. If you want to think 10% more or less its obvious how to do it, unlike changing slope of curve. Although if ponder=on we might want to automatically increase it 50% like sf does while still keeping it the same uci setting. It would be nice to be able to have 100 be the default like SF used to be, even if we have to scale it, but SF drops its default to 88 and such so it's not a big deal.

Smart pruning is something unique that I think we will also always need. Some people will want it to be able to set it to 0 or 1 regardless of elo differences in games because of unique circumstances.

zz4032:
Meanwhile I have started steps 1 + 2.
Tuning runs need a bit more games, but I'd like to share the results:
```
1) Tuning with ponder=off, short TC, Id32247
TimeMidpointPly=101
ImmediateTimeUse=0.528
SmartPruningFactor=1.755
```
```
2) Tuning with ponder=on, short TC, Id32247
TimeMidpointPly=121
ImmediateTimeUse=0.477
SmartPruningFactor=1.853
```
**TimeMidpointPly**: Good news is that TimeMidpointPly seems to work as expected. We could use 100 with ponder=off and 120 with ponder=on.
**ImmediateTimeUse**: That's the third time I'm tuning it and I never got anything else than something between 0.5 and 0.6 and CLOP diagram hardly shows any visible influence on Elo (fading areas of samples). I hope it's working as expected... Maybe some runs with extreme values can show what the influence really is?
**SmartPruningFactor**: I did several tests in the past on that (also @jjoshua2). At low TC the optimum value is around 2.0 while with very long TC the default of 1.33 is probably close to optimum so the estimation of nps is not good enough on lower nodes. Either the parameter can be modified to be a function of nodes or the estimation could be improved...

zz4032:
More test results with PR271:

**Match 1** at short TC
With SmartPruningFactor=1.75, TimeSteepness=9 and slightly different TC.
TC: 14.22s/game+0.079s/move (~4000 nodes/move, increment 1/180th of base time)
```
   # PLAYER       :  RATING  ERROR   GAMES  DRAWS(%)  LOS(%)
   1 lc0-PR271    :       5     16    1000      50.7    74.1
   2 lc0          :       0   ----    1000      50.7     ---
```

**Match 2** at short TC with Slowmover > 1
Parameters same as in match 1, but Slowmover=1.3.
```
   # PLAYER             :  RATING  ERROR   GAMES  DRAWS(%)  LOS(%)
   1 lc0-PR271          :       0   ----    1000      52.5    92.0
   2 lc0-PR271_SM1.3    :     -11     15    1000      52.5     ---
```

**Match 3** at short TC with ponder=on
Parameters same as in match 1.
```
   # PLAYER              :  RATING  ERROR   GAMES  DRAWS(%)  LOS(%)
   1 lc0-PR271           :      75     16    1000      54.0   100.0
   2 lc0                 :       0   ----    1000      54.0     ---

```

My conclusion after tuning and matches:
- PR271 seems to be equal to current master and better with ponder=on
- changing of Slowmover=1 shows no improvement (match 2)
- probably the same can be said about ImmediateTimeUse=0.6 (observed in tuning), going to run one more match with ImmediateTimeUse=1.0
- TimeSteepness=9 may also be best being fixed (with ponder and without), TODO tuning and testing

fli:
Thanks @zz4032.

I think you end up not needing to change slowmover for ponder if immediatetimeuse=1 (needs to be tested though).

Probably a good idea to hide the time curve settings and either have slowmover visible or TimeMidpointPly visible.
My vision of an end goal for time management is to:
- have the slowmover and ImmediateTimeUse parameter removed entirely
- have the smartpruning code also be able to extend the time allocated for a move if visits are close for top 2 moves or Q of top move less than Q of second move. 
- with these changes, have the average time used on all moves still look like the baseline time curve.

careless25:
I also ran a test with the following settings with this PR vs v19.1.1 and id 32194: 
1 min + 1 sec
``` 
--backend=cudnn-fp16 --cpuct=3.1 --slowmover=1.3 --max-collision-events=256 -w 32194.pb.gz --nncache=10000000
```

```
   0 stockfish_10_x64_bmi2          39      22     367   55.6%   59.9%
   1 lc0_32194                     -35      33     180   45.0%   58.9%
   2 lc0_TM_271                    -43      31     187   43.9%   61.0%

367 of 800 games finished.
```

They are about equal.

zz4032:
Match with ImmediateTimeUse=0.6 (default) and ImmediateTimeUse=1.0.
TC: 47.16s/game+0.262s/move (~10,000 nodes/move, increment 1/180th of base time)
Match was played with `concurrency=2`, just in case this parameter needs more time per move (0.8-1.0s/move on average but half as much nps).
```
   # PLAYER              :  RATING  ERROR   GAMES  DRAWS(%)  LOS(%)
   1 lc0-PR271_ITU1.0    :       2     16    1000      52.3    61.6
   2 lc0-PR271           :       0   ----    1000      52.3     ---
```
Based on match results both Slowmover and ImmediateTimeUse can be fixed at default.
Now only both parameters from this PR remaining to test. If the same goes for TimeSteepness, there will be only one remaining.

jjoshua2:
Using zz tuned values there a lot better. Let's me use slowmover 1 and use time immediate 1. `58.45 +/- 46.91` over master with 1.35 at 60s+0.5
```
Score of lc0time vs lc0 v19.1rc2: 10 - 4 - 46 [0.550]
Elo difference: 34.86 +/- 42.12
```
also win compared to slowmover 1.5 at 120+0.75s

mooskagh:
Is this PR ready for review? I keep skipping it because of `wip` tag.

fli:
Yep @mooskagh. It's ready for review. I've made the changes recommended by @jjoshua2 and @zz4032 .

src/engine.cc
```diff
@@ -216,20 +234,12 @@ SearchLimits EngineController::PopulateSearchLimits(
...
-                          time_curve_right_width);
-  // Compute the move time without slowmover.
-  float this_move_time = total_moves_time * this_move_weight /
-                         (this_move_weight + other_move_weights);
+  // Evenly split total time between all moves.
+  float this_move_time = total_moves_time / movestogo;
 
   // Only extend thinking time with slowmover if smart pruning can potentially
   // reduce it.
+  constexpr int kSmartPruningToleranceMs = 200;
```
mooskagh: I'm not sure anymore that kSmartPruningToleranceMs is a good idea, maybe remove it?

src/engine.cc
borg323: This seems to calculate the mean residual time. A web search found a solution in <http://www.lifescienceglobal.com/pms/index.php/ijsmr/article/download/2691/1579> equation (3). The same reference also gives in equation (5) a simpler formula for the median residual time.

src/engine.cc
jjoshua2: I agree, but I think that smart pruning modifications should be separate PR and will need additional testing. Not sure if it can be removed entirely, it seems maybe it should be a function of time budget. We can afford to have it be a second or two in 90 minute games.

jjoshua2:
I will re-test elo gain for master now since 20rc2 with tilps speed and this should be nice. @fli please look into borg's simplification formula and crem's kSmartPruningToleranceMs comment for future PR.

fli:
Thanks. @borg323 that looks very promising, I'll have a look at implementing the median residual time formula.

PR #274, opened 2018-08-19
Fix training data being generated without its flag

dubslow:


src/selfplay/game.cc
```diff
@@ -87,9 +87,11 @@ void SelfPlayGame::Play(int white_threads, int black_threads,
     search_->RunBlocking(blacks_move ? black_threads : white_threads);
     if (abort_) break;
 
-    // Append training data. The GameResult is later overwritten.
-    training_data_.push_back(tree_[idx]->GetCurrentHead()->GetV3TrainingData(
-        GameResult::UNDECIDED, tree_[idx]->GetPositionHistory()));
+    if (training) {
+      // Append training data. The GameResult is later overwritten.
+      training_data_.push_back(tree_[idx]->GetCurrentHead()->GetV3TrainingData(
```
dubslow: Shouldn't this be `emplace_back`?

mooskagh: It's one copy with push_back() and probably one copy with emplace_back() too (surely not less).
I don't really have a preference here actually.

dubslow: Oh right, forgot that the argument to push_back isn't actually a constructor call lol

PR #276, opened 2018-08-19
Make out of order eval standard

dubslow:
I genuinely think this should go in the release (not a new feature!), but also I need a place to store testing results.

This command `./lc0 selfplay --no-share-trees --games=500 --parallelism=8 -w ../../../weights/weights_t9_9155.txt.gz --movetime=5000 --max-prefetch=0 --no-noise --reuse-tree --tempdecay-moves=15 --temperature=2 "player1:" --out-of-order-eval "player2:" --no-out-of-order-eval` produced these results (truncated early): 97-76-63, which shows a not-bad minigain for OOOeval

Meanwhile wajit used a similar command with a test10 network, t=1, tdecay=20, and got 124-126-313, with a much larger sample size showing a wash.

I've started a third run now for further testing. Testing against AB engines is probably also a good idea.

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -577,18 +577,16 @@ void SearchWorker::GatherMinibatch() {
 
       // Only send non-terminal nodes to a neural network.
       if (!node->IsTerminal()) {
-        picked_node.nn_queried = true;
-        picked_node.is_cache_hit = AddNodeToComputation(node, true);
+        picked_node.nn_queried = AddNodeToComputation(node);
       }
     }
 
     // If the latest node doesn't require NN eval (i.e. it's a cache hit or
     // known terminal node), do an out of order eval and immediately remove it.
-    if (node->IsTerminal() || picked_node.is_cache_hit) {
-      FetchSingleNodeResult(&picked_node, computation_->GetBatchSize() - 1);
+    if (!picked_node.nn_queried) {
+      picked_node.v = node->GetQ();
       DoBackupUpdateSingleNode(picked_node);
 
-      if (picked_node.nn_queried) computation_->PopCacheHit();
       minibatch_.pop_back();
       ++number_out_of_order;
     }
@@ -750,15 +748,12 @@ void SearchWorker::ExtendNode(Node* node) {
   node->CreateEdges(legal_moves);
 }
 
-// Returns whether node was already in cache.
-bool SearchWorker::AddNodeToComputation(Node* node, bool add_if_cached) {
+// Returns whether node was added to cache (i.e. if it was a cache miss).
+bool SearchWorker::AddNodeToComputation(Node* node) {
   auto hash = history_.HashLast(search_->kCacheHistoryLength + 1);
   // If already in cache, no need to do anything.
-  if (add_if_cached) {
-    if (computation_->AddInputByHash(hash)) return true;
-  } else {
-    if (search_->cache_->ContainsKey(hash)) return true;
-  }
+  if (search_->cache_->ContainsKey(hash)) return false;
+
   auto planes = EncodePositionForNN(history_, 8);
 
   std::vector<uint16_t> moves;
@@ -780,7 +775,7 @@ bool SearchWorker::AddNodeToComputation(Node* node, bool add_if_cached) {
   }
 
   computation_->AddInput(hash, std::move(planes), std::move(moves));
-  return false;
+  return true;
 }
 
 // 3. Prefetch into cache.
@@ -805,7 +800,7 @@ int SearchWorker::PrefetchIntoCache(Node* node, int budget) {
 
   // We are in a leaf, which is not yet being processed.
   if (!node || node->GetNStarted() == 0) {
-    if (AddNodeToComputation(node, false)) {
+    if (!AddNodeToComputation(node)) {
       // Make it return 0 to make it not use the slot, so that the function
       // tries hard to find something to cache even among unpopular moves.
       // In practice that slows things down a lot though, as it's not always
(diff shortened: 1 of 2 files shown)


dubslow:
Another batch of 500 games completed, and I hate to say it crem but I believe this patch is working about how you expected, on CPU-bound systems at least it seems to be a clear, if slight, winner!

`tournamentstatus final win 100 103 lose 73 84 draw 77 63`, which when combined with my previous results comes to `300 - 233 - 203` under my high-temp = low-draw conditions. (As far as I know, GPU tests still come out a wash, though I don't know how many of the GPU tests have been CPU bound so far.) A shitty online calculator tells me that's about +30 elo on my cpu, though it didn't give error bounds (tho I'm also pretty sure the 95% error is less than 30 given that win-loss gap there!)

Please merge for v0.17.0 final!

Videodr0me:
I just posted this to the commit (but added some rough guestimates here):

We now do out-of-order eval on both cache-hits and terminals and we have no restrictions on the number of out-of-order evals in a row. We should make a mental (or physical) note, to one day collect some stats, so we can establish if doing it only one cache hits, or only on terminals, or not to many times in a row  is "better", and/or has some side-effect on collisions. 

Our theory for this was that it should help as not only n_started but also q is now updated before the next node is added to the batch - and i am hopeful that thats really the case, but one never knows.  For example a winning terminal might raise q along the path, and now attract the next visit (like a magnet) and lead to another out-of-order eval. I also think that this change probably needs so many games to really assess, that we will most likely never know (I guestimate about 20.000 games minimum with SPRT and probably 40.000 games without). 

So this is just a note to keep in mind that while most likely a good idea, we have to keep an eye on real life  performance.

mooskagh:
I don't think we should enable it for v0.17.0, or at least for the version which will be sent to CCCC:

1. We didn't test enough for the situations when all visits go into terminal node and it may stuck in a loop. (there is a check for that, but not tested enough)

2. There are multiple reports that lc0 with tablebases plays worse than without. One of possible explanations for that is that TB query takes longer than normal. Out-of-order eval will attract more tablebase hits and will slow it down even more.

Those may or may not be valid concerns, but we don't have time to test them.

mooskagh:
It does change single-thread visits order, and therefore it does not produce identical results even on a single thread.

dubslow:
Sorry, I meant singled threaded and batchsize 1 MCTS. The theoretic ideal MCTS has no parallel evaluations, which is what batching requires. No parallel evaluations means they can't be reordered, so by definition this doesn't change single threaded and non-batched MCTS, which is the only one that's been mathematically analyzed AFAIK.

As for 1), surely more than 1000 games played to completion, to `bestmove terminalnode`, is test enough that there's no problems with terminal nodes or looping?

I'm not sure if 2) is a valid concern, but that's because I doubt that TB is in fact worse at all.

borg323:
After @mooskagh asked, I did an out-of-order-eval vs no-out-of-order-eval with history-cache-length=0 test on colab with net 10780 and tc=30+0.5:
```
Score of lc0_hist0_out_of_order vs lc0_hist0_no_out_of_order: 52 - 41 - 107  [0.527] 200
Elo difference: 19.13 +/- 32.88
```

Videodr0me:
Evidence so far looks good, still so few games might be just noise. On the other hand, this is one of the safer changes. In another PR i made the suggestion to have a separate branch for all search changes. Search mods are tested and added to this branch, which is "cutting edge" and "experimental" in nature. Then after the changes have been tested in the wild a little, they are restested with a large enough amount of games (>10000, preferably with SPRT and the Fishtest 37400 position openings pgn), and only then added to master. This would reduce the frequent changes to master, so that contributors can use a stable version for longer periods of time.

dubslow:
Add these results to my sample: `tournamentstatus win 83 58 lose 34 47 draw 32 45`

This brings me to `441 - 314 - 280`, that's very clearly an Elo gain with 99.9% confidence on CPU-bound systems, and if you include wajit's games, that's more than 1500 games completed to a terminal state. That's plenty more than enough to conclude that concern 1) is not valid, and frankly reading it again concern 2) has nothing to do with oooeval anyways, that's a concern about the TB implementation.

dubslow:
Sigh... found a crash: https://gist.github.com/dubslow/baf28d20b1ca0c2e82ead188ceee429f Reviewing the output will reveal that it was in the thread with oooeval enabled. Thoughts?

dubslow:
Starting another run with `selfplay --verbose-move-stats --no-share-trees --games=500 --parallelism=8 -w ../../../weights/weights_t9_9155.txt.gz --movetime=1000 --max-prefetch=0 --no-noise --reuse-tree --tempdecay-moves=12 --temperature=3 "player1:" --out-of-order-eval "player2:" --no-out-of-order-eval` in an attempt to reproduce

mooskagh:
"double free or corruption (out)" most likely because we happen to write out of the allocated buffer.
Maybe it worth adding assert. Although it's possible that we write because `size_` is somehow overwritten.

mooskagh:
Also reducing cache size may make it reproduce faster as evictions are more frequent. And I think it deserves a separate github issue.

Videodr0me:
@dubslow I also regard this change as rather safe. Even though the evidence is also points in this direction, the number of games is clearly not enough not assert this with the degree of certainty that for example is used in SF testing. That in my opinion is a sound methodology which we should also adopt to reliably assess changes. But you are correct in that my comment is more general in nature - as we should rather sooner than later implement such a testing regime. Fortunately KillerDuckys OpenBench seems to be almost ready - you can see that thread for some estimates of number of games needed for elo-determination.

Tilps:
Since it doesn't seem to have been mentioned in this thread earlier - note that this will affect training, because training does not use batch size 1 at the current time.

killerducky:
Related to this PR, getting some timeouts with some UCI parameters changed:
http://ec2-34-217-73-2.us-west-2.compute.amazonaws.com:8000/viewTest/9/

Out-of-order_cache_backpropagation=true Length_of_history_to_include_in_cache=0

dubslow:
`option."Out-of-order cache backpropagation"=true` There is no such option in this PR

killerducky:
Yes the OB test is using the UCI option on master, not this PR. It should be similar right?

dubslow:
The point was to test it against master-default.

Dunno about the timeouts. Even the minimax PR got timeouts, and there's no way minimax affects time management.

dubslow:
Ran another 500 to try to reproduce with the smaller cache, still no dice `tournamentstatus final win 100 78 lose 81 103 draw 69 69`

dubslow:
Yet another 500 failed to reproduce the crash. `tournamentstatus final win 95 91 lose 91 94 draw 64 65` It's pretty clear that OOOeval gets bigger gains when there's more cache hits, surprise surprise. I'm giving up on trying to reproduce this for now, further testing will focus on this PR rather than the option in master that all these tests were.

As for influencing training, basically by construction, this won't affect training any more than the cache key length being 7 will affect training, and that magnitude in the opposite direction as well. Nothing to worry about, in other words.

mooskagh:
I'm hesitant to hardcode it on for now, I think it should be default-on but switchable off for at least one entire version.

PR #277, opened 2018-08-20
[WIP] Dynamic pruning agression

dubslow:
(Targeting v0.18)

Instead of a fixed 1.47 or 1.33 early-exit-aggression, compute said aggression dynamically based on the second best Q relative to the top move's Q. If the second best Q is *better*, then be less aggressive, and if the second best Q is *worse*, then be more aggressive. This is currently just a sketch/WIP, there's probably many future improvements.

There are two commits in here, the first is a refactoring I'd like to see, I think it makes a decent chunk of code easier to read. I'll be happy to make a separate PR of it if so desired, but for now I figured I oughta hold off spamming a whole bunch of new PRs.

So treat the first commit separate from the rest.

@fli @jjoshua2 You'll probably be interested in this


killerducky:
Ping @Videodr0me I think you had some code with similar ideas to this? Do you have something up to date that we could compare to this version?

oscardssmith:
Could this ever cause move time to be extended (ie move longer than smartmoving=1)? If so, could this ever cause it to use more time than it has?

dubslow:
Yes, and no, if you read the commit titles you'll see one of them implements a hard cap to allowed bonus time to prevent timeouts.

mooskagh:
Is this change based on some other change or master?
I can see it has GetBestChild() functions moved to Node. It's a good change, but it would be easier to review if it would be separate from the functional change.

killerducky:
Getting some timeouts with this PR:
http://ec2-34-217-73-2.us-west-2.compute.amazonaws.com:8000/viewTest/7/
http://ec2-34-217-73-2.us-west-2.compute.amazonaws.com:8000/viewTest/8/

fli:
This looks like a more dynamic version of the UNST strategy in 
https://dke.maastrichtuniversity.nl/m.winands/documents/Time_Management_for_Monte-Carlo_Tree_Search_in_Go.pdf:

```The UNST strategy, called “unstable evaluation” heuristic in [7], prolongs the search if after the regular search time the most-visited move at the root is not the highest- valued move as well. In this case, the search is continued for the previously used search time multiplied with a factor funstable. The idea is that by searching longer, the highest- valued move could soon become the most-visited and thus change the final move choice.```

Their simple-ish implementation improved strength, so this pr looks promising

PR #282, opened 2018-08-20
V3fix

Error323:
Conversion from lczero to lc0 binary introduced a difference in the trainingdata. All testserver data has been trained without a rule 50 counter as a result. This fix remedies the problem again.


dubslow:
The powers of padding bytes to trick us all

PR #285, opened 2018-08-21
Reduce opencl memory requirements

borg323:
This is my best guess for porting https://github.com/gcp/leela-zero/commit/07c908e1346205c6877f8d0520754ebc5c5ebe10 to the opencl backend. Ideally I would like the opinion of @Ttl and @frpays before merging.


Ttl:
Looks good. `alloc_insize` is the size of buffers used for moving data around between the convolutions. `max_batch_size * max_channels * 8 * 8` elements is enough for that purpose.

PR #293, opened 2018-08-22
Add minimal root probing support.

Tilps:
This is untested, but should probably work to some extent I think.

Its not a 'complete' root probe support as it has two cases where it can fail.
1) If after search GetChildrenVisits is 0 the move selected may not be optimal.
2) If tree reuse brings over a lot of visits, those visits will not have been restricted to the root probe list, so the new visits to the filtered list might not have time to catch up, and an invalid move will be made.

Both of these could be fixed by adding a root probe call to GetBestChild* and applying filtering in there just like it currently does for searchmoves.

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -543,6 +543,25 @@ void SearchWorker::InitializeIteration(
   computation_ = std::make_unique<CachingComputation>(std::move(computation),
                                                       search_->cache_);
   minibatch_.clear();
+
+  if (!root_move_filter_populated_) {
+    root_move_filter_populated_ = true;
+    // Search moves overrides tablebase.
+    if (!search_->limits_.searchmoves.empty()) {
+      root_move_filter_ = search_->limits_.searchmoves;
+    } else {
+      auto board = history_.Last().GetBoard();
+      if (search_->syzygy_tb_ && board.castlings().no_legal_castle() &&
+          (board.ours() + board.theirs()).count() <=
+              search_->syzygy_tb_->max_cardinality()) {
+        if (!search_->syzygy_tb_->root_probe(history_.Last(),
+                                             &root_move_filter_)) {
+          search_->syzygy_tb_->root_probe_wdl(history_.Last(),
+                                              &root_move_filter_);
+        }
+      }
+    }
+  }
 }
 
 // 2. Gather minibatch.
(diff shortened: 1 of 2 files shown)


Tilps:
Refactored and added the filtering to GetBestChild like I suggested.

killerducky:
@Tilps maybe it is already this way, but after we enter TB positions, I think we should only use TB for root probe, and nothing else. This allows Lc0 to search and play the moves it likes without weird influences from TB conversions. Will this PR work like that? Or would it be another enhancement?

Tilps:
TB probing is still used during search - but its the same WDL probing, not choice limiting like it is at the root.  Given that TB probing will be used at all root moves from then on, I doubt it matters much whether WDL probing is on during search or not. But feel free to experiment with that separately if you like.

Tilps:
jjosh reports leela drawing by 50 moves where it should be winning with this applied.  I've tested the problematic position as fen input over in #294 and root_probe correctly rejects the example moves.  I'm still investigating how this PR could possibly be giving invalid input to root_probe.

Tilps:
I've not been able to easily identify why jjosh's test showed issues - hopefully he can provide more details to rule out operator error.

src/mcts/search.cc
```diff
@@ -366,6 +366,26 @@ std::pair<Move, Move> Search::GetBestMove() const {
   return GetBestMoveInternal();
 }
 
+bool Search::PopulateRootMoveLimit(MoveList* root_moves) const {
```
mooskagh: Could you add a comment what this function returns?
And also why, it doesn't seem to be used anywhere. :)

src/mcts/search.cc
```diff
@@ -366,6 +366,26 @@ std::pair<Move, Move> Search::GetBestMove() const {
...
+    *root_moves = limits_.searchmoves;
+    return false;
+  }
+  auto board = played_history_.Last().GetBoard();
+  if (!syzygy_tb_ || !board.castlings().no_legal_castle() ||
+      (board.ours() + board.theirs()).count() > syzygy_tb_->max_cardinality()) {
+    return false;
+  }
+  if (!syzygy_tb_->root_probe(played_history_.Last(), root_moves)) {
+    if (!syzygy_tb_->root_probe_wdl(played_history_.Last(), root_moves)) {
```
mooskagh: Why nested ifs  `if (!a) if (!b) return false; ` and not just `if (!a && !b) return false` ?
Or even `return a || b;`

src/mcts/search.cc
Tilps: Its already commented in the header - and its used in the SearchWorker path to know whether to increment tb hit or not.

src/mcts/search.cc
Tilps: The later option seems reasonable, I'll do that.

PR #294, opened 2018-08-22
Some initial root probe tests.

Tilps:


Tilps:
These tests all pass for me with 5 piece syzygy copied into place besides the test exe.

Issue #295, opened 2018-08-22
Add noconfig option to be sent by the client.


Tilps:
The new config support potentially overrides command line arguments which is not ideal for when lc0 is invoked by the client.
I guess this could be accomplished with --config= with a code change to make that legal, but it might be clearer with a noconfig option.

cwbriscoe:
Is this with selfplay?  If so, config files don't load in selfplay mode.

Tilps:
Match games don't use selfplay.

cwbriscoe:
Oh so they use UCI then?  Then I think your suggestion to override config to nothing would be pretty easy to implement.  I will do it unless you want to.

cwbriscoe:
If you do `./lc0 --config=` it will actually not load a configuration file.  It incorrectly reports that it does.  I will fix it so that it does not give the erroneous message.

cwbriscoe:
@Tilps Does my PR fix your concerns?  Also, command line parameters will always override anything in the configuration file.  Lc0 now parses the command line twice.  Here is the order of operations:
1.  Parse command line to get the --config option
2.  Parse whatever is in --config or lc0.config.
3.  Parse the command line again which will override anything set in the config file if also specified on the command line.

mooskagh:
We are switching to matches using `selfplay`, it won't have this problem anymore!

Tilps:
I think the concerns were already addressed by PR301 anyway :)

PR #301, opened 2018-08-23
Fix reporting that config file was loaded when it wasn't

cwbriscoe:
This fixes issue https://github.com/LeelaChessZero/lc0/issues/295.

If you run lc0 with the following command, it reports that the config file was found when in fact it was not.

`./lc0 --config=`

You would run the above command if you wanted to override using the default lc0.config file.  This PR will not print that the config file at ./ was found which is confusing since no config file was actually loaded.

I also changed the ParseFile function signature to use `const std::string& `instead of `const std::string`.

### DIFF
--- a/src/utils/configfile.cc
+++ b/src/utils/configfile.cc
@@ -56,6 +56,11 @@ bool ConfigFile::Init(OptionsParser* options) {
   // Calculate the relative path of the config file.
   OptionsDict dict = options->GetOptionsDict();
   std::string filename = dict.Get<std::string>(kConfigFileStr);
+
+  // If filename is an empty string then return true.  This is to override 
+  // loading the default configuration file.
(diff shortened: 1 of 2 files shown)


src/utils/configfile.cc
```diff
@@ -56,6 +56,11 @@ bool ConfigFile::Init(OptionsParser* options) {
   // Calculate the relative path of the config file.
   OptionsDict dict = options->GetOptionsDict();
   std::string filename = dict.Get<std::string>(kConfigFileStr);
+
+  // If filename is an empty string then return true.  This is to override 
+  // loading the default configuration file.
+  if (filename == "") return true;
```
SunnyWar: nit: consider:  if (filename.empty()) return true;

PR #305, opened 2018-08-25
Modify EncodePositionForNN to always fill in all history planes

dje-dev:
This PR is a reworked version of my earlier PR 253 (https://github.com/LeelaChessZero/lc0/pull/253), incorporating helpful suggestions to move the change into a lower level (encoder.cc) and also exploit the special case of en passant information. This work also builds on the wide discussion of this general issue in Discord with many contributors, stretching back quite a long time.

As noted in the earlier PR, this enhancements dramatically improves analysis of ad-hoc positions provided via FEN. For example, the score on Strategic Test Suite improves from 8689 to 12290.

Additionally, testing has shown this (slightly) improves performance of LC0 when self playing from the start position. In 6 separate tournaments of 1000 games (at 800 nodes per move), the modified version (slightly) outperformed every time, having an average win rate of 50.6%.

Because of the above, this enhancement could seemingly be incorporated immediately and prove beneficial even before retrained networks are available.

There is also the potential that future training will proceed more quickly or effectively because this modification removes what is essentially highly noisy data from the learning process in the first few moves of each game.

To assist in review and validation, two text files are attached, showing the output from 4 tests cases before and after this PR (including dumping of full board history sent to the NN):
```
go nodes 0
position startpos moves e2e4 e7e6
go nodes 0
position fen 2r3k1/1q1nbppp/r3p3/3pP3/pPpP4/P1Q2N2/2RN1PPP/2R4K b - b3 0 23
go nodes 0
position fen 2r4k/2rn1ppp/p1q2n2/PpPp4/3Pp3/R3P3/1Q1NBPPP/2R3K1 w - b6 0 23
go nodes 0
```
[before.txt](https://github.com/LeelaChessZero/lc0/files/2321100/before.txt)
[after.txt](https://github.com/LeelaChessZero/lc0/files/2321099/after.txt)


dje-dev:
Two further matches of 2000 games each were run, using slightly different temperature settings in the first two moves (to inject variety). The resulting points won by the version with this PR applied were 49.7% and 49.9%. Thus although the total evidence still slightly favors the PR, non-inferiority seems probably the best that we can claim.

src/neural/encoder.cc
```diff
@@ -54,32 +74,35 @@ InputPlanes EncodePositionForNN(const PositionHistory& history,
...
-    result[base + 8].mask = (board.theirs() * board.bishops()).as_int();
-    result[base + 9].mask = (board.theirs() * board.rooks()).as_int();
-    result[base + 10].mask = (board.theirs() * board.queens()).as_int();
-    result[base + 11].mask = (board.their_king()).as_int();
-
-    const int repetitions = position.GetRepetitions();
-    if (repetitions >= 1) result[base + 12].SetAll();
+    for (int i = 0; i < fill_in_positions; i++) {
+      SetInputPlanesFromBoard(result, history_positions + i, fill_in_board,
+                              last_board_repititions);
```
Tilps: Is there any scenario where last_board_repetitions would actually be non-zero?  FEN can't encode repetitions, and if we have the actual moves to show its a repetition then the history will be filled. I'm wondering if its clearer to just force them to 0.

borg323:
I am a bit uneasy with filling the history before the start position, unless this is done during training as well. 

It should be fairly easy to add a check of the bitboard against the start position mask (0xc3c3c3c3c3c3c3c3 unless I got this wrong), or separately white and black (also rooks if we are worried for knight/rook transpositions) and stop filling more than one plane. This check would be easy to disable for the next training run, if we also train with this.

Tilps:
My main thought for this PR is that it should be behind a flag and default off for training until we decide we want it for training. (I think its quite probably a good change for training, but unlikely to be massively significant either way.)

Cyanogenoid:
Personally I would like to see this in training for the upcoming reset. It seems like a change that should easily be beneficial to training and unlikely to cause issues.
One potential issue that I can think of is that a net trained with this might rely on history more because it doesn't see positions without history in training anymore. I don't think that's not much of a problem since it would always get history with this PR anyway, though this might justify raising cache-history-length in UCI mode.

roy7:
After seeing ASilver discuss a set of tests he had on Discord I offered to run them both with and without this PR. This was with 10 seconds per move on a 1080TI. It was a 200 position EPD file (WACsilver200.epd).

/master results:
```
108 of 200 matching moves
30-8-18 01:02:37, Total time: 00:37:34
Rated time: 16:49 = 1009 Seconds
```

/master with this PR results:
```
185 of 200 matching moves
30-8-18 00:19:32, Total time: 00:37:35
Rated time: 03:29 = 209 Seconds
```

ASilver:
I tested this thoroughly in WAC with history and without and compared and the result was perfect. If there is no evidence that it is in any way worse in play, I would recommend it as the default setting. As to the opening play, as of two moves, in my experience, the play is identical to full history, so the worry here is 2 moves. I don't see why seeing how it would play the opening moves with history can ever be wrong. If it starts suggesting 1. b4 as the top choice, then the concern might have merit.

oscardssmith:
@ASilver are you saying that with fill-in history provides no benefit? Otherwise, what does perfect mean?

oscardssmith:
Given that fake histories do not improve ability any more and we are about to start training a new net, is there any reason to not disable history for the new run to train it to not need history?

Sciolto66:
Would be nice if this PR is reviewed and merged.

Tilps:
I'll reiterate - I'm not going to approve this PR unless the change is behind a flag and default off for selfplay mode. (Someone else might have other opinions, but that is my criterion.)

oscardssmith:
@Tilps why? We have evidence it produces better play, why wouldn't we use it?

ddobbelaere:
@oscardssmith If you merge this into master and it is used in selfplay, you have to reset the training run basically, release new version and enforce that all clients use it in next run I guess. It's a pretty significant change during the first few opening moves.

oscardssmith:
I understand it's an intuitively big chance, but why do we think it wouldn't learn the right thing on the fly?

ASilver:
I honestly don't think this is a big deal. For one thing, the new
insta-history only affects the first 2 moves or so if at all, the rest will
not be affected. Second of all, the PR literally does not have a single
downside that I can think of, only upsides. It is a fact that the NN cannot
analyze properly without history, so saying that its first 2 opening moves
will now be analyzed as if they had history can only be a good thing, no?

Albert

Tilps:
The first 2 moves make up 5% of training samples - a massive change there is going to be significant overall.

Any change that could be reasonably expected to affect training should be behind a flag so that we can a) control when to turn it on, and b) turn it off when everything goes horribly wrong for unexpected reasons.

ASilver:
Three things to note:

1) This PR could be used only for the playing client, and not used for
training.

2) A much easier test would be to simply run it at 800 nodes (and more) and
compare the two to see if there is indeed a big change in the opening
moves. I'd test this on just a few, since if nothing stands out, there is
no reason to suspect anything exceptional.

3) Run matches between two builds at 800 nodes, same settings, one with PR
and one without and see. If the version with the PR suffers for some
reason, you have information right there.

Tilps:
On for playing client and not for training is exactly what I'm asking for - adding a flag which is default off for selfplay mode.  It can be default on for uci mode as far as I am concerned.

dje-dev:
Thanks for all the comments,  I'm trying to better understand the concerns here....

Regarding ASilver's comment 3, I did report the result of running 10,000 games at 800 nodes with versus without the PR (first comment above), and the result (insignificantly) favored the PR. Independent verification would of course be welcomed. 

Am I correct that @Tilps concern is that introducing this now into training games would be create an undesirable confounding variable in the current run? Therefore even if it indeed tested better and seemed superior based on theory, we should wait? I do see the logic in that. (Though one could also argue we are still in somewhat exploratory stages of the project and this particular network, and perhaps there is a risk to being too change averse.)

In any case, my main hope is change makes it into training runs before too long (perhaps not this one) because it corrects a flaw (subset of training data is highly incongruous with the distribution of the true space) that could confuse the network and impede learning (admittedly conjectural). Is there agreement this is the ultimate goal, and can we envision the mechanics of how that would eventually happen?

cwbriscoe:
I posted this example a couple of times in Discord but it is probably worth mentioning in this PR as well.  Using the A1A1 null move hack is not sound because it places a "phantom" knight on A1.  Hopefully this can be merged after addressing Tilps concerns.  Here is an example of the "phantom" knight:

```
position fen 2r4k/R6P/8/2K5/8/8/8/8 w - - 0 0 moves a1a1 a1a1
>go nodes 100
<info depth 1 seldepth 2 time 724 nodes 5 score cp -27 hashfull 0 nps 6 tbhits 0 pv c5b5 a1c2
<info depth 2 seldepth 3 time 1079 nodes 16 score cp -54 hashfull 0 nps 14 tbhits 0 pv c5d6 a1b3 d6e5
<info depth 3 seldepth 4 time 1542 nodes 48 score cp -52 hashfull 0 nps 31 tbhits 0 pv c5d5 a1c2 d5e4 c2b4
<info depth 3 seldepth 5 time 2006 nodes 80 score cp -50 hashfull 0 nps 39 tbhits 0 pv c5d6 a1c2 a7c7 c8d8 d6e7
<info depth 4 seldepth 6 time 2469 nodes 113 score cp -51 hashfull 0 nps 45 tbhits 0 pv c5d6 a1b3 d6e5 b3d2 e5f6 d2e4
<info depth 4 seldepth 6 time 2469 nodes 113 score cp -51 hashfull 0 nps 45 tbhits 0 pv c5d6 a1b3 d6e5 b3d2 e5f6 d2e4
<bestmove c5d6 ponder a1b3
```

src/chess/board.cc
```diff
@@ -449,6 +449,39 @@ bool ChessBoard::ApplyMove(Move move) {
...
 
+bool ChessBoard::UndoMoveToPriorBoardIfPossible() {
+  // Check if any en passant bits set, making undo possible.
+  const uint64_t ep_bits_right = pawns_.as_int() & 255;
+  const uint64_t ep_bits_left = pawns_.as_int() >> 56;
+  if (!ep_bits_right && !ep_bits_left) return false;
+
+  if (ep_bits_right) {
+    int ep_column = 0;
+    while (ep_column < 8 && !(ep_bits_right & (1 << ep_column))) {
```
mooskagh: Factor out GetLowestBit() from BitIter::Operator*, and use it.

src/neural/encoder.cc
```diff
@@ -36,6 +36,26 @@ const int kPlanesPerBoard = 13;
...
+  planes[base + 7].mask = (board.their_knights()).as_int();
+  planes[base + 8].mask = (board.theirs() * board.bishops()).as_int();
+  planes[base + 9].mask = (board.theirs() * board.rooks()).as_int();
+  planes[base + 10].mask = (board.theirs() * board.queens()).as_int();
+  planes[base + 11].mask = (board.their_king()).as_int();
+
+  if (repetitions >= 1) planes[base + 12].SetAll();
+}
+
 InputPlanes EncodePositionForNN(const PositionHistory& history,
```
mooskagh: Instead of making a copy of a board and modifying it, do something like:
(untested)
```cpp
InputPlanes EncodePositionForNN(const PositionHistory& history,
                                int history_planes) {
  InputPlanes result(kAuxPlaneBase + 8);

  {
    const ChessBoard& board = history.Last().GetBoard();
    const bool we_are_black = board.flipped();
    if (board.castlings().we_can_000()) result[kAuxPlaneBase + 0].SetAll();
    if (board.castlings().we_can_00()) result[kAuxPlaneBase + 1].SetAll();
    if (board.castlings().they_can_000()) result[kAuxPlaneBase + 2].SetAll();
    if (board.castlings().they_can_00()) result[kAuxPlaneBase + 3].SetAll();
    if (we_are_black) result[kAuxPlaneBase + 4].SetAll();
    result[kAuxPlaneBase + 5].Fill(history.Last().GetNoCaptureNoPawnPly());
    // Plane kAuxPlaneBase + 6 used to be movecount plane, now it's all zeros.
    // Plane kAuxPlaneBase + 7 is all ones to help NN find board edges.
    result[kAuxPlaneBase + 7].SetAll();
  }

  bool flip = false;
  int history_idx = history.GetLength() - 1;
  for (int i = 0; i < std::min(history_planes, kMoveHistory);
       ++i, --history_idx) {
    if (history_idx < 0 && !fill_empty_history) break;
    const Position& position =
        history.GetPositionAt(history_idx < 0 ? 0 : history_idx);
    const ChessBoard& board =
        flip ? position.GetThemBoard() : position.GetBoard();

    const int base = i * kPlanesPerBoard;
    result[base + 0].mask = (board.ours() * board.pawns()).as_int();
    result[base + 1].mask = (board.our_knights()).as_int();
    result[base + 2].mask = (board.ours() * board.bishops()).as_int();
    result[base + 3].mask = (board.ours() * board.rooks()).as_int();
    result[base + 4].mask = (board.ours() * board.queens()).as_int();
    result[base + 5].mask = (board.our_king()).as_int();

    result[base + 6].mask = (board.theirs() * board.pawns()).as_int();
    result[base + 7].mask = (board.their_knights()).as_int();
    result[base + 8].mask = (board.theirs() * board.bishops()).as_int();
    result[base + 9].mask = (board.theirs() * board.rooks()).as_int();
    result[base + 10].mask = (board.theirs() * board.queens()).as_int();
    result[base + 11].mask = (board.their_king()).as_int();

    const int repetitions = position.GetRepetitions();
    if (repetitions >= 1) result[base + 12].SetAll();

    if (history_idx < 0 && board.en_passant()) {
      const idx = GetLowestBit(board.en_passant);
      if (idx < 8) {  // "Us" board
        result[base + 0].mask +
            ((0x0000000000000100ULL - 0x0000000001000000ULL) << idx);
      } else {
        result[base + 6].mask +
            ((0x0001000000000000ULL - 0x0000000100000000ULL) << (idx - 56));
      }
    }
    if (history_idx > 0) flip = !flip;
  }

  return result;
}
```

as diff:
```diff
diff --git a/src/neural/encoder.cc b/src/neural/encoder.cc
index d25f6c8..0784bcf 100644
--- a/src/neural/encoder.cc
+++ b/src/neural/encoder.cc
@@ -57,9 +57,10 @@ InputPlanes EncodePositionForNN(const PositionHistory& history,
   bool flip = false;
   int history_idx = history.GetLength() - 1;
   for (int i = 0; i < std::min(history_planes, kMoveHistory);
-       ++i, flip = !flip, --history_idx) {
-    if (history_idx < 0) break;
-    const Position& position = history.GetPositionAt(history_idx);
+       ++i, --history_idx) {
+    if (history_idx < 0 && !fill_empty_history) break;
+    const Position& position =
+        history.GetPositionAt(history_idx < 0 ? 0 : history_idx);
     const ChessBoard& board =
         flip ? position.GetThemBoard() : position.GetBoard();
 
@@ -80,6 +81,18 @@ InputPlanes EncodePositionForNN(const PositionHistory& history,
 
     const int repetitions = position.GetRepetitions();
     if (repetitions >= 1) result[base + 12].SetAll();
+
+    if (history_idx < 0 && board.en_passant()) {
+      const idx = GetLowestBit(board.en_passant);
+      if (idx < 8) {  // "Us" board
+        result[base + 0].mask +
+            ((0x0000000000000100ULL - 0x0000000001000000ULL) << idx);
+      } else {  // "Them" board
+        result[base + 6].mask +
+            ((0x0001000000000000ULL - 0x0000000100000000ULL) << (idx - 56));
+      }
+    }
+    if (history_idx > 0) flip = !flip;
   }
 
   return result;
```

borg323:
I'm not sure what is the best thing to do with <https://github.com/LeelaChessZero/lc0/blob/master/src/mcts/node.cc#L301>

src/mcts/params.cc
```diff
@@ -131,6 +140,11 @@ void SearchParams::Populate(OptionsParser* options) {
   options->Add<IntOption>(kMaxCollisionVisitsId, 1, 1000000) = 1;
   options->Add<BoolOption>(kOutOfOrderEvalId) = false;
   options->Add<IntOption>(kMultiPvId, 1, 500) = 1;
+  std::vector<std::string> history_fill_opt;
+  history_fill_opt.push_back("no");
+  history_fill_opt.push_back("fen_only");
+  history_fill_opt.push_back("always");
+  options->Add<ChoiceOption>(kHistoryFillId, history_fill_opt) = "fen_only";
```
mooskagh: I think `{"no", "fen_only", "always"}` will work without need of a temporary vector.

src/neural/encoder.h
```diff
@@ -32,8 +32,11 @@
 
 namespace lczero {
 
+enum FillEmptyHistory {NO, FEN_ONLY, ALWAYS};
```
mooskagh: Better `enum class` than old style `enum` as enum puts pretty generic name `NO` into the global namespace.

src/neural/encoder.h
```diff
@@ -32,8 +32,11 @@
...
+
 // Encodes the last position in history for the neural network request.
 InputPlanes EncodePositionForNN(const PositionHistory& history,
-                                int history_planes);
+                                int history_planes,
+                                FillEmptyHistory fill_empty_history = NO);
```
mooskagh: In general it's not recommended to have default parameters. Styleguide disallows them, but we still use them sometimes. But here it doesn't seem to be justified case to use it.

src/mcts/params.cc
```diff
@@ -109,6 +109,15 @@ const OptionId SearchParams::kMultiPvId{
     "multipv", "MultiPV",
     "Number of game play lines (principal variations) to show in UCI info "
     "output."};
+const OptionId SearchParams::kHistoryFillId{
+    "history-fill", "HistoryFill",
+    "When to fill unknown history by repeating the earliest known position."};
+
+FillEmptyHistory SearchParams::GetHistoryFill() const {
```
mooskagh: It's better to have `kHistoryFill` of type `FillEmptyHistory` in order not to convert it every time.
To convert string to enum in constructor, some function in an anonymous namespace will work.

src/mcts/params.cc
```diff
@@ -109,6 +109,15 @@ const OptionId SearchParams::kMultiPvId{
     "multipv", "MultiPV",
     "Number of game play lines (principal variations) to show in UCI info "
     "output."};
+const OptionId SearchParams::kHistoryFillId{
+    "history-fill", "HistoryFill",
+    "When to fill unknown history by repeating the earliest known position."};
```
mooskagh: `Neural network uses 7 previous board positions in addition to the current one. During the first moves of the game such historical positions don't exist, but they can be synthesized. This parameter defines when to synthesize them (always, never, or only at non-standard fen position)`.

src/neural/encoder.cc
```diff
@@ -57,11 +58,16 @@ InputPlanes EncodePositionForNN(const PositionHistory& history,
...
-    if (history_idx < 0) break;
-    const Position& position = history.GetPositionAt(history_idx);
+       ++i, --history_idx) {
+    const Position& position =
+        history.GetPositionAt(history_idx < 0 ? 0 : history_idx);
     const ChessBoard& board =
         flip ? position.GetThemBoard() : position.GetBoard();
+    if (history_idx < 0 && fill_empty_history == NO) break;
+    if (history_idx < 0 && fill_empty_history == FEN_ONLY &&
+        (board.ours() + board.theirs()).as_int() == 0xFFFF00000000FFFFULL) {
```
mooskagh: In tests of chess960 positions, Leela shown that it didn't play optimally without history (e.g. played e4 when the best move would be `1. Ne3`). I think chess960 positions should be counted as "fen".
As well as different exotic positions like "knightmare", "startpos without castling rights", etc.
Or we can just enable it for all positions, and not distinguish fen from non-fen.

src/neural/encoder.cc
```diff
@@ -80,6 +86,18 @@ InputPlanes EncodePositionForNN(const PositionHistory& history,
 
     const int repetitions = position.GetRepetitions();
     if (repetitions >= 1) result[base + 12].SetAll();
+
+    if (history_idx < 0 && !board.en_passant().empty()) {
+      const auto idx = GetLowestBit(board.en_passant().as_int());
```
mooskagh: `// If en passant flag is set, undo last pawn move by removing the pawn from the new square and putting into pre-move square`

src/mcts/params.cc
```diff
@@ -109,6 +109,15 @@ const OptionId SearchParams::kMultiPvId{
...
+
+FillEmptyHistory SearchParams::GetHistoryFill() const {
+  if (kHistoryFill == "fen_only") return FEN_ONLY;
+  if (kHistoryFill == "always") return ALWAYS;
+  return NO;
```
mooskagh: `assert(kHistoryFill == "no");`

src/mcts/params.cc
borg323: I get errors with that, made it
```
  std::vector<std::string> history_fill_opt {"no", "fen_only", "always"};
```

evalon32: I think that's the right thing to do. In absolute terms, a temporary vector is necessary simply because ChoiceOption's constructor requires one. In a perfect world, you wouldn't need to spell it out (the vector could be implicitly created from an initializer_list), but that doesn't work with std::make_unique (a known limitation).

borg323:
I thought EncodePositionForNN is always called under a lock. Am I missing something?

Tilps:
I hadn't traced all the possible paths - but a quick look suggests that it can be called by independent training games simultaneously as part of writing the training data.  So definitely for training if not for UCI flows.

src/mcts/params.cc
mooskagh: Ah, uniform initialization can interfere with `Add()`'s perfect forwarding, but for initializer list it should work..
Anyway, it's fine as it is.

PR #307, opened 2018-08-26
Exponential moving average of q updates

Cyanogenoid:
This PR is just for experimentation for now.

As proposed by Commander8 in the Discord, this uses a different weighting scheme for q updates. Their intention is to improve scaling for high node counts (1M+). In particular, this PR implements a bias-corrected exponential moving average instead of the arithmetic mean. This closely matches how the bias-corrected EMA is written in the [Adam optimiser](https://arxiv.org/abs/1412.6980). The bias correction makes it so that it is not biased towards 0 due to the initialisation of q to 0.

This adds the param --backpropagate-momentum in the range [0, 1): The higher the value, the slower q changes with new updates from v. 0 means that q = v, 1-eps should be close to an arithmetic mean, but it probably runs into numerical precision issues.

### DIFF
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -235,7 +235,8 @@ void Node::CancelScoreUpdate() { --n_in_flight_; }
 
 void Node::FinalizeScoreUpdate(float v, float momentum) {
   // Recompute Q.
-  q_ = momentum * q_ + (1 - momentum) * v / (1 - std::pow(momentum, static_cast<float>(n_)));
+  q_buff_ = momentum * q_buff_ + (1 - momentum) * v;
+  q_ = q_buff_ / (1 - std::pow(momentum, static_cast<float>(n_)));
   // If first visit, update parent's sum of policies visited at least once.
   if (n_ == 0 && parent_ != nullptr) {
     parent_->visited_policy_ += parent_->edges_[index_].GetP();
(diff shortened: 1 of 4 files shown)


mooskagh:
That was tried and was reverted here https://github.com/LeelaChessZero/lc0/pull/203 as it didn't really work.

Cyanogenoid:
This differs from #203 in that #203 is not a proper average. As Tilps pointed out in Discord, beta there simply has the effect of dividing the final q by beta (roughly). That means that if beta < 1, q can go outside -1 and 1 which doesn't seem sensible to me. The implementation here is a proper exponential moving average, so it's qualitatively different.

Commander8:
Values of 1 or sufficiently close to 1 appear to crash the engine.  .999 worked but I did not test exactly how many decimal places before it failed.

Tilps:
I would agree this is not the same as beta or gamma - but the comment referencing me is a little out of context - someone was proposing multiplying the whole term by a parameter, which would have definitely had no effect but to  change the scale of Q.  Beta was actually a bit different to that. (Although as n gets large its behaviour may still have had the odd effect mentioned.)

mooskagh:
Just to get more opinions, maybe @Videodr0me can also tell his opinion whether that's worthwhile to add?

Cyanogenoid:
Since there hasn't been interest in testing this and I only made this PR by request, I'm closing this.

PR #309, opened 2018-08-26
Have a separate thread which watches time and `stop` command.

mooskagh:
Have a separate thread which watches time and `stop` command to reduce delay of reporting the move.

Should fix timeout problems that we have.

Time between sending `stop` and getting the `bestmove`:

Before:
```
Cudnn:   30.27± 5.03ms
Blas:    36.30± 5.46ms
OpenCl: 136.30±52.34ms
```
After:
```
Cudnn:   0.14± 0.03ms
Blas:    0.49± 1.75ms
OpenCl:  0.47±1.65ms
```

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -275,6 +275,8 @@ NNCacheLock Search::GetCachedFirstPlyResult(EdgeAndNode edge) const {
 void Search::MaybeTriggerStop() {
   SharedMutex::Lock nodes_lock(nodes_mutex_);
   Mutex::Lock lock(counters_mutex_);
+  // Already responded bestmove, nothing to do here.
+  if (responded_bestmove_) return;
   // Don't stop when the root node is not yet expanded.
   if (total_playouts_ == 0) return;
   // If smart pruning tells to stop (best move found), stop.
@@ -458,7 +460,12 @@ EdgeAndNode Search::GetBestChildWithTemperature(Node* parent,
 
 void Search::StartThreads(size_t how_many) {
   Mutex::Lock lock(threads_mutex_);
-  while (threads_.size() < how_many) {
+  // First thread is a watchdog thread.
+  if (threads_.size() == 0) {
+    threads_.emplace_back([this]() { WatchdogThread(); });
+  }
+  // Start working threads.
+  while (threads_.size() <= how_many) {
     threads_.emplace_back([this]() {
       SearchWorker worker(this);
       worker.RunBlocking();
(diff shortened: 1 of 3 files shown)


killerducky:
I tested this PR with the command line below. Setting Aversion to 0 means we will use all our time instead of exiting earlier. Before this PR I got ~75% of games lost on time. Now I didn't see any timeouts. I can see we get into a steady state of having 1899ms left, TC code allocates 1899-100=1799ms to think. We overshoot that by 1ms, again leaving us with 1899ms after the 1800ms increment is applied.

`./cutechess -repeat -srand 1535100190 -resign movecount=3 score=400 -draw movenumber=40 movecount=8 score=10 -concurrency 1 -games 250 -recover -wait 10 -engine cmd=Engines/pr309 proto=uci tc=1.0+1.8 option."Number of worker threads"=2 option."Network weights file path"=/home/aolsen/projects/OpenBench/OpenBench/Client/Networks/10925 option."Aversion to search if change unlikely"=0 name=test -engine cmd=Engines/pr309 proto=uci tc=1.0+1.8 option."Number of worker threads"=2 option."Network weights file path"=/home/aolsen/projects/OpenBench/OpenBench/Client/Networks/10925 option."Aversion to search if change unlikely"=0 name=base -openings file=8moves_v3.pgn format=pgn order=random plies=16 -debug | tee timeout6.log`

src/mcts/search.cc
```diff
@@ -458,26 +460,43 @@ EdgeAndNode Search::GetBestChildWithTemperature(Node* parent,
...
+  return !stop_;
+}
+
+void Search::WatchdogThread() {
+  while (IsSearchActive()) {
+    {
+      using namespace std::chrono_literals;
+      constexpr auto kMaxWaitTime = 100ms;
+      Mutex::Lock lock(counters_mutex_);
+      auto remaining_time = (limits_.time_ms - GetTimeSinceStart()) * 1ms;
```
Tilps: Should this check time_ms > 0, or do a max to ensure remaining_time is not less than 0?

src/mcts/search.cc
```diff
@@ -458,26 +460,43 @@ EdgeAndNode Search::GetBestChildWithTemperature(Node* parent,
...
+  // Start working threads.
+  while (threads_.size() <= how_many) {
     threads_.emplace_back([this]() {
       SearchWorker worker(this);
       worker.RunBlocking();
     });
   }
 }
 
-void Search::RunSingleThreaded() {
```
Tilps: What happened to RunSingleThreaded (how does lc0 still link after this change)?

src/mcts/search.cc
Tilps: Or maybe even protect against 0 to ensure this method can't be a hard spin which dominates the lock and makes it slower for the first playout to complete in the worst case?

mooskagh:
PTAL

mooskagh:
Hold on with review for now.
Even though it works, it shouldn't it seems. I still have to call notify_all() even if I hold the same mutex that protects the variable.

mooskagh:
Ready for review again.

killerducky:
I tested this again with my cutechess command above, it still works.

Trying to understand better: WatchdogThread is looking for one of two conditions: 1) time ran out, or 2) stop_ variable changed from false to true. But to me it seems easier to only do 1 and let FireStopInternal call MaybeTriggerStop directly to handle 2? Then you don't need a fancy std::condition_variable.

dubslow:
This could have undesired consequences since it leaves computation running (for several hundred ms) after `bestmove` output. In some cases (such as TCEC's nominal setup) this could lead to lc0 competing with other engines for the same hardware and thus lead to disqualification.

Also I think it's a serious indictment of the current concurrency model that such an ugly kludge as this -- an entirely new, dedicated thread whose only purpose is to make one syscall and then exit -- is necessary. (Same thing goes for having a dedicated garbage collection thread, whose only purpose is to deconstruct objects. At least it does that more than once though, unlike this.)

mooskagh:
@killerducky Option 2) is to respond immediately to uci `stop` command rather than wait for search threads to finish iteration.

PR #310, opened 2018-08-27
Very Verbose Move Stats

gsobala:
I've locally added an option to display very verbose move stats (outputting stats every time uci info is sent). This is quite useful in command-line mode in watching how leela thinks and distributes visits amongst moves in a dynamic fashion. Currently I have just added another command-line option ``--very-verbose-move-stats`` as a separate flag to ``--verbose-move-stats`` but I can see that as being potentially quite stupid and I wonder what the consensus view is on how this flag should be implemented?


gsobala:
I have renamed the flag to ``--high-stat-frequency`` and now it modifies the behaviour of the ``verbose-move-stats`` flag

gsobala:
Now obsolete

PR #313, opened 2018-08-27
First commit for next release

Cyanogenoid:
Dubslow forgot this when the version was bumped to 0.17.0.


dubslow:
I did indeed forget! And I didn't realize it til after my computer was off...

PR #315, opened 2018-08-28
Create executables without the popcnt instruction

borg323:
This adds an alternative implementation of the bitboard count() method that isn't using the popcnt instruction and meson support for it, specifically a popcnt binary option, default true to use the popcnt instruction as usual, or build the lc0_nopopcnt executable if false. Appveyor uses this to package both lc0.exe and lc0_nopopcnt.exe in the release zip files.

### DIFF
--- a/src/chess/bitboard.h
+++ b/src/chess/bitboard.h
@@ -96,10 +96,18 @@ class BitBoard {
   std::uint64_t as_int() const { return board_; }
   void clear() { board_ = 0; }
   int count() const {
+#ifdef NO_POPCNT
+    std::uint64_t x = board_;
+    x -= (x >> 1) & 0x5555555555555555;
+    x = (x & 0x3333333333333333) + ((x >> 2) & 0x3333333333333333);
+    x = (x + (x >> 4)) & 0x0F0F0F0F0F0F0F0F;
+    return (x * 0x0101010101010101) >> 56;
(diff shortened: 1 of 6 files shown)


DanielUranga:
Had not seen this and created: https://github.com/LeelaChessZero/lc0/pull/327 which is quite similar, maybe we can work on merging both PRs in a single one?

src/chess/bitboard.h
```diff
@@ -96,10 +96,18 @@ class BitBoard {
...
   void clear() { board_ = 0; }
   int count() const {
+#ifdef NO_POPCNT
+    std::uint64_t x = board_;
+    x -= (x >> 1) & 0x5555555555555555;
+    x = (x & 0x3333333333333333) + ((x >> 2) & 0x3333333333333333);
+    x = (x + (x >> 4)) & 0x0F0F0F0F0F0F0F0F;
+    return (x * 0x0101010101010101) >> 56;
+#else
 #ifdef _MSC_VER
```
mooskagh: There is `#elifdef`, but I don't know if it's standard enough to be supported by all compilers.
What is standard though is `#if defined NO_POPCNT` or `#if defined(NO_POPCNT)`

borg323:
The simpler version (#341) was merged so closing this.

PR #324, opened 2018-09-01
Update generic Linux instructions

oscardssmith:
Makes reqs be based on desired goal rather than library names.


oscardssmith:
Thanks

Issue #297, opened 2018-08-23
UCI option for syzygy not standard


killerducky:
The defacto standard is "SyzygyPath". We have "List of Syzygy tablebase directories".

cwbriscoe:
Are suggesting just changing the name or also not allowing multiple paths?

mooskagh:
Blocked by https://github.com/LeelaChessZero/lc0/issues/353

PR #326, opened 2018-09-01
Fix non-standard and too long UCI option

zeFresk:
As mentioned by @killerducky in #297, the standard option's name for syzygy directory should be "SyzygyPath" even if multiple directories are possible (and should be separated by a semicolon).

It's a very (very) small change but currently the string is so long that it doesn't fit in Arena.

### DIFF
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -52,7 +52,7 @@ const char* kMoveOverheadStr = "Move time overhead in milliseconds";
 const char* kTimeCurvePeak = "Time weight curve peak ply";
 const char* kTimeCurveRightWidth = "Time weight curve width right of peak";
 const char* kTimeCurveLeftWidth = "Time weight curve width left of peak";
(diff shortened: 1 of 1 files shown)


mooskagh:
I agree that Uci options have to be changed so shorter and possibly spaceless.
But as they are also help text when `--help` is specified, I think we should separate them first. (Have separate text as help and as Uci option). Also this will make it possible to hide certain params from Uci and make them command flag only.

zeFresk:
Yes I agree a separation is needed, but I don't think it's the number 1 priority at the moment. 
I submitted this quick patch because half the sentence was cut in arena and someone mentioned it in the issues. It was a good way for me to get used to the differents modules too.

If the change is too small to justify a merge I may try to separate the two later but right now I would rather experiment a bit with time control which is IMO a bigger problem.

mooskagh:
Should be done after https://github.com/LeelaChessZero/lc0/issues/353 is implemented.

PR #327, opened 2018-09-01
Add Win32-compatible version of "popcount" and "bitscan"

DanielUranga:
Add win32-compatible versions of those functions so it doesn't fail to compile when using 32 bits Visual Studio.

### DIFF
--- a/src/chess/bitboard.h
+++ b/src/chess/bitboard.h
@@ -97,7 +97,11 @@ class BitBoard {
   void clear() { board_ = 0; }
   int count() const {
 #ifdef _MSC_VER
+#ifdef _WIN64
     return _mm_popcnt_u64(board_);
+#else
+    return __popcnt(board_) + __popcnt(board_ >> 32);
+#endif
 #else
     return __builtin_popcountll(board_);
 #endif
--- a/src/chess/bitboard.h
+++ b/src/chess/bitboard.h
@@ -96,12 +96,10 @@ class BitBoard {
   std::uint64_t as_int() const { return board_; }
   void clear() { board_ = 0; }
   int count() const {
-#ifdef _MSC_VER
-#ifdef _WIN64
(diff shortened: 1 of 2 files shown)


src/utils/bititer.h
```diff
@@ -44,9 +44,20 @@ class BitIterator {
   void operator++() { value_ &= (value_ - 1); }
   T operator*() const {
 #ifdef _MSC_VER
+#ifdef _WIN64
     unsigned long result;
     _BitScanForward64(&result, value_);
     return result;
+#else
+    unsigned long result;
+    if (result & 0xFFFFFFFF) {
```
borg323: This should be `value_ & 0xFFFFFFFF`

DanielUranga: Whoops!

mooskagh:
Not commenting anything else, it's better to have
```cpp
#elif defined(BLAH)
```
instead of nested `#else #ifdef BLAH #endif #endif`s

src/chess/bitboard.h
```diff
@@ -97,7 +97,11 @@ class BitBoard {
   void clear() { board_ = 0; }
   int count() const {
 #ifdef _MSC_VER
+#ifdef _WIN64
```
borg323: Can you rearrange it to avoid nested #ifdef blocks?

src/utils/bititer.h
```diff
@@ -44,9 +44,20 @@ class BitIterator {
   void operator++() { value_ &= (value_ - 1); }
   T operator*() const {
 #ifdef _MSC_VER
+#ifdef _WIN64
```
borg323: Here as well

DanielUranga:
I see this is approved, but I can't merge. For some reason all of my PRs always have CircleCI in "pending" status.

PR #337, opened 2018-09-04
fix move selection bug at low temperatures

kiilas:
the bug was caused by all cumulative sum elements being 0 if temperature was low enough

fixed by scaling the elements so that their maximum will always be 1.0

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -453,7 +453,7 @@ EdgeAndNode Search::GetBestChildWithTemperature(Node* parent,
   assert(parent->GetChildrenVisits() > 0);
   std::vector<float> cumulative_sums;
   float sum = 0.0;
-  const float n_parent = parent->GetN();
+  uint32_t max_n = 0;
 
   for (auto edge : parent->Edges()) {
     if (parent == root_node_ && !root_limit.empty() &&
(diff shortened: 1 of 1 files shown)


kiilas:
the bug can realistically occur even at relatively high temperatures and thus affects functioning in normal scenarios

for example, at temperature=0.1 with 5 equally best moves (by N), the exponentiation would fall under FLT_EPSILON and the bug would get triggered

mooskagh:
Was this seen in the wild? How urgent is that PR?

killerducky:
kiilas said he hit this while doing some experiments that included other code changes. I don't think there is a confirmed sighting in normal use cases. But it maybe it could happen in matches where we set tempdecay?

kiilas:
I'd guess it's urgent to fix bugs that cause unexpected behaviour when using reasonable parameters, but unfortunately I don't have the time to hunt for instances of it appearing under any particular settings.

The bug will eventually occur if temperature (or initial_temperature / tempdecay_moves) < 0.02 [give or take], such as on the demo website last time I checked (temp=0.1, moves=10 IIRC).

If for whatever reason you don't want to fix it, please at least consider asserting against it (sum > 0).

PR #321, opened 2018-08-31
[WIP] Cpuct and FPU pow exponent scaling

jjoshua2:
Needs testing to find elo gain, but several tests have shown that default 0.5 power is not optimal. Probably something like 0.55 or 0.6 is good instead with a lower cpuct of 2 or 1 respectively.

PR #339, opened 2018-09-05
[WIP] Tune scaling

jjoshua2:
ZZ did the original one of these but had some bugs so I fixed them hopefully. Needs retesting now.


jjoshua2:
#321 should be referenced. But those tests are only helpful for knowing puct at different node counts when pow is 0.5.

jjoshua2:
I tested around 100k nodes
```
Score of lc0 scale 0.386 vs lc0 17.11250: 7 - 12 - 59 [0.468]
Elo difference: -22.30 +/- 37.99
```
another test with also FPU changes from ZZ clop tune
```
Score of lc0 scale 0.386 vs lc0 default: 1 - 4 - 8 [0.385]
Elo difference: -81.65 +/- 119.97
```

oscardssmith:
Close now that we adjust by cpuct by log(n)?

Issue #304, opened 2018-08-25
Avoid temperature moves with only one (or two) visits


gonzalezjo:
If this is considered a valid issue, I'm willing to write a PR. 

Leela Zero does not play "temperature moves" that only have one visit. This should prevent some of the more egregious moves from entering training game data.

killerducky:
Some earlier related discussion (OP of this issue suggests we do the opposite of what you suggest)
https://github.com/glinscott/leela-chess/issues/559

gonzalezjo:
That’s... interesting. Definitely makes me rethink things. I’m firmly on the side of increasing exploration at almost all costs, but I’m also on the side that temperature can have some negative side effects. I wish we had more data. It’s hard for me to hold both beliefs.



My prior reasoning for this comes from this issue, where gcp discusses the impact of temperature on Leela Chess: https://github.com/gcp/leela-zero/issues/1355



> I can definitely point at positions where Leela Chess is sometimes doing badly because of high temperature, i.e. where it plays to drawn endings from a better position, only because the resulting positions makes it much more likely for the worse player to blunder. Something similar happens to giving very high evaluation to making some threats around the king, and that sometimes causes lost games.



And also: 



> The reasoning is that by selecting moves with 1 visit, you are actually neglecting the result of the tree search.



You commented on it, which was cool! So now I’m wondering what insights you might have on this, given that you’ve probably had it in your mind for w longer than I have.




We know that the Leelas need exploration. We also know that they benefit from having higher quality games to work work with. Both are *likely* to help raise the learning ceiling, but I guess it’s a tradeoff.

jjoshua2:
I think it is necessary for noise alone to be able to cause a move to be
considered and played in training occasionally.
But it also should not normally happen in the average endgame otherwise it
will expect that you can win merely by not blundering. Yes there is some
chance but it is a very small chance against even a mid level engine.

jjoshua2:
So the way I would balance this is to do enough visits where noise alone
gives something like 1/1000 of two visits which would mean over 50 moves
there is a small chance of a blunder and not allow moves with one or zero
visits. This is what leela go did.

gonzalezjo:
Do you think that’s worth the decrease in game rate? Are there any estimates on the game generation speedup that adaptive resign rate can provide?

dubslow:
[...]

PR #340, opened 2018-09-05
Do not play temperature moves with <n visits.

gonzalezjo:
This pull request introduces one new parameter. The intention of adding this parameter is to make things changeable via the server. It is a change that can be undone easily.

This addresses issue #304. 

I have no idea how to validate the behavior of this patch, I found GetBestChildWithTemperature to be confusing, and I wrote this after staying up for quite a while to watch Leela vs. Houdini in the CCCC. Therefore, it's more likely than not that this PR does *not* do what it says on the tin. Hopefully, someone can tell me if I did something really stupid here.

Edit: Chad suggested that I add clarification for what the parameter does. 

I actually don't know what it should do (there are conflicting opinions on this) but right now, it's set such that 0 is off, which is [how Leela Zero does it](https://github.com/gcp/leela-zero/pull/1376).

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -465,9 +465,9 @@ EdgeAndNode Search::GetBestChildWithTemperature(Node* parent,
       continue;
     }
 
-    if (edge.GetN() <= static_cast<unsigned int>(kMinimumTemperatureVisits)) {
+    if (edge.GetN() >= static_cast<unsigned int>(kMinimumTemperatureVisits)) {
       accepted_edges.push_back(edge);
-    } else {
+    } else { 
       all_edges.push_back(edge);
     }
   }
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -466,7 +466,7 @@ EdgeAndNode Search::GetBestChildWithTemperature(Node* parent,
             root_limit.end()) {
       continue;
     }
-    if (edge.GetN() < static_cast<unsigned int>(kMinimumTemperatureVisits))
+    if (edge.GetN() <= static_cast<unsigned int>(kMinimumTemperatureVisits))
       continue;
 
     sum += std::pow(edge.GetN() / n_parent, 1 / temperature);
@@ -484,7 +484,7 @@ EdgeAndNode Search::GetBestChildWithTemperature(Node* parent,
             root_limit.end()) {
       continue;
     }
-    if (edge.GetN() < static_cast<unsigned int>(kMinimumTemperatureVisits))
+    if (edge.GetN() <= static_cast<unsigned int>(kMinimumTemperatureVisits))
       continue;
     if (idx-- == 0) return edge;
   }
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -456,7 +456,7 @@ EdgeAndNode Search::GetBestChildWithTemperature(Node* parent,
   }
 
   assert(parent->GetChildrenVisits() > 0);
-  std::vector<EdgeAndNode> accepted_edges, all_edges;
+  std::vector<EdgeAndNode> accepted_edges, filtered_edges;
 
   for (auto edge : parent->Edges()) {
     if (parent == root_node_ && !root_limit.empty() &&
@@ -468,12 +468,12 @@ EdgeAndNode Search::GetBestChildWithTemperature(Node* parent,
     if (edge.GetN() >= static_cast<unsigned int>(kMinimumTemperatureVisits)) {
       accepted_edges.push_back(edge);
     } else { 
-      all_edges.push_back(edge);
+      filtered_edges.push_back(edge);
     }
   }
 
   if (accepted_edges.empty()) {
-    accepted_edges = all_edges;
+    accepted_edges = filtered_edges;
   }
 
   const float n_parent = parent->GetN();
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -493,7 +493,7 @@ EdgeAndNode Search::GetBestChildWithTemperature(Node* parent,
       std::lower_bound(cumulative_sums.begin(), cumulative_sums.end(), toss) -
       cumulative_sums.begin();
 
-  return accepted_edges[idx];
+  return accepted_edges.at(idx);
 }
 
 void Search::StartThreads(size_t how_many) {
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -470,9 +470,9 @@ EdgeAndNode Search::GetBestChildWithTemperature(Node* parent,
 
     if (edge.GetN() <= static_cast<unsigned int>(kMinimumTemperatureVisits)) {
       accepted_edges.push_back(edge);
+    } else {
+      all_edges.push_back(edge);
     }
-
-    all_edges.push_back(edge);
   }
 
   if (accepted_edges.empty()) {
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -465,7 +465,7 @@ EdgeAndNode Search::GetBestChildWithTemperature(Node* parent,
   // Load edges to vector.
   for (auto& edge : edge_iterator) edges.push_back(edge);
 
-  // Filter empty edges
+  // Filter edges that do not have enough visits.
   edges.erase(std::remove_if(edges.begin(), edges.end(),
                              [&](auto edge) {
                                return edge.GetN() <=
@@ -474,9 +474,10 @@ EdgeAndNode Search::GetBestChildWithTemperature(Node* parent,
                              }),
               edges.end());
 
-  // Avoids crash if no edges left.
-  if (edges.empty())
+  // Avoids crash if all edges have been filtered.
+  if (edges.empty()) {
     for (auto& edge : edge_iterator) edges.push_back(edge);
+  }
 
   for (auto edge : edges) {
     if (parent == root_node_ && !root_limit.empty() &&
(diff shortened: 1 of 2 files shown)


cn4750:
So, this is the equivalent of [this](https://github.com/gcp/leela-zero/pull/1376) Leela Zero PR?

gonzalezjo:
If it works, yup.

src/mcts/search.cc
```diff
@@ -77,6 +79,8 @@ void Search::PopulateUciParams(OptionsParser* options) {
   options->Add<FloatOption>(kCpuctStr, 0.0f, 100.0f, "cpuct") = 1.2f;
   options->Add<FloatOption>(kTemperatureStr, 0.0f, 100.0f, "temperature") =
       0.0f;
+  options->Add<IntOption>(kMinimumTemperatureVisitsStr, 0, 1000,
```
Tilps: The lower bound probably should be one here.

src/mcts/search.cc
```diff
@@ -77,6 +79,8 @@ void Search::PopulateUciParams(OptionsParser* options) {
...
+                          "minimum-temperature-visits") = 2;
```
Tilps: Set the default to 1 - we can then decide when to turn the feature on.

src/mcts/search.cc
sergiovieri: Isn't 0 for disabling this feature?

oscardssmith:
To me this setting would should be for <= n visits. That way 0 is off which seems more natural.

src/mcts/search.cc
```diff
@@ -461,6 +466,9 @@ EdgeAndNode Search::GetBestChildWithTemperature(Node* parent,
             root_limit.end()) {
       continue;
     }
+    if (edge.GetN() < static_cast<unsigned int>(kMinimumTemperatureVisits))
+      continue;
```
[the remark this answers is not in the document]
Mardak: Could probably initialize a local vector from `parent->Edges()` and filter, and if it's empty, restore to `parent->Edges()`. Then use it for generating `cumulative_sums` and random selection without needing to check min visits multiple times.

src/mcts/search.cc
gonzalezjo: I made some changes, based on @oscardssmith and @sergiovieri's comments. Zero seems like what off should be.

src/mcts/search.cc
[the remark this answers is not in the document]
gonzalezjo: Yikes, yeah, this is the sort of thing that I was afraid would slip. Thanks for catching it. @Mardak's proposal sounds reasonable. I will write an implementation. If anyone objects to that, I will also change the implementation.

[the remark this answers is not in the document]
gonzalezjo: Should be fixed. Thanks again for catching this!

gonzalezjo:
I *think* that I have fixed everything, but this definitely needs a lot of eyes looking over it. I don't trust myself a ton here, but it seems like it works.

src/mcts/search.cc
```diff
@@ -452,10 +457,29 @@ EdgeAndNode Search::GetBestChildWithTemperature(Node* parent,
...
+                               return edge.GetN() <=
+                                      static_cast<unsigned int>(
+                                          kMinimumTemperatureVisits);
+                             }),
+              edges.end());
+
+  // Avoids crash if all edges have been filtered.
+  if (edges.empty()) {
+    for (auto& edge : edge_iterator) edges.push_back(edge);
+  }
```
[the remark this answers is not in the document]
gonzalezjo: I tried initializing the vector from the iterator at one point, and after failing, tried using a combination of std::copy and std::copy_if. I failed that, too. I'm not really a C++ programmer, but I spent about two hours trying to find a cleaner way of doing things, and got nowhere. I'm super sorry for not being able to write better code than that.

Can you elaborate on this? 

> You can filter directly at after root limit check. Then later instead of checking root limit again and idx--, you can random access at index idx instead.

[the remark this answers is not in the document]
sergiovieri: After limit root check (still inside the loop) push into filterededges if <= n, and to edges else. If edges is empty, set edges = filterededges. Then at the second limit root check, you can entirely remove the for loop, and use edges[idx] instead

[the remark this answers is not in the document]
gonzalezjo: That's genius. I will do that.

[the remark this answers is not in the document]
gonzalezjo: Okay, in an attempt to implement that, I've come to the conclusion that it requires maintaining an additional sum vector. Is there a way of avoiding that?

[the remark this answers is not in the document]
sergiovieri: make sum after choosing which edge vector to use

src/mcts/search.cc
sergiovieri: I just realised option is minimum temperature visits, not visits to ignore in temperature, so currently description is not consistent with code behaviour

src/mcts/search.cc
```diff
@@ -461,23 +467,30 @@ EdgeAndNode Search::GetBestChildWithTemperature(Node* parent,
...
+      accepted_edges.push_back(edge);
+    } else {
+      all_edges.push_back(edge);
+    }
+  }
+
+  if (accepted_edges.empty()) {
+    accepted_edges = all_edges;
+  }
+
```
sergiovieri: you can move cummulative_sums declaration here and directly specify the size to the constructor

sergiovieri: can also move sum and n_parent here

gonzalezjo: Just when I thought I was done... 


I *cannot* specify the size in the constructor, but I can reserve a size, which should be good enough.


Changes done, will push soon. Thank you again for checking these things.

src/mcts/search.cc
```diff
@@ -461,23 +467,30 @@ EdgeAndNode Search::GetBestChildWithTemperature(Node* parent,
             root_limit.end()) {
       continue;
     }
+
+    if (edge.GetN() <= static_cast<unsigned int>(kMinimumTemperatureVisits)) {
+      accepted_edges.push_back(edge);
```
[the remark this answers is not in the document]
gonzalezjo: Whoops. I actually didn't have either of those issues in earlier commits. Both were introduced during refactors. Thank god you caught them. Those are glaringly bad...

gonzalezjo:
Okay, I ran a very basic test on commit be02202, and I believe it to be stable. I logged cases in which `accepted_edges` was empty, and then checked to see when these cases occurred. It seems to occur at the right moments.

src/mcts/search.cc
```diff
@@ -451,33 +456,43 @@ EdgeAndNode Search::GetBestChildWithTemperature(Node* parent,
...
-        std::find(root_limit.begin(), root_limit.end(), edge.GetMove()) ==
-            root_limit.end()) {
-      continue;
-    }
-    if (idx-- == 0) return edge;
-  }
+  return accepted_edges[idx];
+
   assert(false);
   return {};
```
sergiovieri: this assert is now no longer needed

gonzalezjo: Done. I cleaned up the return values and replaced the assertion with something along the same spirit.

src/mcts/search.cc
```diff
@@ -451,35 +456,44 @@ EdgeAndNode Search::GetBestChildWithTemperature(Node* parent,
...
-    if (parent == root_node_ && !root_limit.empty() &&
-        std::find(root_limit.begin(), root_limit.end(), edge.GetMove()) ==
-            root_limit.end()) {
-      continue;
-    }
-    if (idx-- == 0) return edge;
-  }
-  assert(false);
-  return {};
+  return accepted_edges.at(idx);
```
Tilps: assert cumulative_sums and accepted_edges has the same size? (Its trivially true, but its an assumption this line is making.)

gonzalezjo: std::vector::at does bounds checking, so it should be the same thing as the assertion. Or am I misunderstanding?

Tilps: Right you are, I always forget since I never use at :P

src/mcts/search.cc
```diff
@@ -451,35 +456,44 @@ EdgeAndNode Search::GetBestChildWithTemperature(Node* parent,
...
+  const float toss = Random::Get().GetFloat(cumulative_sums.back());
+  const int idx =
       std::lower_bound(cumulative_sums.begin(), cumulative_sums.end(), toss) -
       cumulative_sums.begin();
 
-  for (auto edge : parent->Edges()) {
-    if (parent == root_node_ && !root_limit.empty() &&
-        std::find(root_limit.begin(), root_limit.end(), edge.GetMove()) ==
-            root_limit.end()) {
-      continue;
```
Mardak: This root_limit checking code was added in #293 and only touched here.. but @Tilps, wouldn't the visit count for root filtered moves be 0 anyway, so there shouldn't be a need to additionally filter for both with and without temperature best child selection?

Tilps: its true there shouldn't need to be a filter for the with temperature best child selection if searching is done with filtering - but without temperature its required as without temperature is supposed to work with only root node visited - thus filtering is needed to protect against the max policy edge may not be in the filter set.  Seems cleaner to just have the filtering applied consistently even if it does nothing.

gonzalezjo: So should I leave this alone?

gonzalezjo: So should I leave this alone?

oscardssmith:
Should this be closed? We basically have it implemented via temp offset

mooskagh:
I think temperature offset addressed this issue?

gonzalezjo:
Yup.

PR #343, opened 2018-09-06
add cudnn max_batch parameter

borg323:
After a discussion we had with @ankan-ban on discord about a month ago, this looks like the easiest way to reduce the memory requirements of the cudnn backend, for cards with limited memory.

### DIFF
--- a/src/neural/network_cudnn.cu
+++ b/src/neural/network_cudnn.cu
@@ -99,9 +99,6 @@ void CudaError(cudaError_t status, const char *file, const int &line) {
 #define ReportCUBLASErrors(status) CublasError(status, __FILE__, __LINE__)
 #define ReportCUDAErrors(status) CudaError(status, __FILE__, __LINE__)
 
-// Hard-coded for now, no point in going above this anyway (can possibly save
-// memory by reducing this).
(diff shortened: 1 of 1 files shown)


ankan-ban:
the changes look good to me.

gonzalezjo:
Using this, with these arguments: ./client_linux --backend-opts="cudnn(max_batch=5)"



I get this error: 


```
Creating backend [cudnn]...
terminate called after throwing an instance of 'lczero::Exception'
  what():  CUDA error: an illegal memory access was encountered (../../src/neural/network_cudnn.cu:1238) 
```

borg323:
@gonzalezjo do I understand correctly (from discord) that 256 fails only for matches? The minimum max_batch value should be at least equal to minibatch-size, so 256 should work. The error happens at the `cudnnCreate` call, so some kind of resource limit is likely.

gonzalezjo:
Correct, it only fails for matches. ./lc0 selfplay with batch size set to 256 is fine. The setting is definitely being read in ./lc0 selfplay, too, because if I can raise the number and force an out of memory error. Moreover, I can even use 128 in ./lc0 selfplay. It's just in matches that I get the access violation.

PR #346, opened 2018-09-09
Tree detaching

mooskagh:
Discussion:
https://groups.google.com/forum/#!topic/lczero-devs/6d1YacfiRyI


RedDenver:
I recommend using the default memory order for initial implementations until it has successfully been tested on multiple operating systems, then go back and add memory ordering like acquire and release and retest across multiple OS. Otherwise, those can be really hard bugs to track down.

src/mcts/node.h
```diff
@@ -418,6 +440,62 @@ class Node::NodeRange {
...
+
+  // Stores whether there is a worker assigned to this subtree.
+  std::atomic<bool> is_used_{false};
+
+  // Latest counters from subtree's root.
+  std::atomic<float> q_{0.0f};
+  // How many completed visits this node had.
+  std::atomic<uint32_t> n_{0};
+
+  std::atomic<uint32_t> parent_n_{0};
```
RedDenver: The constructor initializes the atomic values, so no need to also do it here.

hans-ekbrand:
I tested this PR in Amazon cloud account, where I had access to 8 x Tesla K80. Before I go into the results I must say that I was confused that there were no longer any visible differences in the output produced by lc0 with this PR and normal lc0. In the previous version of this PR I tested, I could easily see that the tree-detaching patch was active, since it said so in the output. That said, I'm pretty sure I actually ran a binary with this patch applied.

Unfortunately, the performance was worse than 'release' (v.17 still, I believe).

Actually, I was surprised that vanilla lc0 scaled so good as it did. The testing is not completed, I decided to stop and try to confirm that I was actually running the right thing, but here are the results.

Please tell me what you want tested. In this account I have up to 8 gpus and 32 CPUs

image

jjoshua2:
Apparently we don't have issues until 30k nps and 8 threads. One v100 gets
30k but two gets 40k?

hans-ekbrand:
I've finally got access to reasonable instance on AWS, and tested how lc0 v0.17.0 (current 'release') scales when given more gpu:s. The GPU here is Tesla K80, and I tested 1-8 GPU:s for v0.17.0 and 1, 4, 6, and 8 GPU:s for PR 346 'tree detaching'. This system has 32 cores Intel(R) Xeon(R) CPU E5-2686 v4 @ 2.30GHz, and each test used twice as many threads as GPU:s. Also, I used --minibatch-size=512 --nncache=2000000 (copied from the current CCCC config). I measured NPS from the startposition using go nodes 2000000. I have the log-files, if you want them. I was suprised over how well lc0 'release' actually scaled. As you can see, it scales linearly with number of GPU:s. Using this test, PR 346 didn't perform very well, unfortunately. If you have ideas about what I should test, please let me know.
lc0-scaling_per_gpu

hans-ekbrand:
@mooskagh I have a machine with 6 x GTX 1060 which you are welcome to develop and test this PR on. (Fersbery already has an account in this box), but I think there is room for both of you. If this sounds interesting, post your public ssh-key here or PM me in discord (@hanse).

mooskagh:
This attempt didn't work. If there will be a new attempt, if will be in a new PR.

PR #348, opened 2018-09-11
Fix data race in MuxingComputation::NotifyReady with heavy threading

TheAnswer:
notify_one in MutexComputing::NotifyReady in some cases ends up running after/while its object is being destroyed after SearchWorker::ExecuteOneIteration inside SearchWorker::InitializeIteration  by a different thread corrupting memory.

*Force dataready_cv_.notify_one(); to finish before the lock is released to a different thread
*Fixes Crash in multiplexed setup with blas, opencl and heavy threading #98

### DIFF
--- a/src/neural/network_mux.cc
+++ b/src/neural/network_mux.cc
@@ -62,10 +62,8 @@ class MuxingComputation : public NetworkComputation {
   }
 
   void NotifyReady() {
-    {
-      std::unique_lock<std::mutex> lock(mutex_);
-      dataready_ = true;
-    }
+    std::unique_lock<std::mutex> lock(mutex_);
+    dataready_ = true;
     dataready_cv_.notify_one();
   }
 


src/neural/network_mux.cc
```diff
@@ -62,10 +62,8 @@ class MuxingComputation : public NetworkComputation {
   }
 
   void NotifyReady() {
-    {
-      std::unique_lock<std::mutex> lock(mutex_);
-      dataready_ = true;
-    }
+    std::unique_lock<std::mutex> lock(mutex_);
+    dataready_ = true;
```
SunnyWar: why not make this atomic?

TheAnswer: dataready_cv_.notify_one() below notifies the waiter to continue and the problem would be back, this thread can be suspended in notify_one and the waiter thread given priority to continue destroying the object before notify_one finishes in the other thread.

PR #350, opened 2018-09-11
Use Q=0 for self-play following AZ paper behavior while keeping FPU reduction tuning

### LINKED ISSUE #344: Initialize Q = 0 instead of parent Q for self-play to match AGZ paper
From AGZ paper:
screen shot 2018-09-07 at 11 40 03 am

I realize initializing to parent Q is common and then additionally reducing with first play urgency is used by Leela Zero and copied to lczero/lc0, and doing that and tuning can improve match strength.

However, self-play training data quality can be reduced when using "match settings" leading to different default values for cpuct, softmax, fpu reduction, etc. as well as turning on/off code specifically for self-play, e.g., https://github.com/gcp/leela-zero/pull/1083

The behavior of Q = 0 instead of parent Q makes it so that in winning positions, the first found "good enough" move will likely have a dominating Q > 0 relative to unvisited Q = 0 moves. Similarly, from a losing position, search will naturally go wider as the usual "best" move probably still has a Q < 0 leading to visits of seemingly Q = 0 moves.

Almost all of the positions in #8 happen to be from losing positions with one good move, so Q = 0 ends up finding all of the expected tactical moves not even needing 800 visits as once search starts going wide, it realizes the "hidden"-very-low-prior tactical move is actually the best of all possible moves.

But even ignoring the fact that Q = 0 happens to improve tactical training in those select positions, it sounds like a main project goal is to "reproduce AZ as closely as possible." There are no details of Q = 0 or other Q in AZ paper itself, so falling back to the previous AGZ paper, it would seem to imply unvisited moves should have Q = 0 for self-play.

Mardak:
Fix #344. As documented in the issue, https://github.com/LeelaChessZero/lc0/issues/344#issuecomment-419722833 neither lc0 or lczero ever used Q=0 diverging from the learning behavior of AZ.

Looking even further back in lczero history into leela-zero, Q/FPU was originally set to 1.1 (win rate on a scale of [0, 1]) based on gcp's experience with leela, so it seems that never in leela-related history has Q=0 been used for generating self-play for training.

https://github.com/gcp/leela-zero/blob/2f7463d2cfba1b4617b3bd73bbdf3e1f52382429/UCTNode.cpp#L291

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -193,7 +193,7 @@ int64_t Search::GetTimeSinceStart() const {
 }
 
 void Search::SendMovesStats() const {
-  const float parent_q =
+  const float parent_q = kNoise ? 0.0f :
       -root_node_->GetQ() -
       kFpuReduction * std::sqrt(root_node_->GetVisitedPolicy());
   const float U_coeff =
@@ -712,9 +712,12 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend() {
         search_->kCpuct * std::sqrt(std::max(node->GetChildrenVisits(), 1u));
     float best = -100.0f;
     int possible_moves = 0;
+    // Initialize Q=0 for self-play games played with noise following AZ paper's
+    // behavior of searching wider for losing positions and deeper for winning
+    // positions. Allow tuning first play urgency of parentQ to optimize search.
     float parent_q =
-        ((is_root_node && search_->kNoise) || !search_->kFpuReduction)
-            ? -node->GetQ()
+        is_root_node && search_->kNoise
+            ? 0.0f
             : -node->GetQ() -
                   search_->kFpuReduction * std::sqrt(node->GetVisitedPolicy());
     for (auto child : node->Edges()) {


killerducky:
I think what we do now is an improvement over the paper, and there are other things I'd prefer to test before this.

Mardak:
For even more context, killerducky asked DeepMind to clarify FPU:
http://computer-go.org/pipermail/computer-go/2017-December/010550.html

Aja Huang responded:
http://computer-go.org/pipermail/computer-go/2017-December/010567.html
> All I can say is that first-play-urgency is not a significant technical detail, and what's why we didn't specify it in the paper.

One could infer that AGZ/AZ did not try different values for FPU/reduction, and just used Q=0 as documented in the AGZ paper to get good results for AZ chess. Additionally assuming DeepMind used Q=0 for match games, potentially this search inefficiency was overcome by just having significantly higher visits than the 800 used for self-play.

As I noted in the code comments, searching wider for losing positions and deeper for winning positions allows for two different learning targets for future networks trained on those self-play games. The current behavior of using parentQ seems to be optimized for match strength leading to just one learning target at the cost of the inability to learn some types of (tactical) moves.

Mardak:
Just to be clear for your analysis, did you do FPU 0 for all nodes or only root nodes? Non-root FPU favoring wider search ends up skewing accurate NN+search average action value towards losing.

Tilps:
hmm looks like I may have indeed accidentally applied it to all nodes - will retry!

Tilps:
Tested again with 0 only for root node - same outcome of 0 mapping to more than 3 visits at 8000, maybe even worse that with using 0 all the time although my new dataset isn't especially big yet, so the difference is quite possibly in the noise.

PR #351, opened 2018-09-14
Remove --sticky-checkmate flag.

mooskagh:
Noone used that, usefulness not proven.


oscardssmith:
Didn't we have data showing it performing strictly better?

mooskagh:
No, we had data that it performed slightly worse but within error margins so it was declared "neutral".
E.g. https://github.com/LeelaChessZero/lc0/pull/211#issuecomment-408633752

PR #352, opened 2018-09-14
Enable out of order eval by default. Make cache history length 0 by d…

mooskagh:
…efault.

That is only for UCI mode. For self play it's as it used to be.


killerducky:
Some tests of this (using UCI option):
http://ec2-34-217-73-2.us-west-2.compute.amazonaws.com:8000/viewTest/6/
TC=30.0+0.1s
19.17 +- 9.39

http://ec2-34-217-73-2.us-west-2.compute.amazonaws.com:8000/viewTest/11/
TC=180+1.8s  
26.78 +- 11.61

Issue #299, opened 2018-08-23
Error on bad UCI parameter name


killerducky:
We do not give any error message when doing e.g. "setoption badoptionname value 0".

mooskagh:
That used to work, it's a bug. It should throw an exception which is caught in UCI loop.

PR #355, opened 2018-09-15
Fix Error on bad setoption UCI parameter name #299

TheAnswer:
Fix Error on bad setoption UCI parameter name #299

### DIFF
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -339,7 +339,10 @@ void EngineLoop::CmdIsReady() {
 
 void EngineLoop::CmdSetOption(const std::string& name, const std::string& value,
                               const std::string& context) {
-  options_.SetOption(name, value, context);
+  if (!options_.SetOption(name, value, context)) {
+    throw Exception("No such option: " + name);
+  }
+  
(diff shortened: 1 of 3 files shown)


src/engine.cc
```diff
@@ -339,7 +339,10 @@ void EngineLoop::CmdIsReady() {
 
 void EngineLoop::CmdSetOption(const std::string& name, const std::string& value,
                               const std::string& context) {
-  options_.SetOption(name, value, context);
+  if (!options_.SetOption(name, value, context)) {
+    throw Exception("No such option: " + name);
```
mooskagh: There is another similar PR: https://github.com/LeelaChessZero/lc0/pull/246
which returns false for SetOption on bad parameter value.
I think we at least have to have a plan how to support both. (Here if SetOption returns false, it will be not clear whether it's bad name or bad value).
I like throwing exceptions directly from inside of SetOption(), not sure if others consider that ugly. :)

mooskagh:
I think #246 makes this obsolete.

TheAnswer:
Yep its fixed there

PR #356, opened 2018-09-16
WIP: Filter nps

borg323:
This uses a simple low pass filter for the nps value to avoid rapid changes. Just for testing.

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -315,6 +315,13 @@ void Search::UpdateRemainingMoves() {
       auto nps = (1000LL * total_playouts_ + kSmartPruningToleranceNodes) /
                      (time_since_start - kSmartPruningToleranceMs) +
                  1;
+      if (nps_ < 0) {
+        nps_ = nps;
+      } else {
+        // Simple IIR low pass filter: H(z)=0.5z^-1/(1-0.5z^-1)
+        nps_ = 0.5 * (nps + nps_);
+        nps = nps_;
+      }
       int64_t remaining_time = limits_.time_ms - time_since_start;
       // Put early_exit scaler here so calculation doesn't have to be done on
       // every node.
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -312,7 +312,7 @@ void Search::UpdateRemainingMoves() {
   if (limits_.time_ms >= 0) {
     auto time_since_start = GetTimeSinceStart();
     if (time_since_start > kSmartPruningToleranceMs) {
-      auto nps = (1000LL * total_playouts_ + kSmartPruningToleranceNodes) /
+      float nps = (1000.0f * total_playouts_ + kSmartPruningToleranceNodes) /
                      (time_since_start - kSmartPruningToleranceMs) +
                  1;
       if (nps_ < 0) {
--- a/src/mcts/search.h
+++ b/src/mcts/search.h
@@ -172,6 +172,7 @@ class Search {
   int64_t total_playouts_ GUARDED_BY(nodes_mutex_) = 0;
   int remaining_playouts_ GUARDED_BY(nodes_mutex_) =
       std::numeric_limits<int>::max();
+  float nps_ GUARDED_BY(nodes_mutex_) = -1.0f;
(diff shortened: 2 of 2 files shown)


borg323:
Test result on colab with net 11089, tc=30+0.5:
```
Score of lc0_master vs lc0_pr356: 41 - 39 - 120  [0.505] 200
Elo difference: 3.47 +/- 30.49
```

mooskagh:
I think that if we don't have proper fix for nps for v0.18, we should at least adjust kSmartPruningToleranceMs (maybe 400 instead of 200) and kSmartPruningToleranceNodes (currently due to bug it's 0.1, maybe make it 1000?).

borg323:
Longer tc (60+1) doesn't look very good, but colab timed out early so not very conclusive yet:
```
Score of lc0_master vs lc0_pr356: 14 - 8 - 64  [0.535] 86
Elo difference: 24.28 +/- 37.05
```

jjoshua2:
our futility pruning is already not very aggressive. If you really want
this to work I would retune futility mover with this patch which should
allow it to be higher like 1.33 to 1.5 and still gain elo, especially with
roy7/crem patch to use time soon which will reduce the downside of too much
volatility in that messing up the curve to use way too much or too little
time. I think ZZ is working on some time tuning, along these lines, but not
with this patch.

borg323:
More games at tc=60+1, similar result:
```
Score of lc0_master vs lc0_pr356: 41 - 34 - 117  [0.518] 192
Elo difference: 12.67 +/- 30.74
```

src/mcts/search.cc
```diff
@@ -312,9 +312,16 @@ void Search::UpdateRemainingMoves() {
...
     if (time_since_start > kSmartPruningToleranceMs) {
-      auto nps = (1000LL * total_playouts_ + kSmartPruningToleranceNodes) /
+      float nps = (1000.0f * total_playouts_ + kSmartPruningToleranceNodes) /
                      (time_since_start - kSmartPruningToleranceMs) +
                  1;
+      if (nps_ < 0) {
+        nps_ = nps;
+      } else {
+        // Simple IIR low pass filter: H(z)=0.5z^-1/(1-0.5z^-1)
+        nps_ = 0.5 * (nps + nps_);
```
Tilps: This formula to me looks like a it just slows the rate of change of the nps estimate value.  Whether the new nps value is lower or higher than before, its replaced with the average.

Since the nps isn't preserved across moves (since a new search instance is created for every move) I don't see why this would result in an improvement.  My understanding is that testing shows that early nps estimates, even after accounting for the tolerances, are never actually too large - so this will only slow the convergence to the actual long term nps value by keeping the value too low.

borg323: Some tests show it to be an improvement in a system with a periodic heavy load: blas backend and simultaneous integer programming solver using all threads for 1 min and off for another 1,

However, it also seems to be leaning towards being negative under normal conditions, so no joy.

ddobbelaere: @Tilps: are you sure early nps estimates are always below the long term value? Maybe this is the case for the UCI reported value, which is different from the formula used for smart pruning:
`auto nps = (1000LL * total_playouts_ + kSmartPruningToleranceNodes) /                  (time_since_start - kSmartPruningToleranceMs) + 1;`

We divide the total number of nodes since the start of search (**including** `kSmartPruningToleranceMs` interval) + `kSmartPruningToleranceNodes` by the time **since** `kSmartPruningToleranceMs`.

Hence, to me, this seems that if the tolerance window is sufficiently large to allow transient responses to disappear, the early nps estimates will be higher than the actual long term nps value.

Tilps: the tolerance nodes is mistakenly outside of the multiply by 1000, so its actually 0.1 nodes of tolerance.
and the initial batches can be so slow that the first time you sample past the tolerance in time, its not sufficient to offset it.  Now there are probably times when the first sample past the tolerance in time is really close, and that gives it a pretty large nudge upwards - but I'm pretty sure you would find that most times the first samples past the initial tolerance are indeed below the long term NPS.

borg323:
Closing since #371 does the job for now. Even just the parenthesis fix seems sufficient.

PR #357, opened 2018-09-16
add benchmark mode

borg323:
This adds a simple benchmark mode. This is equivalent to typing `isready` and `go nodes <nodes>` in uci mode, with <nodes> 30000 by default and changed with with the `--nodes` option. At the end prints the time taken and exits. All uci mode search parameters are available, but some don't do anything.


src/main.cc
```diff
@@ -40,11 +40,16 @@ int main(int argc, const char** argv) {
...
   CommandLine::RegisterMode("selfplay", "Play games with itself");
+  CommandLine::RegisterMode("benchmark", "Quick benchmark");
 
   if (CommandLine::ConsumeCommand("selfplay")) {
     // Selfplay mode.
     SelfPlayLoop loop;
     loop.RunLoop();
+  } else if (CommandLine::ConsumeCommand("benchmark")) {
+    // Selfplay mode.
+    EngineLoop loop;
```
mooskagh: It seems that almost no of EngineLoop/EngineController functionality is not used here.
Maybe a separate class/separate file(or directory) from BenchmarkController?

Also `// Selfplay mode` comment doesn't look correct.

borg323: I'll give it a go - my first effort required a lot of code duplication, but now that I have a working example I may be able to simplify it.

src/engine.cc
```diff
@@ -324,6 +328,25 @@ void EngineLoop::RunLoop() {
...
 
+void EngineLoop::Benchmark() {
+  const char* kNodesStr = "Number of nodes for benchmark";
+  options_.Add<IntOption>(kNodesStr, -1, 999999999, "nodes") = 30000;
+  if (!ConfigFile::Init(&options_) || !options_.ProcessAllFlags()) return;
+  try {
+    CmdIsReady();
+    GoParams go_params;
+    go_params.nodes = options_.GetOptionsDict().Get<int>(kNodesStr);
+    auto start = std::chrono::steady_clock::now();
```
SunnyWar: For benchmarking, I've usually seen std::chrono::high_resolution_clock being used instead.

cwbriscoe:
In the future, it would be nice to have command line options for this mode like:
./lc0 bench --fen="<fen string here>" --nodes=x
./lc0 bench --fen="<fen string here>" --movetime=x
./lc0 bench perft(this is a submode) --depth=x
./lc0 bench --file=<file.txt list of fens with expected results> --movetime=x (or) --nodes=x

I prefer bench instead of benchmark just for brevity but that is minor and I am sure others would disagree.  I was actually going to work on this mode as my next project but you beat me to the punch.

borg323:
I made it a separate class, but I think enough of EngineController is used, so I kept it all together in engine.cc. Also added a `--fen` option.

src/engine.h
```diff
@@ -139,4 +141,18 @@ class EngineLoop : public UciLoop {
...
+ public:
+  Benchmark();
+
+  void Run();
+  void OnBestMove(const BestMoveInfo& move);
+  void OnInfo(const std::vector<ThinkingInfo>& infos);
+
+ private:
+  OptionsParser options_;
+  EngineController engine_;
```
mooskagh: It doesn't seem to me that EngineController is needed at all. Even selfplay doesn't use it, why would it be needed just to populate one position and run search on it?..

But it seems that the code for creating backend and handling options --backend, --backend-opts, --weights indeed exists in 2 times already, and with benchmark it will be three, so it looks like all that functionality should be moved to `NetworkFactory` (e.g. as a helper function)

mooskagh: Also if you find factoring out --backend params too complicated and wish someone else to do that, I can do that!

borg323:
@cwbriscoe calling the search is a given, but why do you think you will need all the uci options? The options you described earlier don't need anything more.

PR #358, opened 2018-09-16
Spend (some portion of) saved time (due to smart pruning) immediately.

mooskagh:
Supersedes https://github.com/LeelaChessZero/lc0/pull/257.

### DIFF
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -102,7 +102,7 @@ void EngineController::PopulateOptions(OptionsParser* options) {
   // This option is currently not used by lc0 in any way.
   options->Add<BoolOption>("Ponder", "ponder") = false;
   options->Add<FloatOption>(kSpendSavedTime, 0.0f, 1.0f, "immediate-time-use") =
-      0.2f;
+      0.0f;
 
   Search::PopulateUciParams(options);
   ConfigFile::PopulateOptions(options);
@@ -175,6 +175,9 @@ SearchLimits EngineController::PopulateSearchLimits(int ply, bool is_black,
   if (slowmover < 1.0 ||
       this_move_time * slowmover > kSmartPruningToleranceMs) {
     this_move_time *= slowmover;
+    // If time is planned to be overused because of slowmover, remove excess
+    // of that time from spared time.
+    time_spared_ms_ -= this_move_time * (slowmover - 1);
   }
 
   // Use `time_to_squander` time immediately.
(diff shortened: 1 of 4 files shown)


mooskagh:
New command line flag is --squander-saved-time
0.0 is old behaviour: allocate saved time for the rest of the game
1.0 is use entire time during the next move.
0.0 < x < 1.0 is something in between

src/engine.h
```diff
@@ -114,6 +114,9 @@ class EngineController {
   // actually search the position one move earlier.
   optional<CurrentPosition> current_position_;
   GoParams go_params_;
+
+  // How much less time was used by search than what was allocated.
+  int64_t time_spared_ms_ = 0;
```
Tilps: This should be reset for ucinewgame, or any other circumstance that we we believe we've transitioned to a new game.

mooskagh: Indeed. Done!

roy7:
I notice you aren't changing the time curve or the slowmover settings in this one. It's important to note if people test this refactor out they should be setting slowmover to 1.0 and adjusting the time curve similar to my original version (or using the new time curve system from @fli). Part of the intent of this change is to remove the slowmover option completely.

roy7:
The options I used for testing were :
`arg="--slowmover=1.0" arg="--time-curve-peak=1" arg="--time-curve-left-width=60" arg="--time-curve-right-width=60"
`

src/engine.cc
```diff
@@ -100,6 +101,8 @@ void EngineController::PopulateOptions(OptionsParser* options) {
   // Add "Ponder" option to signal to GUIs that we support pondering.
   // This option is currently not used by lc0 in any way.
   options->Add<BoolOption>("Ponder", "ponder") = false;
+  options->Add<FloatOption>(kSpendSavedTime, 0.0f, 1.0f, "immediate-time-use") =
+      0.2f;
```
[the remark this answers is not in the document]
mooskagh: Done.

mooskagh:
Fixed it also to support non-zero slowmover. We surely need to simplify it. Not sure if before v18 or not.

roy7:
A suggestion from borg on Discord:

> Large slowmover (e.g the 2.4 default) needs immediate-time-use=0, and immediate-time-use=1 needs slowmover=1. But the optimum may be somewhere in the middle. So the options are either retain current behavior for now, or make slowmover=1 and immeditate-time-use=1 until someone clops it.
> We can also change it to:

`slowmover=2.4-1.4*options_.Get<float>(kSpendSavedTime);`

PR #359, opened 2018-09-16
Add ability to offset visits going in to temperature calculation.

Tilps:
I am thinking that this with -0.7 , or maybe -0.4 and a slight temperature reduction (0.95) could be a good test to run on parallel training after tb rescoring - so it would be good if we could get this in for 0.18.

Note that the assert(max_n > 0.0) can fire if the delta is less than or equal to -1 and the visits are not sufficient.  I would hope there is no actual use case where someone would be trying to test such strange combinations...

Note that delta == -1 is almost, but not quite equivalent the separate proposal to discard single visits.

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -47,6 +47,7 @@ const char* Search::kMaxPrefetchBatchStr = "Max prefetch nodes, per NN call";
 const char* Search::kCpuctStr = "Cpuct MCTS option";
 const char* Search::kTemperatureStr = "Initial temperature";
 const char* Search::kTempDecayMovesStr = "Moves with temperature decay";
+const char* Search::kTemperatureVisitOffsetStr = "Delta to be applied to all visits before calculating temperature";
 const char* Search::kNoiseStr = "Add Dirichlet noise at root node";
 const char* Search::kVerboseStatsStr = "Display verbose move stats";
 const char* Search::kAggressiveTimePruningStr =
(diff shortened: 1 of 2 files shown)


gonzalezjo:
How does this work? I know that discarding single visits doesn't work very well in endgames with few legal moves. Does this PR address that?

src/mcts/search.cc
```diff
@@ -47,6 +47,7 @@ const char* Search::kMaxPrefetchBatchStr = "Max prefetch nodes, per NN call";
 const char* Search::kCpuctStr = "Cpuct MCTS option";
 const char* Search::kTemperatureStr = "Initial temperature";
 const char* Search::kTempDecayMovesStr = "Moves with temperature decay";
+const char* Search::kTemperatureVisitOffsetStr = "Delta to be applied to all visits before calculating temperature";
```
mooskagh: That UCI parameter will overflow any chess GUI. :)
We'll separate help from UCI params eventually, but for now something shorter would be nicer.

src/mcts/search.cc
```diff
@@ -450,27 +454,30 @@ EdgeAndNode Search::GetBestChildWithTemperature(Node* parent,
...
       continue;
     }
-    if(edge.GetN() > max_n) {
-      max_n = edge.GetN();
+    if(edge.GetN() + offset > max_n) {
+      max_n = edge.GetN() + offset;
     }
   }
-  assert(max_n);
+  assert(max_n > 0.0);
```
mooskagh: Is it intended to crash if the offset parameter is too low?

Tilps: I've changed the bounds for the offset parameter to make it clear that it shouldn't be set lower than or equal to -1 so this should only ever crash if someone violates the parameter constraints.

src/mcts/search.cc
Tilps: Shorter it is.

PR #361, opened 2018-09-16
adjust slowmover with immediate-time-use

borg323:
This linearly scales slowmover so that it is 1.0 when immediate-time-use=1 (i.e. when all saved time is allocated to the next move)


borg323:
Note that when immediate-time-use=0 (the default), slowmover is not affected.

roy7:
If you set immediate-time-use to 1 but don't change slowmover, I have no idea what the behavior will be. The original PR immediate-time-use is based on changed the slowmover default to 1.0 very intentionally. You should not this on and have slowmover at a 2.4 default.

borg323:
We need testing to see whether this is a good idea or not. But it is probably too late for v0.18.

borg323:
Closing: This should not be considered for v0.18 at this late stage.

borg323:
This was a quick hack, I'm sure we can do a lot better.

PR #383, opened 2018-09-24
Fix OpenCL tuner matrix size

Ttl:
OpenCL backend tuned the sgemm for wrong matrix size. With the correct dimension nps increases from 630 to 930 using batch size of 16.

### DIFF
--- a/src/neural/opencl/OpenCLTuner.cc
+++ b/src/neural/opencl/OpenCLTuner.cc
@@ -455,10 +455,10 @@ std::string Tuner::load_sgemm_tuners(const int m, const int n, const int k,
         auto tuners = sgemm_tuners_from_line(line, m, n, k, batch_size);
         if (tuners.size() != 0) {
           if (m_params.verbose) {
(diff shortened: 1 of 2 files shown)


src/neural/opencl/OpenCLTuner.cc
```diff
@@ -455,8 +455,12 @@ std::string Tuner::load_sgemm_tuners(const int m, const int n, const int k,
         auto tuners = sgemm_tuners_from_line(line, m, n, k, batch_size);
         if (tuners.size() != 0) {
           if (m_params.verbose) {
-            std::cerr << "Loaded existing SGEMM tuning for batch_size "
-                      << batch_size << "." << std::endl;
+            /* batch_size argument is the number of batched sgemm calls, which
```
mooskagh: We use `//` style comments in our files.

gsobala:
Just confirming that this patch increases nps from 1539 to 1980 after 60sec on a Vega 56 with NN 11250

mooskagh:
Causes #464

PR #385, opened 2018-09-24
Minor tweak to Raspberry Pi build instructions

gsobala:
Referenced .bashrc relative to home directory rather than current


mooskagh:
Is this PR still actual?

PR #387, opened 2018-09-25
Define __i386__ for 32 bit Windows

borg323:
With this fix in node.h it is possible to build the code for 32bit windows with the blas and opencl backends. The blas backend passes all tests and works (but has not been thoroughly tested).

### DIFF
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -253,6 +253,11 @@ class Node {
   friend class Edge;
 };
 
+// Define __i386__ also for 32 bit Windows.
+#if defined(_WIN32) && !defined(_WIN64)
+#define __i386__ 1
+#endif
+
 // A basic sanity check. This must be adjusted when Node members are adjusted.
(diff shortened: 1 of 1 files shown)


src/mcts/node.h
```diff
@@ -253,6 +253,11 @@ class Node {
   friend class Edge;
 };
 
+// Define __i386__ also for 32 bit Windows.
+#if defined(_WIN32) && !defined(_WIN64)
+#define __i386__ 1
```
mooskagh: Why `1`? Just `#define __i386__` should work.

borg323: No reason. However this is not doing the right thing for 32 bit arm processors, so I need to revise it.

mooskagh: Is there windows on arm?

borg323:
The latest patch also defines __arm__ when appropriate, but this part can be dropped.

PR #390, opened 2018-09-26
Make sure Search destructor fires first (#388)

### LINKED ISSUE #388: Exception thrown
Run CUDA build with "go infinite" to a seldepth of about 15 (or less)  Type "quit". An exception is thrown. 

The build was done in VS from the solution provided not from the command line. I tried changing the compiler for Lc0 to use the intel compiler and got the same exception.

vrfcore.dll!00007ffe957125f7()	Unknown
 	vfbasics.dll!00007ffe807f4eaa()	Unknown
 	vfbasics.dll!00007ffe807f3644()	Unknown
 	vfbasics.dll!00007ffe80809952()	Unknown
 	cudnn64_7.dll!00007ffe1180a274()	Unknown
 	cudnn64_7.dll!00007ffe114dc255()	Unknown
 	cudnn64_7.dll!00007ffe104d5663()	Unknown
 	lc0.exe!lczero::CudnnNetwork<float>::~CudnnNetwork<float>(void)	C++
 	lc0.exe!lczero::CudnnNetwork<float>::`vector deleting destructor'(unsigned int)	C++
>	[Inline Frame] lc0.exe!std::default_delete<lczero::Network>::operator()(lczero::Network *) Line 1195	C++
 	[Inline Frame] lc0.exe!std::unique_ptr<lczero::Network,std::default_delete<lczero::Network> >::{dtor}() Line 1397	C++
 	lc0.exe!lczero::EngineController::~EngineController() Line 60	C++
 	lc0.exe!main(int argc, const char * * argv) Line 53	C++

Synthetica9:
Should fix #388

cc @mooskagh

### DIFF
--- a/src/engine.h
+++ b/src/engine.h
@@ -99,8 +99,10 @@ class EngineController {
   std::unique_ptr<NodeTree> tree_;
   std::unique_ptr<SyzygyTablebase> syzygy_tb_;
   std::unique_ptr<Network> network_;
-  std::unique_ptr<Search> search_;
   NNCache cache_;
+
+  // Order here is important. This needs to be destructed last (see #388).
+  std::unique_ptr<Search> search_;
 
   // Store current TB and network settings to track when they change so that
   // they are reloaded.
--- a/src/engine.h
+++ b/src/engine.h
@@ -102,7 +102,7 @@ class EngineController {
   std::unique_ptr<SyzygyTablebase> syzygy_tb_;
   std::unique_ptr<Network> network_;
 
-  // Order here is important. This needs to be destructed last (see #388).
+  // Order here is important. This needs to be destructed first (see #388).
   std::unique_ptr<Search> search_;
 
   // Store current TB and network settings to track when they change so that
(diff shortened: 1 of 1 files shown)


src/engine.h
```diff
@@ -96,10 +96,10 @@ class EngineController {
   RpSharedMutex busy_mutex_;
   using SharedLock = std::shared_lock<RpSharedMutex>;
 
-  std::unique_ptr<Search> search_;
   std::unique_ptr<NodeTree> tree_;
   std::unique_ptr<SyzygyTablebase> syzygy_tb_;
   std::unique_ptr<Network> network_;
+  std::unique_ptr<Search> search_;
```
mooskagh: Also after `cache_`. Maybe better to have `cache_` first in the list for symmetry.

borg323: Please add a comment that the order is important for the destructor.

mooskagh:
```cpp
// Note that the order is important here. Search has to be destructed before
// variables it depends on (e.g. tree_, syzygy_tb_, network_ and cache_), so it
// should be listed last.
```

mooskagh:
Ok.. with this line there, bug is somewhere else. :( Wrong fix.

mooskagh:
Turned out it's not a bug.

PR #404, opened 2018-09-30
pass start time to search

borg323:
This implements the idea by @Tilps to pass the start time to the search constructor, so any time lost during setup can be accounted for.


Tilps:
I was thinking this morning - this might be bad for the nps prediction going in to pruning.  So we might need to keep two start time fields - one for limits and one for nps prediction. (And corresponding two 'GetTimeSinceStart' functions.)

borg323:
When looking at the smart pruning code, [subtracting kSmartPruningToleranceMs from time_since_start](https://github.com/LeelaChessZero/lc0/blob/master/src/mcts/search.cc#L342) seems to be for eliminating such overheads. And this is checked only after 2*kSmartPruningToleranceMs has elapsed, which seems safe.
That being said, having a separate counter for nps cant hurt, so will do it.

src/mcts/search.h
```diff
@@ -167,6 +169,7 @@ class Search {
   Network* const network_;
   const SearchLimits limits_;
   const std::chrono::steady_clock::time_point start_time_;
+  const std::chrono::steady_clock::time_point nps_time_;
```
Tilps: Maybe name this nps_start_time_?

Tilps:
Looks good to me - just the member name could be a bit clearer.

The subtracting and waiting is to offset the 'spin up time' where the first call to the backend can be extraordinarily long compared to all subsequent calls.  I don't think it was intended to offset what we're doing here which is shutting down a previous search.

borg323:
Agreed regarding the purpose of subtracting kSmartPruningToleranceMs. I was wondering whether we could make nps_start_time_ (renamed) also include this 'spin up time'.

src/mcts/search.h
```diff
@@ -57,7 +57,9 @@ class Search {
          BestMoveInfo::Callback best_move_callback,
          ThinkingInfo::Callback info_callback, const SearchLimits& limits,
          const OptionsDict& options, NNCache* cache,
-         SyzygyTablebase* syzygy_tb);
+         SyzygyTablebase* syzygy_tb,
+         std::chrono::steady_clock::time_point start_time =
```
mooskagh: There is a zoo of various start times now...
What do you think about doing this instead:
1. Replace `SearchLimits::time_ms` with `SearchLimits::search_deadline`, make it `std::optional<chrono::steady_clock::time_point>`, and populate in `EngineController::PopulateSearchLimits` (and pass start_time to that function)
2. Make `nps_start_time_` also std::optional and set it after the first iteration which had real NN eval.

borg323: `SearchLimits::time_ms` is also used for selfplay with `--movetime=x` so it isn't easy to remove completely.

mooskagh: There we can also compute end time in game.cc and pass it.

mooskagh:
Which of #404, #427, #450 are intended for merge and in which order?

borg323:
superseded by #450

PR #405, opened 2018-10-01
Rename to SyzygyPath

maelic13:
For lots of GUI, the default name for List of Syzygy Directories is instead SyzygyPath. This helps to automate giving Syzygy tablebases to engines without need to manually write in the path (Chessbase and Fritz GUIs specifically). As such, the variable name should be by default standarts for ease of use.


Tilps:
This was fixed during the merge of #425

PR #407, opened 2018-10-01
Add timestamps to debug logs.

fli:
Prefixes each log line with a Unix-style timestamp.

### DIFF
--- a/src/chess/uciloop.cc
+++ b/src/chess/uciloop.cc
@@ -128,10 +128,9 @@ std::string GetTimestamp() {
   auto ms =
       duration_cast<milliseconds>(time_now.time_since_epoch()).count() % 1000;
   auto timer = system_clock::to_time_t(time_now);
-  std::tm bt = *std::localtime(&timer);
   std::ostringstream oss;
-  oss << std::put_time(&bt, "%F %T") << '.' << std::setfill('0') << std::setw(3)
-      << ms;
+  oss << std::put_time(std::localtime(&timer), "%F %T") << '.'
+      << std::setfill('0') << std::setw(3) << ms;
   return oss.str();
 }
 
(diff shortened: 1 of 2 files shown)


fli:
How's this look? Lines are in format
```
year-month-day hours:minutes:seconds.milliseconds debugging-line
```

src/chess/uciloop.cc
```diff
@@ -215,6 +228,13 @@ void UciLoop::SetLogFilename(const std::string& filename) {
   }
 }
 
+void UciLoop::WriteDebugLogLine(const std::string& line) {
+  static std::mutex logging_mutex;
+  std::lock_guard<std::mutex> lock(logging_mutex);
+  if (debug_log_)
+    debug_log_ << GetTimestamp() << ' ' << line << std::endl << std::flush;
```
mooskagh: `std::endl` does `std::flush` itself (unlike outputting '\n'), so no need to flush explicitly.

src/chess/uciloop.cc
```diff
@@ -122,13 +122,26 @@ bool ContainsKey(const std::unordered_map<std::string, std::string>& params,
   return params.find(key) != params.end();
 }
 
+std::string GetTimestamp() {
+  using namespace std::chrono;
+  auto time_now = system_clock::now();
+  auto ms =
+      duration_cast<milliseconds>(time_now.time_since_epoch()).count() % 1000;
+  auto timer = system_clock::to_time_t(time_now);
+  std::tm bt = *std::localtime(&timer);
```
mooskagh: Maybe feed `std::localtime` directly to `put_time` instead of creating a temporary copy of the structure?

fli:
Thanks, made changes.

PR #408, opened 2018-10-01
fix syzygy draw by repetition bug

RedDenver:
Set any 3-fold repetition position to have a draw score of zero, which will score less than a winning but for 50move draw but score higher than a losing but for 50 move draw. Certain wins score 1000 and certain losses score -1000, so the zero draw score should work just fine.

### DIFF
--- a/src/syzygy/syzygy.cc
+++ b/src/syzygy/syzygy.cc
@@ -1647,15 +1647,17 @@ bool SyzygyTablebase::root_probe(const Position& pos,
       dtz = 1;
     }
     if (result == FAIL) return false;
-    // Better moves are ranked higher. Certain wins are ranked equally.
-    // Losing moves are ranked equally unless a 50-move draw is in sight.
-    int r = dtz > 0
-                ? (dtz + cnt50 <= 99 ? 1000 : 1000 - (dtz + cnt50))
-                : dtz < 0 ? (-dtz * 2 + cnt50 < 100 ? -1000
-                                                    : -1000 + (-dtz + cnt50))
-                          : 0;
-    if (next_pos.GetRepetitions() > 1)
-      r = 0; //draw by repetition
+    int r = 0; //draw by repetition
+    if (next_pos.GetRepetitions() < 2)
+    {
+      // Better moves are ranked higher. Certain wins are ranked equally.
+      // Losing moves are ranked equally unless a 50-move draw is in sight.
+      int r = dtz > 0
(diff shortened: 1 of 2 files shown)


src/syzygy/syzygy.cc
```diff
@@ -1649,13 +1647,17 @@ bool SyzygyTablebase::root_probe(const Position& pos,
...
     if (result == FAIL) return false;
-    // Better moves are ranked higher. Certain wins are ranked equally.
-    // Losing moves are ranked equally unless a 50-move draw is in sight.
-    int r = dtz > 0
-                ? (dtz + cnt50 <= 99 && !rep ? 1000 : 1000 - (dtz + cnt50))
-                : dtz < 0 ? (-dtz * 2 + cnt50 < 100 ? -1000
-                                                    : -1000 + (-dtz + cnt50))
-                          : 0;
+    int r = 0; //draw by repetition
+    if (next_pos.GetRepetitions() < 2)
```
Tilps: This doesn't work as GetRepetitions is only populated when you push the position on to the history stack, which isn't done here.
This logic change is also wrong - as it loses the concept that if the opponent has repeated you must perform the best option.

RedDenver:
> I think the 'correct' fix is that in search when we use tb root filtering - we should also subsequently check each filtered move to see if its a 2 (or 3) fold rep by pushing it onto the history stack and popping it off. Any such moves must be additionally filtered **if we are winning**.
> 
> I didn't realize that this would be a problem as I assumed the logic from SF (which matches ours) would be sufficient, failing to take into consideration that SF uses 2 fold draw in search - so it would _never_ select a 2 fold draw move while its wining - but leela will not be so careful.

Yeah, also the PositionHistory needs the move to be appended in order to correctly check the repetition count, so this way isn't going to work. Close this PR?

Tilps:
Sure - if you want to start a fresh PR (or wait until I have time to write up my proposed change).

RedDenver:
> Sure - if you want to start a fresh PR (or wait until I have time to write up my proposed change).

I'll wait for your proposed change.

joergoster:
Just fyi: CFish does this check in uci.c when receiving the game history. See https://github.com/syzygy1/Cfish/blob/master/src/uci.c#L101-L118
Stockfish has an extra function in position.cpp, see https://github.com/official-stockfish/Stockfish/blob/master/src/position.cpp#L1129-L1155

Tilps:
Yes - the logic for that function has now been (hopefully) correctly implemented in PR #412.

PR #410, opened 2018-10-02
Apply a secondary filter for TB positions to avoid playing repetition while winning.

Tilps:
Untested at this point - but it compiles and shows the logic I think should work.


Tilps:
This was insufficient and has been superseded by #412

PR #412, opened 2018-10-02
Correct implementation of root_probe to consider all prior potential repeats, not just current move.

Tilps:
This was an oversight on my part when converting from cfish movegen to our movgen.


Tilps:
I think you mean has_repeated = true.  The test case I'd most like to do is for the new function, let me see what I can whip up.

PR #416, opened 2018-10-02
fixed repetition init bug

RedDenver:
Found this bug when working on #408. Causes the repetition to not be initialized to zero correctly.

### DIFF
--- a/src/chess/position.cc
+++ b/src/chess/position.cc
@@ -32,6 +32,7 @@ namespace lczero {
 
 Position::Position(const Position& parent, Move m)
     : no_capture_ply_(parent.no_capture_ply_ + 1),
+      repetitions_(0),
       ply_count_(parent.ply_count_ + 1) {
   them_board_ = parent.us_board_;
   bool capture = them_board_.ApplyMove(m);
--- a/src/chess/position.h
+++ b/src/chess/position.h
@@ -76,7 +76,7 @@ class Position {
   // How many half-moves without capture or pawn move was there.
   int no_capture_ply_ = 0;
   // How many repetitions this position had before. For new positions it's 0.
-  int repetitions_;
+  int repetitions_ = 0;
   // number of half-moves since beginning of the game.
   int ply_count_ = 0;
 };


src/chess/position.cc
```diff
@@ -32,6 +32,7 @@ namespace lczero {
 
 Position::Position(const Position& parent, Move m)
     : no_capture_ply_(parent.no_capture_ply_ + 1),
+      repetitions_(0),
```
DanielUranga: What if the Move `m` was actually a repetition?

RedDenver: Then the caller would be responsible for setting the correct repetition. This bug is mostly just assuring that `repetitions_` isn't some arbitrary uninitialized value. But could change to `Position::Position(const Position& parent, Move m, int repetitions = 0)` for more flexibility.

DanielUranga: Then https://github.com/LeelaChessZero/lc0/pull/416/files#diff-2b6cf6f06881eb37c1ebee83cfd70353R79 should already take care of it.

RedDenver: I see what you're saying. I just did it to match the other constructor. Could take the `repetitions_(0)` out of both, but I'd like the constructors to match whichever way we choose to do it.

mooskagh:
It's not a bug, the initialization is omitted as optimization (as this constructor is typically called millions times per second). Maybe it worth adding a comment instead.

On the other hand, the performance hit is probably negligible. Could you nest nps difference on random backend? (with fpu-reduction 60 or so, to encourage more realistic tree shape)

RedDenver:
> It's not a bug, the initialization is omitted as optimization (as this constructor is typically called millions times per second). Maybe it worth adding a comment instead.
> 
> On the other hand, the performance hit is probably negligible. Could you nest nps difference on random backend? (with fpu-reduction 60 or so, to encourage more realistic tree shape)

I don't have a GPU, so my ability to test is limited.

mooskagh:
Then it's probably safer to leave a comment, warning that in this constructor repetition count is not initialized and should be set before using.

PR #417, opened 2018-10-02
quicker test builds by first making a lib from lc0 sources

borg323:
This shaves 4-5 minutes from appveyor builds. Should also improve circleci build speed.


borg323:
Cicleci build was about 1 min shorter.

PR #421, opened 2018-10-05
Exception during lc0 startup - using backend opts

MartinRe63:
When backend options are used like: 
lc0.exe selfplay --backend-opts="cudnn(threads=1, minibatch-size=512)"
the variable dict used before will get invalid, when the methode FromString will end, which lead to an exception in the debug mode and might lead to bigger issues later in the runtime mode, when the options are overwritten by random.

I'm sorry if something is wrong with my pull request. It's my first try.

### DIFF
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -225,8 +225,8 @@ void EngineController::UpdateTBAndNetwork() {
   }
   Weights weights = LoadWeightsFromFile(net_path);
 
-  OptionsDict network_options =
-      OptionsDict::FromString(backend_options, &options_);
+  OptionsDict network_options(&options_);
+  network_options.AddSubdictFromString(backend_options);
 
   network_ = NetworkFactory::Get()->Create(backend, weights, network_options);
 }
--- a/src/selfplay/tournament.cc
+++ b/src/selfplay/tournament.cc
@@ -146,8 +146,8 @@ SelfPlayTournament::SelfPlayTournament(const OptionsDict& options,
     std::string backend_options = options.GetSubdict(kPlayerNames[idx])
                                       .Get<std::string>(kNnBackendOptionsStr);
 
-    OptionsDict network_options = OptionsDict::FromString(
-        backend_options, &options.GetSubdict(kPlayerNames[idx]));
+   OptionsDict network_options(&options.GetSubdict(kPlayerNames[idx]));
+   network_options.AddSubdictFromString(backend_options);
 
     networks_[idx] =
         NetworkFactory::Get()->Create(backend, weights, network_options);
--- a/src/utils/optionsdict.cc
+++ b/src/utils/optionsdict.cc
@@ -317,12 +317,18 @@ class Parser {
 
 }  // namespace
 
+<<<<<<< HEAD
 OptionsDict OptionsDict::FromString(const std::string& str,
                                     const OptionsDict* parent) {
   OptionsDict *dict = new OptionsDict(parent);
   Parser parser(str);
   parser.ParseMain(dict);
   return *dict;
+=======
+void OptionsDict::AddSubdictFromString(const std::string& str) {
+  Parser parser(str);
+  parser.ParseMain(this);
+>>>>>>> master
 }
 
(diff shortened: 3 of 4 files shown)


mooskagh:
Could you provide more details what the exception is? It should work without any change, as copy of `OptionsDict` is returned form that function, not the original one.
Also if we would need to pass pointers, we'd have to use unique_ptr rather than creating an object which has never destroyed.

MartinRe63:
The issue happened with Windows 10 using Microsofts C++ Compiler. 

image

This is the view into the network_options variable direct after the methode Optionsdict::FromString returned, but the debugger is still in the same line of code. 

image

and the second image will show you the content of the variable network_options.subdicts, after I made the step to the next line of code. 

My understanding is, that the debugger marked memory, which is not allocated any more with the 0xccccccccccccccccc values to ensure, that invalid memory is not used further by the application.

MartinRe63:
Because the parent pointer in the ["cudnn"] area has changed, the exeption later will occur. Of cause I'm not able to ensure, that this issue is not a compiler bug or anything else in the setting I'm using. At least I've loaded the complete branch, but the issue still occured using the compiler settings for x64 environment. 
Might also well be, that when the OptionsDict network_options is copied, something went wrong and the parent content is destroyed, while the dict_ content is still ok. 

It's already late today. Tomorrow I'll try the same change with the use of unique_ptr.

MartinRe63:
Ok. I'm clear now what happened and thank you mooskagh you are fully right. 

First the variable dict is initialized within the stack of the static function OptionsDict::FromString. Later within ParseMain a new subdict_ is created having the property second with a parent pointer refering to the dict variable on the stack. 
When OptionsDict::FromString ends dict will be copied to network_options, but the included parent pointer is still refering to the stack variable. The debug code will identify an invalid pointer and fill this with the content 0xcccccccc ... to ensure that the application will fail asap.

I do see 3 options to solve this issue: 
1. Use a std::unique_ptr and the system will move this pointer, when it will leave OptionsDict::FromString. 
2. I'm not sure why this is a static function. Rewrite FromString and use a normal member function. 
3. Add an additional constructor to OptionsDict reading it from a string and remove "FromString".

I'll show the 3 options within the 3 next comments. Please come back to which of the versions I should change the code. 

Thx.

MartinRe63:
Option 1: 

std::unique_ptr<OptionsDict> OptionsDict::FromString(const std::string& str,
                                    const OptionsDict* parent) {
  std::unique_ptr<OptionsDict> dict = std::make_unique<OptionsDict>(parent);
  Parser parser(str);
  parser.ParseMain(dict.get());
  return dict;
}

and tournament.cc to be changed like this: 

    std::unique_ptr<OptionsDict> network_options = OptionsDict::FromString(
        backend_options, &options.GetSubdict(kPlayerNames[idx]));
    networks_[idx] =
        NetworkFactory::Get()->Create(backend, weights, *network_options);

MartinRe63:
Option 2: Change FromString into a "normal" member function e.g. AddSubdictFromString 

void OptionsDict::AddSubdictFromString(const std::string& str) {
  Parser parser(str);
  parser.ParseMain(this);
}

and call this in two steps, which would allow a local variable ...

OptionsDict network_options(&options.GetSubdict(kPlayerNames[idx]));
network_options.AddSubdictFromString(backend_options);

MartinRe63:
Option 3: Add an other constructor: 

OptionsDict::OptionsDict(const std::string& str, const OptionsDict* parent) : parent_(parent) {
  Parser parser(str);
  parser.ParseMain(this);
} 

and call it like this: 

OptionsDict network_options(backend_options, &options.GetSubdict(kPlayerNames[idx]));

Please come back how this should be changed and I'll provide the changes.

mooskagh:
That's on my radar, I just didn't have time this week yet.

mooskagh:
It was static function because it was meant to be "named constructor".
I don't like much option 3 (although it's not that bad either), options 1 and 2 I like more.
Another option would be to make all them shared_ptr.

PR #426, opened 2018-10-07
Batch collisions and terminal node visits.

mooskagh:
If there is a collision, estimate how much more visits will be collisions and apply them at once.
The same for terminal node visits.


src/mcts/params.cc
```diff
@@ -99,7 +102,8 @@ void SearchParams::Populate(OptionsParser* options) {
   options->Add<FloatOption>(kFpuReductionId, -100.0f, 100.0f) = 0.0f;
   options->Add<IntOption>(kCacheHistoryLengthId, 0, 7) = 7;
   options->Add<FloatOption>(kPolicySoftmaxTempId, 0.1f, 10.0f) = 1.0f;
-  options->Add<IntOption>(kAllowedNodeCollisionsId, 0, 1024) = 0;
+  options->Add<IntOption>(kAllowedNodeCollisionEventsId, 0, 1024) = 0;
+  options->Add<IntOption>(kAllowedTotalNodeCollisionsId, 0, 1000000) = 10000;
```
Tilps: Why not default to 0 here?

mooskagh: Done.

PR #427, opened 2018-10-08
Make engine controller track move start time in ways which more closely match how UCI hosts measure time.

Tilps:
Consider this untested - and its got a big TODO in it that I'm not sure what is the ideal behaviour.

Also this is intended to augment #404, and hopefully should be easy to merge the two together.

### DIFF
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -73,7 +73,8 @@ EngineController::EngineController(BestMoveInfo::Callback best_move_callback,
                                    const OptionsDict& options)
     : options_(options),
       best_move_callback_(best_move_callback),
-      info_callback_(info_callback) {}
+      info_callback_(info_callback),
+      move_start_time_(std::chrono::steady_clock::now()) {}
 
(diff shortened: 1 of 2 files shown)


src/engine.cc
```diff
@@ -236,9 +237,15 @@ void EngineController::SetCacheSize(int size) { cache_.SetCapacity(size); }
 void EngineController::EnsureReady() {
   UpdateTBAndNetwork();
   std::unique_lock<RpSharedMutex> lock(busy_mutex_);
+  // If a UCI host is waiting for our ready response, we can consider the move
```
mooskagh: That's not what arena does, at least.
Before we had a separate thread for tree GC, `position startpos moves ...` started to free unused parts of a tree, and we didn't reply `readyok` until it's done, but still timed out a lot because clock ticked.

Tilps: arena doesn't even call isready after set position at all.
cutechess does and doesn't start the timer until readyok.

PR #428, opened 2018-10-09
Corrected case of min/max to UCI specification

gsobala:
The official UCI specification has min and max in lower case. pr425 changed these to upper case which broke a few GUIs. This pr reverts them to lower case.


mooskagh:
Yeah I only intended to do uppercase for --help, overdid it. :)

PR #429, opened 2018-10-10
Make hiding options possible

borg323:
Adds the code to hide options. The options remain accessible through the command line and the uci interface, but a gui won't show them. Also adds a (hidden) `--show-hidden` option to expose them so they can be used (e.g. for tuning using clop). 

As an example the time curve options are hidden. `lc0 --help` won't show them, but `lc0 --show-hidden --help` will.


src/utils/optionsparser.cc
```diff
@@ -41,14 +41,14 @@ const int kUciLineIndent = 32;
 const int kHelpWidth = 72;
 }  // namespace
 
-OptionsParser::Option::Option(const OptionId& id) : id_(id) {}
+OptionsParser::Option::Option(const OptionId& id) : id_(id), hidden_(false) {}
```
mooskagh: nit: in C++11 it's cleaner to initialize such variables right in declaration: `bool hidden_ = false;`.

src/utils/optionsparser.cc
```diff
@@ -67,6 +67,13 @@ void OptionsParser::SetUciOption(const std::string& name,
   throw Exception("Unknown option: " + name);
 }
 
+void OptionsParser::HideOption(const OptionId& id) {
+  auto option = FindOptionByUciName(id.uci_option);
```
mooskagh: `FindOptionById(id.GetId())`
This will work even if it's command-line flag only option, without UCI.
(the `GetId()` part will be not needed after future refactorings, and FindOptionById() will get OptionId instead)

src/utils/optionsparser.cc
```diff
@@ -67,6 +67,13 @@ void OptionsParser::SetUciOption(const std::string& name,
...
+  if (option) {
```
mooskagh: Possible to write as one-liner:
```
if (option) option->hidden_ = true;
```

src/utils/optionsparser.cc
borg323: I saw the FindOptionById prototype and tried it, but it failed to build. I thought it was a placeholder, but maybe it got lost in some refactoring?

mooskagh: Yeah, it's `FindOptionById(id.GetId())`, not `FindOptionById(id)`.

borg323: I mean the implementation of FindOptionById() is missing: 
```
undefined reference to `lczero::OptionsParser::FindOptionById(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) const'
```

PR #434, opened 2018-10-11
Do not parse flags twice (just to find config file)

borg323:
The -n short option works as a toggle. This means that ProcessAllFlags() should not be called twice, since then the value is reset to the default. Unfortunately, this is exactly what is happening since we call it once to get the config file to open, and then once more to actually do the final option parsing. This patch introduces a ProcessConfigFlag function to just look for this specific option and get the filename.

### DIFF
--- a/src/utils/configfile.cc
+++ b/src/utils/configfile.cc
@@ -47,15 +47,38 @@ void ConfigFile::PopulateOptions(OptionsParser* options) {
   options->Add<StringOption>(kConfigFileId) = kDefaultConfigFile;
 }
 
+// This is needed to get the config file from the parameters without calling
+// ProcessAllFlags() that should be called only once, and needs the config file.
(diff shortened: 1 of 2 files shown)


cwbriscoe:
I think -n should be changed to not be a toggle as well.  Seems kind of strange.  `./lc0 -n` will still have unintended effects if `--noise` is in the config file.  It was intended that command line arguments would override anything also defined in the config file.  I did not know bool options worked this way.

borg323:
Agreed on the short flag meaning, will prepare a new  PR for this.
A less hacky way to handle this (but changes behavior) may be to handle the config file while parsing options. Then the config file will override flags before `--config`  on the command line but not after it.

borg323:
@cwbriscoe already made #446 to fix the short flags toggling.

PR #435, opened 2018-10-11
Do something useful with a "go depth N" value.

CarloWood:
The value passed to a 'go depth' UCI command
is parsed but not used. The result is that
'go depth 1' is equal to a 'go infinite'.

For a GUI that doesn't support 'go nodes', like
pychess, this means that if you set the 'difficulty
slider' to any value - the engine doesn't move
until it is stop-ped. To have a better experience
it seems logical to more or less stop once the
printed uci info 'depth' value reaches the requested
depth.

Example engine output after patch:

ucinewgame
position startpos moves e2e4
go depth 3
info depth 1 seldepth 2 time 194 nodes 3 score cp -5 hashfull 0 nps 15 tbhits 0 pv c7c5 d2d4
info depth 2 seldepth 3 time 310 nodes 9 score cp -10 hashfull 0 nps 29 tbhits 0 pv c7c5 c2c3 g8f6
info depth 3 seldepth 4 time 401 nodes 20 score cp -7 hashfull 0 nps 49 tbhits 0 pv c7c5 b1c3 b8c6 g1e2
info depth 3 seldepth 4 time 401 nodes 20 score cp -7 hashfull 0 nps 49 tbhits 0 pv c7c5 b1c3 b8c6 g1e2
bestmove c7c5 ponder b1c3

Note that the last info line is duplicated (this was already the case
before this patch).

The search stops as soon as the (average) depth is reached, because it
is always possible to increment the depth, while on the other end you
can't go lower than 1.

ucinewgame
position startpos moves e2e4
go depth 1
bestmove c7c5

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -319,6 +319,9 @@ void Search::MaybeTriggerStop() {
   if (limits_.time_ms >= 0 && GetTimeSinceStart() >= limits_.time_ms) {
     FireStopInternal();
   }
+  if (limits_.depth >= 0 && cum_depth_ / (total_playouts_ ? total_playouts_ : 1) >= (unsigned int)limits_.depth) {
+    FireStopInternal();
+  }
   // If we are the first to see that stop is needed.
   if (stop_ && !responded_bestmove_) {
     SendUciInfo();
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -319,7 +319,10 @@ void Search::MaybeTriggerStop() {
   if (limits_.time_ms >= 0 && GetTimeSinceStart() >= limits_.time_ms) {
     FireStopInternal();
   }
-  if (limits_.depth >= 0 && cum_depth_ / (total_playouts_ ? total_playouts_ : 1) >= (unsigned int)limits_.depth) {
+  // Stop if average depth reached requested depth.
+  if (limits_.depth >= 0 &&
+      cum_depth_ / (total_playouts_ ? total_playouts_ : 1) >=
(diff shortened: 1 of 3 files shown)


CarloWood:
When using 'go depth 1' no info line is printed (and no ponder either). This is because there wasn't any thinking ;). Compare the play leela zero website at level Easy, in which case there also isn't a "leela zero thinks that it's advantage is..." line.

CarloWood:
Where the git comment says "go depth infinite" that should have been "go infinite" of course :/.

src/mcts/search.cc
```diff
@@ -319,6 +319,9 @@ void Search::MaybeTriggerStop() {
   if (limits_.time_ms >= 0 && GetTimeSinceStart() >= limits_.time_ms) {
     FireStopInternal();
   }
+  if (limits_.depth >= 0 && cum_depth_ / (total_playouts_ ? total_playouts_ : 1) >= (unsigned int)limits_.depth) {
```
mooskagh: According to the style guide:
1. C-style casts are not allowed (use static_cast)
2. Lines longer than 80 characters are not allowed.   It's recommended to run clang-format to format the code as it also does other changes.

I've done it for this snippet though:
```suggestion
if (limits_.depth >= 0 &&
    cum_depth_ / (total_playouts_ ? total_playouts_ : 1) >=
        static_cast<unsigned int>(limits_.depth)) {
}
```

Also add a comment before the `if` statement.

CarloWood: Seems you added a '}' too much?

PR #436, opened 2018-10-12
[Needs Testing] Contempt

jjoshua2:
first commit was static contempt, this one is dynamic

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -1009,7 +1009,7 @@ void SearchWorker::FetchSingleNodeResult(NodeToProcess* node_to_process,
   }
   // For NN results, we need to populate policy as well as value.
   // First the value...
-  node_to_process->v = -computation_->GetQVal(idx_in_computation);
+  node_to_process->v = -computation_->GetQVal(idx_in_computation) + 0.05;
   // ...and secondly, the policy data.
   float total = 0.0;
   for (auto edge : node->Edges()) {
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -1013,9 +1013,9 @@ void SearchWorker::FetchSingleNodeResult(NodeToProcess* node_to_process,
   auto board = search_->played_history_.Last().GetBoard(); 
   //assuming white player has contempt for black player (should be leela has contempt for opponent)
   auto contempt = 0.05;
-  if(search_->is_black_ && search_->depth % 2 == 0)
+  if(search_->is_black_ && depth % 2 == 0)
     contempt = -contempt;
-  else if (!search_->is_black_ && search_->depth % 2 == 1)
+  else if (!search_->is_black_ && depth % 2 == 1)
     contempt = -contempt;
   node_to_process->v = -computation_->GetQVal(idx_in_computation) + contempt;
   // ...and secondly, the policy data.
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -1011,11 +1011,9 @@ void SearchWorker::FetchSingleNodeResult(NodeToProcess* node_to_process,
   // For NN results, we need to populate policy as well as value.
   // First the value...
   auto board = search_->played_history_.Last().GetBoard(); 
-  //assuming white player has contempt for black player (should be leela has contempt for opponent)
+  //assuming leela is searching from root
   auto contempt = 0.05;
-  if(search_->is_black_ && node_to_process->depth % 2 == 0)
-    contempt = -contempt;
-  else if (!search_->is_black_ && node_to_process->depth % 2 == 1)
+  if(node_to_process->depth % 2 == 0)
     contempt = -contempt;
   node_to_process->v = -computation_->GetQVal(idx_in_computation) + contempt;
   // ...and secondly, the policy data.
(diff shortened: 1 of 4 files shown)


RedDenver:
Why is this based on total pieces on the board? At start position, leela will add 0.01*32=0.32 to its expected outcome. In other words, leela will think there's a 32% better chance of winning just because all the pieces are on the board.

Shouldn't contempt be and additional term that only applies when leela thinks its winning? For example:
```
node_to_process->v = -computation_->GetQVal(idx_in_computation); //could add a small amount like 0.01 here
if (node_to_process->v > 0)
  node_to_process->v *= 1.01; //or use a contempt term for the multiplier
```

src/mcts/search.cc
```diff
@@ -1009,7 +1010,12 @@ void SearchWorker::FetchSingleNodeResult(NodeToProcess* node_to_process,
   }
   // For NN results, we need to populate policy as well as value.
   // First the value...
-  node_to_process->v = -computation_->GetQVal(idx_in_computation);
+  auto board = search_->played_history_.Last().GetBoard(); 
+  //assuming leela is searching from root
+  auto contempt = 0.05;
+  if(node_to_process->depth % 2 == 0)
+    contempt = -contempt;
+  node_to_process->v = -computation_->GetQVal(idx_in_computation) + contempt;
```
Tilps: should contempt be a saturating operation - not result in a v outside of the range +/-1 ?

jjoshua2: I don't think it should matter much, except for interactions with EGTB, in which case its probably good for a checkmate to score higher than a DTZ. But it needs testing both ways probably. I think the level of contempt and some sort of dynamic contempt are going to be much bigger deals.

jjoshua2: It should be tested to make sure it will still take a mate instead of going for the mate in 1 as well...

jjoshua2:
stock leela
```
Score of lc0 v18 vs Laser 010618 2CPU: 125 - 8 - 69
```
5% contempt (still running)
```
Score of lc0contempt vs Laser 010618 2CPU: 78 - 7 - 27 [0.817]
Elo difference: 259.87 +/- 65.89
```

zz4032:
vs. asmFish (asmFish on 1 thread and lower TC):
```
   # PLAYER          :  RATING  ERROR   GAMES  WON  DRAWN  LOST  DRAWS(%)
   1 lc0             :     219     31     400  245    132    23      33.0
   2 lc0-contempt    :     209     28     400  236    142    22      35.5
   3 asmFish         :       0   ----     800   45    274   481      34.2
```

jjoshua2:
```Score of lc0contempt vs Laser 010618 2CPU: 297 - 36 - 159 [0.765]
Elo difference: 205.27 +/- 26.93
```
Ran too far in my opening book, so now i need to run the other one farther so they are using the same openings...

zz4032:
@jjoshua2 suggested CLOP tuning, which resulted in contempt value 0.15 vs. same asmFish below.
Match result:
```
   # PLAYER              :  RATING  ERROR   GAMES  WON  DRAWN  LOST  DRAWS(%)
   1 lc0                 :     220     30     400  245    132    23      33.0
   2 lc0-contempt0.05    :     210     32     400  236    142    22      35.5
   3 lc0-contempt0.15    :     200     30     400  231    144    25      36.0
   4 asmFish             :       0   ----    1200   70    418   712      34.8
```
All Elo values inside error bounds and probably not stronger than current lc0.

zz4032:
Tuned the formula with CLOP:
`contempt = 0.0013 * (board.ours() + board.theirs()).count() + 0.029`
Test match:
```
   # PLAYER       :  RATING  ERROR   GAMES  WON  DRAWN  LOST  DRAWS(%)
   1 lc0          :     236     34     400  259    114    27      28.5
   2 lc0-tuned    :     204     31     400  237    133    30      33.2
   3 asmFish      :       0   ----     800   57    247   496      30.9
```
No improvement visible.

zz4032:
Tested the updated code with:
```
  //if(node_to_process->depth % 2 == 0)
  //  contempt = -contempt;
```

"lc0-updated" is the current implementation:
auto contempt = 0.003 * (board.ours() + board.theirs()).count() - 0.003 * 6;

"lc0-tuned" is with both parameters CLOP-tuned:
auto contempt = -0.00191 * (board.ours() + board.theirs()).count() + 0.0193;
No idea why the signs became reversed but that's what it is.

```
   # PLAYER                  :  RATING  ERROR   GAMES  WON  DRAWN  LOST  DRAWS(%)
   1 lc0                     :     233     31     400  259    114    27      28.5
   2 lc0_2nd-run             :     202     28     400  240    127    33      31.8
   3 lc0-tuned               :     197     29     400  225    153    22      38.2
   4 lc0-updated             :     192     32     400  228    143    29      35.8
   5 asmFish                 :       0   ----    1600  111    537   952      33.6
```

I thought maybe the first run with unmodified lc0 was lucky so I did a 2nd-run ("lc0_2nd-run"), still strongest...

So unfortunately I can't see a detectable improvement.
I heard about better match results in discord, maybe they can be posted here.

src/mcts/search.cc
PowerMove72: > should contempt be a saturating operation - not result in a v outside of the range +/-1 ?

we can keep V inside the range with an IF THEN statement checking the V value after the computation.

PR #439, opened 2018-10-12
Implement better logging.

mooskagh:


mooskagh:
It's not that easy to have it bound to variable when initializing using macro:

```cpp
auto log = LOGFILE;  // Error: no copy constructor.
auto& log = LOGFILE;  // Error: cannot bind temporary to non-const reference.
```

PR #440, opened 2018-10-13
Correct grammar and/or orthography

PedanticHacker:
Fixed a lot of text with either bad grammar or bad orthography – or both!

### DIFF
--- a/FLAGS.md
+++ b/FLAGS.md
@@ -7,7 +7,7 @@ Currently, `lc0` has the following modes:
 
 | Mode |   Description |
 |------|---------------|
-| uci (*default*) | Acts as an UCI chess engine |
+| uci (*default*) | Acts as a UCI chess engine |
 | selfplay | Plays one or multiple games with itself and optionally generates training data |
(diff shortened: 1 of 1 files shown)


cwbriscoe:
`an UCI engine`  This one doesn't sound right.  I was curious so I looked it up and found this:

https://english.stackexchange.com/questions/1016/do-you-use-a-or-an-before-acronyms-initialisms

```The important point to remember is the following:

Written language is a representation of the spoken word.

Thus, the answer is "If the word following the indefinite article begins with a vowel sound, use an; if it begins with a consonant sound, use a."

In the case of initialisms and acronymns, use the exact rule above. For initialisms (e.g. "US"), the individual letters are pronounced. With what sound does the first pronounced letter begin? In the example "US", the first sound is /j/ (or "y"). This is a consonant sound, despite the letter "U" being a vowel; thus, you use a, as in a US dollar.

Contrast this with the initialism "RPM", which begins with the consonant "R" but is pronounced starting with /a/; thus, you use an, as in an RPM counter.```

PR #443, opened 2018-10-13
Support <stderr> as special name of a logfile.

mooskagh:
Also increase move overhead limit to ~27 hours to support playing by
fax.

### DIFF
--- a/src/utils/logging.cc
+++ b/src/utils/logging.cc
@@ -27,12 +27,14 @@
 
 #include "utils/logging.h"
 #include <iomanip>
+#include <iostream>
 #include <thread>
 
 namespace lczero {
 
 namespace {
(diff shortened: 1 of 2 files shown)


src/utils/logging.cc
```diff
@@ -54,7 +58,7 @@ void Logging::SetFilename(const std::string& filename) {
   Mutex::Lock lock_(mutex_);
   if (filename_ == filename) return;
   filename_ = filename;
-  if (filename.empty()) {
+  if (filename.empty() || filename == kStderrFilename) {
```
Tilps: if filename is kStderrFilename - the buffer should be pumped to std::cerr.

mooskagh: Indeed!

PR #450, opened 2018-10-14
More accurate search timing.

borg323:
This is an alternative to #404 following a suggestion by @mooskagh (https://github.com/LeelaChessZero/lc0/pull/404#discussion_r223188055). Also the nps estimation for smart pruning starts later simplifying the code there a bit.

### DIFF
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -98,7 +98,8 @@ EngineController::EngineController(BestMoveInfo::Callback best_move_callback,
                                    const OptionsDict& options)
     : options_(options),
       best_move_callback_(best_move_callback),
-      info_callback_(info_callback) {}
+      info_callback_(info_callback),
+      move_start_time_(std::chrono::steady_clock::now()) {}
 
 void EngineController::PopulateOptions(OptionsParser* options) {
   using namespace std::placeholders;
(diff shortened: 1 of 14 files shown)


src/engine.cc
```diff
@@ -138,9 +138,9 @@ void EngineController::PopulateOptions(OptionsParser* options) {
 }
 
 SearchLimits EngineController::PopulateSearchLimits(int ply, bool is_black,
-                                                    const GoParams& params) {
+    const GoParams& params, std::chrono::steady_clock::time_point start_time) {
   SearchLimits limits;
-  limits.time_ms = params.movetime;
+  limits.movetime = params.movetime;
```
Tilps: I think rather than this - limits.search_deadline should be set to start_time + params.movetime in the return on line 154, if infinite is not true.

borg323: I had it like that initially, but according to the uci spec, the meaning of `movetime` is "search exactly x mseconds" so I changed it.

borg323: At least [cutechess](https://github.com/cutechess/cutechess/blob/25e6ed6204698932b6db577316d2c088d2a0a0a6/projects/lib/src/uciengine.cpp#L266) uses `movetime` so this requires further analysis.

Tilps: Given the existing lack of lag time subtraction from movetime - I think we can say this is at least not a regression and so leave it to a follow up PR at some future point.

src/mcts/search.cc
```diff
@@ -65,7 +65,15 @@ Search::Search(const NodeTree& tree, Network* network,
       initial_visits_(root_node_->GetN()),
       best_move_callback_(best_move_callback),
       info_callback_(info_callback),
-      params_(options) {}
+      params_(options) {
+  if (limits_.movetime >= 0) {
```
mooskagh: It feels like movetime should be removed from limits struct completely, and calling code should instead compute deadline.

PR #454, opened 2018-10-17
Honor 'go nodes X' and do actually evalute X number of nodes.

hans-ekbrand:
[EDIT: it seems I did not make this patch against the latest version, so it's not current. I'm sorry for that, and I'd like to cancel it is make a new PR, but I'm not sure how to delete this PR.]

Honor 'go nodes X' and do actually evalute X number of nodes even if `--futile-search-aversion` is non-zero i.e. do not return early when the current best move cannot be surpassed by other moves.

This makes it possible to re-play a game, which is crucial for the analysis of NPS `lc0` delivers under different circumstances.

I have written a script that automatically replicates any previously played game, using the log file and UCI commands, and extract the NPS for each move, while simultaneously monitoring the GPU(s). Using that script I have been able to create data about how NPS changes with hardware and with the position, see the graphs below for examples of the kind of analysis made possible by this patch.

image

The graph above shows an analysis of the first 14 moves of the 16th game Lc0 vs Ethereal in the TCEC cup, @scs-ben provided the log file. The game was replicated on a system with 6 x GTX 1060 under different number of GPU(s) active. Missing data points are due to diverging bestmoves.

k80

This is the same game, replayed on a system of K80s. I'm also planning to recreate the game using various number of V100s.

The script for automatic replay of games is available [here](http://hansekbrand.se/code/re-play-lc0.pl). If you want to include that script into the lc0 repo, are most welcome to do so. I am still working on the data analysis of the GPU usage, and will post graphs which includes GPU-usage data overlayed on the NPS graphs.

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -43,12 +43,60 @@
 
 namespace lczero {
 
+const char* Search::kMiniBatchSizeStr = "Minibatch size for NN inference";
+const char* Search::kMaxPrefetchBatchStr = "Max prefetch nodes, per NN call";
+const char* Search::kCpuctStr = "Cpuct MCTS option";
+const char* Search::kTemperatureStr = "Initial temperature";
+const char* Search::kTempDecayMovesStr = "Moves with temperature decay";
+const char* Search::kTemperatureVisitOffsetStr = "Temperature visit offset";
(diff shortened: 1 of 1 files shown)


hans-ekbrand:
I'll redo it against the current version of lc0.

PR #455, opened 2018-10-17
Honor 'go nodes X' when --futile-search-aversion is non-zero

hans-ekbrand:
Honor 'go nodes X' and do actually evalute X number of nodes even if --futile-search-aversion is non-zero i.e. do not return early when the current best move cannot be surpassed by other moves.

This patch makes it possible to re-play a game, which is crucial for the analysis of NPS lc0 delivers under different circumstances.

I have written a script that _automatically replicates any previously played game_, using extracts from the log file and UCI commands, and extracts the NPS for each move, while simultaneously monitoring the GPU(s). Using that script I have been able to create data about how NPS changes with hardware and with the position, see the graphs below for examples of the kind of analysis made possible by this patch.

image
The graph above shows an analysis of the first 14 moves of the 16th game Lc0 vs Ethereal in the TCEC cup, @scs-ben provided the log file. The game was replicated on a system with 6 x GTX 1060 under different number of GPU(s) active. Missing data points are due to diverging bestmoves.

foo
This is the same game, replayed on a system of K80s, and _including GPU utilisation data_ - shown with the dashed lines. I'm also planning to recreate the game using various number of V100s.

The script for automatic replay of games is available [here](http://hansekbrand.se/code/re-play-lc0.pl). If you want to include that script into the lc0 repo, you are most welcome to do so. 

The graphs suggests a number of hypotheses about lc0:s performance, as measured by NPS, in general and regards to multiGPU-scaling in particular.

- Different chess positions leads to differerent NPS.
- The earliest positions in this game are relatively fast positions.
- In this game the position at move 7, and the positions a few moves after that, are relatively slow positions (those positions came after a rather unexpected capture on b2 by Ethereal).
- NPS at a given position is non-random, when repeated, almost the same NPS will be reached.
- MultiGPU-scaling seems to vary so that fast positions scales better than slow positions.
- GPU utilisation is negatively correlated with number of GPUs.
- It looks as if GPU utilisation is also negatively correlated with NPS - at least when there are many GPUs in use. Is NPS calculated in a way that could explain that rather odd correlation?
- Even if there are similarities between on the one hand V100 (cudnn-fp16) and on the other GTX 1060 and K80 (both cudnn), further research on the V100 is needed.

### DIFF
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -210,14 +210,15 @@ class Node {
   // Debug information about the node.
   std::string DebugString() const;
 
+  // TODO: shrink the padding on this somehow? It takes 16 bytes even though
+  // only 10 are real! Maybe even merge it into this class??
+  EdgeList edges_;
+
  private:
   // To minimize the number of padding bytes and to avoid having unnecessary
   // padding when new fields are added, we arrange the fields by size, largest
   // to smallest.
 
-  // TODO: shrink the padding on this somehow? It takes 16 bytes even though
-  // only 10 are real! Maybe even merge it into this class??
-  EdgeList edges_;
 
   // 8 byte fields.
   // Pointer to a parent node. nullptr for the root.
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -210,15 +210,14 @@ class Node {
   // Debug information about the node.
   std::string DebugString() const;
 
-  // TODO: shrink the padding on this somehow? It takes 16 bytes even though
-  // only 10 are real! Maybe even merge it into this class??
-  EdgeList edges_;
-
  private:
   // To minimize the number of padding bytes and to avoid having unnecessary
   // padding when new fields are added, we arrange the fields by size, largest
   // to smallest.
 
+  // TODO: shrink the padding on this somehow? It takes 16 bytes even though
+  // only 10 are real! Maybe even merge it into this class??
+  EdgeList edges_;
 
   // 8 byte fields.
   // Pointer to a parent node. nullptr for the root.
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -754,9 +754,12 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend(
         // best_move_node_ could have changed since best_node_n was retrieved.
         // To ensure we have at least one node to expand, always include
         // current best node.
+	// However, if given a certain node count to evaluate with `go nodes X`,
+	// then do not make this shortcut.
         if (child != search_->best_move_edge_ &&
             search_->remaining_playouts_ <
-                best_node_n - static_cast<int>(child.GetN())) {
+                best_node_n - static_cast<int>(child.GetN()) &&
+	    search_->limits_.visits == 0) {
           continue;
         }
         // If root move filter exists, make sure move is in the list.
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -36,8 +36,6 @@
 #include <sstream>
 #include <thread>
 
-#include <cstdlib>
-
 #include "mcts/node.h"
 #include "neural/cache.h"
 #include "neural/encoder.h"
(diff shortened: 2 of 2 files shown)


hans-ekbrand:
The V100 is very different from the others I've tested. The difference between 1 and 2 is small but significant. 2 vs 3 is no difference at all, and there is a small difference between 2 and 4, but only on the fast positions early in this game, between position 5 and onwards (move 7 in the real game, since there were two opening book moves) 2 is basically as fast as 4.

foo



Next graph includes GPU-utilisation data using dashed lines. The scale for those are on the right hand side.

foo

The GPU-utilisation is very different from the K80 and GTX 1060 in that only the single GPU case has a decent level of GPU-utilisation. I monitored CPU usage while the test was running, and the CPU:s were never even close to saturated, most of the time top showed 230-250% CPU usage on the lc0 binary, which had 3 or 4 threads, so it should have been 300% or 400% if it were saturated.

hans-ekbrand:
> I may be missing something, but what is the problem with setting `--futile-search-aversion=0` (`--smart-pruning-factor=0` for v0.19) in your use case?

It fails to replicate the original game, since it comes up with other bestmoves.

This is the output of my script when run with this PR applied and the default value 1.33 on `futile-search-aversion`:
```With PR 455 applied

$ perl ~/re-play-lc0.pl nps.csv
Started to monitor the GPU, using process ID 994
Started lc0, using process ID 1000
       _
|   _ | |
|_ |_ |_| v0.18.1 built Oct 15 2018
Found configuration file: lc0/build/release/lc0.config
Found pb network file: lc0/build/release/networks/994de36c13ddf2540bf588817fcca0cb4e279e61c7d118c3bccc9c74419caecd
Creating backend [multiplexing]...
Creating backend [cudnn]...
Creating backend [cudnn]...
Creating backend [cudnn]...
Creating backend [cudnn]...
Creating backend [cudnn]...
Creating backend [cudnn]...
nps: 28632; bestmove: c4d5
nps: 36760; bestmove: b1c3
nps: 27335; bestmove: c1f4
nps: 16961; bestmove: e2e3
nps: 11020; bestmove: f1b5
nps: 12227; bestmove: g1e2
nps: 13060; bestmove: e1g1
nps: 11948; bestmove: f2f3
nps: 14650; bestmove: b5a4
nps: 13764; bestmove: e3e4
nps: 13376; bestmove: f3e4
nps: 13851; bestmove: g1h1
nps: 16393; bestmove: e2c3
nps: 12799; bestmove: g1h1
nps: 13913; bestmove: d4d5
nps: 15177; bestmove: d5c6
```
And this is vanilla lc0, with `futile-search-aversion=0`
```
$ perl ~/re-play-lc0.pl nps.csv
Started to monitor the GPU, using process ID 1502
Started lc0, using process ID 1504
       _
|   _ | |
|_ |_ |_| v0.18.1 built Oct 16 2018
Found configuration file: lc0/build/release/lc0.config
Found pb network file: lc0/build/release/networks/994de36c13ddf2540bf588817fcca0cb4e279e61c7d118c3bccc9c74419caecd
Creating backend [multiplexing]...
Creating backend [cudnn]...
Creating backend [cudnn]...
Creating backend [cudnn]...
Creating backend [cudnn]...
Creating backend [cudnn]...
Creating backend [cudnn]...
nps: 29039; bestmove: c4d5
nps: 37064; bestmove: b1c3
nps: 27842; bestmove: c1f4
nps: 16472; bestmove: e2e3
nps: 12173; bestmove: f1b5
nps: 12478; bestmove: g1e2
nps: 12454; bestmove: e1g1
nps: 12647; bestmove: a2a3
nps: 11156; bestmove: b5a4
nps: 18010; bestmove: e3e4
nps: 13929; bestmove: f3e4
nps: 13509; bestmove: c3e4
nps: 12043; bestmove: e2c3
nps: 13620; bestmove: g1h1
nps: 13036; bestmove: d4d5
nps: 14384; bestmove: d5c6
```
As you can see, here a2-a3 is considered the bestmove after castles, but in the original game f2-f3 was played.

borg323:
Are the last two commits related to this PR?

hans-ekbrand:
@borg323 No, sorry about that. We're working on other things, I didn't know that changes to my fork would propagate back here. How can I avoid that?

borg323:
In general it is a good idea to work on a separate branch for each pr/feature.

mooskagh:
As discussed, there's no good way to get exact nodes when smart-pruning-factor is not 1.0.
Closing this for now, feel free to reopen if you don't agree.

hans-ekbrand:
Nah, it's OK. It was useful to debug NPS variation, but I don't need it anymore.

PR #459, opened 2018-10-22
SE support for OpenCL backend

Ttl:
TODO for SE-Resnet support:

- [x] OpenCL support. This PR.
- [x] Cudnn support. @ankan-ban is working on this.
- [x] Blas support. I have working implementation in LZ that should be relatively easy to port.
- [x] Training code support. At https://github.com/Ttl/lczero-training/tree/se version that writes .txt weights.
- [x] ProtoBuf support.

This is completely backwards compatible with the old networks. Currently supports reading weights in .txt format. 128x10 CCRL dataset weight trained with value head weight 1.0 available at: http://hforsten.com/leelaz/128x10-film-sig-200000.pb.gz

Test against 128x10 CCRL baseline trained with value weight 1.0 at 1600 nodes TC (Equivalent except for the SE-units):
```
Score of lc0_se vs lc0: 204 - 90 - 106 [0.642]
Elo difference: 101.84 +/- 30.01, LOS: 100.00 %
```


gsobala:
This compiles and runs ok on Macos.

I do note that all is fine if the se-resnet is specifically identified by a `-w` command-line flag or by uci option, but it is not auto-discovered.

Ttl:
Blas support added. I don't know much about protobufs and I'm not working on it at the moment.

mooskagh:
I'm working on adding support of multiple network architectures (it will be protobuf which goes directly to backends, and I will remove text weights support completely).
This PR should only be merged after that PR is merged.

ankan-ban:
My WIP cudnn backend changes are here: https://github.com/ankan-ban/lc0/tree/se
still some bugs to figure out...

ankan-ban:
Thanks to Ttl's help, cudnn/cudnn-fp16 backends now produce identical results as Blas/OpenCL. I can create a pull request after Ttl's changes have been merged.

Ttl:
Protobuf support needs updated protobuf definition from: https://github.com/LeelaChessZero/lczero-common/pull/5

Training code is done, but I need to test and clean it up some more before making a pull request.

ankan-ban:
I will update cudnn backend to support FiLM-like SE-unit. 

Btw, I just pushed a fused implementation of (regular) SE to my fork (https://github.com/ankan-ban/lc0/commit/1931d8ebc4ed260938b51ea5bd29fedf858ac0d9). 

Right now only fp16 is supported (also need some refactoring to make the compiled binary work correctly on all GPUs).

RAW NPS benchmarks:
`Size      implementation     %slower     notes
128x10    SE-old             39%         128x10-baseline-vw-1-200000 vs 128x10-se2-200000_quantized
128x10    SE-fused           19%         -same as above-

256x20    SE-old             19%         test10-11248 vs 256x20-se2-200000_quantized
256x20    SE-fused            7%         -same as above-`

ankan-ban:
I pushed a change to support se-FiLM for cudnn backend (https://github.com/ankan-ban/lc0/commit/a197478f63b13271a02a33140c6d6e8db9880b41). 

Right now only the un-fused path works (both fp32 and fp16). 

@Ttl , I am seeing huge slowdown with se-FiLM (>80knps to 20knps for fp16). This was not happening with regular SE.

The problem seems to be due to massive reduction of average batch size coming to the backend. When I do 'go nodes 1000000' with default settings (minibatch-size=256), normally I see batches of  ~200. With the SE-FiLM network (128x10-film-sig-200000.pb.gz), I am seeing avg batch size ~60. 

(In the cudnn backend, you can enable DEBUG_RAW_NPS compile time switch to see these stats).

ankan-ban:
My code for se-FiLM had a bug which was causing the huge-slowdown (it was also producing incorrect result for batch sizes > 1). Thanks Ttl for spotting the bug. I have pushed a fix into my fork. 

The performance raw NN eval performance for 128 filter fp16 network is ~40% lower than baseline (no-SE) network which is similar to what I was seeing with plain SE.

Will try to implement a fused kernel now.

ankan-ban:
I created a PR with cudnn backend changes: https://github.com/LeelaChessZero/lc0/pull/496
Also did some refactoring of cudnn/cuda code to make it more manageable.

borg323:
Thanks @Ttl for pointing me to the pb network file. https://github.com/borg323/lc0/commit/b5c2500f4d5c9d7ca1c56b0918646ed27cc24ff3 fixes the merge failures and now runs OK with the se net.

src/neural/blas/se_unit.cc
```diff
@@ -0,0 +1,94 @@
...
+void SEUnit::global_avg_pooling(const size_t channels,
+                                const float* input,
+                                float* output) {
+
+  for (auto c = size_t{0}; c < channels; c++) {
+    auto acc = 0.0f;
+    for (auto i = size_t{0}; i < kSquares; i++) {
+      acc += input[c * kSquares + i];
+    }
+    output[c] = acc/kSquares;
```
mooskagh: This file surely needs clang-format

src/neural/blas/se_unit.h
```diff
@@ -0,0 +1,58 @@
...
+
+namespace lczero {
+
+// Squeeze-excitation unit
+class SEUnit {
+ public:
+  SEUnit() = delete;
+
+  // Batched forward inference.
+  static void Forward(const size_t batch_size, const size_t channels,
```
mooskagh: What's the point of having a class with a single static function? Why not make it just a free-standing function (and to put what's now in private section into anonymous namespace of .cc file)

src/neural/cuda/network_cudnn.cc
```diff
@@ -173,6 +173,9 @@ class CudnnNetwork : public Network {
     // 0. Process weights.
     processConvBlock(weights.input, true);
     for (auto i = size_t{0}; i < numBlocks_; i++) {
+      if (weights.residual[i].has_se) {
```
mooskagh: It should not be checked at this level. There is `network_format()` check, and LegacyWeights should not be changed anyway.

src/neural/network_legacy.cc
```diff
@@ -21,6 +21,7 @@
 #include <cmath>
 #include <algorithm>
 #include "utils/weights_adapter.h"
+#include "utils/logging.h"
```
mooskagh: Not in alphabetical order.

borg323: Actually this is a leftover from a debugging session, should be removed.

src/neural/network_legacy.h
```diff
@@ -45,14 +45,26 @@ struct LegacyWeights {
...
+    Vec b1;
+    Vec w2;
+    Vec b2;
+  };
+
   struct Residual {
     explicit Residual(const pblczero::Weights::Residual& residual);
     ConvBlock conv1;
     ConvBlock conv2;
+    SEunit se;
```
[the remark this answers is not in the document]
borg323: I agree with @Ttl here. The changes would be easier to do separately for each backend. Also technically this pr predates LegacyWeights.

PR #460, opened 2018-10-23
Allow use of old flag names as alternatives to current ones, and print warning if used.

borg323:
As an example, `--futile-search-aversion`, `--debuglog` and `allowed-node-collisions` are added as alternatives to the new flags.

### DIFF
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -377,6 +377,7 @@ EngineLoop::EngineLoop()
               options_.GetOptionsDict()) {
   engine_.PopulateOptions(&options_);
   options_.Add<StringOption>(kLogFileId);
+  options_.AddDeprecatedFlag(kLogFileId.GetId(), "debuglog");
 }
 
 void EngineLoop::RunLoop() {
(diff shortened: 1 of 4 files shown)


mooskagh:
I think it's better not to have deprecated flags.
It totally fine that Lc0 will fail to start, a person looks why it happens, and changes command line, rather than introducing a mess with command lines.

hans-ekbrand:
I was recently hit by the change from debuglog to logfile, still I think it's perfectly fine to not support old names.

borg323:
My idea was to have a simple low overhead mechanism for supporting old flags, and do so just for a single release. Something that is easy to write, like `options_.AddDeprecatedFlag(kLogFileId.GetId(), "debuglog");` (from the patch) and localized to be easy to delete before the next release.

mooskagh:
I think that makes more harm than good.
If flags changed, we want users to be aware of that as soon as possible. It's not that we have some external legacy code which runs lc0 and we should work around. If you think it is a problem, I'd rather just keep old command line flags.

For hiding options I'm also rather skeptical that it has any positive use.
If option is useless for a user and noone touches it ever (like for move curve params), we should replace it with a constant in a source code and kill 10 lines of code, instead of having additional 10 lines of code to trick with hiding.

Adding code is bad, removing code is great.

borg323:
I wasn't arguing, just explaining my motivation.

Regarding the option hiding, I agree that anything hidden is a candidate for turning into a constant. Option hiding allows it to be tuned (e.g. using clop) using the exact same build and when we have good defaults the can be removed permanently. I expect the hiding mechanism to be unnecessary (and removed) after 2-3 releases.

PR #461, opened 2018-10-24
make stop_ atomic and check it in GatherMinibatch()

borg323:
Also remove mutex from IsSearchActive() and GUARDED_BY annotation from stop_.

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -280,7 +280,7 @@ void Search::MaybeTriggerStop() {
     FireStopInternal();
   }
   // If we are the first to see that stop is needed.
-  if (stop_ && !responded_bestmove_) {
+  if (stop_.load(std::memory_order_acquire) && !responded_bestmove_) {
     SendUciInfo();
     if (params_.GetVerboseStats()) SendMovesStats();
     best_move_ = GetBestMoveInternal();
@@ -519,7 +519,7 @@ void Search::RunBlocking(size_t threads) {
 }
 
 bool Search::IsSearchActive() const {
-  return !stop_;
+  return !stop_.load(std::memory_order_acquire);
 }
 
 void Search::WatchdogThread() {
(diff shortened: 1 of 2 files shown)


src/mcts/search.cc
```diff
@@ -519,7 +519,6 @@ void Search::RunBlocking(size_t threads) {
 }
 
 bool Search::IsSearchActive() const {
-  Mutex::Lock lock(counters_mutex_);
   return !stop_;
```
mooskagh: `!stop_.load()`
or `!stop_.load(std::memory_order_acquire)` to make it even faster, at all places where `stop_.load()` is called (no performance penalty due to being atomic at all on x86)

Similarly, all places where value is stored into the atomic, should be `stop_.store(value)`, or `stop_.store(value, std::memory_order_release)`.

src/mcts/search.cc
```diff
@@ -692,6 +688,8 @@ void SearchWorker::GatherMinibatch() {
       --minibatch_size;
       ++number_out_of_order;
     }
+    // Check for stop at the end so we have at least one node.
+    if (search_->stop_) return;
```
mooskagh: `if (search_->stop_.load())`

src/mcts/search.cc
```diff
@@ -692,6 +689,8 @@ void SearchWorker::GatherMinibatch() {
       --minibatch_size;
       ++number_out_of_order;
     }
+    // Check for stop at the end so we have at least one node.
+    if (search_->stop_.load(std::memory_order_acquire)) return;
```
Tilps: I think this check should also be added above the continue for processing a collision.

Tilps:
I wonder if this may not be very effective due to prefetching then going and filling the remaining slots in the early out batch.

mooskagh:
Indeed, we also should not prefetch anything if `stop_`, and maybe stop prefetching in the middle too.

src/mcts/search.cc
```diff
@@ -542,17 +541,17 @@ void Search::WatchdogThread() {
...
+      watchdog_cv_.wait_for(lock.get_raw(), remaining_time, [this]() {
+        return stop_.load(std::memory_order_acquire);
+      });
     }
     MaybeTriggerStop();
   }
   MaybeTriggerStop();
 }
 
 void Search::FireStopInternal() REQUIRES(counters_mutex_) {
```
mooskagh: Doesn't really require `counters_mutex_` anymore, so this annotation can be removed.

borg323: Done

PR #463, opened 2018-10-24
remove movetime from limits

borg323:
I'm not entirely happy with the selfplay changes. Any suggestions for improvement are welcome.
Also it isn't very easy to test this interactively, since time is running while you type. The best way is to write some uci commands in a text editor and paste them as a bloc, something like:
```
position startpos
go movetime 1000
```

### DIFF
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -396,9 +396,7 @@ void EngineController::Go(const GoParams& params) {
 
   if (limits.search_deadline) {
     LOGFILE << "Timer started at "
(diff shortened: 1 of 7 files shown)


mooskagh:
`movetime` is mainly used for debugging and selfplay, so I'm starting to think that the way it works now (without this change) is actually better..
Search deadline is an "external" thing, while movetime is something internal to the search.

borg323:
I thought so as well, but I checked and at least cutechess uses movetime on some occasions.

mooskagh:
Looks good!

PR #465, opened 2018-10-26
Refactor network loading

borg323:
Using 2 helper functions in NetworkFactory, one to add the options and one to do the actual loading.

### DIFF
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -45,15 +45,6 @@ const OptionId kThreadsOptionId{"threads", "Threads",
 const OptionId kLogFileId{"logfile", "LogFile", "Write log to that file.", 'l'};
 const OptionId kNNCacheSizeId{"nncache", "NNCache",
                               "Number of positions to store in cache."};
-const OptionId kWeightsId{"weights", "WeightsFile",
-                          "Path from which to load network weights.\n"
-                          "Setting it to <autodiscover> makes it search "
-                          "in ./ and ./weights/ subdirectories for the latest "
-                          "(by file date) file which looks like weights.",
-                          'w'};
-const OptionId kNnBackendId{"backend", "Backend", "NN backend to use."};
-const OptionId kNnBackendOptionsId{"backend-opts", "BackendOptions",
-                                   "NN backend parameters."};
 const OptionId kSlowMoverId{
     "slowmover", "Slowmover",
     "Parameter value X means that whole remaining time is split in such a way "
@@ -80,8 +71,6 @@ const OptionId kSpendSavedTimeId{"immediate-time-use", "ImmediateTimeUse",
                                  "Fraction of saved time to use immediately."};
 const OptionId kPonderId{"ponder", "Ponder", "This option is ignored."};
 
-const char* kAutoDiscover = "<autodiscover>";
-
 float ComputeMoveWeight(int ply, float peak, float left_width,
                         float right_width) {
   // Inflection points of the function are at ply = peak +/- width.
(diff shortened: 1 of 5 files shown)


src/engine.cc
```diff
@@ -236,31 +220,8 @@ void EngineController::UpdateFromUciOptions() {
...
-    std::cerr << "Loading weights file from: " << net_path << std::endl;
-  }
-  Weights weights = LoadWeightsFromFile(net_path);
-
-  OptionsDict network_options(&options_);
-  network_options.AddSubdictFromString(backend_options);
-
-  network_ = NetworkFactory::Get()->Create(backend, weights, network_options);
+  std::unique_ptr<Network> tmp_net = NetworkFactory::LoadNetwork(options_);
+  if (tmp_net) network_.swap(tmp_net);
```
mooskagh: Why is that? I think if network failed to load, it's better to do something than ignore that. E.g. throw an exception?

Also with unique_ptrs, it's more idiomatic to write `network_ = std::move(tmp_net);` or directly `network_ = NetworkFactory::LoadNetwork(options_);`

src/neural/factory.h
```diff
@@ -80,6 +93,11 @@ class NetworkFactory {
 
   std::vector<Factory> factories_;
   friend class Register;
+
+  // Cached values.
+  static std::string network_path_;
```
mooskagh: Why is this static? What happens in selfplay when player1 and player2 use different network files and/or backends?

src/engine.cc
borg323: Here, nullptr is not an indication of failure to load but that we are reusing the already loaded network, so it souldn't change. This was the cleaner way I found to refactor the load logic from engine.cc and tournament.cc. I'll do the change to move(), it has to be conditional.

src/neural/factory.h
mooskagh: Maybe store `BackendConfiguration network_configuration_` (see comment in tournament.h what it is) in `EngineController` (in addition to network_), and then load new network if `network_configuration_ != BackendConfiguration(options)`.
E.g. let calling code watch what is loaded and when there's need to reload network, rather than do that inside factory.

src/neural/factory.h
```diff
@@ -51,13 +52,37 @@ class NetworkFactory {
...
+  static const OptionId kBackendId;
+  static const OptionId kBackendOptionsId;
+
+  struct BackendConfiguration {
+    BackendConfiguration() {}
+    BackendConfiguration(const OptionsDict& options);
+    std::string weights_path;
+    std::string backend;
+    std::string backend_options;
+    bool operator==(const BackendConfiguration& other);
```
mooskagh: `const` (and operator!= also const)

src/neural/factory.h
```diff
@@ -51,13 +52,37 @@ class NetworkFactory {
...
+  // if no network options changed since the previous call.
+  static std::unique_ptr<Network> LoadNetwork(const OptionsDict& options);
+
+  // Parameter IDs.
+  static const OptionId kWeightsId;
+  static const OptionId kBackendId;
+  static const OptionId kBackendOptionsId;
+
+  struct BackendConfiguration {
+    BackendConfiguration() {}
```
mooskagh: In C++11 it's better to have `= default` instead of `{}`.
In some cases writing `{}` prevents the class from autogenerating a move constructor while `= default` doesn't.
Doesn't make a difference in this case but still a good habit.

src/neural/factory.h
borg323: I missed this comment before merging. I got some compile error when I tried it earlier and gave up. I can make a follow-up if you think it is important.

mooskagh: Probably not worth a separate PR, but I think I'll change that together with some other change, if not forget.

borg323: I can do it with #357 later today.

PR #466, opened 2018-10-27
Tunable Contempt

jjoshua2:
Seems useful to have as an adjustable parameter with defaults at 0.


src/mcts/params.cc
```diff
@@ -92,6 +102,8 @@ void SearchParams::Populate(OptionsParser* options) {
   options->Add<IntOption>(kMiniBatchSizeId, 1, 1024) = 1;
   options->Add<IntOption>(kMaxPrefetchBatchId, 0, 1024) = 32;
   options->Add<FloatOption>(kCpuctId, 0.0f, 100.0f) = 1.2f;
+  options->Add<FloatOption>(kContemptId, -1.0f, 1.0f) = 0.005f;
+  options->Add<FloatOption>(kContempt2Id, -1.0f, 1.0f) = 0.015f;
```
borg323: The default for both options should be zero, and then changed in engine.cc (as is done already). This will make it safe for training.

zz4032:
I applied the suggested changes. I would further suggest to rename the parameter names in something like 1) ContemptFactor and ContemptShift or 2) PiecesContemptFactor and PiecesContemptShift or 3) AvoidTradingFactor and AvoidTraiding Shift.

I couldn't detect any measurable improvement with this code myself, so I don't think this is likely to be commited unless there was something wrong in my tests.

jjoshua2:
@zz4032 would you mind taking out the `tree_->IsBlackToMove())` and bool is_black stuff? it didn't end up getting used here.

5+2 results against weaker engines
```
   0 lc0contempt                   585     nan      15   96.7%    6.7%
   1 Vajolet2_2.5 30CPU           -417     nan       6    8.3%   16.7%
   2 Nirvanachess 2.4 30CPU       -inf     nan       4    0.0%    0.0%
   3 wasp260-x64_no_bmi 30CPU     -inf     nan       5    0.0%    0.0%
```

jjoshua2:
```@jjosh , @Swynndla Here are results of 5820k (12 threads) vs 2080Ti, 5m+2s games:
Lc0-baseline - stockfish_dev : 35.0/78 4-12-62 (==1=00===============0=====0==============1========0===00=1==0======00=====010)  45%   -35


Trade Tariff (default: 0.003)
Lc0-contempt - stockfish_dev : 38.5/79 7-9-63 (1==0===0===0===========011==01=======0=====00=1====0==1===============1========)  49%    -7



Testing with higher trade tariff. Note: Different conditions (1m+1s per game, OC 2070 vs 10 threads of overclocked 5820k)
Lc0-contempt-0.006 - stockfish_dev       : 85.5/220 9-58-153  39%   -78```

**Comment by jjoshua2** [comment]
```Rank Name                          Elo     +/-   Games   Score   Draws
   1 lc0v19 11248                   15      15     612   52.1%   69.0%
   2 lc0contempt 16                 -1      15     610   49.9%   69.7%
   3 lc0contempt                    -1      16     612   49.8%   65.7%
   4 lc0contempt .006              -13      16     610   48.1%   67.4%```
So contempt isn't hurting self elo that much, but .006 is starting to be a lot, so is a sane max value. But sitll against an opponent 300 elo lower it doesn't matter if elo is 50 worse against SF if it significantly lowers draw rate.

**Comment by jjoshua2** [comment]
```   # PLAYER                 : RATING  ERROR   POINTS  PLAYED    (%)
   1 lc0contempt .0015 8    :   2537    194     23.0      24   95.8%
   2 lc0contempt 0          :   2462    207     22.5      24   93.8%
   3 lc0contempt .006 8     :   2288    157     19.5      23   84.8%
   4 lc0contempt .003 8     :   2240    138     19.5      24   81.3%
   5 Vajolet2_2.5 30CPU     :   1973     72     10.5      95   11.1%
```
that's 10s+0.5s games. i think someone else tested 1+0.5s in the cloud recently? @Swynndla
```  # PLAYER                 : RATING  ERROR   POINTS  PLAYED    (%)
   1 lc0contempt .0015 8    :   2490    157     32.0      34   94.1%
   2 lc0contempt .006 8     :   2372    113     28.5      32   89.1%
   3 lc0contempt 0          :   2334    107     29.5      34   86.8%
   4 Vajolet2_2.5 30CPU     :   2004     65     10.0     100   10.0%
```
this is 30s+1s games

mooskagh:
I don't understand this change.

1. It adds "contempt" to all nodes at even depth and subtracts "contempt" from all nodes at odd depth?
Why would it be improving play? It just biases Q at root depending on whether most of nodes are at even depth or odd depth.

2. Why is amount of "contempt" proportional to number of pieces at root of the search? After tree reuse, V of the a becomes different, so it makes a difference whether the node is evaled before or after the move, which breaks lots of assumptions.

3. During opponent's move (ponder or e.g. just analyzing with `go infinite`) "contempt" changes its sign (as even depth nodes become odd depth and vice versa). Again, old nodes from tree reuse have opposite sign of "contempt" from the fresh nodes.

jjoshua2:
I have to admit it's not doing exactly what I thought it was doing at first, and @swylandia has tried many variations that are more what I was planning on originally that address a lot of your concerns, but so far they have not done as well in testing. Here is more results at 30+0.5 that show 99% LOS over default.
```
ResultSet-EloRating>los
                    lc lc Va
lc0contempt .0015      99100
lc0contempt 0        0   100
Vajolet2_2.5 30CPU   0  0
ResultSet-EloRating>ratings
Rank Name                 Elo    +    - games score oppo. draws
   1 lc0contempt .0015    132   40   40   392   90%  -199   18%
   2 lc0contempt 0         67   35   35   391   85%  -199   25%
   3 Vajolet2_2.5 30CPU  -199   26   26   783   13%   100   21%
```

mooskagh:
I think what it is currently doing is roughly increasing FPU reduction (but instead of reducing eval of unvisited nodes it increases eval of visited, if subtree is shallow). Tweaking FPU reduction can indeed change strength, but that's totally different from the original idea of contempt to avoid draws.

If there is some effect which indeed improves strength, it has to be understood and that idea has to be implemented then.

In the current state it's both not does what it was intended to do, and has known issues with implementation (as far as I understand it), so it should not be merged.

jjoshua2:
I think the only real problem with is that it shouldn't be used in ponder without more testing or a negation if currently pondering. (And after 30 quick games it looks like currently its not helping when ponder is on, but it wasn't really supposed to anyway, as it's designed for TCEC/CCCC matches, not analysis or other things) When default contempt is 0 it will work with ponder just fine.

SF has contempt similarly where it adds to its own score and subtracts from opponent's score. Also SF has dynamic contempt so depending on the current score (when it's good) it will use more contempt. Piece count is a proxy for that. We could use our score instead, but we can't trust our score, since it can be quite high in endgames for no reason, or quite neutral or even negative in opening, when we know we have a 80% win chance just because of a weak opponent.

I can try a match with slightly higher FPU reduction and see if that also has strength gains, but I doubt it will help as much if at all. FPUR was already tuned for test10 and found 1.2 to be optimal, so we are unlikely to be able to increase it for strength gain.

jjoshua2:
Turns out my declaration of no improvement after 30 games was premature!
```ResultSet-EloRating>los
                    lc lc Va
lc0contempt .0015      94 99
lc0contempt          5    99
Vajolet2_2.5 30CPU   0  0
ResultSet-EloRating>ratings
Rank Name                 Elo    +    - games score oppo. draws
   1 lc0contempt .0015    151   87   87    94   91%  -212   13%
   2 lc0contempt 0           61   71   71    93   87%  -212   26%
   3 Vajolet2_2.5 30CPU  -212   55   55   187   11%   106   19%```
Ponder ON 20s + 0.33s. The contempt version seems to have a problem with losing drawn games in endgame fortresses that I think would be solved by using a bigger increment.

**Comment by jjoshua2** [comment]
Stopped 60s+1s at a tad over 95% LOS
```
                    lc lc Va
lc0contempt .0015      95100
lc0contempt 0        4   100
Vajolet2_2.5 30CPU   0  0
ResultSet-EloRating>ratings
Rank Name                 Elo    +    - games score oppo. draws
   1 lc0contempt .0015    129   46   46   294   90%  -206   17%
   2 lc0contempt 0         76   41   41   293   87%  -206   25%
   3 Vajolet2_2.5 30CPU  -206   31   31   587   11%   103   21%
```

jjoshua2:
More good results (for both PRs) posted in other ticket here. https://github.com/LeelaChessZero/lc0/pull/536#issuecomment-441878162

PR #467, opened 2018-10-27
Bring protobufs directly to backends

mooskagh:
It's still converted to the legacy structure in backends constructor, because backends depend on that too much.
Still few things that doesn't work.
1. encode_test fails to compile as it doesn't see proto headers.
2. I wasn't able to build tensorflow backend (that same protobuf version mismatch problem as before)

### DIFF
--- a/src/utils/weights_adapter.cc
+++ b/src/utils/weights_adapter.cc
@@ -25,24 +25,28 @@
   Program grant you additional permission to convey the resulting work.
 */
 
-#pragma once
-
-#include <string>
-#include <vector>
-
-#include "neural/network.h"
+#include "src/utils/weights_adapter.h"
 
 namespace lczero {
+float LayerAdapter::Iterator::ExtractValue(const uint16_t* ptr,
+                                           const LayerAdapter* adapter) {
+  return *ptr / static_cast<float>(0xffff) * adapter->range_ + adapter->min_;
+}
 
-using FloatVector = std::vector<float>;
-using FloatVectors = std::vector<FloatVector>;
+LayerAdapter::LayerAdapter(const pblczero::Weights_Layer& layer)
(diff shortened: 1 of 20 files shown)


src/utils/weights_adapter.cc
```diff
@@ -0,0 +1,52 @@
...
+  return *ptr / static_cast<float>(0xffff) * adapter->range_ + adapter->min_;
+}
+
+LayerAdapter::LayerAdapter(const pblczero::Weights_Layer& layer)
+    : data_(reinterpret_cast<const uint16_t*>(layer.params().data())),
+      size_(layer.params().size() / sizeof(uint16_t)),
+      min_(layer.min_val()),
+      range_(layer.max_val() - min_) {}
+
+std::vector<float> LayerAdapter::as_vector() const {
```
Error323: Why not LayerAdapter::AsVector()?

borg323:
Comparison between blas and tensorflow-cpu backends:
```
Absolute error histogram for a batch of 68
      |                                                                                         |
 0.15 +                 ##                                                                      +
      |                 ###                                                                     |
      |                 ###                                                                     |
      |                ####                                                                     |
      |                ####                                                                     |
  0.1 +                ####                                                                     +
      |                #####                                                                    |
      |               ######                                                                    |
      |               ######                                                                    |
      |               #######                                                                   |
 0.05 +              ########                                                                   +
      |             ##########                                                                  |
      |           # ##########                                                                  |
      |#         ###############                                                                |
      |#      ###############################################                                   |
      +----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+
   -inf  -15  -14  -13  -12  -11  -10   -9   -8   -7   -6   -5   -4   -3   -2   -1    0    1  +inf 

```

PR #469, opened 2018-10-27
remove hash usage display

borg323:


borg323:
Instead of this we will reset the cache usage display to 0 before every move.

PR #472, opened 2018-10-28
Prevent search from overflowing by setting limits.visits to 4B by default.

mooskagh:
Also while `go infinite` never responds bestmove by itself, it still stops search when reaches the limit.
E.g. it's possible to do `go nodes 100000 infinite` now. (Not sure whether useful, but just a side effect)


mooskagh:
Done.

PR #473, opened 2018-10-28
More logging

mooskagh:


src/utils/logging.cc
```diff
@@ -65,17 +65,36 @@ void Logging::SetFilename(const std::string& filename) {
...
+        << ' ' << file << ':' << line << "] ";
 }
 
 LogMessage::~LogMessage() { Logging::Get().WriteLineRaw(str()); }
 
+StderrLogMessage::StderrLogMessage(const char* file, int line)
+    : log_(file, line) {}
+
+StderrLogMessage::~StderrLogMessage() {
+  std::cerr << str() << std::endl;
```
borg323: now that #443 is in, maybe this can be conditional?

src/mcts/search.cc
```diff
@@ -48,8 +48,18 @@ const int kSmartPruningToleranceNodes = 300;
 const int kSmartPruningToleranceMs = 200;
 // Maximum delay between outputting "uci info" when nothing interesting happens.
 const int kUciInfoMinimumFrequencyMs = 5000;
+using std::literals::chrono_literals::operator""ms;
```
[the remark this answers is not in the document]

src/utils/logging.cc
mooskagh: It's not very trivial (I think it's better to output without timestamp in that case), so I'll do that later.

borg323:
To make the older vc compiler happy (for appveyor), stack overflow suggests:
```
std::chrono::time_point<std::chrono::system_clock> SteadyClockToSystemClock(
    std::chrono::time_point<std::chrono::steady_clock> time) {
  return std::chrono::system_clock::now() +
         std::chrono::duration_cast<std::chrono::system_clock::duration>(
             time - std::chrono::steady_clock::now());
}
```

PR #479, opened 2018-10-30
Implement RAM limit.

mooskagh:


src/engine.cc
```diff
@@ -176,6 +186,17 @@ SearchLimits EngineController::PopulateSearchLimits(
...
+  float ram_limit = options_.Get<float>(kRamLimitMbId.GetId());
+  if (ram_limit) {
+    const auto cache_size =
+        options_.Get<int>(kNNCacheSizeId.GetId()) * kAvgCacheItemSize;
+    int64_t limit = (ram_limit * 1000000 - cache_size) / kAvgNodeSize;
+    LOGFILE << "RAM limit " << ram_limit << "MB. Cache takes "
+            << cache_size / 1000000 << "MB. Remaining memory is enough for "
+            << limit << " nodes.";
+    if (limit < 0) limit = 0;
+    if (limit < limits.visits) limits.visits = limit;
```
Tilps: This doesn't account for the ram savings of repeated visits to terminals? Maybe worth explicitly calling that out in the help text?

mooskagh: Added more help.

mooskagh:
@borg323 Fixed! Only check for possible search stop if stop_ is false.

PR #483, opened 2018-11-03
add benchmark features

borg323:
Prints bestmove at the end, which was requested for use in scripts, and adds the `--movetime` option.


ankan-ban:
In some cases, the raw NN eval nps might also be desirable. Currently for cudnn backend, it can be enabled using a compile time switch (DEBUG_RAW_NPS in network_cudnn.cu). It prints out the raw average NPS and average batch size coming to the backend (after every 100 calls). 

Maybe not the default for benchmark mode, but a command line option?

src/benchmark/benchmark.cc
```diff
@@ -38,8 +38,10 @@ const OptionId kNNCacheSizeId{
     "nncache", "NNCacheSize",
     "Number of positions to store in a memory cache. A large cache can speed "
     "up searching, but takes memory."};
-const OptionId  kNodesId{"nodes", "", "Number of nodes to run as a benchmark"};
-const OptionId  kFenId{"fen", "", "Benchmark initial position FEN"};
+const OptionId kNodesId{"nodes", "", "Number of nodes to run as a benchmark"};
```
mooskagh: Now we have a period at the end of every help message. :)

PR #490, opened 2018-11-04
Fix EncodePositionForNN test failure

borg323:
This is in two parts, first moves kStartBoardPos to ChessBoard and second sets Appveyor to always build with WholeProgramOptimization to avoid failing to notice tings like this before the release.

### DIFF
--- a/src/chess/board.h
+++ b/src/chess/board.h
@@ -42,8 +42,8 @@ class ChessBoard {
   ChessBoard() = default;
   ChessBoard(const std::string& fen) { SetFromFen(fen); }
 
-  static const char* kStartingFen;
-  static const ChessBoard kStartBoardPos;
+  static const char* kStartposFen;
+  static const ChessBoard kStartposBoard;
 
   // Sets position from FEN string.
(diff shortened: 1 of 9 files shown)


src/chess/board.h
```diff
@@ -43,6 +43,7 @@ class ChessBoard {
   ChessBoard(const std::string& fen) { SetFromFen(fen); }
 
   static const std::string kStartingFen;
```
mooskagh: Let's make it `const char*` rather than `std::string`. I think this makes it not need initializer at all (e.g. it won't depend on initialization order). I'm not sure it will work but at least I heard that recommendation to have global variables as `const char*` rather than strings.

src/chess/board.h
```diff
@@ -43,6 +43,7 @@ class ChessBoard {
...
+  static const ChessBoard kStartBoardPos;
```
mooskagh: Why not startposBoard?. Maybe for consistency name them `kStartposFen` and `kStartposBoard`?

PR #497, opened 2018-11-08
WIP Tactics Mode

jjoshua2:
To test on WAC


jjoshua2:
Last version scored same on WAC but worse in match play. I think this version will help with both.

jjoshua2:
I think cpuctbase fixes this mostly, and would need changes anyway to try adding noise to cpuct in new formula.

Issue #507, opened 2018-11-14
v0.19.0-rc4 caused an exception(games stopped), loses on time, freezes gui = bugs


all12:
It **freezes** the **Aquarium 2018 gui**, when playing matches or tournaments with ponder **on**.
At 1 minute per game + **1 second per move** time control.
[Aquarium 2018, ponder on, 1+1 LogFile.txt](https://github.com/LeelaChessZero/lc0/files/2579604/Aquarium.2018.ponder.on.1%2B1.LogFile.txt)

It **loses sometimes on time** with ponder **off**, at 1 minute per game + **1 second per move** time control.
But it **doesn't freezes** the gui.
[Aquarium 2018, ponder off, 1+1 LogFile.txt](https://github.com/LeelaChessZero/lc0/files/2579611/Aquarium.2018.ponder.off.1%2B1.LogFile.txt)

**Works fine:)**
It **doesn't freezes** the gui.
It **doesn't loses on time** with ponder **off**, at 1 minute per game + **60 seconds per move** time control.
[Aquarium 2018, ponder off, 1+60 LogFile.txt](https://github.com/LeelaChessZero/lc0/files/2579618/Aquarium.2018.ponder.off.1%2B60.LogFile.txt)

When using the **Fritz 16 gui**, it **caused an exception** in every game.
With ponder **on** and 1 minute per game + **1 second per move** time control.
[Fritz 16, ponder on, 1+1 LogFile.txt](https://github.com/LeelaChessZero/lc0/files/2579628/Fritz.16.ponder.on.1%2B1.LogFile.txt)

It also works fine when using the ChessBase 15 gui:)

Infinite analysis runs fine in every gui.

Help wanted
I don't like release blocker
Let us kill the bugs:)

all12:
@borg323,

it doesn't freezes the gui but it loses on time.
[LogFile.txt](https://github.com/LeelaChessZero/lc0/files/2580619/LogFile.txt)

borg323:
Running out of time with 1 second increment is somewhat likely with the current search overhead, so I haven't looked at the logs yet. Changing `HistoryFill` fixes the symptom, but not the problem - only confirms what was overwritten.

borg323:
@all12 can you try a few of the following lc0 versions and see tell us whether they work or not? You don't need to do all of them. 
[v0.19-rc1](https://ci.appveyor.com/api/buildjobs/9oalbrjdwe6w2prs/artifacts/build%2Flc0.exe)
[...]

Issue #500, opened 2018-11-08
With amd gpus v0.19.0-rc3 is crashing after some time


all12:
@borg323,
the bug is still there.

I already discussed it with many chess players which also has the bug and it turns out that we all have the AMD R9 ... GPUs (different R9 designs).

mooskagh:
@all12 could you maybe record a log file? (--logfile command line option, or LogFile UCI param)

all12:
How can I do these things?

I need a step by step explanation please.

I can use the Fritz 16 gui and CB14 gui and Aquarium 2018 gui if needed.

borg323:
In that case, the easiest thing is probably to make a log file. The GUI should have an engine configuration dialog, where you will find an option named `LogFile`. Depending on the GUI, it may have a file selector, so make a new file and then run as usual. Starting with v0.19 the log file records very detailed information on the internal workings of lc0, which will hopefully help us determine what is happening.

mooskagh:
See also http://talkchess.com/forum3/viewtopic.php?p=778703

all12:
Thx for the link, maybe this would help me later.
I used the Aquarium 2018 gui and here is the logfile:
[LogFile.txt](https://github.com/LeelaChessZero/lc0/files/2563432/LogFile.txt)

borg323:
@all12 do you get crashes even with pondering disabled?

mooskagh:
Seems that something bad happens when `ponderhit` arrives before search starts to search.
I'm not able to reproduce it yet though.

all12:
@borg323,

yes.

Should I do a logfile with ponder disabled?

mooskagh:
```ruby
1108 21:05:29.558816 6648 ..\src/chess/uciloop.cc:218] << bestmove d2d6 ponder d7d6
1108 21:05:29.558843 6648 ..\src/mcts/search.cc:615] End a watchdog thread.
1108 21:05:29.559963 6372 ..\src/chess/uciloop.cc:131] >> setoption name UCI_Chess960 value false
1108 21:05:29.560012 6372 c:\projects\lc0\src\utils/exception.h:39] Exception: Unknown option: UCI_Chess960
1108 21:05:29.560054 6372 ..\src/chess/uciloop.cc:218] << error Unknown option: UCI_Chess960
[...]

Issue #126, opened 2018-06-30
Crash in OpenCL


frpays:
Happened once, but let's keep an eye on it.
It was pr87, but pr87 did not change any logic in opencl (except a really big bug).

https://ghostbin.com/paste/8oqes

```
bestmove e2e4
ASAN:DEADLYSIGNAL
=================================================================
==42381==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x7fffc60ae457 bp 0x700006f6bbe0 sp 0x700006f6bbd0 T5)
==42381==The signal is caused by a READ memory access.
==42381==Hint: address points to the zero page.
    #0 0x7fffc60ae456  (GeForceGLDriver:x86_64+0x2ff456)
    #1 0x7fffc60ac9e3  (GeForceGLDriver:x86_64+0x2fd9e3)
    #2 0x7fffd8502a82 in gldDestroyQueue (libGPUSupportMercury.dylib:x86_64+0x2a82)
    #3 0x7fffcec05783  (OpenCL:x86_64+0x6783)
    #4 0x7fffcec057a9  (OpenCL:x86_64+0x67a9)
    #5 0x7fffcec212be in clReleaseCommandQueue (OpenCL:x86_64+0x222be)
    #6 0x105baf354 in cl::detail::ReferenceHandler<_cl_command_queue*>::release(_cl_command_queue*) cl2.hpp:1534
    #7 0x105baf31a in cl::detail::Wrapper<_cl_command_queue*>::release() const cl2.hpp:1724
    #8 0x105baf270 in cl::detail::Wrapper<_cl_command_queue*>::~Wrapper() cl2.hpp:1656
    #9 0x105baf214 in cl::CommandQueue::~CommandQueue() cl2.hpp:6620
    #10 0x105baea64 in cl::CommandQueue::~CommandQueue() cl2.hpp:6620
    #11 0x105c0f502 in ThreadData::~ThreadData() OpenCL.h:61
    #12 0x105be78e4 in ThreadData::~ThreadData() OpenCL.h:61
    #13 0x7fffdffbe82a in tlv_finalize (libdyld.dylib:x86_64+0x382a)
    #14 0x7fffe01db354 in _pthread_tsd_cleanup (libsystem_pthread.dylib:x86_64+0x5354)
    #15 0x7fffe01db0d8 in _pthread_exit (libsystem_pthread.dylib:x86_64+0x50d8)
[...]

PR #516, opened 2018-11-16
OpenCL: replace thread_local with a resource pool.

frpays:
The per-thread OpenCL resources (OpenCL command queue and buffers) were under `thread_local` management.  When the search threads were created/destroyed, so were these resources, including the OpenCL command queue. Under some architectures, the ASAN (the address sanitiser) reported that destroying the OpenCL command queue was corrupting memory.

This pull request replaces the `thread_local` with a resource pool. The resources will be recycled between threads and the OpenCL command queues never destroyed.

It should fix crashes reported in these issues:

https://github.com/LeelaChessZero/lc0/issues/98
https://github.com/LeelaChessZero/lc0/issues/126
https://github.com/LeelaChessZero/lc0/issues/500
https://github.com/LeelaChessZero/lc0/issues/507

### DIFF
--- a/src/neural/opencl/network_opencl.cc
+++ b/src/neural/opencl/network_opencl.cc
@@ -57,7 +57,7 @@ struct OpenCLWeights {
 
 class OpenCLComputation : public NetworkComputation {
  public:
-  OpenCLComputation(OpenCL_Network& opencl_net, const OpenCLWeights& weights)
+  OpenCLComputation(const OpenCL_Network& opencl_net, const OpenCLWeights& weights)
       : opencl_net_(opencl_net), weights_(weights), policies_(), q_values_() {
     buffers_ = opencl_net.acquire_buffers();
   }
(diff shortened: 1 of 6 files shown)


src/neural/opencl/network_opencl.cc
```diff
@@ -57,11 +57,14 @@ struct OpenCLWeights {
 
 class OpenCLComputation : public NetworkComputation {
  public:
-  OpenCLComputation(const OpenCL_Network& opencl_net,
-                    const OpenCLWeights& weights)
-      : opencl_net_(opencl_net), weights_(weights), policies_(), q_values_() {}
+  OpenCLComputation(OpenCL_Network& opencl_net, const OpenCLWeights& weights)
```
mooskagh: Non-const reference parameters are not allowed. Replace with a pointer. (also in other functions)

frpays: Done for the two constructor-calls with non-const ref. parements: OpenCLComputation and OpenCLBuffers.

Issue #109, opened 2018-06-23
Remove the BLAS dependency from OpenCL backend.


frpays:
This is to be investigated as it would make the build easier for AMD users.

There are 2 BLAS dependencies:

- sgemm reference timing in the OpenCL tuning.  Can it be removed ?

- sgemv at the end of the inference. It can definitely be replaced by a for loop that the compiler will happily vectorize. This implementation will be marginally slower than the blas one, and is not critical for overall backend performance.

Ttl:
sgemm in tuning shouldn't be removed as it is needed to check that the output of the GPU is correct. Some parameters can give incorrect output with certain GPUs. It is possible to replace it with any other sgemm though. It's performance doesn't matter at all, it is only run once at the start of tuning.

Last sgemv of the value head is very small (128 multiplications). Loop would work here just fine without having any measurable effect on performance.

oscardssmith:
@frpays should this be closed then?

PR #521, opened 2018-11-18
Removed BLAS dependency from OpenCL backend.

frpays:
Building lc0 with OpenCL does not necessitate  BLAS anymore.

Thanks @Ttl  for sgemmBatched_ref code in Leela go.

I left the neural/shared c++ files into the build even when neither Blas nor OpenCL backend are built. As to not complicate purposelessly the meson build.

This is issue https://github.com/LeelaChessZero/lc0/issues/109.


meson.build
```diff
@@ -251,11 +244,20 @@ if get_option('build_backends')
 
   if get_option('blas') and has_blas
 
-    blas_files += [
+    blas_files = [
+    'src/neural/blas/fully_connected_layer.cc',
```
mooskagh: Sort files alphabetically.

src/neural/shared/activation.h
```diff
@@ -0,0 +1,34 @@
...
+
+namespace lczero {
+
+class Activation {
+ public:
+  Activation() = delete;
+
+  // Softmax activation
+  static void Softmax(const size_t size, const float* input, float* output);
+};
```
mooskagh: Why class with a static function? Why not just free standing function?

Issue #519, opened 2018-11-17
v0.19.0-rc5, ponder on loses on time bug


all12:
Loses on time:
[Aquarium 2018, ponder on, 1+1, LogFile.txt](https://github.com/LeelaChessZero/lc0/files/2592332/Aquarium.2018.ponder.on.1%2B1.LogFile.txt)

all12:
It also loses on time with ponder off:
[Aquarium 2018, ponder off, 1+1, LogFile.txt](https://github.com/LeelaChessZero/lc0/files/2592348/Aquarium.2018.ponder.off.1%2B1.LogFile.txt)

borg323:
Can you check if #522 helps? The executable to test is [this one](https://ci.appveyor.com/api/buildjobs/uc3ne262dhyubt51/artifacts/build%2Flc0.exe).

borg323:
Can you try to reproduce this crash using rc5? The additional code in the test executable is only called later in the search, so this is strange.

all12:
I tried it again with ponder off and much more games.

We have no crash:)
And no loses on time:)

I splitted the logfile in 2 parts because it was to big for upload.
[Aquarium 2018, ponder off, 1+1, LogFile.txt](https://github.com/LeelaChessZero/lc0/files/2593144/Aquarium.2018.ponder.off.1%2B1.LogFile.txt)
[Aquarium 2018, ponder off, 1+1, LogFile - part 2.txt](https://github.com/LeelaChessZero/lc0/files/2593149/Aquarium.2018.ponder.off.1%2B1.LogFile.-.part.2.txt)

borg323:
Looking at the latest log with ponder on I see the following pattern (worst case shown):
```
1118 12:10:17.430507 2404 ..\src/chess/uciloop.cc:131] >> go ponder btime 15263 binc 1000
...
1118 12:10:17.430870 2404 ..\src/chess/uciloop.cc:131] >> stop
...
1118 12:10:17.430973 4704 ..\src/chess/uciloop.cc:218] << info depth 4 seldepth 4 time 0 nodes 3015 score cp -261 hashfull 755 nps 0 tbhits 0 pv b2b5 c7a7 b5h5 c1g1 g6f6 a7a6 f6e7 g1g7 h5h3 g7g2 h3f3 g2e2 e7f8
...
1118 12:10:17.432356 4704 ..\src/chess/uciloop.cc:218] << bestmove e2d2 ponder b2b5
...
[...]

PR #522, opened 2018-11-18
remove Wait() from EngineController::Stop()

borg323:


borg323:
I'll wait a bit before merging - there was a strange hang in #519 when testing this.

PR #525, opened 2018-11-20
optional GoParams

borg323:
Continuation of #515, making the remaining integer GoParam fields optional.


src/engine.cc
```diff
@@ -186,11 +186,12 @@ SearchLimits EngineController::PopulateSearchLimits(
...
   }
-  limits.depth = params.depth;
+  if (params.depth) limits.depth = *params.depth;
   if (limits.infinite || !time) return limits;
-  int increment = std::max(int64_t(0), is_black ? params.binc : params.winc);
+  const optional<int64_t>& inc = is_black ? params.binc : params.winc;
+  int increment = inc ? std::max(int64_t(0), *inc) : 0;
 
-  int movestogo = params.movestogo < 0 ? 50 : params.movestogo;
+  int movestogo = !params.movestogo ? 50 : *params.movestogo;
```
evalon32: Can be: `params.movestogo.value_or(50)`

borg323: We use a custom optional implementation where value_or() is not implemented.

mooskagh: I think it's cleaner to implement it. No need for rvalue references like in stl's value_or(), something like that instead would be enough:
```cpp
 T value_or(const T& def) const { return has_value_ ? value_ : def; }
```

PR #526, opened 2018-11-20
updated include dirs for cuda 10

Ghotrix:
Updated include dirs for the latest cuda. Needed to be approved that location changed not only for Arch Linux. In other case this change must to be more complicated than changing 2 lines of the text file :-)

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -20,7 +20,7 @@ option('opencl_include',
 
 option('cuda_include',
        type: 'array',
-       value: ['/usr/local/cuda/include/'],
+       value: ['/opt/cuda/include/'],
        description: 'Paths to CUDA include directories')
 
 option('tensorflow_libdir',
@@ -60,7 +60,7 @@ option('mkl_include',
 
 option('cudnn_include', 
        type: 'array',
-       value: ['/usr/local/cuda/include/'],
+       value: ['/opt/cuda/include/'],
        description: 'Paths to cudnn include directory')
 
 option('build_backends',


Tilps:
I suspect #508 is the better option here.

borg323:
The fix is a bit more complicated, needs a loop in `meson.build` like [this](https://github.com/LeelaChessZero/lc0/blob/master/meson.build#L188).

meson.build
```diff
@@ -334,10 +334,30 @@ if get_option('build_backends')
...
+  required_cuda_header = 'cuda.h'
+  if not cc.has_header(required_cuda_header)
+    cuda_header_found = false
+  endif
+
+  required_cudnn_header = 'cudnn.h'
+  if not cc.has_header(required_cudnn_header)
+    cudnn_header_found = false
+  endif
+
```
borg323: I think the above block should be split and each half added before the respective `foreach` loop. In most cases `cudnn.h` will be in the same directory with `cuda.h` so it will be found if the check is after the first `foreach` loop.

borg323:
Can't find a way to make this work with meson - looks like the best option is to do it in `build.sh`.

borg323:
OK, found an approach that works, using an external python script to verify that the directories exist. See https://github.com/borg323/lc0/commit/940ac655b81a764e0bc92fc88e26284853b5ce48

borg323:
This is just for documentation: There was a suggestion from #mesonbuild to drop the script by doing `if compiler.has_header(join_paths(path, 'some-header-that-should-be-there.h'))`. This works fine on linux but on appveyor the build fails because we get:
```
Has header "C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v10.0/include/cudnn.h": NO
Has header "C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v10.0/include/cuda.h": YES
Has header "C:/cache/cuda/include/cudnn.h": NO
Has header "C:/cache/cuda/include/cuda.h": NO
```
The second to last check should have succeeded, since the path exists and contains cudnn.h:
```
 Directory of C:\cache\cuda\include
11/20/2018  10:31 PM    <DIR>          .
11/20/2018  10:31 PM    <DIR>          ..
09/01/2018  05:12 AM           100,962 cudnn.h
               1 File(s)        100,962 bytes
               2 Dir(s)  45,695,709,184 bytes free
```
The exact same command line works with the python script (this pr).

PR #527, opened 2018-11-20
compile fix for older cuda

borg323:


ankan-ban:
looks good. Thanks for fixing this!

ankan-ban:
instead of memcpy, you can also do:
alpha = *((half*) &one_h);
beta = *((half*) &zero_h);
but probably memcpy is less confusing. I am ok with either.

borg323:
I'm quite sure the casts violate aliasing rules.

ankan-ban:
maybe std::bit_cast is the most modern way of doing this:
https://en.cppreference.com/w/cpp/numeric/bit_cast

borg323:
Unfortunately it is c++20, so we can't  use it.

ankan-ban:
yeah. I think memcpy is fine. For types of same size, the casting the pointer works but using them is not a good idea in general.

PR #530, opened 2018-11-20
ignore position commands while searching

### LINKED ISSUE #506: Engine UCI interface should ignore 'position' while in infinite search.
If you type:

```
position startpos
go infinite
(... pv output)
position startpos e2e4
```

It *stops* the search. (you can restart the search with `go ...`). I believe that the command `position startpos e2e4` should be ignored until `stop` is received. 

```
position startpos
go infinite
(... pv output)
stop
(... bestmove output)
position startpos e2e4
```

Here is the protocol specification.

http://wbec-ridderkerk.nl/html/UCIProtocol.html

borg323:
This closes #506

### DIFF
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -305,6 +305,7 @@ void EngineController::NewGame() {
 
 void EngineController::SetPosition(const std::string& fen,
                                    const std::vector<std::string>& moves_str) {
+  if (search_ && search_->IsSearchActive()) return;
   // Some UCI hosts just call position then immediately call go, while starting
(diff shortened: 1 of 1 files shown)


mooskagh:
I don't think it's worth fixing (and I'm not convinced that the current behavior is not correct).

E.g. stockfish in similar cases behaves worse.
E.g.
```
position startpos
go infinite
position startpos
go infinite
stop  (does nothing)
stop
stop
stop (still does nothing)
```

borg323:
Closing since #506 was also closed.

PR #535, opened 2018-11-24
[Experiment] Attempt to fight with "another move is better" situation.

mooskagh:
--policy-persistence parameter

Default: 0.5
Range: 0.0 to 1.0

Higher values mean it will more likely to look into less promising moves (kind of similar to cpuct though)

I expect something slightly larger than 0.5 (e.g. 0.51) will be better for long time controls.


oscardssmith:
What is the advantage of this compared to making policy eval get updated by value?

mooskagh:
What is making policy eval get updated by value?

oscardssmith:
policy is the net evaluating the chance that a given move is the best move. Unlike value, this estimate is never updated by the information of the search.

PR #538, opened 2018-11-25
cudnn backend warning fixes

borg323:


src/neural/cuda/layers.h
```diff
@@ -39,6 +39,7 @@ class BaseLayer {
   int GetW() const { return W; }
 
   BaseLayer(int c, int h, int w, BaseLayer* ip);
+  virtual ~BaseLayer() {}
```
mooskagh: C++11 way of doing that is `=default` instead of `{}`.
That way move constructor generation is now suppressed.

PR #539, opened 2018-11-27
ensure bestmove is sent when aborting search

borg323:
Winboard with polyglot, when stopping pondering, sends the `position` command very quickly after `stop` and then sends `go` while at the same time waiting for `bestmove` before continuing. This leads to time losses: In some cases lc0 doesn't send `bestmove` at all when stopping, so the one sent after the final `go` is taken to be the one after `stop` and both sides end up waiting until time runs out.

This ensures that a `bestmove` will always be sent when stopping the search (provided `ok_to_respond_bestmove_` is true).

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -630,8 +630,10 @@ void Search::Stop() {
 
 void Search::Abort() {
   Mutex::Lock lock(counters_mutex_);
-  bestmove_is_sent_ = true;
-  FireStopInternal();
+  if (!stop_.load(std::memory_order_acquire)) {
+    bestmove_is_sent_ = true;
+    FireStopInternal();
+  }
   LOGFILE << "Aborting search, if it is still active.";
 }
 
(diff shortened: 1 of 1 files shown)


borg323:
As suggested on discord, this is replaced by a change to the search destructor that now calls Abort() only when `stop_` is false.

src/mcts/search.cc
```diff
@@ -644,7 +644,7 @@ void Search::Wait() {
 }
 
 Search::~Search() {
-  Abort();
+  if (!stop_.load(std::memory_order_acquire)) Abort();
```
mooskagh: After some thought, I think it's even better to do that in Abort(). If `stop_` is already true, just exit (maybe also output log message).
That way if Stop() was called before Abort(), stop wins. Otherwise Abort() wins.

But we don't have situations like this in the code now, so current way is totally fine too.

borg323: I'm not sure which version I prefer.

PR #545, opened 2018-12-01
U+Q softmax sampled visits

mooskagh:


zz4032:
Tuning latest code at long TC (50K nodes on opening moves) vs. Stockfish resulted in VisitSoftmax=0.00863.
On very short TC the parameter had no effect. Plausible?
Starting a selfplay match next. Games take very long, so it'll take a while.

zz4032:
lc0-tuned with VisitSoftmax=0.003 vs. lc0
```
Score of lc0-tuned vs lc0: 10 - 23 - 87  [0.446] 120
Elo difference: -37.8 +/- 32.3, LOS: 1.2 %, DrawRatio: 72.5 %
```
Trying VisitSoftmax=0.001 next...

zz4032:
lc0-0.001 with VisitSoftmax=0.001 vs. lc0 (VisitSoftmax=0.0)
```
Score of lc0-0.001 vs lc0: 15 - 51 - 104  [0.394] 170
Elo difference: -74.7 +/- 32.1, LOS: 0.0 %, DrawRatio: 61.2 %
```
Less drawish book was used, therefore Elo difference not comparable to previous matches. But 0.001 still very likely weaker.

mooskagh:
That was an attempt to make multipv more useful, but it probably has to be done otherwise (only change visit decision at root).

PR #550, opened 2018-12-04
Combine Trade Penalty with jjosh's Contempt

nblaxall:
Combine Trade Penalty with jjosh's Contempt to see if it tests with higher elo


nblaxall:
So this is called with: --trade-penalty=0.003 --trade-penalty2=32.0 --contempt=0.0015
or whatever values.
There is no contempt2, as this offset is fixed at 12 times the contempt flag value.

For example, using --contempt=0.0015 will mean the same as:
contempt = 0.0015 * #pieces-at-root - 0.0015 * 12
Code is:
```
auto contempt = params_.GetContempt() * (board.ours() + board.theirs()).count() - params_.GetContempt() * 12;
```
 ... and then that contempt calculated value is added to the trade-penalty value which is added to the value for the position.

PR #552, opened 2018-12-04
Added frequent-verbose-move-stats flag

gsobala:
A fresh attempt at a flag to increase the frequency of verbose move stats output. If --verbose-move-stats is set then --frequent-verbose-move-stats will output N/P/Q/U values as an info string every time an info string is sent to the UCI interface. This may be useful in debugging unusual MCTS search behaviour eg in the position ``2r1r3/2pR3R/1p3kpp/p3p3/1nP1PbP1/BP3B2/5P2/5K2 b - - 3 35`` which is currently topical


gsobala:
Rebased, as I did not know what I was doing, but I think that was a bad thing, so deleted to avoid confusion.

PR #555, opened 2018-12-05
revert IsSearchActive() and better fix for one of #500 crashes

borg323:


src/mcts/search.h
```diff
@@ -198,9 +198,9 @@ class SearchWorker {
   // Runs iterations while needed.
   void RunBlocking() {
     LOGFILE << "Started search thread.";
-    while (search_->IsSearchActive()) {
+    do {
```
mooskagh: Would be nice to add a comment why it is `do {} while`.

PR #558, opened 2018-12-06
remove UpdateFromUciOptions() from EnsureReady()

### LINKED ISSUE #547: Cutechess fails to fetch UCI params when there is no network file
Steps to reproduce:
1. Start cutechess
2. Ensure that there's no network next to ./lc0 file
3. Add ./lc0 as a new engine
4. Go to the parameters tab
5. Note that it is empty.

The reason for that is that cutechess sends `isready` but never receives `readyok`.

borg323:
This fixes #547 as outlined in https://github.com/LeelaChessZero/lc0/issues/547#issuecomment-444700691

### DIFF
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -432,8 +432,6 @@ void EngineLoop::CmdUci() {
     SendResponse(option);
   }
   SendResponse("uciok");
-  // Not all GUIs send a ucinewgame command.
-  engine_.NewGame();
 }
 
 void EngineLoop::CmdIsReady() {
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -284,6 +284,7 @@ void EngineController::UpdateFromUciOptions() {
 }
 
 void EngineController::EnsureReady() {
+  // TODO(borg) The call to UpdateFromUciOptions() should be removed.
   if (options_changed_) UpdateFromUciOptions();
   std::unique_lock<RpSharedMutex> lock(busy_mutex_);
   // If a UCI host is waiting for our ready response, we can consider the move
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -257,7 +257,6 @@ SearchLimits EngineController::PopulateSearchLimits(
 // Updates values from Uci options.
 void EngineController::UpdateFromUciOptions() {
   SharedLock lock(busy_mutex_);
-  options_changed_ = false;
 
   // Syzygy tablebases.
   std::string tb_paths = options_.Get<std::string>(kSyzygyTablebaseId.GetId());
@@ -284,8 +283,6 @@ void EngineController::UpdateFromUciOptions() {
 }
 
 void EngineController::EnsureReady() {
-  // TODO(borg) The call to UpdateFromUciOptions() should be removed.
-  if (options_changed_) UpdateFromUciOptions();
   std::unique_lock<RpSharedMutex> lock(busy_mutex_);
   // If a UCI host is waiting for our ready response, we can consider the move
   // not started until we're done ensuring ready.
@@ -435,6 +432,8 @@ void EngineLoop::CmdUci() {
     SendResponse(option);
   }
   SendResponse("uciok");
+  // Not all GUIs send a ucinewgame command.
+  engine_.NewGame();
 }
 
 void EngineLoop::CmdIsReady() {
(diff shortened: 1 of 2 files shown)


mooskagh:
Hmm looks pretty artificial. Why should it load network on `isready` if it's set from uci param, but should not if it's set from command line flag?.

What was the reason to load it on `isready` rather than on `position`?

borg323:
The reason was to finish the time consuming initialization before starting the timer.  With this I'm trying to avoid the specific case of #547.

Long term I think we should move towards another solution. From the example section of the uci spec:
```
// init tbs
setoption name NalimovCache value 1
setoption name NalimovPath value d:\tb;c\tb

// waiting for the engine to finish initializing
// this command and the answer is required here!
isready

// engine has finished setting up the internal values
		readyok
```
This implies that `setoption` triggers the loading of TB's and by extension the weights file. I think the best long term solution would be to split UpdateFromUciOptions() into several methods, called individually when the corresponding option is changed and also from UpdateFromUciOptions() as now, and remove the call from EnsureReady().

mooskagh:
There used to be callbacks for that https://github.com/LeelaChessZero/lc0/pull/424
But depending on moment of parameter change was still not very useful, as majority of people use default `<autodiscover>` setting.

I still don't understand why we need it in `isready`, from practical perspective.
It will only cover the case when GUI doesn't send `ucinewgame` or `position`, but does send `isready`.
Is there any GUI that is known to behave this way?

I'd expect that if a GUI behaves poorly (e.g. not sends `ucinewgame` or `position` before the first move), it's even less likely for it to send `isready`, and we should not rely on it.

If we really want to trigger load on `setoption`, then the best thing to do is probably have  <https://github.com/LeelaChessZero/lc0/pull/424> reverted back. But it won't really help with `<autodiscover>` case.
This will be more conformant with the snippet from the specification though. (Note that in that example `isready` also doesn't trigger any work, it just waits work after `setoption` to finish).

borg323:
I checked a few logs (from other issues) and saw that Fritz 16, chessbase 14 (I think this is a version of the same gui) and Aquarium 2018 are not sending `ucinewgame` (at least for the first game). They do send `isready` though.

I also don't like this patch, but is the best I can come up with for a quick fix. 

An alternative long term solution is to call UpdateFromUciOptions() right after we send `uciok` at the start  but this would require additional testing with lots of guis.

I'll add a TODO comment in EnsureReady() that the call to UpdateFromUciOptions() should be removed.

mooskagh:
But do they send `position`?
Also, how does this fix help in `<autodiscover>` case?

I don't think we should make workarounds for uci-compliant hosts to please uci-incomplaint ones.. If Lc0 takes too much time during first move in Fritz because it loads the network, maybe so be it?..

borg323:
The whole thing is a mess, but it is not our mess...

I checked a few more logs, every one of them always sends `position` right before `go` so doing the initialization in `go` (as we do now as a last resort) is at worst a few ns late.

However, of all the logs I saw, even the guis than send `ucinewgame` ignore the uci spec that explicitly states:
```
As the engine's reaction to "ucinewgame" can take some time the GUI should always send "isready"
after "ucinewgame" to wait for the engine to finish its operation.
```
I haven't seen a single one with an `isready` after `ucinewgame` and I think I have logs from most major guis (except Arena).

This pr doesn't help with `<autodiscover>` in general, only works around #547.

borg323:
The latest alternative removes the UpdateFromUciOptions() from EnsureReady() and does the equivalent of `ucinewgame` right after sending the `uciok` reply to the gui. Then if the gui doesn't send a `ucinewgame` the next UpdateFromUciOptions() will be called from `go`.

borg323:
I asked around for an Arena log (thanks Albert), and it sends `isready` after `ucinewgame`. Also cutechess does send `isready` after the next `position` so is compliant (rereading the spec, it doesn't say it must be sent directly after `ucinewgame`). I believe this covers most of our users, so we should be OK.

mooskagh:
But that will search and load network during `uci` response, even though after that that network is pretty likely to be changed by setoption, and even though GUI may not have intended to start a game (e.g. only wanted to get list of params).

I don't understand, what's wrong with loading network only at `ucinewgame` and `position` (and `go`)?

mooskagh:
In general I think our strategy should be:
- Be nice if GUI is UCI conformant (i.e. sends `ucinewgame`, then `isready`, then waits). We don't load network until the game is actually started, don't trigger any work on `isready`, etc.

- If GUI fails to support proper UCI, we still work, but we are not that nice. E.g. it's totally fine to load network during our first move.

PR #563, opened 2018-12-06
Updated cpuct formula from alphazero paper.

mooskagh:


borg323:
I'm not sure about changing the defaults at this stage.

zz4032:
Line 58 in src/mcts/params.cc should be changed to
"cpuctbase", "CPuctBase",

PR #567, opened 2018-12-07
Rework Cpuct related params.

mooskagh:
Cpuct 3.4 -> 3.0 for play
Cpuct 1.25 -> back to 1.2 for training

Introduce CpuctFactor parameter (2 for play, 0 for training), because we have larger value space (-1..1) than alphazero (0..1)

Cut a few zeros in max CpuctBase value (went above what float could represent)
Minimal value for CpuctBase is now 1, not 0.

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -183,14 +183,23 @@ int64_t Search::GetTimeToDeadline() const {
       .count();
 }
 
+namespace {
+
+inline float ComputeCpuct(const SearchParams& params, uint32_t N) {
+  const float init = params.GetCpuct();
+  const float k = params.GetCpuctFactor();
+  const float base = params.GetCpuctBase();
+  return init + (k ? k * std::log((N + base) / base) : 0.0f);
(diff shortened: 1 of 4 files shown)


src/mcts/search.cc
```diff
@@ -183,14 +183,23 @@ int64_t Search::GetTimeToDeadline() const {
       .count();
 }
 
+namespace {
+
+inline float ComputeCpuct(const SearchParams& params, uint32_t N) {
+  const float init = params.GetCpuct();
+  const float k = params.GetCpuctFactor();
+  const float base = params.GetCpuctBase();
+  return init + (k ? k * std::log((N + base) / base) : 0.0f);
```
Tilps: As borg mentioned, deepmind version has a +1 in here. (I really don't think its significant unless someone tries to test very small 'base' parameters.)

mooskagh: Yeah I didn't understand that 1+ so decided to simplify.

PR #568, opened 2018-12-07
Temperature and FPU related params.

mooskagh:
Introduce
--temp-cutoff-move (default: 0, i.e. switched off).
  Starting from the move specified in this parameter, the temperature is taken from --temp-endgame param.

--temp-value-cutoff
   When selecting move using temperature, only consider moves within X% (1% = 0.02 of Q value) from bestmove. Default is 100%, e.g. all moves.

FPU:
--fpu-strategy  Can be "reduction" (as before, value is taken from --fpu-reduction flag) or "absolute" (value is taken from --fpu-value flag)

### DIFF
--- a/src/mcts/params.cc
+++ b/src/mcts/params.cc
@@ -68,6 +68,18 @@ const OptionId SearchParams::kTempDecayMovesId{
     "Reduce temperature for every move from the game start to this number of "
     "moves, decreasing linearly from initial temperature to 0. A value of 0 "
     "disables tempdecay."};
+const OptionId SearchParams::kTemperatureCutoffMoveId{
+    "temp-cutoff-move", "TempCutoffMove",
+    "Move number, starting from which endgame temperature is used rather "
+    "than initial temperature. Setting it to 0 disables cutoff."};
+const OptionId SearchParams::kTemperatureEndgameId{
+    "temp-endgame", "TempEndgame",
+    "Temperature used during endgame (starting from cutoff move). Endgame "
+    "temperature doesn't decay."};
+const OptionId SearchParams::kTemperatureWinpctCutoffId{
+    "temp-value-cutoff", "TempValueCutoff",
+    "When move is selected using temperature, bad moves (with win "
+    "probability less than X than the best move) are not considered at all."};
 const OptionId SearchParams::kTemperatureVisitOffsetId{
     "temp-visit-offset", "TempVisitOffset",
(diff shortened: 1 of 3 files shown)


src/mcts/search.cc
```diff
@@ -447,8 +456,11 @@ void Search::EnsureBestMoveKnown() REQUIRES(nodes_mutex_)
...
   float temperature = params_.GetTemperature();
-  if (temperature && params_.GetTempDecayMoves()) {
-    int moves = played_history_.Last().GetGamePly() / 2;
+  const int cutoff_move = params_.GetTemperatureCutoffMove();
+  const int moves = played_history_.Last().GetGamePly() / 2;
+  if (cutoff_move && (moves + 1) >= cutoff_move) {
+    temperature = params_.GetTemperatureEndgame();
+  } else if (temperature && params_.GetTempDecayMoves()) {
     if (moves >= params_.GetTempDecayMoves()) {
       temperature = 0.0;
```
Tilps: Why not use GetTemperatureEndgame here as well.

src/mcts/search.cc
```diff
@@ -1038,11 +1052,11 @@ int SearchWorker::PrefetchIntoCache(Node* node, int budget) {
 
   float puct_mult = cpuct * std::sqrt(std::max(node->GetChildrenVisits(), 1u));
   // FPU reduction is not taken into account.
```
Tilps: This comment is no longer correct?

src/mcts/search.cc
mooskagh: Decided not to do now. (unless you are willing to send a patch).

src/mcts/params.cc
```diff
@@ -90,13 +102,27 @@ const OptionId SearchParams::kSmartPruningFactorId{
     "promising moves from being considered even earlier. Values less than 1 "
     "causes hopeless moves to still have some attention. When set to 0, smart "
     "pruning is deactivated."};
+const OptionId SearchParams::kFpuStrategyId{
+    "fpu-strategy", "FpuStrategy",
+    "How is an eval of unvisited node determined. \"reduction\" more subtracts "
```
borg323: Delete `more`

borg323:
Long term I think the best way is to make temperature take a list of temperatures and use a second option to set the breakpoints (maybe even reuse `--temp-cutoff-move`)

PR #575, opened 2018-12-10
try fmath.hpp from https://github.com/herumi/fmath

borg323:
This is an experiment to use some (hopefully) faster code for log() and exp(), Also implements pow(x, y) as exp(y * log(x)).

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -184,6 +184,8 @@ int64_t Search::GetTimeToDeadline() const {
 }
 
 namespace {
+#include "fmath.hpp"
+
 inline float GetFpu(const SearchParams& params, Node* node, bool is_root_node) {
   return params.GetFpuAbsolute()
              ? params.GetFpuValue()
@@ -198,7 +200,12 @@ inline float ComputeCpuct(const SearchParams& params, uint32_t N) {
   const float init = params.GetCpuct();
   const float k = params.GetCpuctFactor();
   const float base = params.GetCpuctBase();
-  return init + (k ? k * std::log((N + base) / base) : 0.0f);
+  return init + (k ? k * fmath::log((N + base) / base) : 0.0f);
+}
+
+inline float PowPositive(const float x, const float y) {
+  if (x <= 0.0) return 0.0;
+  return fmath::exp(y * fmath::log(x));
 }
 }  // namespace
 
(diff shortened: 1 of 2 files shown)


oscardssmith:
For most of the places where we use log, couldn't we just use bitlength? We only care that it increases at roughly the right rate, not the exact value, right?

borg323:
You mean the fp exponent? Possibly, there are various approximations that can be used.

oscardssmith:
oh yeah, fp, not integers, But yeah the exponent gives a good 0th order approximation that has the benefit of being as fast as addition to get.

oscardssmith:
We could also do a similar thing with non-critical sqrts, of just halving the exponent and base.

ddobbelaere:
I fear that such zeroth or first order approximations might be too crude and lose elo.

I found an alternative library that might be just as fast (or even faster?) and also provides relative precision and a power function: http://www.machinedlearnings.com/2011/06/fast-approximate-logarithm-exponential.html
- logarithm is factor 4 or 5 times faster w.r.t. full-precision `std::logf` (you can choose between `fastlog` and `fasterlog`, whose relative precision seems good enough for us: 2.09348e-05 resp. 0.0130367)
- exponential is factor 35 or 50 times faster than `std::expf` (you can choose between `fastexp` and `fasterexp` with similar relative precision as their log-counterparts)

I was able to reproduce the claimed speedup for the logarithm, but only got about factor 3x speedup for `fasterexp`.

There's also a function `fastpow` which is about 5x faster than `std::pow`.

It has BSD license and you can use all approximate functions (including other funky functions as LambertW, erf...) by including a single header file (or several seperate header files): https://github.com/etheory/fastapprox/tree/master/fastapprox/src

borg323:
I was looking at this library earlier today, but as an experiment I decided to try a minimal approach with fp exponent tricks and simple linear extrapolation for the mantissa. This has an error up to 8% but can be improved by adding a few more terms. Will make a separate wip pr for this later.

oscardssmith:
the reason I don't think horrible approximations are likely to lose elo is that there is no proof that what we have is optimal, and it just seems to be chosen as "lets pick a concave function and mess with some params till it plays well enough".

borg323:
The simple approximations are in https://github.com/LeelaChessZero/lc0/pull/580

PR #577, opened 2018-12-10
v0.19.1 tp

nblaxall:
I had to create a new PR to update instead of updating the old tp from v0.19.0 to v0.19.1 as I had all sorts of problems otherwise (although I'm very inexperienced with git).

This supports both tp and contempt, although they are both off by default.

Call tp with eg:
```
--trade-penalty=0.005 --trade-penalty2=16.0
```
Call contempt with eg:
```
--contempt=0.001
```
(there is no contempt2 as that is hardset to 12)

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -1143,6 +1143,8 @@ void SearchWorker::FetchSingleNodeResult(NodeToProcess* node_to_process,
     node_to_process->v = -computation_->GetQVal(idx_in_computation) + penalty;
     // penalty shouldn't put v outside (-1, 1), and we will clip if it is:
     node_to_process->v = std::max(-0.9999f, std::min(0.9999f, node_to_process->v));
+  } else {
+    node_to_process->v = -computation_->GetQVal(idx_in_computation);
   }
   // ...and secondly, the policy data.
   float total = 0.0;
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -1133,16 +1133,16 @@ void SearchWorker::FetchSingleNodeResult(NodeToProcess* node_to_process,
   }
   // For NN results, we need to populate policy as well as value.
   // First the value...
-  auto penalty = params_.GetTradePenalty() * (node_to_process->piececount - params_.GetTradePenalty2());
-  // We flip penalty sign for Leela's moves (odd depths)
-  // (opponent depth is even depths and has opposite sign)
-  if(node_to_process->depth % 2 == 1)
-    penalty = -penalty;
-  node_to_process->v = -computation_->GetQVal(idx_in_computation) + penalty;
-  if(node_to_process->v < -1.0) {
-    node_to_process->v = -0.9999;
-  } else if (node_to_process->v > 1.0) {
-    node_to_process->v = 0.9999;
+  // Trade Penalty defaults to off, ie 0.0, so we only calculate penalty if it's set to != 0:
+  if (params_.GetTradePenalty() != 0.0f) {
+    auto penalty = params_.GetTradePenalty() * (node_to_process->piececount - params_.GetTradePenalty2());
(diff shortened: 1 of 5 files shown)


killerducky:
The following code doesn't work well for pondering or when using lc0 to make moves for both sides:
```
  if(node_to_process->depth % 2 == 1)
    penalty = -penalty;
```

One solution is to remember if it's White or Black that is trying to avoid trading/draws. You can have two modes: Manually specify what color is avoiding trades/draws, or auto-detect. auto-detect means you see what color is to move in `go` commands, and auto set that color to be the one avoiding trades/draws. But if the color changes you have to throw away the tree and invalidate the N

In auto-detect, when `go` is called, check if it's the same color to avoid draws (in normal games it will be). If not, simplest solution is to throw away the tree (I think penalty is applied after nncache, so you don't have to invalidate it?). 

In auto-detect, if `go ponder` is called, you can still guess which side "you" are playing and apply trade/draws to the correct side.

For live analysis where you want to see Lc0's opinion of both sides, either turn off TP, or manually set which color has it.

I'm still afraid some GUIs or use cases will be problematic...

killerducky:
Another problem is this code adds an offset to the value, breaking the assumption that -1 and 1 are the worst/best scores. This can cause the engine to avoid playing checkmate, since there are now positions that are better than checkmate.

To fix this the transform should shift values up/down, but in some more complicated way that preserves the [-1,1] range.

killerducky:
@mooskagh The other PR is long, and I didn't follow all the Discord chat. Can you update this PR with your view on how to move forward?

nblaxall:
Thank you killerducky.  I'll add the cap, and as discussed in # events-discuss I'm remove the contempt (so it's just pure tp so as not to complicate matters).
For ponder, I'll need help with the code from someone.

nblaxall:
Just before I update, here is the latest of my still-running rc2 tp vs SF for 2m+2s (120+2) TC ... as longer TC means variable cpuct kicks in more than the 30s+0.5s TC ... also my theory is that tp should kick in more at deeper depths too:
```
   # PLAYER                         :  RATING  PLAYED   (%)  CFS(%)    W    D    L  D(%)
   1 11248-v0.19.1-rc2-tp_005_16    :      82     114  61.4      98   33   74    7  64.9
   2 11248-v0.19.1-rc2              :      37     114  55.3      99   16   94    4  82.5
   3 SF9-bmi                        :       0     228  41.7     ---   11  168   49  73.7
```
... games take about 8 or 9 mins each game.

src/mcts/search.cc
```diff
@@ -1139,6 +1139,11 @@ void SearchWorker::FetchSingleNodeResult(NodeToProcess* node_to_process,
   if(node_to_process->depth % 2 == 1)
     penalty = -penalty;
   node_to_process->v = -computation_->GetQVal(idx_in_computation) + penalty;
+  if(node_to_process->v < -1.0) {
+    node_to_process->v = -0.9999;
```
killerducky: If we go with simple clipping like this, I think it should be:
 `if (v < -0.9999) v = -0.9999;... etc`
and maybe should use the min/max style the code uses elsewhere:
`v = std::max(-0.9999, std::min(0.9999, v))`

killerducky: Also this code should only execute if the penalty is enabled.

nblaxall: I made those changes as you suggested.

src/mcts/search.cc
```diff
@@ -1130,7 +1133,17 @@ void SearchWorker::FetchSingleNodeResult(NodeToProcess* node_to_process,
   }
   // For NN results, we need to populate policy as well as value.
   // First the value...
-  node_to_process->v = -computation_->GetQVal(idx_in_computation);
```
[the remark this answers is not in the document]
nblaxall: Updated with the else statement - thanks!

PR #580, opened 2018-12-11
use fast log and pow approximations

borg323:
This uses fast approximations to calculate log and pow in search. Mainly for discussion.

The approximations used are:
log2(2^N * (1 + f)) ~ N + f 
2^(N + f) ~ 2^N * (1 + f)
N is the integer and f the fractional part, f>=0.

The absolute error of the above approximations for f in [0,1] is less than 0.08.

Better approximations (not tested) are:
log2(2^N * (1 + f)) ~ N + f * (1.34267 - 0.342671 * f)
2^(N + f) ~ 2^N*(1 + f * (0.656366 + 0.343634 * f))

Those are chosen to have zero error for f=0 and f=1 so that there are no discontinuities (also true for the simple ones).

### DIFF
--- a/src/utils/fastmath.h
+++ b/src/utils/fastmath.h
@@ -1,0 +1,66 @@
+/*
+  This file is part of Leela Chess Zero.
+  Copyright (C) 2018 The LCZero Authors
+
+  Leela Chess is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  Leela Chess is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with Leela Chess.  If not, see <http://www.gnu.org/licenses/>.
+
+  Additional permission under GNU GPL version 3 section 7
+
+  If you modify this Program, or any covered work, by linking or
+  combining it with NVIDIA Corporation's libraries from the NVIDIA CUDA
+  Toolkit and the NVIDIA CUDA Deep Neural Network library (or a
+  modified version of those libraries), containing parts covered by the
+  terms of the respective license agreement, the licensors of this
+  Program grant you additional permission to convey the resulting work.
+*/
+
+#pragma once
+
+namespace lczero {
+// These stunts are performed by trained professionals, do not try this at home.
+
+// Fast approximate log2(x). Does no range checking.
+// The approximation used here is log2(2^N*(1+f)) ~ N+f*(1.342671-0.342671*f)
+// where N is the integer and f the fractional part, f>=0.
+inline float FastLog2(const float a) {
+  int32_t tmp;
+  std::memcpy(&tmp, &a, sizeof(float));
+  int expb = (tmp >> 23);
+  tmp = (tmp & 0x7fffff) | (0x7f << 23);
+  float out;
+  std::memcpy(&out, &tmp, sizeof(float));
+  return out * (2.028011f - 0.342671f * out) - 128.68534f + expb;
+}
+
+// Fast approximate 2^x. Does only limited range checking.
+// The approximation used here is 2^(N+f) ~ 2^N*(1+f*(0.656366+0.343634*f))
+// where N is the integer and f the fractional part, f>=0.
+inline float FastPow2(const float a) {
+  if (a < -126) return 0.0;
+  int exp = floor(a);
+  float out = a - exp;
+  out = 1.0f + out * (0.656366f + 0.343634f * out);
+  int32_t tmp;
+  std::memcpy(&tmp, &out, sizeof(float));
+  tmp += exp << 23;
+  std::memcpy(&out, &tmp, sizeof(float));
+  return out;
+}
+
+// Fast approximate ln(x). Does no range checking.
+inline float FastLog(const float a) {
+  return 0.6931471805599453f * FastLog2(a);
+}
+
+}  // namespace lczero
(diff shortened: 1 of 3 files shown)


ddobbelaere:
Hi borg,

I am getting faster results with `fasterlog` mentioned in #575. Here are the execution times for the following loop:

`float dummy = 2.0;
for(int i=0; i<NUM_ITER; i++) {
		dummy += fasterlog(dummy);
	}`

Note that nothing can be SSE-optimized as the results depends serially on one another.

- `fasterlog`: 1.87449e+08 operations/second
-  `flog`: 1.23341e+08 operations/second

borg323:
Thanks for doing this. The result is probably processor/compiler dependent but `fasterlog` is indeed very fast. The final `- 1` in `flog2` can be merged with a previous calculation, but I was hoping the compiler will do it. I'll check the generated assembly.

borg323:
In my tests `flog2` is about 5% faster than `fasterlog2` (probably compiler/processor dependent). `flog` and `fasterlog` just multiply with `1/ln(2)` so the timing should be similar. 

Both versions use the exact same approximation, so you can make `fasterlog(1)=0` by changing the last constant in fasterlog2 to 127. This was a design decision from the start, to have exact values for powers of 2 (also true for the more accurate quadratic approximations given above). Otherwise a small constant term can be used to shift the (absolute) error and the value used for fasterlog2 is not the best in this respect.

ddobbelaere:
Nice, it took me a few minutes, but now I see why `fasterlog2` and `flog2` are exactly the same approximation (up to the different term `127`). Thanks for the clarification!

borg323:
Did the math and 126.942695 is the constant to minimize the RMS error of the approximation.

ddobbelaere:
The approximation is much better for the practically more relevant case of exponents larger than 1. Probably the case above can be neglected for our applications, if all considered temperatures we will ever use are always smaller than or equal to 1.

figure_1

borg323:
The best I have so far for the more accurate quadratic approximations is:
```
inline float flog2(const float a) {
  int32_t tmp;
  std::memcpy(&tmp, &a, sizeof(float));
  int expb = (tmp >> 23);
  tmp = (tmp & 0x7fffff) | (0x7f << 23);
  float out;
  std::memcpy(&out, &tmp, sizeof(float));
  return out * (2.028011f - 0.342671f * out) - 128.68534f + expb;
}

inline float fpow2(const float a) {
  if (a < -126) return 0.0;
  int exp = floor(a);
  float out = a - exp;
  out = 1.0f + out * (0.656366f + 0.343634f * out);
  int32_t tmp;
  std::memcpy(&tmp, &out, sizeof(float));
  tmp += exp << 23;
  std::memcpy(&out, &tmp, sizeof(float));
  return out;
}

```
Absolute log2 error plot in the [1,2] interval for linear (red) and quadratic (blue) approximations.
image

Relative pow2 error plot in the [1,2] interval for linear (red) and quadratic (blue) approximations.
image

oscardssmith:
that looks a lot better (which only really matters for pow2). How's the relative speed?

borg323:
A bit slower than linear, pow2 more than log2 but it is hard to measure exactly.
`./lc0 benchmark -w 11248.pb.gz --nodes=1000000 --smart-pruning-factor=0 --backend=random -t 1` gives the following nps (best of 3 runs):
master: 132809
linear: 168117
quadratic: 164251

ddobbelaere:
Hi borg,

I arrive at the same constants in the proposed quadratic approximations for `flog2` and `fpow2` after minimizing the RMS error under the constraint of zero error for powers of 2.

Here are the relative speeds w.r.t. `std::log` for the loop `for (float i = 1.0; i < 2.0; i += 1e-6f) { dummy += flog(dummy); }`

function | speed relative to `std::log`
 ---- | ----
`fasterlog` | 2.83
`fastlog` | 1.20
`flog_linear` | 2.23
`flog_quadratic` | 1.56

Concerning the speedup in terms of nps of `lc0`, I have to say from my own experience that the main reason for speedup is the introduction of `PowPositive` in the following code fragment:

```
if (params_.GetPolicySoftmaxTemp() != 1.0f) {
    p = PowPositive(p, 1 / params_.GetPolicySoftmaxTemp());
}
```
This is because the policy softmax temperature is set to `2.2` in benchmark mode. So this huge speedup seems to only make a difference for selfplay at the moment, or am I missing something?

jjoshua2:
Selfplay uses 1 and matches use 2.2. Also we could compute the inverse just
once instead of every time. Compiler is probably smart enough to do that
already...

ddobbelaere:
Yeah, probably the compiler will do that.

The plot for `PowPositive` using quadratic approximations (policy softmax temperature equal to 2.2) is looking really good (maximum absolute error is equal to 0.00343152):

pow_positive

ddobbelaere:
I can confirm the huge speedup with this PR reported by  @borg323 above (on Linux Fedora 29).

Tested by running `./lc0 benchmark --nodes=1000000 --smart-pruning-factor=0 --backend=random -t 1` 25 times for each case. A 95% confidence interval is also given.

`lc0`  |  nps 
------ | -----
master | 119819 ± 501 
#580 0 | 148882 ± 528

borg323:
Can you try also together with #590?

oscardssmith:
Locks have compounding effect. The lock contention, the more time spent doing real work.

borg323:
I get about a 7% speedup from #590. Maybe due to lto.

src/utils/fastmath.h
```diff
@@ -0,0 +1,66 @@
...
+
+namespace lczero {
+// These stunts are performed by trained professionals, do not try this at home.
+
+// Fast approximate log2(x). Does no range checking.
+// The approximation used here is log2(2^N*(1+f)) ~ N+f*(1.342671-0.342671*f)
+// where N is the integer and f the fractional part, f>=0.
+inline float FastLog2(const float a) {
+  int32_t tmp;
+  std::memcpy(&tmp, &a, sizeof(float));
```
mooskagh: memcpy needs  `#include <cstring>`

PR #584, opened 2018-12-13
 pgo build on release 

borg323:
Using `lc0 benchmark` with the random backend for profile generation was found to lead to faster pgo executables than the usual lto builds. Unfortunately, doing this takes an extra couple of minutes so this patch only adds it for the release builds.


borg323:
The profile generation runs for 10 seconds and about 30K nodes. Maybe this should be increased?

PR #586, opened 2018-12-14
fix cuda debug build and warning

borg323:


ankan-ban:
Looks good. Why was the assert causing warning?

borg323:
The assert was giving the error, not a warning, because I commented out input2 to avoid the "unused" warning in release builds. Now it is `const float* /*input2*/`
I was mostly worried about the `num_weights2`, whether we missed something in the refactoring.

ankan-ban:
Ok. I see. 

Looks like num_weights2 isn't used anywhere. Instead computing weight_size2 from it we simply compute it as twice of weight_size1 which is same so we should be fine.

PR #590, opened 2018-12-17
Reserve container space

SunnyWar:
Add reserve for containers that are used often.
1. GeneratePseudolegalMoves populated result with the number of valid moves.  A comment on discord indicated that for the average game this number is "30". It was recommended to double it. 
2. ApplyDirichletNoise always populates "noise" with the number of edges so it makes sense to reserve it.
3.  SearchWorker populates minibatch_. I make a guess as to a reasonable number.

Note: I don't like that two of these are magic numbers. Recommendations on what to call consts to represent them and where to put the constants is appreciated.

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -83,6 +83,7 @@ namespace {
 void ApplyDirichletNoise(Node* node, float eps, double alpha) {
   float total = 0;
   std::vector<float> noise;
+  noise.reserve(node->GetNumEdges());
 
   for (int i = 0; i < node->GetNumEdges(); ++i) {
     float eta = Random::Get().GetGamma(alpha, 1.0);
@@ -989,7 +990,8 @@ bool SearchWorker::AddNodeToComputation(Node* node, bool add_if_cached) {
 
   if (node && node->HasChildren()) {
     // Legal moves are known, use them.
-    for (auto edge : node->Edges()) {
+    moves.reserve(node->GetNumEdges());
+    for (const auto& edge : node->Edges()) {
       moves.emplace_back(edge.GetMove().as_nn_index());
     }
   } else {
--- a/src/mcts/search.h
+++ b/src/mcts/search.h
@@ -193,7 +193,9 @@ class Search {
 class SearchWorker {
  public:
   SearchWorker(Search* search, const SearchParams& params)
-      : search_(search), history_(search_->played_history_), params_(params) {}
+      : search_(search), history_(search_->played_history_), params_(params) {
+    minibatch_.reserve(32);
+  }
 
   // Runs iterations while needed.
   void RunBlocking() {
--- a/src/chess/board.cc
+++ b/src/chess/board.cc
@@ -188,6 +188,7 @@ BitBoard ChessBoard::en_passant() const { return pawns_ - pawns(); }
 
 MoveList ChessBoard::GeneratePseudolegalMoves() const {
   MoveList result;
+  result.reserve(60);
   for (auto source : our_pieces_) {
     // King
     if (source == our_king_) {


src/mcts/search.cc
```diff
@@ -83,6 +83,7 @@ namespace {
 void ApplyDirichletNoise(Node* node, float eps, double alpha) {
   float total = 0;
   std::vector<float> noise;
+  noise.reserve(node->GetNumEdges());
```
borg323: My tests indicate this doesn't make a difference.

src/mcts/search.cc
```diff
@@ -989,7 +990,8 @@ bool SearchWorker::AddNodeToComputation(Node* node, bool add_if_cached) {
 
   if (node && node->HasChildren()) {
     // Legal moves are known, use them.
-    for (auto edge : node->Edges()) {
+    moves.reserve(node->GetNumEdges());
+    for (const auto& edge : node->Edges()) {
       moves.emplace_back(edge.GetMove().as_nn_index());
     }
   } else {
```
borg323: There is another reserve() a few lines later that also doesn't seem to help.

src/mcts/search.h
```diff
@@ -193,7 +193,9 @@ class Search {
 class SearchWorker {
  public:
   SearchWorker(Search* search, const SearchParams& params)
-      : search_(search), history_(search_->played_history_), params_(params) {}
+      : search_(search), history_(search_->played_history_), params_(params) {
+    minibatch_.reserve(32);
```
borg323: This also doesn't help in my tests

ddobbelaere:
I have also noticed a few percentages speedup with this PR (on Linux Fedora 29).

Tested by running `./lc0 benchmark --nodes=1000000 --smart-pruning-factor=0 --backend=random -t 1` 25 times for each case. A 95% confidence interval is also given.

`lc0`  |  nps 
------ | -----
master | 119819 ± 501 
#590 | 124201 ± 463

SunnyWar:
I ran the same params that @ddobbelaere ran and captured the maximum size of minibatch_. It went up to 287. So my guess was way off. I tried setting the reserve to 300 and got a bit of a perf boost.

oscardssmith:
@SunnyWar what's the effect on memory use by doing this? 300 is a lot.

SunnyWar:
@oscardssmith 
std::vector<NodeToProcess> minibatch_;
class lczero::SearchWorker::NodeToProcess	size(24):
9>	+---
9> 0	| node
9> 8	| v
9>12	| multivisit
9>16	| depth
9>18	| nn_queried
9>19	| is_cache_hit
9>20	| is_collision
9>  	| <alignment member> (size=3)
9>	+---

Therefore 300 * 24 = 7.2k bytes. + vector overhead. I tried to confirm this but Leela appears to presently have a memory leak.

SunnyWar:
@oscardssmith I'm not sure. I can tell you that the container is a vector and the size of the object is exactly 24 bytes. From the comments in the code it appears that only one SearchWorker object is created so that would make the size of the container 7200 bytes. (300 * 24) plus some overhead for the vector which usually also 24 bytes, which gives us 7224 bytes.  
I was going to compare Master build against this one to compare the working set but I got _wildly_ different memory usage each time I ran it (master). I don't understand this.

src/mcts/search.cc
SunnyWar: I was able to measure an improvement on my low end laptop but not on my high end desktop.

src/mcts/search.cc
SunnyWar: Again, I was able to measure an improvement on my low end laptop but not on my high end desktop.

src/mcts/search.h
SunnyWar: I output the actual size the minibatch grows to, which usually went above 250, and changed this to a much larger number in updated code (300). After making this change the performance different, for me, was significant.

borg323: This is probably compiler/system dependent (still doesn't help here).

ddobbelaere:
Memory is indeed not an issue. The number of CPU worker threads that spawn one `SearchWorker` each is given by the flag `-t` and is a small constant. Moreover, who cares about a few kilobytes?

Now that #580 has been merged into master, let's redo the measurements. Tested by running `./lc0 benchmark --nodes=1000000 --smart-pruning-factor=0 --backend=random -t 1` 25 times for each case. A 95% confidence interval is also given. Note that the default minibatch size is 256.

`lc0`  |  nps | speedup
------ | ----- | -----------
master | 152935 ± 503 | -
#590 | 160870 ± 558 | 5.2%
#590 without `minibatch_.reserve(300)` | 162327 ± 586 | 6.1%

Like @borg323 I can attest that `minibatch_.reserve(300)` has no positive effect on my system (slightly negative even). I think this can be explained as follows: even if no fixed number of elements are reserved, the minibatch grows to the configured minibatch size (256 in the default case) during the first iteration (**this only happens once for each search**). All subsequent iterations clear the vector but don't destroy it.

src/mcts/search.cc
mooskagh: That's a very slow laptop. Dirichlet noise is only applied once per search, it wouldn't be noticeable even on computer which is 25 years old.

src/mcts/search.cc
mooskagh: Here it's a bit more plausible than with applying Dirichlet noise, but as it seems that there's something wrong with your measurements (they should have not shown anything there), it would be better if someone else retested/reconfirmed it. Or at least if you could post numbers from your tests.

In general it's good idea only to optimize where it gives benefits, adding reserve() line just for the case doesn't do any good.

ddobbelaere:
Let us the label the four contributions as follows:

1       `GeneratePseudolegalMoves result.reserve(60);`
2       `ApplyDirichletNoise noise.reserve(node->GetNumEdges());`
3       `SearchWorker::AddNodeToComputation moves.reserve(node->GetNumEdges());`
4       `SearchWorker::SearchWorker minibatch_.reserve(300);`

Then I get the following back-to-back measurements:

`lc0`  |  nps | speedup
------ | ----- | -----------
master | 152552 ± 976 | -
1 | 156649 ± 506 | 2.6%
2 | 152639 ± 457 | 0.06%
3 | 156763 ± 606 | 2.8%
4 | 152124 ± 1112 | -0.3%
1+2+3+4 | 161210 ± 466 | 5.7%
1+3 | 162271 ± 434 | 6.4%

Hence, only 1 and 3 contribute to the speedup on my system (for the given benchmark). The reasons why 2 (Dirichlet noise) and 4 (SearchWorker constructor) don't have any effect (slightly negative effect in fact) is that they happen only once per search, as argued above by @mooskagh and myself, respectively.

borg323:
My tests are similar (also on linux). There is also a pre-existing reserve in search.cc that doesn't seem to help in my tests. @ddobbelaere can you verify this?

ddobbelaere:
Sure, let's call this existing reserve 3* (as its also in `SearchWorker::AddNodeToComputation`).

Then these are my back-to-back measurements:

`lc0`  |  nps | speedup
------ | ----- | -----------
1+3+3* | 162402 ± 292 | 6.5%
1+3-3* | 161582 ± 728 | 5.9%

So, the pre-existing 3* has its effect. On top of the lower average, the variance of the second case is also higher due to several outliers, which I think clearly hints at memory reallocation that is worse in some runs (depending on the memory layout at that point). Note that the case 1+3+3* = 1+3 is consistent with the previous measurement in the above table (so very reproducible, as it should be).

src/mcts/search.cc
SunnyWar: I think I'm going to give up benchmarking anything of this laptop. I think has some dramatic temperature controls that make it unreliable for benchmarking.

SunnyWar:
Thank you _all_ for your excellent insight and investigation. I'm removed the noise and minibatch reserve statements.

ankan-ban:
With v20 dev, here are the CPU side hotspots (running with a very small 32x4 network, with 2 threads on GTX 970 on Windows 10):
```
Sorted by Self time:
--------------------
Function Name	                                                          Total (ms) Total (%)	Self(ms) Self (%)
  lczero::SearchWorker::PickNodeToExtend	                                1977	37.76%	1079	20.61%	
  lczero::Edge::GetP	                                                         181	3.46%	181	3.46%	
  lczero::Edge_Iterator<0>::Actualize	                                         104	1.99%	104	1.99%	
  lczero::Node::FinalizeScoreUpdate	                                          93	1.78%	93	1.78%	
  lczero::CudnnNetworkComputation<float>::GetPVal	                          74	1.41%	74	1.41%	
  lczero::ChessBoard::GeneratePseudolegalMoves	                                 264	5.04%	73	1.39%	
  lczero::LruCache<unsigned __int64,lczero::CachedNNRequest>::LookupAndPin	  76	1.45%	44	0.84%	
  lczero::ChessBoard::IsLegalMove	                                          45	0.86%	37	0.71%	


Total Time function call tree:
------------------------------
Function Name	Total CPU (ms)	Total CPU (%)	Self CPU (ms)	Self CPU (%)	Module
 - lczero::SearchWorker::RunBlocking	5172	98.78%	1	0.02%	lc0.exe
| - lczero::SearchWorker::GatherMinibatch	3003	57.35%	5	0.10%	lc0.exe
|| - lczero::SearchWorker::PickNodeToExtend	1977	37.76%	1079	20.61%	lc0.exe
||| + std::shared_timed_mutex::lock	304	5.81%	3	0.06%	lc0.exe
|||   lczero::Edge::GetP	154	2.94%	154	2.94%	lc0.exe
||| + lczero::Node::Edges	109	2.08%	6	0.11%	lc0.exe
||| + std::shared_timed_mutex::unlock	107	2.04%	1	0.02%	lc0.exe
||| + lczero::PositionHistory::Append	87	1.66%	4	0.08%	lc0.exe
||| + lczero::Edge_Iterator<0>::GetOrSpawnNode	50	0.95%	7	0.13%	lc0.exe
|||   [External Code]	39	0.74%	39	0.74%	Multiple modules
||| + lczero::EdgeAndNode::GetVisitsToReachU	32	0.61%	21	0.40%	lc0.exe
||| + lczero::`anonymous namespace'::GetFpu	12	0.23%	6	0.11%	lc0.exe
|||   lczero::Node::TryStartScoreUpdate	3	0.06%	3	0.06%	lc0.exe
|||   sqrt	1	0.02%	1	0.02%	lc0.exe
|| - lczero::SearchWorker::ExtendNode	445	8.50%	2	0.04%	lc0.exe
||| + lczero::ChessBoard::GenerateLegalMoves	370	7.07%	16	0.31%	lc0.exe
||| + lczero::Node::CreateEdges	65	1.24%	1	0.02%	lc0.exe
|||   [External Code]	7	0.13%	7	0.13%	Multiple modules
|||   lczero::ChessBoard::HasMatingMaterial	1	0.02%	1	0.02%	lc0.exe
|| + lczero::SearchWorker::AddNodeToComputation	343	6.55%	8	0.15%	lc0.exe
|| + lczero::SearchWorker::FetchSingleNodeResult	63	1.20%	8	0.15%	lc0.exe
|| + lczero::SearchParams::GetMiniBatchSize	56	1.07%	0	0.00%	lc0.exe
|| + std::shared_timed_mutex::lock	53	1.01%	0	0.00%	lc0.exe
|| + lczero::SearchWorker::DoBackupUpdateSingleNode	47	0.90%	3	0.06%	lc0.exe
|| + std::vector<lczero::CachingComputation::WorkItem,std::allocator<lczero::CachingComputation::WorkItem> >::pop_back	9	0.17%	1	0.02%	lc0.exe
||   std::vector<lczero::SearchWorker::NodeToProcess,std::allocator<lczero::SearchWorker::NodeToProcess> >::emplace_back<lczero::SearchWorker::NodeToProcess>	3	0.06%	3	0.06%	lc0.exe
|| + std::shared_timed_mutex::unlock	1	0.02%	0	0.00%	lc0.exe
||   lczero::CudnnNetworkComputation<float>::GetBatchSize	1	0.02%	1	0.02%	lc0.exe
| - lczero::CachingComputation::ComputeBlocking	1783	34.05%	23	0.44%	lc0.exe
|| + lczero::CudnnNetwork<float>::forwardEval	1592	30.40%	0	0.00%	lc0.exe
|| + lczero::LruCache<unsigned __int64,lczero::CachedNNRequest>::Insert	86	1.64%	35	0.67%	lc0.exe
||   lczero::CudnnNetworkComputation<float>::GetPVal	55	1.05%	55	1.05%	lc0.exe
|| + std::make_unique<lczero::CachedNNRequest,unsigned __int64>	26	0.50%	0	0.00%	lc0.exe
||   lczero::CudnnNetworkComputation<__half>::GetQVal	1	0.02%	1	0.02%	lc0.exe
| - lczero::SearchWorker::DoBackupUpdateSingleNode	193	3.69%	4	0.08%	lc0.exe
|| + lczero::Search::GetBestChildNoTemperature	99	1.89%	0	0.00%	lc0.exe
||   lczero::Node::FinalizeScoreUpdate	89	1.70%	89	1.70%	lc0.exe
||   lczero::Node::CancelScoreUpdate	1	0.02%	1	0.02%	lc0.exe
| - lczero::SearchWorker::FetchSingleNodeResult	177	3.38%	24	0.46%	lc0.exe
||   [External Code]	73	1.39%	73	1.39%	ucrtbase.dll
||   lczero::Edge::GetP	19	0.36%	19	0.36%	lc0.exe
||   lczero::CudnnNetworkComputation<float>::GetPVal	19	0.36%	19	0.36%	lc0.exe
||   lczero::Edge::SetP	17	0.32%	17	0.32%	lc0.exe
||   lczero::Move::as_nn_index	8	0.15%	8	0.15%	lc0.exe
||   lczero::Edge::GetMove	8	0.15%	8	0.15%	lc0.exe
||   lczero::Node::Edges	5	0.10%	5	0.10%	lc0.exe
||   lczero::CachingComputation::GetPVal	3	0.06%	3	0.06%	lc0.exe
||   lczero::CachingComputation::GetQVal	1	0.02%	1	0.02%	lc0.exe
```

borg323:
Used the trick from https://stackoverflow.com/questions/16306012/github-pull-request-showing-commits-that-are-already-in-target-branch to fix the rebase/merge issues.

src/mcts/search.h
```diff
@@ -193,8 +193,10 @@ class Search {
 class SearchWorker {
  public:
   SearchWorker(Search* search, const SearchParams& params)
-      : search_(search), history_(search_->played_history_), params_(params) {}
-
+      : search_(search), history_(search_->played_history_), params_(params) {
+    minibatch_.reserve(300);
```
ankan-ban: better reserve params_.GetMiniBatchSize() ?

ankan-ban:
I have created another PR to fix some of the small perf issues I found in CPU limited cases:
https://github.com/LeelaChessZero/lc0/pull/602

PR #593, opened 2018-12-18
`n_in_flight_` was not reverted properly in collision nodes.

mooskagh:
When collision happened, collision code didn't have `n_in_flight_` reverted (`collision_value - 1` leaked into n_in_flight_).
Most of collisions are small in size so most likely it didn't cause large problems, but with large collisions it may happen that it leads to a blunder.

Bug was introduced in commit https://github.com/mooskagh/lc0/commit/636d326697cf88c8fc657e137ef707e1575db3d1

### DIFF
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -234,7 +234,7 @@ void Node::MakeTerminal(GameResult result) {
   }
 }
 
-bool Node::TryStartScoreUpdate() { return n_in_flight_++ == 0 || n_ != 0; }
+bool Node::StartScoreUpdate() { return n_in_flight_++ == 0 || n_ != 0; }
 
 void Node::CancelScoreUpdate(int multivisit) { n_in_flight_ -= multivisit; }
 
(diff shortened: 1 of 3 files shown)


Tilps:
It is possible that the increased liklihood of deviation at leaf level could either be a bad thing, or something that should be tuned by reducing max collision multivisits to avoid it happening quite so much.  Such tuning may make it even more clear that there should be separate max collision counters for terminal hits vs collisions.
The old behavior is however a bit odd as the number of multivisit calculated is based on including the leaves switching, so its possible that it took more collision events than 1 to actually switch things most times, so maybe a lower maximum collision events would also be justified.

zz4032:
Short tuning of PR593 at ~50 nodes on opening moves resulted in:
MaxCollisionEvents = 6
MaxCollisionVisits = 53000

Test match PR594 (current dev) vs. PR593-tuned:
Score of lc0-PR594 vs lc0-PR593-tuned: 171 - 112 - 217  [0.559] 500
Elo difference: 41.2 +/- 22.9, LOS: 100.0 %, DrawRatio: 43.4 %

mooskagh:
While this fix is technically more correct, alternative fix #594 gave better Elo, so closing this.

PR #594, opened 2018-12-19
Alternative 'minimalist' bugfix for same issue PR593 fixes.

Tilps:


zz4032:
This PR shows a good performance against current dev:
```
Score of lc0-dev vs lc0-dev-PR594: 214 - 258 - 528  [0.478] 1000
Elo difference: -15.3 +/- 14.8, LOS: 2.1 %, DrawRatio: 52.8 %
```

While PR593 shows a strong regression, aborted after 50 games:
```
Score of lc0-dev vs lc0-dev-PR593: 41 - 1 - 8  [0.900] 50
Elo difference: 381.7 +/- 137.6, LOS: 100.0 %, DrawRatio: 16.0 %
```

Both matches at very low TC (about 50 nodes/move on opening moves)

Tilps:
Based on results I would propose we submit this while we come up with a better strategy - since this is just a bugfix, while the improved strategy will probably require tuning which will take time.

PR #599, opened 2018-12-21
Add simple round robin network and a basic demuxing network.

Tilps:


Tilps:
Using random backend with 5 backends and 5 threads, I measure betwee 5-10% nps improvement using this backend compared to multiplexing.

src/neural/network_lb.cc
```diff
@@ -0,0 +1,85 @@
...
+
+#include <condition_variable>
+#include <queue>
+#include <thread>
+#include "utils/exception.h"
+
+namespace lczero {
+namespace {
+
+class LoadBalancingNetwork : public Network {
```
mooskagh: Maybe rename it to RoundRobinNetwork? It doesn't really balance load (e.g. it doesn't look which of backends is free).

Tilps: Yeap I was thinking the same thing - done.

Tilps:
I decided to introduce the round robin concept into demux as well, so that I could introduce an optional minimum split size and not lead to too much concentration on a single gpu if there are multiple feeding threads.

src/neural/network_demux.cc
```diff
@@ -0,0 +1,224 @@
...
+  std::mutex mutex_;
+  std::condition_variable dataready_cv_;
+  int dataready_ = 0;
+  int partial_size_ = 0;
+};
+
+class DemuxingNetwork : public Network {
+ public:
+  DemuxingNetwork(const WeightsFile& weights, const OptionsDict& options) {
+    // int threads, int max_batch)
```
mooskagh: Probably kill those two commented lines.

src/neural/network_demux.cc
```diff
@@ -0,0 +1,224 @@
...
+  }
+  dataready_cv_.wait(lock, [this]() { return dataready_ == 0; });
+}
+
+std::unique_ptr<Network> MakeDemuxingNetwork(const WeightsFile& weights,
+                                             const OptionsDict& options) {
+  return std::make_unique<DemuxingNetwork>(weights, options);
+}
+
+REGISTER_NETWORK("demux", MakeDemuxingNetwork, -1000)
```
mooskagh: Maybe some unique values instead of -1000 (-1001, -999?) so that their priority order is stable.

src/neural/network_rr.cc
```diff
@@ -0,0 +1,85 @@
...
+#include <thread>
+#include "utils/exception.h"
+
+namespace lczero {
+namespace {
+
+class RoundRobinNetwork : public Network {
+ public:
+  RoundRobinNetwork(const WeightsFile& weights, const OptionsDict& options) {
+    // int threads, int max_batch)
```
mooskagh: Also those lines.

PR #601, opened 2018-12-21
Add an option to restrict resign in self play to only trigger after a certain number of moves.

Tilps:


gonzalezjo:
Why bother? Isn't part of the purpose of resign to end games quickly after temperature blunders?

src/selfplay/game.cc
```diff
@@ -104,7 +110,9 @@ void SelfPlayGame::Play(int white_threads, int black_threads, bool training,
     float eval = search_->GetBestEval();
     eval = (eval + 1) / 2;
     if (eval < min_eval_[idx]) min_eval_[idx] = eval;
-    if (enable_resign) {
+    if (enable_resign && ply_count / 2 >= (options_[idx].uci_options->Get<int>(
+                                               kResignEarliestMoveId.GetId()) -
+                                           1)) {
```
borg323: I don't like the `-1` here. On the other hand, `resign-earliest-move` is clearer than something like `resign-after-move`. Can you think of a better name for the option in order to remove it?

src/selfplay/game.cc
```diff
@@ -63,6 +67,8 @@ void SelfPlayGame::Play(int white_threads, int black_threads, bool training,
                         bool enable_resign) {
   bool blacks_move = false;
 
+  int ply_count = 0;
```
mooskagh: This is available through `tree_[0]->GetPositionHistory().GetLength()`, probably better than adding a counter.

src/selfplay/game.cc
Tilps: While resolving mooskagh's comment I factored it so it was a + 1 on the other side and assigned to variable with name that should make the logic clearer - hopefully that is good enough.

PR #605, opened 2018-12-23
Fix demux to match cuda expected threading model for computations.

Tilps:


Tilps:
Jesse has tested this - seems its actually a win once it doesn't crash.... :)
Went from 35k nps to 75k nps switching from mux to demux with 4 gpus on a system with cpu + mutex constraints.

PR #606, opened 2018-12-23
Minor changes to `src/chess`

ddobbelaere:
Some minor changes to `src/chess/board.{h,cc}`

If I am not mistaken, `ChessBoard::GenerateLegalMovesAndPositions` is not used anymore? Can this be removed or is it still present to cross-check some possible future optimizations of `IsLegalMove` used in `ChessBoard::GenerateLegalMoves`?

### DIFF
--- a/src/chess/board.h
+++ b/src/chess/board.h
@@ -33,8 +33,6 @@
 
 namespace lczero {
 
-struct MoveExecution;
-
 // Represents a board position.
 // Unlike most chess engines, the board is mirrored for black.
 class ChessBoard {
@@ -75,8 +73,6 @@ class ChessBoard {
   MoveList GenerateLegalMoves() const;
   // Check whether pseudolegal move is legal.
   bool IsLegalMove(Move move, bool was_under_check) const;
-  // Returns a list of legal moves and board positions after the move is made.
-  std::vector<MoveExecution> GenerateLegalMovesAndPositions() const;
 
   uint64_t Hash() const {
     return HashCat({our_pieces_.as_int(), their_pieces_.as_int(),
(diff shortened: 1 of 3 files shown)


ddobbelaere:
Added some hash calculation improvements. Yields small speedup (~0.5%) for default params. Also fixes poor "spreading" of `positions` (that at the moment only seems necessary to allow for future dynamic parameter changes?).

src/chess/board.h
```diff
@@ -81,8 +81,10 @@ class ChessBoard {
   uint64_t Hash() const {
     return HashCat({our_pieces_.as_int(), their_pieces_.as_int(),
                     rooks_.as_int(), bishops_.as_int(), pawns_.as_int(),
-                    our_king_.as_int(), their_king_.as_int(),
-                    castlings_.as_int(), flipped_});
+                    (static_cast<unsigned>(our_king_.as_int()) << 24) |
```
mooskagh: `unsigned` has different length at different platforms, it's safer to cast to something of known size (e.g. `uint64_t` or `uint32_t`)

mooskagh:
`GenerateLegalMovesAndPositions` is from time when every node contained position. Can be removed now!

ddobbelaere:
Reverted modifications to `PositionHistory::HashLast`. I think `uint64_t hash = positions;` is not critical at all (in fact, `uint64_t hash = 0;` would do also), as `positions` is implicitly influencing number of hashed positions.

src/chess/board.h
ddobbelaere: Thanks, you are right!

PR #608, opened 2018-12-24
Set Default threads

SunnyWar:
Set the default threads to the number of supported hardware threads, which is usually the number of cores.


SunnyWar:
dammit.

PR #609, opened 2018-12-24
make uci search parameters the defaults ones

borg323:


src/benchmark/benchmark.cc
```diff
@@ -60,7 +60,8 @@ void Benchmark::Run() {
 
   defaults->Set<int>(SearchParams::kMiniBatchSizeId.GetId(), 256);
   defaults->Set<float>(SearchParams::kFpuReductionId.GetId(), 1.2f);
-  defaults->Set<float>(SearchParams::kCpuctId.GetId(), 3.4f);
+  defaults->Set<float>(SearchParams::kCpuctId.GetId(), 3.0f);
```
mooskagh: Maybe make those defaults and override in selfplay?

borg323: Why not a helper function in SearchParams that changes the defaults?

borg323:
In the end I decided that it is better for all search parameters to default to the uci values with overrides for selfplay only.

PR #617, opened 2018-12-27
Add default number of thread as number of cores.

SunnyWar:
Sets the default number of threads to the hardware supported number, which is normally the number of cores. It the api can't determine this it set's it to 2.

### DIFF
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -36,7 +36,6 @@
 
 namespace lczero {
 namespace {
-const int kDefaultThreads = 2;
 
 const OptionId kThreadsOptionId{"threads", "Threads",
                                 "Number of (CPU) worker threads to use.", 't'};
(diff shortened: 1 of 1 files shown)


jjoshua2:
It's probably better to do 1 plus number of gpus by default. Threads can
make things worse as much as better with all the locks...
And more threads means lower clock speeds these days with pbo xfr
turboboost etc

SunnyWar:
I don't think the pbo xfr stuff has much to do with the number of thread as it is about clock frequency, same with turboboost. 

I think Leelas's defaults should be that which is likely to provide the best performance. If that is not the case currently with this PR then I'm happy to close it.

mooskagh:
While initially it seemed like a good idea (I think it's been in lc0.org/timeline spreadsheet for quite a while), it turned out that our non-parallelized code (node picking) is the bottleneck, so after having enough threads to load GPUs (which is probably 1+number_of_gpus) adding more threads actually make performance worse.
So I think it should not go through until node picking parallelization is solved. (there is ongoing #678 for that btw).

borg323:
Please close this, we know now that this is very rarely the best option.

PR #618, opened 2018-12-27
Add const to locals

SunnyWar:
Added const to local variables that can be const. This will identify non-const variables to insure that they are that way intentionally.


src/syzygy/syzygy.cc
```diff
@@ -421,7 +421,7 @@ void init_indices() {
   }
   // TODO: choose a good seed.
   std::mt19937 gen(123523465);
-  std::uniform_int_distribution<Key> dist(std::numeric_limits<Key>::lowest(),
+  const std::uniform_int_distribution<Key> dist(std::numeric_limits<Key>::lowest(),
```
borg323: g++/clang on linux complain about this (and the other `const` on distributions in `random.cc`)

SunnyWar: I'll look into it.

SunnyWar: I tried a variety of way to repro this in windows, including using the intel compiler, and attempting to set up a Linux sub system, but to no avail. So, I'm simply fixing the issues as demonstrated in the build log.

borg323: You missed the one in `syzygy.cc`.

mooskagh:
This looks good, but I'll merge it a bit later (after RC season) as such a widespread change may make cherrypicking harder.

SunnyWar:
sure thing

mooskagh:
@SunnyWar 
Note that that the code does not meet our style guide requirements (e.g. there are lines longer than 80 characters), but  I think it's better to merge it, otherwise it's always in "needs merging" state.

Please reformat code with `clang-format` and submit fix.

PR #101, opened 2018-06-20
CircleCI Linux automatic builds

DanielUranga:
Initial CircleCI support. Builds using Ubuntu Linux 18.04.
Build log here: https://circleci.com/gh/DanielUranga/lc0/10
This is an attempt to start solving https://github.com/LeelaChessZero/lc0/issues/50

PR #621, opened 2018-12-27
Easily run latest releases of lc0 and client using NVIDIA docker

vochicong:
I've make a repo
https://github.com/vochicong/lc0-docker
that help you easily run the latest versions of lc0 and the client under docker.
Both CPU and GPU versions are available.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,8 +102,8 @@ Make sure that `~/.local/bin` is in your `PATH` environment variable. You can no
 
 #### Docker
 
-Follow https://github.com/vochicong/lc0-nvidia-docker
-to run lc0 and the client under NVIDIA docker.
+Use https://github.com/vochicong/lc0-nvidia-docker
+to run latest releases of lc0 and the client under NVIDIA docker.
 
 ### Windows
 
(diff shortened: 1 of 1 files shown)


DanielUranga:
Consider re-using the `danieluranga/leela_chess_zero-lc0_ubuntu_builder:0.0.4` image. Also probably it would be better to just add instructions to the wiki, and add a link to the nvidia docker documentation also.

vochicong:
Do you have the link to Dockerfile of `danieluranga/leela_chess_zero-lc0_ubuntu_builder`?

vochicong:
@DanielUranga It seems that the `danieluranga/leela_chess_zero-lc0_ubuntu_builder` image is created from https://github.com/LeelaChessZero/lc0/blob/master/.circleci/Dockerfile #101. This one looks good for CircleCi to build and test lc0, but not god for someones to run lc0 and the client to contribute for lc0 training.

vochicong:
> Don't know what it is all about, but probably something useful. :)

Yes it is. It's actually running as a top **active user** at https://lczero.org/active_users.

Issue #627, opened 2018-12-29
Lc0 missing mate in #1 in tight endgame


MelleKoning:
I've posted this bug in the forum:
https://groups.google.com/forum/#!topic/lczero/O0OWPd3ujE4

The position is visible on lichess: https://lichess.org/8j8a0wQy
The blunder move is 47. Qa8-a7, while it should be 47. Bf1-g2

How to reproduce:

1. Start lc0 with a recent weightsfile, example:
>lc0 --weights=nets\32300

2. set the position right at the position the mate can still be prevented with this command:

`position startpos moves e2e4 c7c5 g1f3 d7d6 d2d4 c5d4 f3d4 g8f6 b1c3 a7a6 c1e3 e7e5 d4b3 f8e7 f2f3 c8e6 d1d2 h7h5 c3d5 f6d5 e4d5 e6f5 b3a5 b7b6 a5c4 b8d7 a2a4 e8g8 f1e2 a8c8 e1g1 h5h4 h2h3 f5g6 b2b4 d8c7 f1c1 c7b7 c2c3 f8e8 e2f1 c8b8 e3f2 f7f5 d2e1 b7c7 f2h4 e7h4 e1h4 f5f4 a4a5 b6b5 c4d2 d7f6 c3c4 b5c4 d2c4 b8b4 c4b6 c7e7 f1a6 e5e4 h4f4 f6h5 f4e3 e7e5 f3f4 h5f4 a6f1 f4d3 c1d1 e5b2 b6d7 b4b3 a5a6 d3e5 e3g5 b3h3 g5d2 b2b3 d7e5 b3g3 e5g4 g3g4 d2e1 h3h8 a6a7 e4e3 a7a8Q g4f4 g2g3 f4h6`

3. now let leela think for 10.000 nodes and you will see this:
```
go nodes 10000
Loading weights file from: nets\32300
Creating backend [cudnn]...
info depth 1 seldepth 2 time 123 nodes 5 score cp 270 hashfull 0 nps 40 tbhits 0 pv a8e8 g6e8
info depth 2 seldepth 3 time 196 nodes 6 score cp 406 hashfull 0 nps 30 tbhits 0 pv a8e8 g8h7 g3g4
info depth 2 seldepth 3 time 252 nodes 11 score cp 95 hashfull 0 nps 43 tbhits 0 pv f1g2 g8f7 g3g4
[...]

PR #629, opened 2018-12-29
Calculate limits for tb hits differently to collisions.

Tilps:
To fix Issue #627 maybe.

The new code as it currently stand contains a huge mess of assumptions that are probably wrong.  I'm doing some quick Elo testing to see how badly wrong they are....

### DIFF
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -253,7 +253,8 @@ void Node::FinalizeScoreUpdate(float v, int multivisit) {
   best_child_cached_ = nullptr;
 }
 
-void Node::UpdateBestChild(const Iterator& best_edge, int visits_allowed) {
+void Node::UpdateBestChild(const Iterator& best_edge, int visits_allowed,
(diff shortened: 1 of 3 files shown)


Tilps:
I could make it less conservative if there are any noticeable regressions, but the logic will get a lot more complicated.

Tilps:
5second + 0.1 run shows possible regression but within error bars.
Score of lc0 dev local build vs lc0 PR619: 26 - 35 - 139 [0.477]
Elo difference: -15.65 +/- 26.57

mooskagh:
The current terminal multivisits are clearly broken, but I'm a bit hesitant to grow `Node` structure even more.

And probably we should have this fixed before TCEC superfinal. Maybe just disable multivisits for terminal nodes (and keep them only for collisions)?..

killerducky:
@Tilps do we need this now the #683 is in?

Tilps:
No need to keep it open - Can ressurrect the logic in future if desired.

PR #630, opened 2018-12-30
Enhancement to move selection by using Q along with N

szacer:
This PR is a modification to how Leela chooses the best move to return. MCTS proceeds as before, expect at the very end when submitting a final move to play. Currently, Leela submits the move with the most visits (N). This change would cause Leela to slightly scale the number of visits based on the Q value of that move. So if a certain move had slightly less visits than the move with the highest visits, but a better Q value, it could be chosen as the best move.  

If Q is the Q value of a certain move and N is its visits, and Qtop is the Q value of the move with most visits, then Leela scores each move accordingly:

If Q >= Qtop: modified_N = N * exp( 1.06 * sqrt(Q - Qtop))
Else : modified_N = N * exp( -1.06 * sqrt(Qtop - Q))

The move with the highest modified N is chosen. 

This PR has performed well in testing, scoring [+119, -89, =192] in a 400 game match against release v0.20.0-rc2 using ID32332. That corresponds to +26.1 (±25.0) elo. LOS = 1.88%. Time control was 5sec +0.25sec increment on a GTX1050. Leela was getting 500-1000 nodes per move. PGN file attached. [Main_vs_Qbias2.txt](https://github.com/LeelaChessZero/lc0/files/2716794/Main_vs_Qbias2.txt)

Match conditions are at a low node count. However, from a theoretical point of view, I do not think node count should have an impact on performance of the PR. Also, the coefficient of 1.06 in the formula is probably not the ideal number to use, and performance could possibly be improved further by adjusting this. 

From a theoretical point of view, here is what I was thinking when I decided to experiment with this.  Leela uses its policy and value network to decide where in its tree it should explore next. It then plays the move it explored the most, as this is what its policy and value heads found most interesting. However, I felt relatively speaking, the value head's importance should increase in deciding upon the final move versus during tree exploration.

I hope that this summary was clear for you. Feel free to ask me any questions.

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -520,9 +520,83 @@ std::vector<EdgeAndNode> Search::GetBestChildrenNoTemperature(Node* parent,
   return res;
 }
 
-// Returns a child with most visits.
+// Chooses an edge from parent with the highest N value, and returns this edge's
+// Q value
+float Search::GetQ_with_highest_N(Node* parent) const {
+  MoveList root_limit;
+  if (parent == root_node_) {
+    PopulateRootMoveLimit(&root_limit);
+  }
+
+  int N = 0;
+  float Q = -3;
+
+  for (auto edge : parent->Edges()) {
+    if (parent == root_node_ && !root_limit.empty() &&
+        std::find(root_limit.begin(), root_limit.end(), edge.GetMove()) ==
+            root_limit.end()) {
+      continue;
+    }
+
+    // Very simple code to find the maximum N value of an edge
+    // and record the corresponding Q value
+    if (edge.GetN() >= N) {
+      N = edge.GetN();
+      Q = edge.GetQ(-1);
+    }
+  }
+
+  return Q;
+}
+
+// Very similar to GetBestChildrenWithNoTemperature, expect it chooses
+// the best child with a bit more emphasis on the Q value of each child.
+// Also takes in the return value of GetQ_with_highest_N as Qtop
+std::vector<EdgeAndNode> Search::GetQBiasedMove(Node* parent,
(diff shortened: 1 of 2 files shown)


gsobala:
After about 80 games at 2min+1sec vs SF10, this pr is not obviously superior to old master, though error bars are obviously very wide.

```
Program                        Elo      +      -   Games   Score   Av.Op.  Draws
lc0 32319                  : 3442.0   97.3   97.3    82    45.1 %   3476   73.2 %
lc0-pr630                  : 3432.2   96.8   96.8    80    43.8 %   3476   60.0 %
```

szacer:
That is a very interesting result. Thank you for taking the time to run that test. It might be due longer time controls or comparing against stockfish instead of Leela. When I calculate the error bars I get that pr630 is -10 ± 65 elo compared to main. That means if pr630 was 26 elo stronger than main, there is only a 13% chance you would have obtained the result you did or worse. 13% could possibly have happened by chance or  be down to poor scaling of pr630. 

My laptop isn't really powerful enough to run a 400 game match at blitz time controls as I have overheating issues. However, I have written a script that compares the move choices of pr630 and the current main to the moves of AlphaZero in the Stockfish 8 match from TCEC positions. This will greatly help me run large-sample tests at long time controls. As when playing a match, Leela might think for 100 plys and at the end I only get one result. This way, if Leela thinks for 100 plys, I get 100 results back. 

I am going to leave my laptop running overnight analysing about 2000 positions at 10k nodes per position.  This nodes per move figure would roughly correspond to Leela playing at 5min+3 on my computer. This will hopefully give some insight into how this PR performs at longer time controls and the sample size should be large enough to have a small error.

PR #632, opened 2018-12-30
Make const'able functions const

SunnyWar:
Make const'able functions const


mooskagh:
I think for most of cases changed here it's not really beneficial to add const, because:

- Sometimes while function happen not to modify member variables, logically it's mutable.
- It breaks consistency (e.g. OnBestMove vs OnInfo).

As function const-ness participates in function resolution, those inconsistencies may introduce bugs.

So I think while missing `const` in some of functions is geniune oversight, most of them are actually more correct as they are.

SunnyWar:
ok

PR #633, opened 2018-12-30
simplify smallarray

SunnyWar:
simplify smallarray by using a vector

### DIFF
--- a/src/neural/cache.h
+++ b/src/neural/cache.h
@@ -33,7 +33,7 @@
 namespace lczero {
 
 struct CachedNNRequest {
-  CachedNNRequest(size_t size) : p(size) {}
+  CachedNNRequest(uint8_t size) : p(size) {}
   typedef std::pair<uint16_t, float> IdxAndProb;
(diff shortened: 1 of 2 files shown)


mooskagh:
The reason SmallArray is used instead of vector, is to save memory (Vector stores pointer+size+capacity, while SmallArray stores pointer+size).
Size is 1 byte, but due to alligment, the structure is probably still 16 (or maybe 12) bytes. Still better than 24.
Anyway, implementing SmallArray with a vector defeats the purpose of SmallArray.

SunnyWar:
>           The reason SmallArray is used instead of vector, is to save memory (Vector stores pointer+size+capacity, while SmallArray stores pointer+size).
> Size is 1 byte, but due to alligment, the structure is probably still 16 (or maybe 12) bytes. Still better than 24.
> Anyway, implementing SmallArray with a vector defeats the purpose of SmallArray.

You don't even know the exact sizes? I do. Why close this without do consideration?

PR #634, opened 2018-12-30
Simplify movestogo approximator to use median residual time.

fli:
cc: @jjoshua2 
Replaces calculating the mean residual time (summing to approximate an integral) with the median residual time (a much simpler function) as per @borg323's suggestion in #271. So it now uses 2 std::pow calls instead of 300.

I've also adjusted the default parameters to match the current tuned curve better. See purple and red lines in graph below. Red = current master curve, purple = this PR curve.

screen shot 2018-12-30 at 8 34 22 pm

I haven't tested this at all apart from the spreadsheet and won't be able to get to it for a while, so would appreciate some help with that.

### DIFF
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -97,15 +97,22 @@ const size_t kAvgCacheItemSize =
     NNCache::GetItemStructSize() + sizeof(CachedNNRequest) +
     sizeof(CachedNNRequest::IdxAndProb) * kAvgMovesPerPosition;
 
-float ComputeSurvivalAtPly(int ply, float midpoint, float steepness) {
-  // This function is the survival function of the log-logistic distribution, it
-  // was chosen as it fit empirical analysis finding P(game ended at ply). We
-  // determine how many moves to plan time management for by summing over this
-  // function from ply to infinity (or some other reasonably large value).
-  // midpoint: The ply where the function is half its maximum value.
+float ComputeMedianMovesToGo(int ply, float midpoint, float steepness) {
+  // An analysis of chess games shows that the distribution of game lengths
+  // looks like a log-logistic distribution. The mean residual time function
+  // calculates how many more moves are expected in the game given that we are
+  // at the current ply. Given that this function can be expensive to compute,
+  // we calculate the median residual time function instead. This is derived and
+  // shown to be similar to the mean residual time in "Some Useful Properties of
+  // Log-Logistic Random Variables for Health Care Simulations" (Clark &
+  // El-Taha, 2015).
+  // midpoint: The median length of games.
   // steepness: How quickly the function drops off from its maximum value,
   // around the midpoint.
-  return 1 / (1 + std::pow(ply / midpoint, steepness));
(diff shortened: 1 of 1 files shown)


borg323:
A bit late to raise the issue, but I just noticed: `time-midpoint-halfmove` seems counter-intuitive, doing it in moves would have been better. Since it is hidden by default this isn't a big issue, but may be worth considering renaming it (and dividing by 2) before the release.

fli:
Agreed. I've changed it.

zz4032:
I tried several tuning sessions on short TC yesterday. It seems parameters can vary a lot depending on conditions (selfplay or vs. AB-engines, TC). So instead of several different tunings I will run only one at long TC (100k nodes/move) vs. Stockfish. This will be most valuable I think.

zz4032:
Here some results after 2 days (didn't expect it to take so long...):

374_clop
Tuning is running with the "old" code with plies instead of moves.
Red big dot is the optimum, one can see a linear correlation so we could make TimeSteepness a function of TimeMidpointMove (blue line in the diagram):
TimeSteepness = -0,18 * TimeMidpointMove + 19
It makes the curve slightly steeper, if values for TimeMidpointMove are decreased.
And maybe also restrict TimeMidpointMove to plausible values like 25 to 100.
What do you think?
@fli @jjoshua2

zz4032:
I would prefer to let it run a bit longer, but I think this is nevertheless a good point of stoppage.
TimeMidpointMove = 121.4 (TimeMidpointMove=60.7)
TimeSteepness = 9.08
I would recommend using these parameters for any long TC tournaments.
plot

fli:
Thanks @zz4032. I would want to see test matches to make sure it is an improvement. When I was tuning these params with CLOP I set them both as GammaParameters because a change from 8-9 isn't the same as 2-3 for example. I wonder how much of a difference that makes.

zz4032:
I forgot about the option GammaParameter. Probably would converge quicker since it's non-linear.

zz4032:
Tuned both parameters as LinearParameters and in a second tuning as GammaParameters at lower TC (26.1s/game+0.145s/move). After 858 games in each of the tunings:

**LinearParameter:**
```
TimeMidpointPly=107.0
TimeSteepness=8.301
```
**GammaParameter:**
```
TimeMidpointPly=106.6
TimeSteepness=7.592
```
TimeMidpointPly is very similar in both, TimeSteepness is not.
CLOP diagrams look completely different too.

**Test match:**
```
   # PLAYER              :  RATING  ERROR  POINTS   GAMES  DRAWS(%)  LOS(%)
   1 lc0-tuned-gamma     :    18.1   15.8   409.5     780      58.1    50.0
   2 lc0-tuned-linear    :    18.1   15.9   409.5     780      53.7    98.7
   3 stockfish-dev       :     0.0   ----   741.0    1560      55.9     ---
```
One more observation is that with longer TC both values are higher (tuned as LinearParameters in result posted above).

PR #635, opened 2018-12-30
WDL value head support

Ttl:
WDL head support for all backends. Includes just the backend support and doesn't do anything yet with the draw information.

Needs updated protobuf: https://github.com/LeelaChessZero/lczero-common/pull/6

Training code for replacing the old value head with WDL in existing old type network: https://github.com/Ttl/lczero-training/tree/wdl_surgery

~11248 with WDL value head: http://hforsten.com/leelaz/11248-wdl.pb.gz~ (Obsolete due to protobuf changes)

```
Score of lc0_11248_wdl vs lc0_11248: 15 - 5 - 45 [0.577]
Elo difference: 53.88 +/- 46.33, LOS: 98.73 %, DrawRatio: 69.2 %

TC: 10+0.5s on GTX 1050 Ti
```


Ttl:
Strength increase in testing is probably from the additional training on TB rescored test10 data instead of WDL head. I wouldn't expect it to be any stronger. All convolutional layers and policy head were set to non-trainable, so the only difference in this network is the fully connected layers of the value head.

Tests fail because protobuf needs to be updated.

Ttl:
Added the tree search part. Previous commit also includes parameter for adjusting the score of the draw, but I removed since it didn't gain any Elo in testing. Verbose-move-stats was modified to report WDL scores for testing purposes.

Also I added WDL head to a test30 net that is much better at evaluating draws than test10 net. It can be downloaded at: http://hforsten.com/leelaz/256x20-32585-wdl-4000.pb.gz

src/mcts/search.cc
```diff
@@ -236,29 +236,22 @@ std::vector<std::string> Search::GetVerboseStats(Node* node,
...
-    oss << "(Q: " << std::setw(8) << std::setprecision(5) << edge.GetQ(fpu)
+    auto Q = edge.GetQ(fpu);
+    auto D = edge.GetD();
+    auto W = 0.5f*(1.0f - D + Q);
+    auto L = 0.5f*(1.0f - D - Q);
+
+    oss << "(Q: " << std::setw(8) << std::setprecision(5) << Q
         << ") ";
 
-    oss << "(U: " << std::setw(6) << std::setprecision(5) << edge.GetU(U_coeff)
```
Tilps: Verbose stats still needs to print U. (And Q+U is convenient.)

Tilps:
Yes, but we can do that as a follow up PR.

TFiFiE:
> Previous commit also includes parameter for adjusting the score of the draw, but I removed since it didn't gain any Elo in testing.

But wouldn't the usefulness of that come from the ability to do stuff like force the engine to play for a draw or for a win or with some other contempt-like factor?

Tilps:
More work will be 

> > Previous commit also includes parameter for adjusting the score of the draw, but I removed since it didn't gain any Elo in testing.
> 
> But wouldn't the usefulness of that come from the ability to do stuff like force the engine to play for a draw or for a win or with some other contempt-like factor?

More work can be done on this aspect after the PR is submitted.

Ttl:
Needs modifications for protobuf changes in: LeelaChessZero/lczero-common#8

TFiFiE:
Maybe change https://github.com/orgs/LeelaChessZero/projects/1#card-10518034 to point to this instead?

Ttl:
Merged with master and implemented the protobuf changes.

Test network with WDL value head and convolutional policy head: http://hforsten.com/leelaz/128x10-az-pol-map-wdl-200000.pb.gz

src/mcts/node.h
```diff
@@ -306,7 +310,7 @@ class Node {
 #if defined(__i386__) || (defined(__arm__) && !defined(__aarch64__))
 static_assert(sizeof(Node) == 48, "Unexpected size of Node for 32bit compile");
```
Tilps: I'm guessing this needs updating to 52.

src/neural/network_random.cc
```diff
@@ -50,16 +51,23 @@ class RandomNetworkComputation : public NetworkComputation {
...
   }
 
   int GetBatchSize() const override { return inputs_.size(); }
+
   float GetQVal(int sample) const override {
     return (int(inputs_[sample] % 200000) - 100000) / 100000.0;
   }
+
+  float GetDVal(int sample) const override {
+    return (int(inputs_[sample] % 200000) - 100000) / 100000.0;
```
Tilps: This will always output the same D value as the Q value - which is wrong for a few reasons.
D should always be in the range 0 to 1.
Furthermore D should only ever be in the range 0 to (1-abs(Q)) since otherwise one of W or L must be negative.
Probably should HashCat in a constant, and then calculate its bounds correctly to know what divisor to use.

src/mcts/node.h
Ttl: I increased it to 52, but can't check if it's correct.

Tilps: Yeah, I'm confident its correct.  Adding a 32 bit field on 32 bit is much easier to guess what is going to happen than adding a 32 bit field on 64 bit.

borg323:
Just tested that 32 bit builds work.

lp200:
iirc, currently move_count is disabled
when it enable, will not it help to judge the draw more accurately?

PR #636, opened 2018-12-31
fix out of memory

SunnyWar:
1.  Prevents memory from growin out of bounds because GC can't keep up.
2. Simplifies the GC logic
3. Reduces the Mutex task switching.

### DIFF
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -70,17 +70,10 @@ class NodeGarbageCollector {
 
  private:
   void GarbageCollect() {
-    while (!stop_.load()) {
-      // Node will be released in destructor when mutex is not locked.
-      std::unique_ptr<Node> node_to_gc;
-      {
-        // Lock the mutex and move last subtree from subtrees_to_gc_ into
-        // node_to_gc.
-        Mutex::Lock lock(gc_mutex_);
-        if (subtrees_to_gc_.empty()) return;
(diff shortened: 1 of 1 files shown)


mooskagh:
That mutex is inside the loop for a reason. GC of a large tree make take a few seconds, we don't want to block main thread for so long. 

I doubt that it helps with running out of memory, but if it does, it's precisely because of that: we block main thread so that it doesn't do search, not because we make GC thread faster.

Also, the queue to GC has a length of number of possible moves from previous root (which is ~30 per move in average). It's unlikely that 30 mutex switches cause considerable slowdown.

SunnyWar:
> I doubt that it helps with running out of memory, but if it does, it's precisely because of that: we block main thread so that it doesn't do search, not because we make GC thread faster.

Ok, guess I'm a lier. I didn't see my machine rise to a working set to 20 GB without this change. I didn't see the memory say a steady 11 GB working set for a solid 2 hours. I guess that didn't happen.

SunnyWar:
>           Have you done per node tests? They would make it easier to tell if it is reducing memory consumption or just slowing down search. Also, have you done an nps comparison? That would be helpful.

I ran the benchmark several times with 1-4 threads and at different node counts and didn't see a slowdown. I honestly don't know what a "per node test" is. I am concerned that this change would break something from my limited viewpoint it appeared to be good.

PR #645, opened 2019-01-02
First move boost

Tilps:
I'm guessing someone will want this to be a configurable parameter.  But I doubt there is a lot of elo in tuning this, so I didn't bother...

I instead 'derived' this value by experiment - looking at the distribution of ratio of initial_nodes to total_playouts at the end of each turn across a bunch of games played under time control.
With default, there is a clear growth of that ratio over the first 3-4 moves out of book.  With the current multiplier of 3, that growth is not clear.
I didn't really expect to need a multiplier greater than 2, but 2 clearly still had a growth pattern happening.

Publishing in case someone wants to try and see if there is any elo improvement.  I don't think I have the patience :P

### DIFF
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -251,9 +251,8 @@ SearchLimits EngineController::PopulateSearchLimits(
   // tree to reuse.
   if (first_move_of_game_) {
     float adj = options_.Get<float>(kFirstMoveTimeAdjId.GetId());
-    LOGFILE << "Giving extra time: New move time is " << this_move_time * adj
-            << " ms";
     this_move_time *= adj;
+    LOGFILE << "Giving extra time: New move time " << this_move_time << " ms";
   }
 
   // Make sure we don't exceed current time limit with what we calculated.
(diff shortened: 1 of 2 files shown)


fli:
I like this change. Should it be scaled by the pruning aggressiveness factor in some way?

jjoshua2:
It probably should be something like 2.0 * smart pruning factor.

I was thinking that maybe ponder multiplier could be in the same scaling function as the first move scalar, but doesn't look too easy here...

mooskagh:
I think it's somehow neater to implement by setting `time_spared_ms_` to something else than 0 in the beginning of the game.

jjoshua2:
I heard @borg323 did some tests and this and it lost elo? Maybe this can be a hidden variable with default 1.0 so we can tune it. And then next release we will put some reasonable value there. What do you think? @Tilps

borg323:
There is no easy/clean way to use `time_spared_ms_` since we need the adjustment to be relative to `this_move_time` that is calculated relatively late. I added some patches introducing the `--first-move-time-adj` hidden option with a lower default of 1.6 that seems OK in my tests.

Naphthalin:
I think this is tackled in crem's smooth time manager which tries to keep nodes per move fixed over the game instead of allocated time, but I think there is no PR yet for it.

PR #646, opened 2019-01-02
Support several games within one thread in selfplay mode.

mooskagh:
* `--parallelism` flag is removed
* Instead `--selfplay-threads` and `--thread-parallelism` are introduced.
* Default minibatch-size is 4 now
* Max prefetch is 8.
* `multiplexing` is no longer the default backend.

This (in theory) allows running training games with minibatch-size=1 with massive parallelism

This needs thorough testing:
* Whether it helps at all
* What is the best parameters. Maybe we should return multiplexing.
* Whether it handles player1/player2 settings/stats correctly
* Whether resign still works as expected.


DanielUranga:
👍 yay for smaller batch-sizes in training games!

src/mcts/search.h
```diff
@@ -192,8 +192,10 @@ class Search {
 // within one thread, have to split into stages.
 class SearchWorker {
  public:
-  SearchWorker(Search* search, const SearchParams& params)
-      : search_(search), history_(search_->played_history_), params_(params) {}
+  SearchWorker(Search* search)
+      : search_(search),
+        history_(search_->played_history_),
```
borg323: For consistency I would make this `search->played_history_`

mooskagh: `history_` is a reference so sometimes it's important from where to initialize.
But here indeed as `search` is a pointer, it's fine.

Changed!

src/selfplay/tournament.cc
```diff
@@ -205,103 +189,156 @@ void SelfPlayTournament::PlayOneGame(int game_number) {
...
-                       ? 1
-                       : game.GetGameResult() == GameResult::WHITE_WON ? 0 : 2;
-      if (player1_black) result = 2 - result;
-      ++tournament_info_.results[result][player1_black ? 1 : 0];
-      tournament_callback_(tournament_info_);
-    }
+// Called when the SelfPlayGame is finished, so that it's processed.
+void SelfPlayTournament::SendGameReport(const SelfPlayGame& game) {
+  const int game_number = game.GetGameNumber();
+  const bool player1_black = (game_number % 2) != first_game_is_flipped_;
```
Tilps: Maybe it would be cleaner if SelfPlayGame knew which was which and carried that info around rather than recomputing it in multiple places?

src/selfplay/tournament.cc
```diff
@@ -98,8 +100,6 @@ void SelfPlayTournament::PopulateOptions(OptionsParser* options) {
   defaults->Set<bool>(SearchParams::kNoiseId.GetId(), true);
   defaults->Set<float>(SearchParams::kFpuReductionId.GetId(), 0.0f);
   defaults->Set<std::string>(SearchParams::kHistoryFillId.GetId(), "no");
-  defaults->Set<std::string>(NetworkFactory::kBackendId.GetId(),
```
Tilps: Why remove the multiplexing default?  Its more important than ever... ?

Tilps: Ahh I see why, theoretically its not needed if we have enough per thread parallelism.  But I think it is still likely to be useful.

src/neural/network_st_batch.cc
```diff
@@ -73,7 +76,7 @@ float SingleThreadBatchingNetworkComputation::GetDVal(int sample) const {
 
 float SingleThreadBatchingNetworkComputation::GetPVal(int sample,
                                                       int move_id) const {
-  return network_->parent_computation_->GetPVal(sample - start_idx_, move_id);
+  return network_->parent_computation_->GetPVal(sample + start_idx_, move_id);
```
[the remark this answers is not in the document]
mooskagh: Done!

src/selfplay/tournament.cc
```diff
@@ -38,11 +39,11 @@ const OptionId kShareTreesId{"share-trees", "ShareTrees",
                              "When on, game tree is shared for two players; "
                              "when off, each side has a separate tree."};
 const OptionId kTotalGamesId{"games", "Games", "Number of games to play."};
-const OptionId kParallelGamesId{"parallelism", "Parallelism",
-                                "Number of games to play in parallel."};
-const OptionId kThreadsId{
-    "threads", "Threads",
-    "Number of (CPU) worker threads to use for every game,", 't'};
+const OptionId kThreadsId{"selfplay-threads", "SelfplayThreads",
```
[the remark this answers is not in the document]
mooskagh: Done.

src/selfplay/tournament.cc
```diff
@@ -166,37 +161,26 @@ SelfPlayTournament::SelfPlayTournament(const OptionsDict& options,
...
     opt.cache = cache_[pl_idx].get();
     opt.uci_options = &player_options_[pl_idx];
     opt.search_limits = search_limits_[pl_idx];
 
     // "bestmove" callback.
-    opt.best_move_callback = [this, game_number, pl_idx, player1_black,
-                              verbose_thinking,
-                              &last_thinking_info](const BestMoveInfo& info) {
-      // In non-verbose mode, output the last "info" message.
-      if (!verbose_thinking && !last_thinking_info.empty()) {
```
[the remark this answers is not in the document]
mooskagh: Yeah it PR was kind of large enough so I thought I'd reintroduce it later.
But I can do that in that PR too if you think it worth it.

(for me it doesn't seem important because
- We use selfplay for training mostly anyway
- It doesn't slow anything down
- It doesn't take that much memory
Yeah it's kind of a waste to build it in memory if we are not going to dump it but I think it's fine. :)
)

src/selfplay/tournament.cc
mooskagh: Returned multiplexing back.

src/selfplay/tournament.cc
mooskagh: Done.

mooskagh:
PTAL

Naphthalin:
This seems to be a rather major change, and at the same time related to Tilps' private fork for the policy/value head ordo tests. How is the performance gain here relative to the updated PRs by ankan #749 and in v0.25?

mooskagh:
It's not really related.
It's a complete change (well, used to be, now it's probably easier to redo than to merge it).
The problem with the idea was that when you have many games in parallel, it takes long until the first game finishes.
I'm closing it for now.

PR #647, opened 2019-01-02
add pgo and lto build types to build.sh

borg323:


ddobbelaere:
Works on Fedora 29 64-bit.

I have no fast GPU. For the default `blas` backend, I get the same benchmarks nps results for `release`, `lto` and `pgo` (namely 455 nps). 

For the random backend I got initially much worse results for `pgo` (which seems normal as the profile for `blas` is totally different). Adding the switch `--backend random` to line 59 of `build.sh` yielded same results as `lto` build:

build type  |  random backend nps (best of 3 runs)
------ | -----
release | 246279
lto | 269587
pgo | 219219
'modified' pgo | 270013

I had to copy a network to the `networks` folder. However, if I start `./build/release/lc0` from the same folder it doesn't detect the network from the same folder. Is this logical? (Should I also copy the network to `./build/release` or start `lc0` from somewhere else?

peterdewit:
multiplexing + fp16  and defaults otherwise

`go nodes 5000000` each test run a couple times, results were similar

-----
weight 32382   

pgo 54888 nps
lgo 54607 nps


-----
weight 35602

rel 110620 nps
pgo 110353 nps
lto 115874 nps

borg323:
@ddobbelaere lc0 searches for a weights file in the directory of the executable and the `networks` subdirectory. I always specify the `-w` option when running form a different directory.
The build.sh script similarly searches the script's directory and the `networks` subdirectory for the newest weights file.

borg323:
@peterdewit what compiler were you using? And what weights file was used for the pgo profile generation?

peterdewit:
@borg323  
gcc 8.2.1  
weight 32382

Naphthalin:
What is the problem this PR attempts to fix? And does it relate in any way to the restriction to newer gcc/clang versions only since switching to C++17?

Naphthalin:
#1224 is more recent than this PR, any chance that this will get done?

borg323:
Currently `lto` is the default for our builds, so we should maybe replace it with a `fast` option (as rebuilds take long with lto). Also we can use the trivial backend for `pgo` profile generation.

borg323:
I simplified the script to work with the trivial backend, but I don't see a performance improvement over the default lto builds. Unless more testing shows otherwise I'll close it. (Keeping the "stale" label as a reminder.)

PR #648, opened 2019-01-03
Check for zero capacity in cache

SunnyWar:
Short circuit: do nothing if there is nothing to do.

lc0.exe benchmark --nodes=10000000 --backend=random -t 4
Benchmark final time 52.9688s calculating 186011 nodes per second.
lc0.exe benchmark --nodes=10000000 --backend=random --nncache=0 -t 4
Benchmark final time 45.2008s calculating 220642 nodes per second.

### DIFF
--- a/src/utils/cache.h
+++ b/src/utils/cache.h
@@ -75,7 +75,7 @@ class LruCache {
     ShrinkToCapacity(capacity_ - 1);
     ++size_;
     ++allocated_;
-    Item* new_item = new Item(key, std::move(val), 0);
+    Item* new_item = new Item(key, std::move(val));
     new_item->next_in_hash = hash_head;
     hash_head = new_item;
     InsertIntoLru(new_item);
@@ -191,8 +191,8 @@ class LruCache {
 
  private:
   struct Item {
-    Item(K key, std::unique_ptr<V> value, int pins)
-        : key(key), value(std::move(value)), pins(pins) {}
+    Item(K key, std::unique_ptr<V> value)
+        : key(key), value(std::move(value)) {}
     K key;
     std::unique_ptr<V> value;
     int pins = 0;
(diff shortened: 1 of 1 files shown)


src/utils/cache.h
```diff
@@ -62,6 +62,8 @@ class LruCache {
   // If @pinned, pins inserted element, Unpin has to be called to unpin.
   // In any case, puts element to front of the queue (makes it last to evict).
   V* Insert(K key, std::unique_ptr<V> val, bool pinned = false) {
+    if (capacity_ == 0) return val.get();
```
mooskagh: std::unique_ptr will be destroyed immediately on exit, so the function will return hanging pointer.

mooskagh: On a second look, this function is never used with pinning.

With that, that functionality can be removed (part with pinned==true), and this function should return void instead of V*.
Then it seems correct.

borg323: After this change there is no need for the `pins` argument in the `Item` constructor.

SunnyWar:
I resolved the issues pointed out and pushed a change.

src/utils/cache.h
```diff
@@ -115,6 +117,8 @@ class LruCache {
   // Unpins the element given key and value. Use of LruCacheLock is recommended
   // to automate this pin management.
   void Unpin(K key, V* value) {
+    if (capacity_ == 0) return;
```
mooskagh: I think it's better not to do that for unpin.
If cache has capacity of 0 during entire lifetime, then Unpin will never be called, and there's no harm in not having the shortcut.

However, in this (allowed but unlikely) situation we'll have a leak:
1. capacity is non-zero
2. Some entry is pinned
3. Capacity is reset to 0
4. Entry is unpinned.

We used to allow cache size change during the search, that made this situation real. Currently I believe we only do that between moves, but still if it's easy to support, we should.

SunnyWar: Makes sense. I'll remove the short circuit here.

mooskagh:
This produces warnings
```cpp
In file included from ../../src/neural/cache.h:30:
../../src/utils/cache.h:87:9: warning: reading variable 'capacity_' requires holding mutex 'mutex_' [-Wthread-safety-analysis]
    if (capacity_ == 0) return false;
        ^
```
I'll revert this for now.

PR #653, opened 2019-01-04
Blas update to 0.3.5

SunnyWar:
This is to update blas from the current 0.3.3 to 0.3.5.
Hopefully this is correct and builds and runs ok.


SunnyWar:
dammit

PR #655, opened 2019-01-04
Update OpenCL

SunnyWar:
Update nuget package from .12 to .77


borg323:
You missed a few places that need to be updated (search for 777.12). The build worked because the old package is still cached, but will fail next time appveyor refreshes the cache.

PR #659, opened 2019-01-05
Don't use Winograd for 1x1 conv.

Tilps:


Tilps:
We want to increase policy head channels to a number >32 and it triggered a crash on cudnn-32.  This fixed it.
Not sure if there is a better option.

cn4750:
I can confirm that this fixes the [crash](https://ghostbin.com/paste/hocjw) on my GTX 960 system.

PR #672, opened 2019-01-11
Use the GNU C Library versioning sysmacros.h to avoid compile time wa…

malcolmlewis:
…rning.

### DIFF
--- a/src/neural/loader.h
+++ b/src/neural/loader.h
@@ -29,6 +29,7 @@
 
 #include <string>
 #include <vector>
+#include <sys/sysmacros.h>
 
 #include "neural/network.h"
 #include "proto/net.pb.h"
--- a/src/neural/loader.h
+++ b/src/neural/loader.h
@@ -29,7 +29,9 @@
 
 #include <string>
 #include <vector>
+#ifdef __unix__
 #include <sys/sysmacros.h>
+#endif
 
 #include "neural/network.h"
 #include "proto/net.pb.h"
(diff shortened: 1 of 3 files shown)


malcolmlewis:
Hi
When building on openSUSE, various warnings are emitted, above commit fixes these.
For example;
[   95s] In file included from ../src/neural/loader.h:34,
[   95s]                  from ../src/neural/factory.h:33,
[   95s]                  from ../src/neural/opencl/network_opencl.cc:19:
[   95s] lc0_lib@sha/proto/net.pb.h:281:13: warning: In the GNU C Library, "minor" is defined
[   95s]  by <sys/sysmacros.h>.

However I cannot get rid of the ones for the meson compile time ones, for example 'lc0_lib@sha/proto/net.pb.h', perhaps this can be resolved at your end?

Prcuvu:
This will definitely fail on Windows. Please wrap the include with OS check macro.

malcolmlewis:
Hi
My bad, should be fixed now.

borg323:
I think this is the wrong fix, because the warning is spurious. My understanding is that we do not use the definitions from sysmacros.h. but the `major` and `minor` fields in the network and thus the generated header file by `protoc` is triggering the warnings. Try adding `#undef __SYSMACROS_DEPRECATED_INCLUSION` instead (see <https://sourceware.org/bugzilla/show_bug.cgi?format=multiple&id=19239>).

mooskagh:
No longer needed, we're not using libprotobuf anymore.

PR #679, opened 2019-01-14
[WIP] Change nps estimation formula to use exponential smoothing and more

fli:
This changes the nps calculation for smart pruning by:
- Using exponential smoothing (formula [here](https://pdfs.semanticscholar.org/882e/93570eae184ae737bf0344cb50a2925e353d.pdf) which deals with irregular time steps) instead of playouts / time. _This makes smart pruning more responsive to changes in nps._
- Turning off smart pruning if there is a trend upwards in nps at the beginning of the move. _Because we don't know how high it will go and don't want to underestimate._
- Using the previous move's average nps to seed the exponential smoothing of the next move. _Helps make next nps estimation more accurate, as long as the npss' are similar, and helps account for any initial trend in the nps._
- Doubles the initial nps estimate for the first move of the game (after the initial trend is finished). _My analysis shows that the first move typically underestimates, so this seems to counteract it well_.

Instead of smart pruning nodes tolerance and smart pruning time tolerance there is now:
- `nps_smoothing_factor = 2000`
- `nps_trend_smoothing_factor = 200`
- `average_move_nps_smoothing_factor = 2/3`

My analysis showed that these are decent values at least for Google Colab. Needs to be tested for other setups.

Code is still convoluted and needs commenting but it seems to work and can be tested for bugs and performance now (I haven't yet tested if it is stronger than master). `SmartPruningFactor` may need to be retuned.

The pictures that follow are an analysis of the first 5 moves of a 15+10 game I tested with.
- The ideal curve uses future information so is unobtainable.
- The y-axis is remaining playouts, x-axis is time (ms).
- Curves show behaviour with `SmartPruningFactor = 1`
- A good result is to match the ideal curve and not go too much under it (or you will underestimate and maybe instamove in a bad situation). It's better to be over than under.
- Parameters can be changed to make the graph smoother like master, but there is a trade off in responsiveness.
- Rest of the moves are [in this album](https://imgur.com/a/AviYe9p).

### DIFF
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -218,16 +218,10 @@ SearchLimits EngineController::PopulateSearchLimits(
   // Evenly split total time between all moves.
   float this_move_time = total_moves_time / movestogo;
 
-  // Only extend thinking time with slowmover if smart pruning can potentially
-  // reduce it.
-  constexpr int kSmartPruningToleranceMs = 200;
-  if (slowmover < 1.0 ||
-      this_move_time * slowmover > kSmartPruningToleranceMs) {
-    this_move_time *= slowmover;
-    // If time is planned to be overused because of slowmover, remove excess
-    // of that time from spared time.
-    time_spared_ms_ -= this_move_time * (slowmover - 1);
-  }
+  this_move_time *= slowmover;
+  // If time is planned to be overused because of slowmover, remove excess
+  // of that time from spared time.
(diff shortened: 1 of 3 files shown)


Naphthalin:
If I understood correctly, we by now improved the nps estimation, which is the purpose of this PR. If yes, this PR would be obsolete and can be closed.

PR #681, opened 2019-01-15
modify default benchmark setting to run for 10 seconds

ankan-ban:
- 30k nodes isn't enough to get representative speeds on fast hardware.
(getting lower than expected speeds here due to this: https://www.phoronix.com/scan.php?page=news_item&px=LCZero-NVIDIA-Benchmarks)
- set default kNodesId to a very high value (50m nodes), otherwise time setting has no effect if nodes is reached first.


jjoshua2:
50M is a bit tacky is all I can think of. maybe a -1 check like for time would be better...

mooskagh:
Looks good, but I'd increase time a but further to 20-30 seconds. Fine as it is, too!

PR #682, opened 2019-01-16
During search, score positions repeated once as draws

DanielUranga:
This is a draft to check if there is interest on this. The idea is to score positions that have happened at least once before in search as terminal node draws. If the position was repeated before it needs two reps to score as draw. This way the search becomes a directed acyclic graph.
More info: https://www.chessprogramming.org/Repetitions

The intention of this is to help with perpetuals and make Leela play a little less troll in general, but could also help with strength a bit.

Test results using a ID 35920:
```
command: lc0.exe selfplay --no-share-trees --games=1500 --visits=1000 --temperature=1 --tempdecay-moves=10 --player1.repetitions-before-draw-score=1
Result: tournamentstatus final win 196 93 lose 80 190 draw 474 467
```
```
command: lc0.exe selfplay --no-share-trees --games=1500 --visits=1200 --temperature=1 --tempdecay-moves=10 --player1.repetitions-before-draw-score=1
Result: tournamentstatus final win 196 73 lose 49 174 draw 505 503
```

If there is interest on having this, then the search history could be improved to have a hash for fast lookup of repeated positions to address this TODO: https://github.com/LeelaChessZero/lc0/blob/f66f69e1bd82509c2582afa3cefc71ea1352dec9/src/chess/position.cc#L108

### DIFF
--- a/src/chess/position.cc
+++ b/src/chess/position.cc
@@ -127,6 +127,15 @@ bool PositionHistory::DidRepeatSinceLastZeroingMove() const {
   return false;
 }
 
+bool PositionHistory::IsContainedInHistorySinceLastZeroingMove(
+    const ChessBoard& board) const {
+  for (int idx = positions_.size(); idx >= 0; idx--) {
+    if (board == positions_[idx].GetBoard()) return true;
+    if (positions_[idx].GetNoCaptureNoPawnPly() == 0) return false;
+  }
+  return false;
+}
+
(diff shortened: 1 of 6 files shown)


Naphthalin:
This is an older iteration of #1161 and still highly relevant. However it is unclear how this works with tree reuse, as a 2-fold repetition of a played move needs to be scored appropriately to avoid blind spots, see examples in the newer PR.
@DanielUranga can you check how the two PRs compare?

PR #683, opened 2019-01-17
no terminal multivisits

borg323:
This is an alternative to #629 (suggested by @mooskagh) that disables multivisits for terminal nodes.

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -857,8 +857,7 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend(
     // Either terminal or unexamined leaf node -- the end of this playout.
     if (!node->HasChildren()) {
       if (node->IsTerminal()) {
-        IncrementNInFlight(node, search_->root_node_, collision_limit - 1);
-        return NodeToProcess::TerminalHit(node, depth, collision_limit);
+        return NodeToProcess::TerminalHit(node, depth, 1);
       } else {
         return NodeToProcess::Extension(node, depth);
       }


jjoshua2:
I hope someone runs elo tests to verify no loss before merge. It seems 629
probably gains so we don't need to lose elo for bugfix

borg323:
Initial tests have been neutral to positive - a bit better than #629 but not conclusive given the error bounds. I would feel more comfortable with some additional testing.

zz4032:
TC 2.4s/game+0.01s/move (~150nodes/move)
```
   # PLAYER           :  RATING  ERROR  LOS(%)  POINTS   GAMES  DRAWS(%)
   1 lc0-dev          :     0.0   ----    92.8  2529.5    5000      61.6
   2 lc0-dev+PR683    :    -4.1    5.5     ---  2470.5    5000      61.6
```

src/mcts/search.cc
```diff
@@ -858,8 +858,7 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend(
     // Either terminal or unexamined leaf node -- the end of this playout.
     if (!node->HasChildren()) {
       if (node->IsTerminal()) {
-        IncrementNInFlight(node, search_->root_node_, collision_limit - 1);
-        return NodeToProcess::TerminalHit(node, depth, collision_limit);
+        return NodeToProcess::TerminalHit(node, depth, 1);
```
MelleKoning: Can you explain a bit? If I understand correctly this change affects the initialization of multivisit to 1, which in turn affects the calculation within FinalizeScoreUpdate to properly take Terminal evaluations to 1, correct? 
Would this change only have an affect when minibatchsize > 1 or is that not related at all?

Issue #666, opened 2019-01-07
Make Lc0 a valuable sparring partner for human players.


Harumpel:
Lc0 extracts knowledge from games of self-play. After watching some games, this knowledge seems slightly different to classic human chess knowledge at times. Thus, Lc0 chess knowledge would be of great value for humans training chess. At least a great alternative to alpha-beta engines. 

Humans need a motivating way to play against Lc0. Some will be OK with playing sparring games against a full-strength-Lc0 engine. But some would like to play against a weakened version. The possibilities to weaken Lc0 would be as follows:

1. adjust temperature / temp value cutoff to randomize moves (UCI settings)
2. limit max nodes computed for one move, instant reply
3. slow down Lc0 engine execution speed so that some very low number of nodes per move is computed
4. use some older networks with diffused knowledge, but still full speed computation, high nodes count.

Please share your ideas about this, I will add and edit this issue.

oscardssmith:
a 10b net at 1 or 2 nodes is a pretty good sparring partner for me. There also has been a patch to make leela aim for even positions which makes her blunder when up.

borg323:
I have made a branch adding an extra option, `--move-delay-ratio` to specify a portion of the move time that lc0 will wait before starting the search. Setting this to 1 makes lc0 process only a single node, and 0.5 wastes half the search time. Test cuda binary: <https://ci.appveyor.com/api/buildjobs/hm1a1q0t6n9mu0eh/artifacts/build%2Flc0.exe>

Suggestions for better option names are welcome.

Naphthalin:
This can probably be closed once #685 is merged.

mooskagh:
Closing as #685 is there, and otherwise it's more like a vague idea rather than concrete feature.

PR #685, opened 2019-01-18
Add option to limit search nps

borg323:
This is an attempt to address #666. It adds the `--move-delay-ratio` option to stall for a portion of the move time before starting the search. Suggestions for a better user interface are welcome.


borg323:
The latest revision uses the `--strength-ratio=` option, where a value *sr* is used to calculate the move delay ratio *mdr* using the formula *mdr=1-sr^2* that gives a somewhat more linear strength reduction. Setting `--strength-ratio=0` gives the minimum strength (equivalent to doing `go nodes 1`) while `--strength-ratio=1` (the default) plays at full strength.

mooskagh:
It it expected to work with `go infinite`? For me it doesn't slow anything down in this mode.

borg323:
No, this only works when there is a time limit. This was the best I could come up with.

borg323:
I merged this into master in the most direct way, but I have the feeling there is a better way.

PS. The way this works is similar to what Carlsen is doing here: <https://www.reddit.com/r/chess/comments/f1sks0/carlson_takes_20_seconds_before_playing_his_first/?utm_source=share&utm_medium=web2x>

borg323:
I made a new version where the delay is done in `SearchWorker::ExecuteOneIteration()`, relative to the time the current iteration took. This way the strength reduction also works for infinite searches and when pondering.

jgkwok:
From a feature perspective, I think a nps limit would be clearer for users. Perhaps using some "max" terminology, --max-nps or something more appropriate. The reason is this would allow for more casual users to have a consistent way of handicapping a Leela's play. Consistency would allow regular people to experiment with settings and share them with each other and bring good publicity to the project.

From an implementation point of view, I'm not familiar with the code structure, but would it be possible to evaluate a batch of positions and then feed them to the search code at the expected nps rate? 

Or an alternate idea maybe closer to what you have implemented, would it be possible to have some sort of "nps throttle" where the node evaluation happens in the regular fashion, but once the actual nps exceeds the specified throttle nps, you sleep for an appropriate period based on the batch size or time management being used? Similar to taking the foot off the gas when exceeding the speed limit, then adding gas when falling below. 

(I posted a similar message on discord, but thought this might be the more appropriate place. Let me know if this isn't the right way to contribute, never contributed to a project before)

Naphthalin:
I vaguely remember this to be mentioned once in a while. If one of our goals with Lc0 is to provide a training partner for human players, merging this should be considered.

borg323:
I reworked it into a configurable nps limit.

src/mcts/search.cc
```diff
@@ -72,15 +72,9 @@ Search::Search(const NodeTree& tree, Network* network,
...
-  float strength_ratio =
-      options.Get<float>(SearchParams::kStrengthRatioId);
-  if (strength_ratio != 1.0f) {
-    // Factor to multiply execution time to get the delay to add, so that the
-    // execution time is strength_ratio^2 of the total. This was found to
-    // approximate linear strength scaling from strength_ratio 0 to 1.
-    delay_factor_ = 1.0f / (strength_ratio * strength_ratio +
-                            std::numeric_limits<float>::epsilon()) -
-                    1.0f;
+  float nps_limit = options.Get<float>(SearchParams::kNpsLimitId);
```
mooskagh, line 75: I think it's better to do in params.cc (and it's fine for GetNpsLimit() to return `optional<float>` if that's needed).

params.cc is where all that happens anyway (and its purpose is actually to cache the parameters so that it won't have a performance impact), and search.cc is already a bit large.

borg323: The optional bit was a remnant from an earlier iteration.

jgkwok:
Would this handle fractional nps values? E.g., something like 0.5 nps? For weaker players, this may add to the entertainment value of the feature.

src/mcts/params.cc
```diff
@@ -257,11 +257,11 @@ const OptionId SearchParams::kDrawScoreWhiteId{
...
-const OptionId SearchParams::kStrengthRatioId{
-    "strength-ratio", "StrengthRatio",
-    "An option to reduce engine strength. Works by delaying the search for a "
-    "portion of the allocated time. Set to zero for minimum and one for "
-    "maximum strength, or anything in-between."};
+const OptionId SearchParams::kNpsLimitId{
+    "nps-limit", "NodesPerSecondLimit",
+    "An option to specify an upper limit to the nodes per second searched. The "
+    "accuracy depends on the minibatch size used, increasing for lower sizes, "
+    "and on the lenght of the search. Zero to disable."};
```
lealgo: Typo: "length"
:wink:

borg323:
@jgkwok yes, it can go down to fractional nps values (but 0 disables the limit). However for this to be accurate you need considerable time per move.

PR #686, opened 2019-01-18
simplify immediate-time-use mechanism

borg323:
This is an effort to simplify the immediate-time-use mechanism by just keeping track of the time to be used immediately. Not tested (but at least it builds).


borg323:
No option changes. This should be a neutral change with default values, but with lower immediate time use, the remaining saved time is not accumulated separately. In contrast, a higher slowmover value would reduce the immediate time available which is not implemented here (but maybe it should).

borg323:
Superseded by #1195

PR #690, opened 2019-01-21
Check Syzygy tablebase file sizes for corruption

cn4750:
Check that Syzygy tablebase file sizes modulo 64 are 16 as a simple check for corruption.
Inspired by a [Stockfish PR](https://github.com/official-stockfish/Stockfish/commit/bb843a00c1ef381162dc9c2491b5436b6cf5563f) by vondele.

### DIFF
--- a/src/syzygy/syzygy.cc
+++ b/src/syzygy/syzygy.cc
@@ -44,6 +44,7 @@
 
 #include "syzygy/syzygy.h"
 
+#include "utils/exception.h"
 #include "utils/logging.h"
 #include "utils/mutex.h"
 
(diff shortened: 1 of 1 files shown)


borg323:
Is exiting the correct response to a corrupt syzygy file? Just printing the warning and returning nullptr seems better to me.

cn4750:
> 
> 
> Is exiting the correct response to a corrupt syzygy file? Just printing the warning and returning nullptr seems better to me.

Stockfish exits. If a mis-sized tablebase file is detected, using a corrupted file for analysis seems dangerous.

mooskagh:
Probably we should report error and stop, but `exit(1)` is still poor way to do that.
It's better to throw Exception() (from utils/exception.h) so that calling code could handle that better.

borg323:
I was not suggesting using a corrupt file, just ignoring it, which `return nullptr;` should do at that point.

cn4750:
I am afraid that if we detect any corruption of tablebase files that we cannot trust any of the other files. That is why we should exit with an error; we are in an exceptionally bad state. I will look at converting the exits to exceptions shortly.

mooskagh:
It's better to throw this https://github.com/LeelaChessZero/lc0/blob/master/src/utils/exception.h exception rather than plain std::runtime_error.
Currently there's no difference rather than outputting error to log file, but maybe one day there will be a difference.

Tilps:
lgtm

PR #691, opened 2019-01-22
fix define

SunnyWar:
use proper syntax

### DIFF
--- a/src/chess/board.cc
+++ b/src/chess/board.cc
@@ -33,7 +33,7 @@
 #include <sstream>
 #include "utils/exception.h"
 
-#if not defined(NO_PEXT)
+#ifndef NO_PEXT
 // Include header for pext instruction.
 #include <immintrin.h>
 #endif


SunnyWar:
This looks like it was a copy-paste error from meson.build.

ddobbelaere:
Good point, looks like the header was never included but it worked with GCC anyway as `_pext_u64` is builtin according to https://gcc.gnu.org/onlinedocs/gcc/x86-Built-in-Functions.html

It would be good if someone could verify that PEXT builds for Windows work now though (there are not built on appveyor).

ddobbelaere:
Looks like `#if not defined(NO_PEXT)` is also valid syntax. I can confirm on Linux 64-bit with both `gcc` and `clang` that it has the same effect as `#ifndef NO_PEXT`. Probably because of https://en.cppreference.com/w/cpp/keyword/not

mooskagh:
Yeah `#if defined()` is usually preferred instead of `#ifdef`, as it allows combining multiple ones in expressions, `#elif`s etc.
See for example https://github.com/LeelaChessZero/lc0/blob/9a6c0a9aec36ad01f03cfc761b68e4c915734981/src/chess/bitboard.h#L107

PR #692, opened 2019-01-23
Fix Cpuct formula to match AlphaZero paper

fli:
I believe the Cpuct factor is defaulted to 2 because our Q is from -1 to 1 and AlphaZero's is 0 to 1. That means that our entire formula for calculating `U(s,a)` needs to be scaled by 2, not just the Cpuct growth part of the formula.

This also means that our Cpuct values are off by a factor of 2. ~~So really we are training the network with Cpuct = 0.6 not Cpuct=1.2.~~ I've adjusted the default self-play Cpuct value to match master's behaviour, ~~but to match AlphaZero it will need to be adjusted back.~~

You'll also notice that the default match Cpuct value is now closer to AlphaZero's chosen Cpuct value (maybe a coincidence).

I've also removed the Cpuct factor parameter. The AlphaZero paper says _C(s) is essentially constant during the fast training games_. They would've said something like _"C(s) is set to be constant during the fast training games"_ if they meant that C(s) is actually set to a constant value. With their CpuctBase value set to 16592, C(s) is essentially constant which matches what they say in the paper and also their pseudocode.

### DIFF
--- a/src/mcts/params.cc
+++ b/src/mcts/params.cc
@@ -58,8 +58,6 @@ const OptionId SearchParams::kCpuctBaseId{
     "cpuct-base", "CPuctBase",
     "cpuct_base constant from \"UCT search\" algorithm. Lower value means "
     "higher growth of Cpuct as number of node visits grows."};
-const OptionId SearchParams::kCpuctFactorId{
-    "cpuct-factor", "CPuctFactor", "Multiplier for the cpuct growth formula."};
 const OptionId SearchParams::kTemperatureId{
     "temperature", "Temperature",
     "Tau value from softmax formula for the first move. If equal to 0, the "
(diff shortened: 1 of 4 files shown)


ddobbelaere:
So basically this PR amounts to the same formula, but makes it less flexible (by removing `kCpuctFactor`), or am I missing something crucial here?

I think it will be confusing after all this time to see the initial cpuct halved all of sudden, but that's just my opinion.

fli:
Yeah it could definitely be confusing.

Depends what you value more:

1. Confusion of why cpuct differs from AlphaZero vs confusion of Cpuct values being halved suddenly.

2. Simplifying Cpuct formula by removing a parameter and matching AlphaZero vs having flexibility to tweak it in this way (this is really our invention, AlphaZero doesn’t factor the growth part of the formula).

We could make a lot of things more flexible by adding parameters, but I think there’s value in simplifying (see how many parameters there are for time management).

IMO, there’s no reason to have cpuctfactor = 0 off for training. The effect is marginal, doesn’t match AlphaZero and I can’t think of a theoretical reason for it to be different than from match play (e.g. if you instead played 800000 node training games instead of 800 node training games do you really want the exploration part off?).

Tilps:
I don't think any win here out weighs the requirement to at some point stop all training while we force everyone to upgrade to the new version before restarting the server with the new puct values.

fli:
@ASilver I don't think so. AlphaZero uses the factor in self-play and they use fixed nodes. But maybe there's another reason it is off.

@Tilps I agree. I imagine if this were to be merged it would be as part of some other change(s) that did require everyone to upgrade the client and restart training (maybe way down the line).

Naphthalin:
As we now agree that cpuct and cpuct-scaling is very important, this PR seems to be the opposite of that. I therefore recommend closing the PR.

Issue #651, opened 2019-01-03
Lc0 responds with `bestmove` if it reaches limit during ponder.


mooskagh:
Reported on discord.
It seems that during ponder, when Lc0 decides to suspend progress, it sends `bestmove` and confuses Arena.

Video where this issue is reproduced:
https://cdn.discordapp.com/attachments/427066771627966466/530507183432138753/arena-lc0-bug-002.mp4
https://cdn.discordapp.com/attachments/427066771627966466/530502146848849930/arena-lc0-bug-WARNING-search-limit.mp4

Arena log:
https://cdn.discordapp.com/attachments/427066771627966466/530509246777917453/arena-protocol.txt

Lc0 log:
https://cdn.discordapp.com/attachments/427066771627966466/530513736964767744/lc0log.txt

mooskagh:
Arena sends:
```
go ponder movetime 15000
```
Those 15000 mean that that timer should start on `ponderhit`. But we start it immediately.
When the timer runs out before `ponderhit`, it hangs.
If timer runs out after `ponderhit`, it's still wrong, as timer should not have counted time before `ponderhit`.

borg323:
There are two issues here:
1. `go ponder` with any additional constraint (except `wtime` and `btime`) honors the constraint. This is correct for `searchmoves` and certainly wrong for `movetime` which should only be honored after `ponderhit` (as mentioned above by @mooskagh). Whether it should be honored in the other cases like `depth` or `nodes` is arguable, although the nodes limit is also used internally to constrain memory usage. 
2. When a limit is used while pondering, lc0 stops responding to uci commands.

The `Bestmove was from pondersearch, ignoring` message from Arena is just indicating that the required `bestmove` after a `stop` was received, this is not a problem.

borg323:
Point 2 above is probably due to <https://github.com/LeelaChessZero/lc0/pull/539> (from reading the code).

PR #693, opened 2019-01-23
better fix for issue 651

borg323:
The previous fix for #651 (#658) made the problem fixed by #539 reappear. The issue is that `stop_` is set when reaching a limit so a `ponderhit` after that (with the original code) will not set `bestmove_is_sent_` that is needed for the search to exit. #658 was also checking if `bestmove_is_sent_` is not set but this allows a race when `Abort` is called too fast after `Stop` and the `bestmove` reply is not sent, thus confusing uci hosts that expect one. This patch adds a check that `ok_to_respond_bestmove_` (that is set by `Stop`) is not set, thus avoiding this race.

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -660,7 +660,8 @@ void Search::Stop() {
 
 void Search::Abort() {
   Mutex::Lock lock(counters_mutex_);
-  if (!stop_.load(std::memory_order_acquire) || !bestmove_is_sent_) {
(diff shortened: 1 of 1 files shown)


jizhongling:
This is really needed since I am using XBoard. I noticed that starting from v0.20.1-rc2 pondering not work and I just removed || !bestmove_is_sent_, but maybe it does not work for Arena.

PR #696, opened 2019-01-26
[WIP] Basic server mode hack.

Tilps:
Currently replaces uci mode, and makes no attempt to share backend or caches.

Requires boost to compile since I used asio for the networking... probably pretty straightforward on linux, but windows maybe not so much.

It listens on port 45312.


Tilps:
Now this is more interesting - if network configuration or tablebase configuration is the same between 2 different connections, they will share them.

Tilps:
The purpose of this is so that someone can run cutechess with net cat as the client engine - run parallel fixed node games in cutechess, and have them share a single gpu.
Using grpc means netcat has to be replaced with a fat client.

Tilps:
I've still not worked out a way around the 1.66 minimum version.

Tilps:
I can't see any way to get below a boost 1.66 minimum without having to write my own streambuf type wrapper code for socket.

Tilps:
Merged latest master so it might work with T50 networks and networks with negative gammas.

Naphthalin:
This seems to require using the `boost` library, which per an old consensus won't be used. If this still stands, this PR should probably be closed (?)

borg323:
I don't think using boost is a hindrance for the intended audience. Maybe adding the `not for merge` label is a good idea.

Tilps:
not for merge is definitely applicable to this - although I'm not sure when I'll next get a chance to merge master.

PR #698, opened 2019-01-26
Informative Tournament Stats - Part 1 of PR487

Videodr0me:
Informative Tournament Stats improves display of selfplay stats::
before:
```
tournamentstatus win 3 3 lose 5 3 draw 8 10
```
after - obviously not same tournament as above ;-):
```
tournamentstatus P1: +815 -825 =1213 Win: 49.82% Elo: -1.22 LOS: 40.25% P1-W: +473 -354 =600 P1-B: +342 -471 =613
```

### DIFF
--- a/src/selfplay/loop.cc
+++ b/src/selfplay/loop.cc
@@ -136,7 +136,7 @@ void SelfPlayLoop::SendTournament(const TournamentInfo& info) {
   if ((winp1 + losep1 + draws) > 0) {
     percentage =
         (static_cast<float>(draws) / 2 + winp1) / (winp1 + losep1 + draws);
-    }
+  }
   // Calculate elo and los if percentage strictly between 0 and 1 (avoids divide
   // by 0 or overflow).
   if ((percentage < 1) && (percentage > 0))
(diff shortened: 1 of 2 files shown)


Tilps:
Just to save anyone who is wondering - client doesn't use this output, so its safe to change I think.

src/selfplay/loop.cc
```diff
@@ -123,13 +123,56 @@ void SelfPlayLoop::CmdSetOption(const std::string& name,
 }
 
 void SelfPlayLoop::SendTournament(const TournamentInfo& info) {
+  const int winp1 = info.results[0][0] + info.results[0][1];
+  const int loosep1 = info.results[2][0] + info.results[2][1];
```
mooskagh: I think it's `lose` not `loose` in English.

src/selfplay/loop.cc
```diff
@@ -123,13 +123,56 @@ void SelfPlayLoop::CmdSetOption(const std::string& name,
...
+  const int draws = info.results[1][0] + info.results[1][1];
+
+  // Initialize variables
+  float percentage = -1;
+  float elo = 10000;
+  float los = 10000;
+
+  // Only caculate percentage if any games at all (avoid divide by 0).
+  if ((winp1 + loosep1 + draws) > 0)
+    percentage =
```
mooskagh: When entire `if` (or `for`) doesn't fit one line, `{` `}` around body is needed.
( https://google.github.io/styleguide/cppguide.html#Conditionals )

(also other cases like that below)

(but if the body itself is still one line (even if separate), it's fine to omit braces, according to the latest edition of style guide.

src/selfplay/loop.cc
```diff
@@ -123,13 +123,56 @@ void SelfPlayLoop::CmdSetOption(const std::string& name,
...
+
+  // Calculate elo and los if percentage strictly between 0 and 1 (avoids divide
+  // by 0 or overflow).
+  if ((percentage < 1) && (percentage > 0))
+    elo = -400 * log(1 / percentage - 1) / log(10);
+  if ((winp1 + loosep1) > 0)
+    los = .5 +
+          .5 * std::erf((winp1 - loosep1) / std::sqrt(2.0 * (winp1 + loosep1)));
+
   std::string res = "tournamentstatus";
```
mooskagh: I think it's cleaner to create `std::ostringstream res` instead of res, and output to it, and then just return `res.str()` from the function.

src/selfplay/loop.cc
Videodr0me: done

src/selfplay/loop.cc
Videodr0me: its just like it was before. If you really insist i can change it though.

mooskagh: Alternatively, you can have a helper function:
```cpp
std::string FormatFixed(double num, int width, int precision);
```
and call it whereever needed.
But current way of mixing two techniques is kind of untidy.

Videodr0me: done

PR #699, opened 2019-01-26
Change U calculation to match deepmind pseudocode

Tilps:
While investigating why fpu mixing was having a stronger effect for leela lite than for lc0 I noticed that leela lite's formula for U was slightly different to the formula in lc0.

I'm still of the opinion that the lc0 formula is an accurate rendering of the formulas presented in papers, but I went back to the deepmind pseudocode published with the alphazero paper and found that they too had the same difference in their calculation of U.

I haven't got any positive tests in favour of this change as of yet, so it doesn't seem super relevant - but maybe it'll help someone testing fpu mixing to see if it has an amplifying effect together.

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -207,8 +207,7 @@ std::vector<std::string> Search::GetVerboseStats(Node* node,
                                                  bool is_black_to_move) const {
   const float fpu = GetFpu(params_, node, node == root_node_);
   const float cpuct = ComputeCpuct(params_, node->GetN());
-  const float U_coeff =
-      cpuct * std::sqrt(std::max(node->GetChildrenVisits(), 1u));
+  const float U_coeff = cpuct * std::sqrt(node->GetN());
 
   std::vector<EdgeAndNode> edges;
   for (const auto& edge : node->Edges()) edges.push_back(edge);
@@ -880,8 +879,7 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend(
     // If we fall through, then n_in_flight_ has been incremented but this
     // playout remains incomplete; we must go deeper.
     const float cpuct = ComputeCpuct(params_, node->GetN());
-    float puct_mult =
-        cpuct * std::sqrt(std::max(node->GetChildrenVisits(), 1u));
+    float puct_mult = cpuct * std::sqrt(node->GetN());
     float best = std::numeric_limits<float>::lowest();
     float second_best = std::numeric_limits<float>::lowest();
     int possible_moves = 0;
@@ -1099,7 +1097,7 @@ int SearchWorker::PrefetchIntoCache(Node* node, int budget) {
   typedef std::pair<float, EdgeAndNode> ScoredEdge;
   std::vector<ScoredEdge> scores;
   const float cpuct = ComputeCpuct(params_, node->GetN());
-  float puct_mult = cpuct * std::sqrt(std::max(node->GetChildrenVisits(), 1u));
+  float puct_mult = cpuct * std::sqrt(node->GetN());
   const float fpu = GetFpu(params_, node, node == search_->root_node_);
   for (auto edge : node->Edges()) {
     if (edge.GetP() == 0.0f) continue;


mooskagh:
If we do this change, we can also remove GetChildrenVisits.
Actually GetChildrenVisits did show up as non-zero in profiler (because of condition probably), so having it removed will also improve nps by a (very) tiny bit.

Tilps:
GetChildrenVisits is still used in the logic for creating training data as its the number of visits to normalize by.

Tilps:
(I still have no improvement to go with this - but I have decided that logically this change probably should go with a slight decrease in cpuct - so I am going to try testing whether it responds better to a lower cpuct than default.)

Tilps:
There was an interim result posted in off-topic which suggests that this change has a positive delta when baseline and comparison use fpu absolute rather than our default fpu strategy.  I might give that a try after my current tests.

Tilps:
Preliminary results suggest baseline is reacting more positive to cpuct 2.7 compared to 3.0 - compared to the alternative U code.  Still running.

Tilps:
Based on my ongoing test I think that what is happening is that this new U value changes the ideal tune - and that at the nodes I am testing at, our default is clearly not optimal for either.  Without deriving ideal tune for both before and after and then comparing them, I don't think I can come to a clear conclusion whether this change is good at all.

Ttl:
Some more test games with self-play settings:

```
Score of lc0_pr699 vs lc0_master: 200 - 181 - 619  [0.509] 1000
Elo difference: 6.60 +/- 13.28, LOS: 83.48 %, DrawRatio: 61.9 %
```

Network 40626, Parameters: `--threads=1 --fpu-strategy=absolute --smart-pruning-factor=0 --minibatch-size=32 -cache-history-length=7 --policy-softmax-temp=1.0 --cpuct=2.5`.

src/mcts/search.cc
```diff
@@ -207,8 +207,7 @@ std::vector<std::string> Search::GetVerboseStats(Node* node,
                                                  bool is_black_to_move) const {
   const float fpu = GetFpu(params_, node, node == root_node_);
   const float cpuct = ComputeCpuct(params_, node->GetN());
-  const float U_coeff =
-      cpuct * std::sqrt(std::max(node->GetChildrenVisits(), 1u));
```
mooskagh: I think now we can delete GetChildrenVisits() completely.

Tilps: Its still used in GetV4TrainingData.

mooskagh: It can be `GetN() - 1` there probably. :)

Naphthalin:
In [node.h](https://github.com/LeelaChessZero/lc0/blob/master/src/mcts/node.h#L360) the comment should be changed as well for this PR.

Tilps:
Looks to me that the comment is already accurate for after this change and misleading for before? :P

Naphthalin:
Oh, good catch :D pretty sure though that wasn't intended and should anyway be differentiated if we introduce a flag there.

src/mcts/params.cc
```diff
@@ -60,6 +60,10 @@ const OptionId SearchParams::kCpuctBaseId{
     "higher growth of Cpuct as number of node visits grows."};
 const OptionId SearchParams::kCpuctFactorId{
     "cpuct-factor", "CPuctFactor", "Multiplier for the cpuct growth formula."};
+const OptionId SearchParams::kNewUEnabledId{
+    "new-u-enabled", "NewUEnabled",
```
mooskagh: Neither flag name, nor description (what corresponds to true and what to false?) are completely clear.

Maybe `--parent-based-u`? `--include-initial-visit-to-u`? `--u-parent-visits`?

Tilps:
After reading 913 I expect to rollback the last patch which added the command line flag.

Naphthalin:
If I understand this correctly, this behavior was never changed, so the PR might still be relevant.

Tilps:
yeah its just so low priority that I've never come back to it.

PR #703, opened 2019-01-26
Add trade penalty

nblaxall:
Add trade penalty to lc0 20.1


Naphthalin:
As trade penalty is non-zero and the possible advantages of this are available through a combination of MLH and draw score which are both merged in v0.25, I recommend closing this PR.

borg323:
Moreover there seems to have been a bad merge at some point, there are lots of code unrelated to trade penalty here.

PR #707, opened 2019-01-27
Add FPU visits parameter

Ttl:
This parameter controls the number of visits that initial Q should have in later Q updates. I clopped the best value at very quick time controls to be around 0.6.

At longer time controls and other values being default it seems to be improvement:

```
Score of lc0_fpu_visits_35258 vs lc0_master_35258: 45 - 25 - 330  [0.525] 400
Elo difference: 17.39 +/- 14.16, LOS: 99.16 %, DrawRatio: 82.5 %

Network 35258, TC: 10s+0.5s, GPU GTX 1080 Ti, 5 man TBs and all other parameters defaults.
```

### DIFF
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -174,7 +174,7 @@ class Node {
   // * Q (weighted average of all V in a subtree)
   // * N (+=1)
   // * N-in-flight (-=1)
-  void FinalizeScoreUpdate(float v, int multivisit);
+  void FinalizeScoreUpdate(float v, int multivisit, float fpu_visits);
   // When search decides to treat one visit as several (in case of collisions
   // or visiting terminal nodes several times), it amplifies the visit by
   // incrementing n_in_flight.
@@ -264,7 +264,8 @@ class Node {
   // subtree. For terminal nodes, eval is stored. This is from the perspective
   // of the player who "just" moved to reach this position, rather than from the
   // perspective of the player-to-move for the position.
-  float q_ = 0.0f;
+  // Initialized to loss (-1).
+  float q_ = -1.0f;
   // Sum of policy priors which have had at least one playout.
   float visited_policy_ = 0.0f;
   // How many completed visits this node had.
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -238,9 +238,9 @@ void Node::CancelScoreUpdate(int multivisit) {
   best_child_cached_ = nullptr;
 }
 
-void Node::FinalizeScoreUpdate(float v, int multivisit) {
+void Node::FinalizeScoreUpdate(float v, int multivisit, float fpu_visits) {
   // Recompute Q.
-  q_ += multivisit * (v - q_) / (n_ + multivisit);
+  q_ += multivisit * (v - q_) / (fpu_visits + n_ + multivisit);
   // If first visit, update parent's sum of policies visited at least once.
   if (n_ == 0 && parent_ != nullptr) {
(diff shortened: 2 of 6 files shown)


src/mcts/node.h
```diff
@@ -264,7 +264,8 @@ class Node {
   // subtree. For terminal nodes, eval is stored. This is from the perspective
   // of the player who "just" moved to reach this position, rather than from the
   // perspective of the player-to-move for the position.
-  float q_ = 0.0f;
+  // Initialized to loss (-1).
+  float q_ = -1.0f;
```
killerducky: I think the code assumes we're running in FpuStrategy = absolute, with FpuValue = -1.0. Let's use GetFpuValue here?

For FpuStrategy = reduction I don't know what to do, will capturing the GetFpuValue at the time the node is constructed work? Or just punt on that when this is enabled? I think most testers are still using this mode right?

Ttl: I clopped and tested this with FPU reduction since it's stronger in play. There isn't really any reason why this should use the same FPU method for initial Q value. FPU decides the 0 visit value and this only affects the value after node has visits. Option name could probably modified to make this more clear.

I did test a version where I modified `GetQ(fpu)` to set the `q_` to fpu if node doesn't have any visits. That version was weaker in testing: `Score of lc0_fpu_visits_fpu_init_35258 vs lc0_fpu_visits_35258: 26 - 42 - 332  [0.480] 400
Elo difference: -13.90 +/- 13.98, LOS: 2.62 %, DrawRatio: 83.0 %`.

killerducky: Ok I was thinking of this as treating the FPU value as a sort of Bayesian prior on q_, which gets mixed into q_. Instead they are considered different, I don't have a good idea for a name right now but maybe some comment here in the code that q_ = -1.0f is not the same as the FPU?

Eddh: It could be named something like "value_initialization"
I was thinking also that fpu_visit is maybe not the right term for the divisor. I guess something like value_divisor_init instead.

Tilps: value_pessimism?

Ttl:
Some more tests with the latest commit. Hopefully no functional changes. It uses the pessimistic Q value only for picking the positions for the batch. The reported score is not affected anymore.

```
Score of lc0_value_pessimism_06 vs lc0_master: 92 - 74 - 634  [0.511] 800
Elo difference: 7.82 +/- 10.95, LOS: 91.88 %, DrawRatio: 79.3 %

800 nodes, MinibatchSize=32, Slowmover=0 
```

```
Score of lc0_value_pessimism_06 vs lc0_master: 68 - 60 - 272  [0.510] 400
Elo difference: 6.95 +/- 19.25, LOS: 76.02 %, DrawRatio: 68.0 %

800 nodes, MinibatchSize=32 
```

```
Score of lc0_value_pessimism_06 vs lc0_master: 92 - 74 - 634  [0.511] 800
Elo difference: 7.82 +/- 10.95, LOS: 91.88 %, DrawRatio: 79.3 %

1+0.5s, GTX 1080 Ti
```

The later tests are all positive, but all are inside the error bounds. Probably still needs more tests to make sure that this gains Elo.

Naphthalin:
This is an interesting approach (and somehow similar in general to the initial AlphaGo paper, where the NN evaluation was treated as 100 MC rollouts). @Ttl do you still think this is a sensible approach, and if yes update the PR and test with recent nets?

PR #711, opened 2019-01-30
 Print cuda and gpu info, warn if mismatches are noticed

borg323:
This will help spot some issues quickly.


ankan-ban:
Looks good. Maybe also add a warning if the GPU supports fp16, but user is running it with default (cudnn fp32) backend?

PR #719, opened 2019-02-07
[WIP] Policy scaling

jjoshua2:
Test to see if this improves tactics at high nodes.


src/mcts/search.cc
```diff
@@ -1190,7 +1190,7 @@ void SearchWorker::FetchSingleNodeResult(NodeToProcess* node_to_process,
       // Flush denormals to zero.
       p = p < 1.17549435E-38
               ? 0.0
-              : FastPow2(FastLog2(p) / params_.GetPolicySoftmaxTemp());
+              : FastPow2(FastLog2(p) / (params_.GetPolicySoftmaxTemp() + FastLog2(node->GetN()+1) ));
```
Tilps: I don't think this does what you think - this is only called once for a fresh node and then never again - so its GetN should always be 0.

jjoshua2:
Yah I found that out. Is there an easy fix?

Videodr0me:
Something like this is a good idea, as Deep Minds scaling c formula has some important implications:

Under some conditions the new DM scaling prevents convergence to the minimax result. Especially if priors are extreme there are situations where the visit distribution converges to the policy distribution regardless of Q. Leading to paradoxical situations where a losing move is preferred over a non-losing move with large enough visits.

Now there are some possible explanations for this: 
(1) DM is fully aware of this, and thinks that through the learning feedback loop, this reduces extreme polices, which they feel is a good thing.
(2) DM just found that this  just improved elo slightly across all games and DM did not further refine this.

I am a little sceptical of (1) and believe that this scaling can be refined. And policy flattening with increasing N is a very viable altermative, so i think this PR is a good idea.

Implementation wise: You really need to move the policy flattening code to picknodetoextend, as you want to change policy with increasing N, and not just initially when policy is fetched (where N is zero anyways). I think there are formulas that can flatten policy without first iterating over all policy values, so using policy softmax might not be the most efficient.

PR #725, opened 2019-02-10
Add a uniform output option for random backend to support a0 type seed data.

Tilps:
As per the a0 pseudocode - they seed using a uniform backend that returns draw for every position and equal weighting to all policies.


Tilps:
I think I would like to use this to generate seed data for T50.

mooskagh:
Strictly speaking, with uniform it's not random anymore. :)

PR #731, opened 2019-02-13
misc (non-functional) changes to cudnn backend

ankan-ban:
- Replace all cudaMemcpyAsync used for loading weights with cudaMemcpy as  source (in CPU memory) could be deleted before the async version of the function actually does the copy. Thanks @Tilps for pointing this out.
- Add comments explaining what the policy map layer does and how the layout conversion from CHW to HWC works.
- replace new/delete with std::vector for the allocation of scratch space used by layout conversion.
- minor naming/style change.


src/neural/cuda/layers.cc
```diff
@@ -300,24 +300,23 @@ void SELayer<float>::LoadWeights(float* w1, float* b1, float* w2, float* b2,
...
   // Weight for the second FC layer.
-  ReportCUDAErrors(
-      cudaMemcpyAsync(w2_, w2, weight_size2, cudaMemcpyHostToDevice));
+  ReportCUDAErrors(cudaMemcpy(w2_, w2, weight_size2, cudaMemcpyHostToDevice));
 
   // Bias for the first FC layer.
-  ReportCUDAErrors(cudaMemcpyAsync(b1_, b1, numFc1Out_ * sizeof(float),
+  ReportCUDAErrors(
+      cudaMemcpy(b1_, b1, numFc1Out_ * sizeof(float),
                                    cudaMemcpyHostToDevice));
```
Ttl: Some of these lines could use clang-format.

Ttl:
I think that `CudaMemcpyAsync` needs the CPU buffer to be page-locked memory for it to be able to copy asynchronously. Since none of these buffers are page-locked, copies are actually synchronous and there wasn't risk of them failing. Using `cudaMemcpy` instead is still a good change to make it explicit.

ankan-ban:
Agree, that was likely the reason `CudaMemcpyAsync` was working. The documentation isn't very clear though so I agree that we should change to `CudaMemcpy` to be safe in case some later version of cuda stops supporting it.

src/neural/cuda/layers.cc
```diff
@@ -591,13 +636,12 @@ void PolicyMapLayer<DataType>::LoadWeights(const short* cpuWeight,
         else
           convertedWeights[hw * Cin + c] = -1;
       }
-    ReportCUDAErrors(cudaMemcpyAsync(weights_, convertedWeights,
-                                     usedSize * sizeof(short),
-                                     cudaMemcpyHostToDevice));
-    delete[] convertedWeights;
+    ReportCUDAErrors(cudaMemcpy(weights_, &convertedWeights[0],
```
Tilps: convertedWeights.data() ?

Issue #728, opened 2019-02-11
Possible improvement to centipawn output?


yuzisee:
In TCEC Superfinal 14, Leela's cenitpawn output seemed a little wonky to most observers (mostly that it's too large)

The formula proposed in https://www.chessprogramming.org/Pawn_Advantage,_Win_Percentage,_and_Elo could reduce the absolute magnitude of centipawns and might make for improved readability with traditional engine ranges / viewer expected ranges.

Right now Leela uses 
  `uci_info.score = 290.680623072 * tan(1.548090806 * edge.GetQ(0));`
from https://github.com/LeelaChessZero/lc0/blob/7e9190a711d570367193cf068bbfcb9e32d3f868/src/mcts/search.cc#L127 which matches https://github.com/LeelaChessZero/lc0/wiki/Technical-Explanation-of-Leela-Chess-Zero
so the suggestion would be to try something like
  `uci_info.score = -400.0 * log10((1 / (edge.GetQ(0) * 0.5 + 0.5) - 1)

oscardssmith:
Attached is a visualization of the 2 results y axis cp, x axis, win pct. Blue is the new proposal.
![screenshot_2019-02-11 desmos graph](https://user-images.githubusercontent.com/11729272/52580729-97937300-2dee-11e9-81ae-e8d1c598bfa0.png)

bunkbail:
Blue in that graph looks great!

bunkbail:
Larry Kaufman of Komodo mentioned a similar thing about lc0 current implementation of centipawn conversion on Talkchess:

"Nothing wrong with converting win probs to scores in pawn units, we do the same in Komodo MCTS. But we don't output nonsense scores like this. Looks like no one bothered to check whether the conversion formula made any sense."

http://talkchess.com/forum3/viewtopic.php?f=2&t=69896

Prcuvu:
| (Q+1)/2 |cp (old)|cp (new)|
| - | - | - |
|0.50|0.00|0.00|
|0.55|0.45|0.35|
|0.60|0.93|0.70|
|0.65|1.46|1.08|
|0.70|2.07|1.47|
|0.75|2.84|1.91|
|0.80|3.89|2.41|
|0.85|5.49|3.01|
|0.90|8.42|3.82|
|0.95|16.20|5.12|
|1.00|128.00|+inf|

It looks a little bit linear. Additionally, +inf is not good to display as score and should be thresholded.

alreadydone:
[...]

PR #734, opened 2019-02-16
Improved centipawn formula

ddobbelaere:
Improved centipawn formula suggested by @yuzisee in #728 to better match other engines meaning.

The suggestion to cap absolute value of score to 128 centipawns (in correspondence to the current tan-formula behavior for |Q| = 1) by @Prcuvu and @alreadydone is also included.


ddobbelaere:
Upon second thought, I think that the proposed formula gives a pessimistic view of the score in centipawns for Leela, mainly because the winning probability of the best move is underestimated due to exploration (cpuct).

To get a score of more than 10 pawns, for example, q must be higher than 0.997, which is irrealistically high imo. So, for high evaluations, the current formula better models the output of other engines, which matters too.

Feel free to open another PR if you disagree :p.

oscardssmith:
Any chance of some testing to see agreement w SF before and after?

ddobbelaere:
Crawling through the history, I found out that the current formula replaced a formula basically the same as the proposed one (with factor -191 instead of -400) by commit https://github.com/LeelaChessZero/lc0/commit/2ee835a7bc6fdec63104aee7f172f742daaccd13

Before redoing work/research, probably @fli can help us out.

PR #738, opened 2019-02-17
Allow --show-hidden to be specified after --help

Ttl:
Also adds `--show-hidden` to help and hides KLD options.


Naphthalin:
This is more relevant than before I think, after we decided to hide all temp parameters in #1212 which is in v0.25.0

mooskagh:
I think we need to have have a "facade" of much simple parameters by default, and then we can unhide the parameters in expert mode.

But it's pretty far in my priority list, so for now it indeed makes sense to improve what we have.

src/utils/optionsparser.cc
```diff
@@ -119,12 +119,9 @@ bool OptionsParser::ProcessAllFlags() {
...
 bool OptionsParser::ProcessFlags(const std::vector<std::string>& args) {
+  auto show_help = false;
   for (auto iter = args.begin(), end = args.end(); iter != end; ++iter) {
     std::string param = *iter;
-    if (param == "-h" || param == "--help") {
-      ShowHelp();
-      return false;
-    }
     if (param == "--help-md") {
       ShowHelpMd();
```
Tilps, line 126: Any reason not to make this consistent with --help?

PR #739, opened 2019-02-17
SE-layer fix for cudnn-fp16 backend

ankan-ban:
Fall back to non-fused implementation for unsupported sizes.


Tilps:
Looks like this needs a merge from master.

ankan-ban:
> Looks like this needs a merge from master.

yeah. done.

src/neural/cuda/layers.cc
```diff
@@ -439,10 +439,12 @@ template <>
...
                          cudnnHandle_t /*cudnn*/, cublasHandle_t cublas) {
+  bool se_done = false;
   if (kUseFusedSELayer) {
-    Se_Fp16_NHWC(N, C, numFc1Out_, output, input2, input, w1_, b1_, w2_, b2_,
-                 bPrev_);
-  } else {
+    se_done = Se_Fp16_NHWC(N, C, numFc1Out_, output, input2, input, w1_, b1_,
+                           w2_, b2_, bPrev_);
+  } 
+  if (!se_done) {
```
simlu: I'm sorry, but why is this not one if statement with and "and"? Would be much more legible...

ankan-ban: Do you mean something like:
```
if (!kUseFusedSELayer || !Se_Fp16_NHWC(N, C, numFc1Out_, output, input2, input, w1_, b1_, w2_, b2_, bPrev_) ) {
   // handle non-fused case
}
```
Personally I find it less legible, but I don't have any strong objection.

simlu:
I'd do a single negation with and an &&. The double negation with || is a bit confusing agreed

I find it more readable, or at least much better LoC.

```
if (!(kUseFusedSELayer && Se_Fp16_NHWC(...))) {
```

But then again I really like concise code and some people even prefer the old Java over the new one with lambda functions (not saying you do). So I guess this is preference ;)

PR #741, opened 2019-02-18
Cudnn update to 7.4.2 + warn if old version is used with RTX cards

borg323:


gcp:
The RC binaries seem to give a warning because they're compiled with 7.3.1 and they detect a mismatch with the shipped version.

borg323:
Is this downloading just the executable? Probably appveyor is still using a cached version but the packaged zip should be OK.

gcp:
The packaged zip of the 0.20.2 release actually ships with 7.3.1 instead of 7.4.2 so it won't give the error, but I doubt that was the intention?

borg323:
Updating the appveyor cache is a hassle, so I avoid it as much as I can. There is no urgency to move to 7.4.2, so I didn't force the update: It usually happens frequently enough.

Issue #743, opened 2019-02-18
Make policy used only for initial guidance


oscardssmith:
Currently policy has a large effect on search even when there are millions of nodes. This causes weird behaviour and can lead to significant oversights. To fix this, I'm proposing that rather than maximizing `Q + c_puct*policy*sqrt(N / n_i)`, we instead maximize `Q + f(policy, N) + c_puct*sqrt(N/n_i)`. I don't have a great idea of what `f` should be, but a reasonable first aproximation is `e^-(aN+policy)`, see the link for the graph. https://www.desmos.com/calculator/g3kmkjp5zs

gcp:
I like this idea (read: I'm experimenting with it as well), it seems so counter-intuitive that policy is still a direct factor in the allocation of search effort even if we've searched thousands of nodes below a tree and have win-rates (which are better guides) for the subtrees.

oscardssmith:
Yeah also, this detail of policy effecting U was invented by Deepmind with no justification.

roy7:
I've never liked heavy weighting of policy in search even when millions of payouts have been done. I like to think of it more as a tool to guide very early search or know which unknown children to expand first.

Eager to hear about the results.

gcp:
>Yeah also, this detail of policy effecting U was invented by Deepmind with no justification.

The original paper about P-UCT had the prior as an additive, not a multiplicative factor. But I'm sure DM was well aware of that :-/

roy7:
Maybe this is a case of them trying an experiment that led to a strength gain and so they kept it without a theoretical basis to do so.

RedDenver:
This is from discord and didn't want to lose it:

> the U computation comes from several places, here's how cpuct is computed: 
> ```
> inline float ComputeCpuct(const SearchParams& params, uint32_t N) {
>   const float init = params.GetCpuct();
>   const float k = params.GetCpuctFactor();
>   const float base = params.GetCpuctBase();
>   return init + (k ? k * FastLog((N + base) / base) : 0.0f);
> }
> ```
> then
> ```
> float puct_mult =
>         cpuct * std::sqrt(std::max(node->GetChildrenVisits(), 1u));
> ```
> then
> ```
> float GetU(float numerator) const {
>     return numerator * GetP() / (1 + GetNStarted());
>   }
> ```
> where `numerator=puct_mult`

kmcrage:
I agree about not maximizing 
`Q + c_puct*policy*sqrt(N / n_i)`
[...]

PR #746, opened 2019-02-19
Make policy go away from search faster

oscardssmith:
This is not a final form, but could be made so if it gains elo. Needs testing. It is an implimentation of #743

### DIFF
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -354,11 +354,11 @@ class EdgeAndNode {
     return edge_ ? edge_->GetMove(flip) : Move();
   }
 
-  // Returns U = numerator * p / N.
+  // Returns U = numerator * decay * p / N + numerator * (1-decay) * limit / N.
   // Passed numerator is expected to be equal to (cpuct * sqrt(N[parent])).
-  float GetU(float numerator, float decay, float t) const {
+  float GetU(float numerator, float decay, float limit) const {
     float U = numerator / (1 + GetNStarted());
-    return decay * U * (GetP() - t) + U * t;
+    return decay * U * (GetP() - limit) + U * limit;
   }
 
   int GetVisitsToReachU(float target_score, float numerator,
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -357,7 +357,7 @@ class EdgeAndNode {
   // Returns U = numerator * p / N.
   // Passed numerator is expected to be equal to (cpuct * sqrt(N[parent])).
   float GetU(float numerator) const {
-    int decay = 800;
+    int decay = params_.GetPolicyDecay();
     float w = decay / (decay + numerator);
     float U = numerator / (1 + GetNStarted());
     return (1-w) * (U * GetP()) + w * U;
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -358,7 +358,7 @@ class EdgeAndNode {
   // Passed numerator is expected to be equal to (cpuct * sqrt(N[parent])).
   float GetU(float numerator, float decay, float limit) const {
     float U = numerator / (1 + GetNStarted());
-    return decay*U*(GetP() - limit) + U*limit;
+    return decay*U*GetP() + (1-decay)*U*limit;
   }
 
   int GetVisitsToReachU(float target_score, float numerator,
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -359,7 +359,8 @@ class EdgeAndNode {
   float GetU(float numerator) const {
     int decay = 800;
     float w = decay / (decay + numerator);
-    return (1-w) * (numerator * GetP() / (1 + GetNStarted())) + w * (numerator / 1 + GetNstarted());
+    float U = (numerator / 1 + GetNStarted());
+    return (1-w)*(U * GetP()) + w * U;
   }
 
   int GetVisitsToReachU(float target_score, float numerator,
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -360,7 +360,7 @@ class EdgeAndNode {
     int decay = 800;
     float w = decay / (decay + numerator);
     float U = (numerator / 1 + GetNStarted());
-    return (1-w)*(U * GetP()) + w * U;
+    return (1-w) * (U * GetP()) + w * U;
   }
 
   int GetVisitsToReachU(float target_score, float numerator,
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -359,7 +359,7 @@ class EdgeAndNode {
   float GetU(float numerator, int decay) const {
     float w = decay / (decay + numerator);
     float U = numerator / (1 + GetNStarted());
-    return (1 - w) * (U * GetP()) + w * U;
+    return w * (U * GetP()) + (1 - w) * U;
   }
 
   int GetVisitsToReachU(float target_score, float numerator,
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -359,7 +359,7 @@ class EdgeAndNode {
   float GetU(float numerator) const {
     int decay = 800;
     float w = decay / (decay + numerator);
-    float U = (numerator / 1 + GetNStarted());
+    float U = numerator / (1 + GetNStarted());
     return (1-w) * (U * GetP()) + w * U;
   }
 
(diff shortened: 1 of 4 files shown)


hans-ekbrand:
I get this build error:

```
../../src/mcts/search.cc:966:63: error: ‘fpu’ was not declared in this scope
           best_edge.GetVisitsToReachU(second_best, puct_mult, fpu);
                                                               ^~~

```

RedDenver:
The code looks fine to me, but I don't know about the f(policy,N) function you chose. If you want something that causes policy to dominate early and then fade to nothing, then I suggest something similar to a step function like:
```
      float Q = child.GetQ(0);
      const float score = child.GetU(puct_mult) + child.GetP()*exp(800 - node->GetN()) + Q;
```
The 800 can be replaced by however many playouts  are desired before policy fades away (e.g. number of playouts during training).

oscardssmith:
e^(800-N) won't work, because then it would start at e^800 which is way way way too big.

RedDenver:
> e^(800-N) won't work, because then it would start at e^800 which is way way way too big. Also playing around with the graph of that, it seems sub-optimal. That creates straight lines of varying slopes.

Yep, e^800 is greater than max of double. But e^700 is within range.

However, it might be easier to just use a linear slope with a floor like:
```
float Q = child.GetQ(0);
const float score = child.GetU(puct_mult) + child.GetP()*max(800 - node->GetN(), 0) + Q;
```

RedDenver:
Actually, I don't think any of our equations will work because we're using the parent node count, so all of the edges will have their f(policy,N) remain the same relative to each other. I think we need to use the child node counts f(policy, n_i, N), so something like:
```
float Q = child.GetQ(0);
const float score = child.GetU(puct_mult) + max(static_cast<int>(child.GetP()*800) - child.GetN(), 0) + Q;
```

oscardssmith:
@RedDenver the point is that even though policy remains the same relative to each other, the impact gets smaller and Q takes over. I very specifically used N, not n_i because the purpose of this is to prevent policy from controlling things at higher nodes, but if n_i was used, you would have a feedback loop where bad policy needed lots of nodes to have the impact of the policy diminished.

RedDenver:
@oscardssmith Are you trying to come up with an equation for both training and normal play or just for normal play after a NN has already been trained?

oscardssmith:
Not sure yet. Normal play is the easier to validate so probably just that at first.

oscardssmith:
Initial blitz results by @hans-ekbrand have this branch up 9-7.

RedDenver:
Then take a look at my last code suggestion as it should work for what you want in normal play because policy will determine the first 800 playouts and then none after that, but it will probably not work for training.

Although thinking about both play and training, maybe what we really want is the old equation at low node counts (`Q_i + c_puct*p_i*sqrt(N/n_i)`) and the same but without policy at high node counts (`Q_i + c_puct*sqrt(N/n_i)`).

which makes me think of a weight that approaches zero as N gets larger:
```
w = 800/(800+N)
Q_i + w*c_puct*p_i*sqrt(N/n_i) + (1-w)*c_puct*sqrt(N/n_i)
```

patrickDBL:
I'm recording here a discord conversation.

If the current formula is 
```Q + c_puct*policy*sqrt(N / n_i)```
I would suggest 
```Q + c_puct*[(alpha*policy + n_i)/(alpha + n_i)] *sqrt(N / n_i).```
I am wary of both adding decay and changing the functional form at the same time. Then you don't know whether the effects (positive or negative) come from the decay, the change from multiplicative to additive, or both.

My suggestion interpolates between the current formula at n_i=0 and a flat policy when n_i is large. Further, you can tune alpha to see the effect of the decay. I would start with alpha quite large (1000 say), and this should produce essentially no change in strength. Then lowering alpha slowly lets you test the effect of decay in a controlled way, with alpha=0 corresponding to having no policy whatsoever.

**Comment by manyoso** [comment]
FWIW, I did something like this for Allie and it lost elo. I was recalculating the policy values for all nodes after they had been scored and received qvals by normalizing and scaling the pvals to the qvals. I have this code if anyone wants to see it, but I am dubious that it will help with ELO. The point is that the Net has these policy priors for a reason and ignoring them when you get a posterior value or minimizing them when you get a posterior value doesn't seem to work in my testing.

**Comment by hans-ekbrand** [comment]
This PR doesn't comile as it is. I fixed that, tested it, improved it got better test results (still not better the master, but better than the original PR). But unfortunately I am not allowed to push my changes here, so I pushed them to my forked repo: https://github.com/hans-ekbrand/lc0/commits/pr746

Testing results with the original PR:
800 nodes per move
```
Score of lc0 master vs lc0 PR746: 29 - 20 - 51  [0.545] 100
Elo difference: 31.35 +/- 47.90
Finished match
```
So that wasn't good, but I took the liberty of change addition to multiplication of the U term, and then this PR is even with master. I've tested at 800 nodes per move, 3000 nodes per move and 9000 nodes per move:
```
With * instead of + 800 nodes
Score of lc0 master vs lc0 PR746: 21 - 23 - 56  [0.490] 100
Elo difference: -6.95 +/- 45.36
Finished match

With * instead of + 3000 nodes
Score of lc0 master vs lc0 PR746: 12 - 11 - 77  [0.505] 100
Elo difference: 3.47 +/- 32.70
Finished match

With * instead of + 9000 nodes
Score of lc0 master vs lc0 PR746: 11 - 12 - 77  [0.495] 100
Elo difference: -3.47 +/- 32.70
Finished match
```

Personally, I think any change that isn't directly worse than our current MCTS is promising.

**Comment by oscardssmith** [comment]
Changing it to multiplication gets rid of the point of this (removing policy from U eventually), but the fact that it's so easy to get even to current behaviour does kind of prove my point that our current search is broken. I'm pretty busy now, (getting close to the end of term), but I would love if you kept working on this. I think trying the approach of @patrickDBL looks good. At low nodes it's the same as master, but it should be better at more nodes.

**Comment by oscardssmith** [comment]
At this point, I need a clop of `cpuct` and `policyDecay`. My initial guess is that `cpuct=.5`, `policyDecay=3000` ish, but those are only very rough estimates.

**Comment by killerducky** [comment]

**Comment by oscardssmith** [comment]
I think this is ready for code review and eventually merge. jhorthos will do an elo test which I will post once it's ready.

**Comment by oscardssmith** [comment]
Jhorthos's tactics results This was all using net 50690.

version nodes     correct    total    percent
PR746  200000  768           1007    76.27    
v21.1     200000  753          1007    74.78    
PR746  500000  792          1007    78.65    
 v21.1    500000  775          1007    76.96

In other words, this PR is equivalent to more than doubling nodes in tactical positions.

**Comment by zz4032** [comment]
Two matches with default parameter values vs. current dev (Id41751).
TC 2.4s/game + 0.01s/move (~120nodes/move)
```
   1 lc0          :       0   ----    99.7     100      70.0
   2 lc0_PR746    :     -71     50     ---     100      70.0
```
TC 9.6s/game + 0.04s/move (~3000nodes/move)
```
   1 lc0          :       0   ----    75.8     100      81.0
   2 lc0_PR746    :     -18     49     ---     100      81.0
```

**Comment by oscardssmith** [comment]
That's really weird. At 120 nodes, search should be completely unchanged.

**Comment by aartdappel** [comment]
@oscardssmith I guess policy-softmax-temp needs to be closer to original value, also maybe speed is affected much at very short TC

**Comment by zz4032** [comment]
@oscardssmith Probably an effect of modified PolicySoftmaxTemp. Going to test again with PolicySoftmaxTemp=2.2 for both at ~3000nodes/move.

**Comment by oscardssmith** [comment]
Aart he was testing on CPU, so the speed shouldn't have any difference.

**Comment by zz4032** [comment]
TC 15.12s/game + 0.063s/move (~4000nodes/move)
lc0_PR746: commit '718a517' with `option.PolicyTemperature=2.2`
```
   1 lc0          :     0.0   ----   100.0    1000      65.0
   2 lc0_PR746    :   -43.5   12.1     ---    1000      65.0

White ELO advantage = 89.4
Draw rate (equal opponents) = 74.0 %
```

aartdappel:
thats odd, must be a way too high policy-limit then, or we must take amount of edges into consideration

Naphthalin:
As the approach of this is similar to #1150 but only produced negative results, it would either need to be retested with newer nets (as the quality of the policies has increased dramatically over the last year) or closed; the general discussion about this kind of deep change should probably happen in #1150.

oscardssmith:
agreed

PR #750, opened 2019-02-21
Add AtRoot versions of FpuStrategy and FpuValue (and convert FpuReduction to FpuValue).

### LINKED ISSUE #8: Improve training data for learning tactics
Porting to lc0 of lczero issues https://github.com/glinscott/leela-chess/issues/698 and https://github.com/glinscott/leela-chess/issues/699 using the same game for analysis:

<details>
<summary>CCLS SCTR vs id359 game 1</summary>

Trying to find Rxh4 https://clips.twitch.tv/NimbleLazyNewtPRChase:
```
position startpos moves d2d4 d7d5 c1f4 g7g6 e2e3 g8f6 c2c4 c7c5 d4c5 f8g7 b1c3 d8a5 c4d5 f6d5 d1d5 g7c3 b2c3 a5c3 e1e2 c3a1 f4e5 a1b1 e5h8 c8e6 d5d3 b1a2 e2f3 f7f6 h8g7 b8d7 f3g3 a8c8 c5c6 c8c6 d3d4 c6d6 d4b4 d6b6 b4h4 d7c5 h2h3 b6b2 g1e2 a2d5 g3h2 d5e5 e2g3 h7h5 h4d4 e5d4 e3d4 c5b3 g7h6 h5h4 g3e4 g6g5 f1d3 b3d4 h1a1 a7a6 e4c5 b2f2 d3e4 e6f5 e4b7 f2c2 a1a4 d4e2 c5e4 f5e4 b7e4 c2c1 e4d3 e2f4 d3a6 f4h5
```
screen shot 2018-05-31 at 10 38 50 am
</details>

Here's the history of networks from 364 going back 10 at a time and what they thought of the winning move Rxh4 / a4h4 (focus on V and P for now):
```
id364 a4h4  (666 ) N:     759 (+ 0) (V: -36.58%) (P:  0.47%) (Q:  0.76777) (U: 0.00021) (Q+U:  0.76798) 
id354 a4h4  (666 ) N:     801 (+ 0) (V: -41.09%) (P:  0.23%) (Q:  0.78342) (U: 0.00010) (Q+U:  0.78353) 
id344 a4h4  (666 ) N:     747 (+ 0) (V: -43.49%) (P:  0.15%) (Q:  0.76876) (U: 0.00007) (Q+U:  0.76883) 
id334 a4h4  (666 ) N:     760 (+ 0) (V: -39.09%) (P:  0.11%) (Q:  0.77163) (U: 0.00005) (Q+U:  0.77168) 
id324 a4h4  (666 ) N:     752 (+ 0) (V: -40.29%) (P:  0.21%) (Q:  0.76674) (U: 0.00010) (Q+U:  0.76683) 
id314 a4h4  (666 ) N:     725 (+ 0) (V: -40.58%) (P:  0.18%) (Q:  0.76648) (U: 0.00009) (Q+U:  0.76656) 
id304 a4h4  (666 ) N:     779 (+ 0) (V: -46.39%) (P:  0.18%) (Q:  0.76410) (U: 0.00008) (Q+U:  0.76418) 
id294 a4h4  (666 ) N:     812 (+ 0) (V: -44.82%) (P:  0.17%) (Q:  0.76923) (U: 0.00007) (Q+U:  0.76930) 
id284 a4h4  (666 ) N:     756 (+ 0) (V:  -3.65%) (P:  0.29%) (Q:  0.76365) (U: 0.00013) (Q+U:  0.76379) 
id274 a4h4  (666 ) N:     775 (+ 0) (V:  16.01%) (P:  0.25%) (Q:  0.78128) (U: 0.00011) (Q+U:  0.78139) 
id264 a4h4  (666 ) N:     708 (+ 0) (V: -47.48%) (P:  0.17%) (Q:  0.74895) (U: 0.00008) (Q+U:  0.74903) 
id254 a4h4  (666 ) N:     721 (+ 0) (V: -19.61%) (P:  0.25%) (Q:  0.73499) (U: 0.00012) (Q+U:  0.73511) 
id244 a4h4  (666 ) N:     718 (+ 0) (V: -34.57%) (P:  0.23%) (Q:  0.69746) (U: 0.00011) (Q+U:  0.69756) 
id234 a4h4  (666 ) N:     750 (+ 0) (V: -21.15%) (P:  0.58%) (Q:  0.68519) (U: 0.00027) (Q+U:  0.68546) 
id224 a4h4  (666 ) N:     762 (+ 0) (V: -16.51%) (P:  0.38%) (Q:  0.73801) (U: 0.00017) (Q+U:  0.73818) 
id214 a4h4  (666 ) N:     745 (+ 0) (V: -13.26%) (P:  0.40%) (Q:  0.74920) (U: 0.00018) (Q+U:  0.74938) 
id204 a4h4  (666 ) N:     729 (+ 0) (V: -28.40%) (P:  0.31%) (Q:  0.53719) (U: 0.00015) (Q+U:  0.53734) 
id194 a4h4  (666 ) N:     741 (+ 0) (V:  -3.47%) (P:  0.51%) (Q:  0.74044) (U: 0.00024) (Q+U:  0.74068) 
id184 a4h4  (666 ) N:     745 (+ 0) (V: -23.81%) (P:  0.44%) (Q:  0.72124) (U: 0.00020) (Q+U:  0.72144) 
id174 a4h4  (666 ) N:     724 (+ 0) (V:  16.63%) (P:  0.25%) (Q:  0.66271) (U: 0.00012) (Q+U:  0.66283) 
id164 a4h4  (666 ) N:     715 (+ 0) (V:  -5.59%) (P:  0.91%) (Q:  0.65085) (U: 0.00043) (Q+U:  0.65129) 
id154 a4h4  (666 ) N:     711 (+ 0) (V: -21.75%) (P:  0.57%) (Q:  0.62195) (U: 0.00027) (Q+U:  0.62222) 
id144 a4h4  (666 ) N:     723 (+ 0) (V: -28.62%) (P:  0.59%) (Q:  0.62137) (U: 0.00028) (Q+U:  0.62165)
id134 a4h4  (666 ) N:     492 (+ 0) (V: -47.86%) (P:  0.61%) (Q:  0.53013) (U: 0.00042) (Q+U:  0.53055) 
id124 a4h4  (666 ) N:     636 (+ 0) (V: -37.09%) (P:  0.39%) (Q:  0.62892) (U: 0.00021) (Q+U:  0.62912) 
```

Generally, the prior for this winning move is very low at under 1%, and the value is also unfavorable for white, so search will normally avoid it. This is tricky for tactics to be learned where playing an initially bad move opens up a better outcome.

That's where noise comes in to trick search into visiting more, and here's 50 runs of `./lc0 --weights=id359 --verbose-move-stats --noise --no-smart-pruning` with `go nodes 800` from the above `position startpos …`:
```
info string a4h4  (666 ) N:       0 (+ 0) (V:   0.00%) (P:  0.25%) (Q: -0.29468) (U: 0.08479) (Q+U: -0.20989) 
info string a4h4  (666 ) N:       1 (+ 0) (V: -38.94%) (P:  0.25%) (Q: -0.38945) (U: 0.04206) (Q+U: -0.34739) 
info string a4h4  (666 ) N:       1 (+ 0) (V: -38.94%) (P:  0.25%) (Q: -0.38945) (U: 0.04234) (Q+U: -0.34711) 
info string a4h4  (666 ) N:       1 (+ 0) (V: -38.94%) (P:  0.25%) (Q: -0.38945) (U: 0.04248) (Q+U: -0.34696) 
info string a4h4  (666 ) N:       1 (+ 0) (V: -38.94%) (P:  0.25%) (Q: -0.38945) (U: 0.04262) (Q+U: -0.34683) 
info string a4h4  (666 ) N:       1 (+ 0) (V: -38.94%) (P:  0.25%) (Q: -0.38945) (U: 0.04310) (Q+U: -0.34634) 
info string a4h4  (666 ) N:       1 (+ 0) (V: -38.94%) (P:  0.25%) (Q: -0.38945) (U: 0.04329) (Q+U: -0.34616) 
info string a4h4  (666 ) N:       1 (+ 0) (V: -38.94%) (P:  0.26%) (Q: -0.38945) (U: 0.04338) (Q+U: -0.34607) 
info string a4h4  (666 ) N:       1 (+ 0) (V: -38.94%) (P:  0.26%) (Q: -0.38945) (U: 0.04345) (Q+U: -0.34600) 
info string a4h4  (666 ) N:       1 (+ 0) (V: -38.94%) (P:  0.27%) (Q: -0.38945) (U: 0.04560) (Q+U: -0.34384) 
info string a4h4  (666 ) N:       1 (+ 0) (V: -38.94%) (P:  0.27%) (Q: -0.38945) (U: 0.04596) (Q+U: -0.34348) 
info string a4h4  (666 ) N:       1 (+ 0) (V: -38.94%) (P:  0.30%) (Q: -0.38945) (U: 0.05126) (Q+U: -0.33818) 
info string a4h4  (666 ) N:       1 (+ 0) (V: -38.94%) (P:  0.33%) (Q: -0.38945) (U: 0.05541) (Q+U: -0.33404) 
info string a4h4  (666 ) N:       1 (+ 0) (V: -38.94%) (P:  0.34%) (Q: -0.38945) (U: 0.05772) (Q+U: -0.33173) 
info string a4h4  (666 ) N:       1 (+ 0) (V: -38.94%) (P:  0.36%) (Q: -0.38945) (U: 0.06209) (Q+U: -0.32735) 
info string a4h4  (666 ) N:       1 (+ 0) (V: -38.94%) (P:  0.38%) (Q: -0.38945) (U: 0.06505) (Q+U: -0.32440) 
info string a4h4  (666 ) N:       1 (+ 0) (V: -38.94%) (P:  0.38%) (Q: -0.38945) (U: 0.06522) (Q+U: -0.32422) 
info string a4h4  (666 ) N:       1 (+ 0) (V: -38.94%) (P:  0.39%) (Q: -0.38945) (U: 0.06538) (Q+U: -0.32406) 
info string a4h4  (666 ) N:       1 (+ 0) (V: -38.94%) (P:  0.40%) (Q: -0.38945) (U: 0.06868) (Q+U: -0.32077) 
info string a4h4  (666 ) N:       1 (+ 0) (V: -38.94%) (P:  0.40%) (Q: -0.38945) (U: 0.06874) (Q+U: -0.32070) 
info string a4h4  (666 ) N:       1 (+ 0) (V: -38.94%) (P:  0.41%) (Q: -0.38945) (U: 0.06946) (Q+U: -0.31998) 
info string a4h4  (666 ) N:       1 (+ 0) (V: -38.94%) (P:  0.41%) (Q: -0.38945) (U: 0.06971) (Q+U: -0.31973) 
info string a4h4  (666 ) N:       1 (+ 0) (V: -38.94%) (P:  0.43%) (Q: -0.38945) (U: 0.07282) (Q+U: -0.31662) 
info string a4h4  (666 ) N:       1 (+ 0) (V: -38.94%) (P:  0.47%) (Q: -0.38945) (U: 0.07992) (Q+U: -0.30952) 
info string a4h4  (666 ) N:       1 (+ 0) (V: -38.94%) (P:  0.48%) (Q: -0.38945) (U: 0.08057) (Q+U: -0.30888) 
info string a4h4  (666 ) N:       1 (+ 0) (V: -38.94%) (P:  0.48%) (Q: -0.38945) (U: 0.08158) (Q+U: -0.30786) 
info string a4h4  (666 ) N:       1 (+ 0) (V: -38.94%) (P:  0.48%) (Q: -0.38945) (U: 0.08243) (Q+U: -0.30702) 
info string a4h4  (666 ) N:       1 (+ 0) (V: -38.94%) (P:  0.53%) (Q: -0.38945) (U: 0.08946) (Q+U: -0.29998) 
info string a4h4  (666 ) N:       1 (+ 0) (V: -38.94%) (P:  0.54%) (Q: -0.38945) (U: 0.09278) (Q+U: -0.29667) 
info string a4h4  (666 ) N:       1 (+ 0) (V: -38.94%) (P:  0.60%) (Q: -0.38945) (U: 0.10182) (Q+U: -0.28763) 
info string a4h4  (666 ) N:       1 (+ 0) (V: -38.94%) (P:  0.61%) (Q: -0.38945) (U: 0.10405) (Q+U: -0.28540) 
info string a4h4  (666 ) N:       1 (+ 0) (V: -38.94%) (P:  0.64%) (Q: -0.38945) (U: 0.10879) (Q+U: -0.28066) 
info string a4h4  (666 ) N:       1 (+ 0) (V: -38.94%) (P:  0.83%) (Q: -0.38945) (U: 0.14161) (Q+U: -0.24783) 
info string a4h4  (666 ) N:       1 (+ 0) (V: -38.94%) (P:  0.88%) (Q: -0.38945) (U: 0.14977) (Q+U: -0.23967) 
info string a4h4  (666 ) N:       1 (+ 0) (V: -38.94%) (P:  0.90%) (Q: -0.38945) (U: 0.15223) (Q+U: -0.23722) 
info string a4h4  (666 ) N:       1 (+ 0) (V: -38.94%) (P:  0.96%) (Q: -0.38945) (U: 0.16305) (Q+U: -0.22639) 
info string a4h4  (666 ) N:       1 (+ 0) (V: -38.94%) (P:  1.00%) (Q: -0.38945) (U: 0.17040) (Q+U: -0.21904) 
info string a4h4  (666 ) N:     132 (+ 0) (V: -38.94%) (P:  1.29%) (Q:  0.76707) (U: 0.00331) (Q+U:  0.77039) 
info string a4h4  (666 ) N:     132 (+ 0) (V: -38.94%) (P:  1.30%) (Q:  0.76707) (U: 0.00336) (Q+U:  0.77043) 
info string a4h4  (666 ) N:     527 (+ 0) (V: -38.94%) (P:  1.39%) (Q:  0.77280) (U: 0.00092) (Q+U:  0.77373) 
info string a4h4  (666 ) N:     215 (+ 0) (V: -38.94%) (P:  1.48%) (Q:  0.77114) (U: 0.00236) (Q+U:  0.77350) 
info string a4h4  (666 ) N:     484 (+ 0) (V: -38.94%) (P:  1.50%) (Q:  0.77353) (U: 0.00107) (Q+U:  0.77460) 
info string a4h4  (666 ) N:     289 (+ 0) (V: -38.94%) (P:  1.71%) (Q:  0.77238) (U: 0.00202) (Q+U:  0.77440) 
info string a4h4  (666 ) N:     536 (+ 0) (V: -38.94%) (P:  1.81%) (Q:  0.77391) (U: 0.00117) (Q+U:  0.77507) 
info string a4h4  (666 ) N:     714 (+ 0) (V: -38.94%) (P:  2.39%) (Q:  0.77523) (U: 0.00123) (Q+U:  0.77645) 
info string a4h4  (666 ) N:     649 (+ 0) (V: -38.94%) (P:  2.97%) (Q:  0.77412) (U: 0.00158) (Q+U:  0.77570) 
info string a4h4  (666 ) N:     643 (+65) (V: -38.94%) (P:  3.22%) (Q:  0.77441) (U: 0.00156) (Q+U:  0.77597) 
info string a4h4  (666 ) N:     714 (+ 0) (V: -38.94%) (P:  3.66%) (Q:  0.77523) (U: 0.00181) (Q+U:  0.77703) 
info string a4h4  (666 ) N:     714 (+ 0) (V: -38.94%) (P:  5.44%) (Q:  0.77523) (U: 0.00263) (Q+U:  0.77786) 
info string a4h4  (666 ) N:     738 (+ 0) (V: -38.94%) (P:  8.65%) (Q:  0.77528) (U: 0.00404) (Q+U:  0.77931) 
```
Here, 13 of 50 games would have produced valuable training data, so noise is indeed working, but the majority is training to avoid the correct move. Averaging this training data for the move across 50 games should cause P to move towards 16.3% (= 6523 / ~800 / 50). But then combined with training data from other games, the networks have learned to keep avoiding this move.

As from the other issue: The premise is that for a self-play to end up in a learnable board state, it seems unfortunate that it misses the opportunity to generate valuable training data for the correct move more often than not. Clearly, AZ's numbers are good enough to eventually generate strong networks, but perhaps training search could be better optimized?

I've rerun the analysis with lc0 and 50 games each configuration from the above board state to measure the average training data for the expected tactic:

<details>
<summary>Testing patches for visit twice and negative fpu</summary>

```diff
diff --git a/src/mcts/search.cc b/src/mcts/search.cc
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -650,4 +650,9 @@ Node* Search::PickNodeToExtend(Node* node, PositionHistory* history) {
     for (Node* iter : node->Children()) {
       if (is_root_node) {
+        if (kNoise && iter->GetN() < 2) {
+          node = iter;
+          possible_moves = 2; // avoid "only one possible move" short circuit
+          break;
+        }
         // If there's no chance to catch up the currently best node with
         // remaining playouts, not consider it.
```
```diff
diff --git a/src/mcts/search.cc b/src/mcts/search.cc
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -645,5 +645,5 @@ Node* Search::PickNodeToExtend(Node* node, PositionHistory* history) {
     float parent_q =
         (is_root_node && kNoise)
-            ? -node->GetQ(0, kExtraVirtualLoss)
+            ? -node->GetQ(0, kExtraVirtualLoss) + kFpuReduction
             : -node->GetQ(0, kExtraVirtualLoss) -
                   kFpuReduction * std::sqrt(node->GetVisitedPolicy());
```
</details>

epsilon | alpha | fpu | twice | average tactic training
-- | -- | -- | -- | --
0.25 | 0.3 | 0.0 | no | 16.3%
0.25 | 3.0 | 0.0 | no | 27.6%
0.5 | 0.3 | 0.0 | no | 33.1%
0.5 | 3.0 | 0.0 | no | 60.0%
0.25 | 0.3 | -0.2 | no | 18.5%
0.25 | 3.0 | -0.2 | no | 28.3%
0.5 | 0.3 | -0.2 | no | 34.8%
0.5 | 3.0 | -0.2 | no | 59.6%
0.25 | 0.3 | 0.0 | yes | 92.1%

I only ran one "visit each root move twice" as even with the default search parameters, it generally searches much deeper after being nudged over with the forced breadth exploration. This is true across all the previously listed networks from id364 to id124 above, and the output with high Ns are with "visit twice."

Is there an appropriate level of average tactic training? It looks like the current 16.3% is too low to outweigh the other training data. A related question is how often are self-play games getting into learnable states, but I don't have a good way to answer that.

### LINKED ISSUE #748: AlphaZero searches much wider than lc0 at low visits
From Game Changer, page 81 has a table showing move priors and eval at "64" visits (seems to actually be 67 total child visits) and the top 19 moves out of 33 possible.
screen shot 2019-02-20 at 1 01 12 pm

```
position startpos moves g1f3 g8f6 c2c4 e7e6 b1c3 f8b4 d1c2 e8g8 a2a3 b4c3 c2c3 a7a5 b2b4 d7d6 e2e3 f6e4 c3c2 e4g5 b4b5 g5f3 g2f3 d8f6 d2d4 f6f3 h1g1 b8d7 f1e2 f3f6 c1b2 f6h4 g1g4 h4h2 g4g3 f7f5 e1c1 f8f7 e2f3 h2h4 d1h1 h4f6 c1b1 g7g6 g3g1 a5a4 b1a1 f7g7 e3e4 f5f4 c4c5 f6e7 g1c1 d7f6 e4e5 d6e5 h1e1 e5e4 f3e4 e7f8
```
screen shot 2019-02-20 at 1 04 29 pm

The table shows that Kb1/a1b1 with 0.07% prior managed to get 3% of visits, which matches up with 2 visits of the "64". It looks like the table is sorted by N then P, so an assumption here is that because a 0.07% prior child was visited twice, the unshown children have 1 visit with prior less than 0.26% and unvisited children have priors less than 0.07%.

To get 67 children visits, the table shows 19 children with 60 visits, so most likely there are 7 unshown children with 1 visit each and 7 more unshown children with 0 visits each.

For reference, here's 32930 (policy-softmax-temp=1) forcing a visit to each child once:
```
info string c2a4  (262 ) N:       1 (+ 0) (P:  0.02%) (Q: -0.85741) (D:  0.000) (U: 0.00209) (Q+U: -0.85532) (V: -0.8574) 
info string e4d5  (795 ) N:       1 (+ 0) (P:  0.05%) (Q: -0.98659) (D:  0.000) (U: 0.00406) (Q+U: -0.98253) (V: -0.9866) 
info string c2b3  (258 ) N:       1 (+ 0) (P:  0.02%) (Q: -0.97691) (D:  0.000) (U: 0.00193) (Q+U: -0.97498) (V: -0.9769) 
info string e4f5  (797 ) N:       1 (+ 0) (P:  0.05%) (Q: -0.96944) (D:  0.000) (U: 0.00445) (Q+U: -0.96500) (V: -0.9694) 
info string e1e3  (111 ) N:       1 (+ 0) (P:  0.05%) (Q: -0.92453) (D:  0.000) (U: 0.00415) (Q+U: -0.92038) (V: -0.9245) 
info string e4c6  (799 ) N:       1 (+ 0) (P:  0.03%) (Q: -0.71045) (D:  0.000) (U: 0.00288) (Q+U: -0.70758) (V: -0.7105) 
info string e4g6  (803 ) N:       1 (+ 0) (P:  0.03%) (Q: -0.47933) (D:  0.000) (U: 0.00292) (Q+U: -0.47641) (V: -0.4793) 
info string b2c3  (231 ) N:       1 (+ 0) (P:  0.13%) (Q: -0.02199) (D:  0.000) (U: 0.01067) (Q+U: -0.01132) (V: -0.0220) 
info string a1b1  (0   ) N:       1 (+ 0) (P:  0.09%) (Q: -0.00074) (D:  0.000) (U: 0.00764) (Q+U:  0.00690) (V: -0.0007) 
info string a1a2  (7   ) N:       1 (+ 0) (P:  0.09%) (Q:  0.03331) (D:  0.000) (U: 0.00781) (Q+U:  0.04112) (V:  0.0333) 
info string c1b1  (48  ) N:       1 (+ 0) (P:  0.11%) (Q:  0.07281) (D:  0.000) (U: 0.00963) (Q+U:  0.08244) (V:  0.0728) 
info string b5b6  (941 ) N:       1 (+ 0) (P:  0.21%) (Q:  0.10347) (D:  0.000) (U: 0.01815) (Q+U:  0.12162) (V:  0.1035) 
info string e1f1  (101 ) N:       1 (+ 0) (P:  0.12%) (Q:  0.11431) (D:  0.000) (U: 0.01032) (Q+U:  0.12463) (V:  0.1143) 
info string c2d2  (252 ) N:       1 (+ 0) (P:  0.26%) (Q:  0.16386) (D:  0.000) (U: 0.02225) (Q+U:  0.18611) (V:  0.1639) 
info string e1d1  (100 ) N:       1 (+ 0) (P:  0.15%) (Q:  0.17411) (D:  0.000) (U: 0.01290) (Q+U:  0.18700) (V:  0.1741) 
info string c2c3  (259 ) N:       1 (+ 0) (P:  0.16%) (Q:  0.19155) (D:  0.000) (U: 0.01336) (Q+U:  0.20491) (V:  0.1916) 
info string c2b1  (246 ) N:       1 (+ 0) (P:  0.21%) (Q:  0.19083) (D:  0.000) (U: 0.01780) (Q+U:  0.20862) (V:  0.1908) 
info string e4g2  (781 ) N:       1 (+ 0) (P:  0.79%) (Q:  0.16545) (D:  0.000) (U: 0.06753) (Q+U:  0.23298) (V:  0.1654) 
info string c2d1  (248 ) N:       1 (+ 0) (P:  0.21%) (Q:  0.22839) (D:  0.000) (U: 0.01758) (Q+U:  0.24597) (V:  0.2284) 
info string e1g1  (102 ) N:       1 (+ 0) (P:  0.48%) (Q:  0.21439) (D:  0.000) (U: 0.04065) (Q+U:  0.25504) (V:  0.2144) 
info string f2f3  (346 ) N:       1 (+ 0) (P:  0.52%) (Q:  0.22276) (D:  0.000) (U: 0.04454) (Q+U:  0.26730) (V:  0.2228) 
info string c2d3  (260 ) N:       1 (+ 0) (P:  0.25%) (Q:  0.24711) (D:  0.000) (U: 0.02098) (Q+U:  0.26809) (V:  0.2471) 
info string e1h1  (103 ) N:       1 (+ 0) (P:  0.32%) (Q:  0.24280) (D:  0.000) (U: 0.02710) (Q+U:  0.26991) (V:  0.2428) 
info string e1e2  (106 ) N:       1 (+ 0) (P:  0.19%) (Q:  0.25640) (D:  0.000) (U: 0.01583) (Q+U:  0.27224) (V:  0.2564) 
info string c1d1  (49  ) N:       1 (+ 0) (P:  0.18%) (Q:  0.28285) (D:  0.000) (U: 0.01569) (Q+U:  0.29854) (V:  0.2829) 
info string e4h1  (776 ) N:       1 (+ 0) (P:  1.19%) (Q:  0.19882) (D:  0.000) (U: 0.10067) (Q+U:  0.29949) (V:  0.1988) 
info string c2e2  (253 ) N:       1 (+ 0) (P:  0.47%) (Q:  0.29207) (D:  0.000) (U: 0.03964) (Q+U:  0.33172) (V:  0.2921) 
info string c2c4  (264 ) N:       1 (+ 0) (P:  1.20%) (Q:  0.28042) (D:  0.000) (U: 0.10174) (Q+U:  0.38217) (V:  0.2804) 
info string e4b7  (804 ) N:       1 (+ 0) (P:  0.03%) (Q:  0.38252) (D:  0.000) (U: 0.00285) (Q+U:  0.38537) (V:  0.3825) 
info string e4f3  (785 ) N:       1 (+ 2) (P: 15.74%) (Q:  0.26857) (D:  0.000) (U: 0.66853) (Q+U:  0.93710) (V:  0.2686) 
info string c5c6  (973 ) N:       1 (+ 5) (P: 23.27%) (Q:  0.37320) (D:  0.000) (U: 0.56480) (Q+U:  0.93801) (V:  0.3732) 
info string e4d3  (783 ) N:       1 (+ 5) (P: 35.75%) (Q:  0.10710) (D:  0.000) (U: 0.86772) (Q+U:  0.97482) (V:  0.1071) 
info string d4d5  (761 ) N:       1 (+ 8) (P: 17.61%) (Q:  0.69821) (D:  0.000) (U: 0.29924) (Q+U:  0.99745) (V:  0.6982) 
```

The 7 unvisited children most likely match up with the first 7 moves above with a very negative Q while other children are at least draw-ish and matching up with that is very low P at most 0.05%. So it seems quite likely AZ's P for these 7 children is very close to 0.00%.

The table shows averaged Q values near 0.657 with smallest Q in table 0.508, which mapping to a range of [-1, 1] should be average Q=.314.

Assuming AZ search indeed initialized unvisited children Q=-1, then for AZ search to be visiting children with priors as low as 0.07% means:
```
0.07% child Q+U > max(Q+U)
-1 + U-of-0.07% > ~.314 + U-of-other
U-of-0.07% > ~1.3 + U-of-other
```

However, 32930 with policy-softmax-temp=1 FPU=-1 and 67 child visits only visits the top 4 highest prior moves:
```
…
info string c2e2  (253 ) N:       0 (+ 0) (P:  0.47%) (Q: -1.00000) (D:  0.000) (U: 0.11486) (Q+U: -0.88514) (V:  -.----) 
info string e1g1  (102 ) N:       0 (+ 0) (P:  0.48%) (Q: -1.00000) (D:  0.000) (U: 0.11777) (Q+U: -0.88223) (V:  -.----) 
info string f2f3  (346 ) N:       0 (+ 0) (P:  0.52%) (Q: -1.00000) (D:  0.000) (U: 0.12903) (Q+U: -0.87097) (V:  -.----) 
info string e4g2  (781 ) N:       0 (+ 0) (P:  0.79%) (Q: -1.00000) (D:  0.000) (U: 0.19564) (Q+U: -0.80436) (V:  -.----) 
info string e4h1  (776 ) N:       0 (+ 0) (P:  1.19%) (Q: -1.00000) (D:  0.000) (U: 0.29167) (Q+U: -0.70833) (V:  -.----) 
info string c2c4  (264 ) N:       0 (+ 0) (P:  1.20%) (Q: -1.00000) (D:  0.000) (U: 0.29477) (Q+U: -0.70523) (V:  -.----) 
info string d4d5  (761 ) N:       7 (+ 0) (P: 17.61%) (Q:  0.03270) (D:  0.000) (U: 0.54185) (Q+U:  0.57455) (V:  0.6982) 
info string e4f3  (785 ) N:       7 (+ 0) (P: 15.74%) (Q:  0.12017) (D:  0.000) (U: 0.48421) (Q+U:  0.60438) (V:  0.2686) 
info string e4d3  (783 ) N:      14 (+ 0) (P: 35.75%) (Q: -0.01088) (D:  0.000) (U: 0.58658) (Q+U:  0.57570) (V:  0.1071) 
info string c5c6  (973 ) N:      39 (+ 1) (P: 23.27%) (Q:  0.29288) (D:  0.000) (U: 0.13969) (Q+U:  0.43257) (V:  0.3732)
```

Here Q+U of 1.2% prior move is -0.7 and nowhere close to being picked as the current max(Q+U) is 0.6. So somehow AZ is picking these very low prior moves.

This behavior of AZ being able to search wide assuming the behavior exists in self-play would seem to fix #8 for finding tactics as often times the network would have known a sacrifice tactic is actually a good position with a single visit, but lc0's search ends up ignoring low prior moves.

Mardak:
r?@mooskagh ~Wasn't sure if this should be conditional on self-play / noise. Or if it needs another param -- without, the logic simplifies a little bit.~ ~Edit: Added `fpu-strategy=root-absolute` to do absolute at root and reduction otherwise. Self-play maintains current previous default `fpu-strategy=reduction` and still sets `fpu-reduction=0`, so that's unaffected by the code simplification of not checking noise (but would prevent wanting to use reduction with noise except at root -- do we want to keep that?). To turn on new behavior for training, server would set `fpu-strategy=root-absolute` (with `fpu-value=1 fpu-reduction=0` still set on the client). Added `--early-root-widening` flag which changes FPU behavior at root only for search (leaving verbose, temperature, other `GetFPU` callers unaffected).~

Edit: Added the suggested FpuStrategyAtRoot and FpuValueAtRoot; and at the same time converted FpuReduction to FpuValue. I made "same" ignore FpuValueAtRoot and documented help as such to avoid complexities of thinking "same" might mean.. "if strategy is reduction, reduce eval by --fpu-value for most nodes except reduce root children by --fpu-value-at-root instead."

Analyzing the position from page 81 (fix #748) again with 32930 with `minibatch-size=1 smart-pruning-factor=0` (expecting to explore more than top 4):
page81
```
position startpos moves g1f3 g8f6 c2c4 e7e6 b1c3 f8b4 d1c2 e8g8 a2a3 b4c3 c2c3 a7a5 b2b4 d7d6 e2e3 f6e4 c3c2 e4g5 b4b5 g5f3 g2f3 d8f6 d2d4 f6f3 h1g1 b8d7 f1e2 f3f6 c1b2 f6h4 g1g4 h4h2 g4g3 f7f5 e1c1 f8f7 e2f3 h2h4 d1h1 h4f6 c1b1 g7g6 g3g1 a5a4 b1a1 f7g7 e3e4 f5f4 c4c5 f6e7 g1c1 d7f6 e4e5 d6e5 h1e1 e5e4 f3e4 e7f8
go nodes 68

before:
…
c2c4  (264 ) N:   0 (+ 0) (P:  3.59%) (Q: -0.66368) (U: 0.88393) (Q+U:  0.22025) (V:  -.----) 
e4d3  (783 ) N:   9 (+ 0) (P: 16.80%) (Q:  0.01363) (U: 0.41335) (Q+U:  0.42698) (V:  0.1071) 
d4d5  (761 ) N:  10 (+ 0) (P: 12.22%) (Q:  0.09005) (U: 0.27343) (Q+U:  0.36348) (V:  0.6982) 
e4f3  (785 ) N:  22 (+ 0) (P: 11.60%) (Q:  0.29520) (U: 0.12411) (Q+U:  0.41931) (V:  0.2686) 
c5c6  (973 ) N:  26 (+ 0) (P: 13.87%) (Q:  0.28326) (U: 0.12639) (Q+U:  0.40965) (V:  0.3732) 

after:
c2b3  (258 ) N:   1 (+ 0) (P:  0.59%) (Q: -0.97691) (U: 0.07363) (Q+U: -0.90328) (V: -0.9769) 
e4d5  (795 ) N:   1 (+ 0) (P:  0.83%) (Q: -0.98659) (U: 0.10328) (Q+U: -0.88331) (V: -0.9866) 
e4f5  (797 ) N:   1 (+ 0) (P:  0.86%) (Q: -0.96944) (U: 0.10715) (Q+U: -0.86229) (V: -0.9694) 
e1e3  (111 ) N:   1 (+ 0) (P:  0.84%) (Q: -0.92453) (U: 0.10422) (Q+U: -0.82031) (V: -0.9245) 
c2a4  (262 ) N:   1 (+ 0) (P:  0.55%) (Q: -0.85741) (U: 0.06828) (Q+U: -0.78913) (V: -0.8574) 
e4c6  (799 ) N:   1 (+ 0) (P:  0.71%) (Q: -0.71045) (U: 0.08829) (Q+U: -0.62217) (V: -0.7105) 
e4g6  (803 ) N:   1 (+ 0) (P:  0.72%) (Q: -0.47933) (U: 0.08895) (Q+U: -0.39038) (V: -0.4793) 
a1b1  (0   ) N:   1 (+ 0) (P:  1.11%) (Q: -0.00074) (U: 0.13756) (Q+U:  0.13682) (V: -0.0007) 
b2c3  (231 ) N:   1 (+ 0) (P:  1.29%) (Q: -0.02199) (U: 0.15988) (Q+U:  0.13789) (V: -0.0220) 
a1a2  (7   ) N:   1 (+ 0) (P:  1.12%) (Q:  0.03331) (U: 0.13893) (Q+U:  0.17224) (V:  0.0333) 
c1b1  (48  ) N:   1 (+ 0) (P:  1.23%) (Q:  0.07281) (U: 0.15246) (Q+U:  0.22526) (V:  0.0728) 
e1f1  (101 ) N:   1 (+ 0) (P:  1.27%) (Q:  0.11431) (U: 0.15737) (Q+U:  0.27169) (V:  0.1143) 
b5b6  (941 ) N:   1 (+ 0) (P:  1.64%) (Q:  0.10347) (U: 0.20353) (Q+U:  0.30699) (V:  0.1035) 
e1d1  (100 ) N:   1 (+ 0) (P:  1.41%) (Q:  0.17411) (U: 0.17497) (Q+U:  0.34907) (V:  0.1741) 
c2c3  (259 ) N:   1 (+ 0) (P:  1.44%) (Q:  0.19155) (U: 0.17794) (Q+U:  0.36950) (V:  0.1916) 
c2d2  (252 ) N:   1 (+ 0) (P:  1.80%) (Q:  0.16386) (U: 0.22263) (Q+U:  0.38649) (V:  0.1639) 
c2b1  (246 ) N:   1 (+ 0) (P:  1.63%) (Q:  0.19083) (U: 0.20183) (Q+U:  0.39265) (V:  0.1908) 
c2d1  (248 ) N:   1 (+ 0) (P:  1.62%) (Q:  0.22839) (U: 0.20069) (Q+U:  0.42908) (V:  0.2284) 
e1e2  (106 ) N:   1 (+ 0) (P:  1.55%) (Q:  0.25640) (U: 0.19218) (Q+U:  0.44858) (V:  0.2564) 
c2d3  (260 ) N:   1 (+ 0) (P:  1.75%) (Q:  0.24711) (U: 0.21667) (Q+U:  0.46379) (V:  0.2471) 
e4b7  (804 ) N:   1 (+ 0) (P:  0.71%) (Q:  0.38252) (U: 0.08786) (Q+U:  0.47038) (V:  0.3825) 
c1d1  (49  ) N:   1 (+ 0) (P:  1.54%) (Q:  0.28285) (U: 0.19137) (Q+U:  0.47423) (V:  0.2829) 
e1h1  (103 ) N:   1 (+ 0) (P:  1.97%) (Q:  0.24280) (U: 0.24410) (Q+U:  0.48690) (V:  0.2428) 
e1g1  (102 ) N:   1 (+ 0) (P:  2.36%) (Q:  0.21439) (U: 0.29309) (Q+U:  0.50748) (V:  0.2144) 
f2f3  (346 ) N:   1 (+ 0) (P:  2.47%) (Q:  0.22276) (U: 0.30605) (Q+U:  0.52881) (V:  0.2228) 
e4g2  (781 ) N:   1 (+ 0) (P:  2.99%) (Q:  0.16545) (U: 0.37093) (Q+U:  0.53638) (V:  0.1654) 
c2e2  (253 ) N:   2 (+ 0) (P:  2.34%) (Q:  0.19584) (U: 0.19319) (Q+U:  0.38903) (V:  0.2921) 
e4h1  (776 ) N:   2 (+ 0) (P:  3.57%) (Q:  0.11043) (U: 0.29533) (Q+U:  0.40576) (V:  0.1988) 
c2c4  (264 ) N:   2 (+ 0) (P:  3.59%) (Q:  0.18332) (U: 0.29684) (Q+U:  0.48016) (V:  0.2804) 
e4f3  (785 ) N:   5 (+ 0) (P: 11.60%) (Q:  0.05569) (U: 0.47931) (Q+U:  0.53500) (V:  0.2686) 
d4d5  (761 ) N:   6 (+ 0) (P: 12.22%) (Q:  0.07366) (U: 0.43289) (Q+U:  0.50655) (V:  0.6982) 
e4d3  (783 ) N:   8 (+ 0) (P: 16.80%) (Q:  0.03414) (U: 0.46271) (Q+U:  0.49684) (V:  0.1071) 
c5c6  (973 ) N:  17 (+ 0) (P: 13.87%) (Q:  0.35679) (U: 0.19100) (Q+U:  0.54779) (V:  0.3732)
```

This patch also happens to avoid @sergiovieri’s stalemate capture to queen promotion with t40 (41000 used here with `policy-softmax-temp=1`; expecting to find non-stalemate move):
stalemate
```
position fen 5n2/4P2k/8/8/8/2K3Q1/8/8 w - - 0 1
go nodes 800

before:
…
e7f8r (1832) N:   0 (+ 0) (P:  0.05%) (Q: -1.19817) (U: 0.04482) (Q+U: -1.15335) (V:  -.----) 
e7f8q (1831) N: 799 (+ 0) (P: 99.90%) (Q:  0.00000) (U: 0.10854) (Q+U:  0.10854) (V:  0.0000) (T) 

after:
…
e7f8q (1831) N:  86 (+ 0) (P: 99.90%) (Q:  0.00000) (U: 0.99810) (Q+U:  0.99810) (V:  0.0000) (T) 
g3g1  (600 ) N: 115 (+ 1) (P:  0.00%) (Q:  0.99785) (U: 0.00000) (Q+U:  0.99785) (V:  0.9981) 
e7f8r (1832) N: 134 (+ 0) (P:  0.05%) (Q:  0.99622) (U: 0.00033) (Q+U:  0.99656) (V:  0.9994) 
e7e8q (1828) N: 244 (+ 0) (P:  0.00%) (Q:  0.99420) (U: 0.00000) (Q+U:  0.99420) (V:  0.9985) 
```

Bonus 32930 trying to find this “houdini tactic” from tactics list (fix #8; expecting Qxe7 sacrifice):
tactic
```
position startpos moves d2d4 e7e6 c2c4 f8b4 c1d2 b4e7 e2e4 d7d5 e4e5 c7c5 d1g4 e7f8 d4c5 h7h5 g4g3 h5h4 g3a3 b8d7 g1f3 f8c5 b2b4 c5b6 d2g5 g8e7 a3b2 h8h5 c4d5 e6d5 f1b5 e8f8 e1g1 d7e5 b2e5 f7f6 e5f4 b6c7 f4e3 f6g5 b1c3 d8d6 b5d3 c7b6 e3e2 h4h3 f1e1 g5g4 f3e5 h5g5 e5g6 g5g6 d3g6 c8d7 g6h5 a8c8 a1c1 c8c4
go nodes 800

before:
…
e2e7  (330 ) N:   0 (+ 0) (P:  0.81%) (Q: -1.44808) (U: 0.70426) (Q+U: -0.74382) (V:  -.----) 
…
c3a4  (483 ) N:  21 (+ 0) (P:  8.25%) (Q: -0.58938) (U: 0.32603) (Q+U: -0.26336) (V: -0.4660) 
e1d1  (100 ) N:  22 (+ 0) (P:  2.83%) (Q: -0.38214) (U: 0.10682) (Q+U: -0.27532) (V: -0.5075) 
g2h3  (375 ) N:  98 (+ 0) (P: 13.42%) (Q: -0.38267) (U: 0.11784) (Q+U: -0.26484) (V: -0.1010) 
e2d2  (311 ) N: 540 (+ 0) (P:  3.21%) (Q: -0.26595) (U: 0.00516) (Q+U: -0.26079) (V: -0.5581) 

after:
…
c3a4  (483 ) N:  16 (+ 0) (P:  8.25%) (Q: -0.46731) (U: 0.42192) (Q+U: -0.04539) (V: -0.4660) 
e1d1  (100 ) N:  21 (+ 0) (P:  2.83%) (Q: -0.24851) (U: 0.11168) (Q+U: -0.13683) (V: -0.5075) 
g2h3  (375 ) N:  38 (+ 0) (P: 13.42%) (Q: -0.37856) (U: 0.29913) (Q+U: -0.07944) (V: -0.1010) 
e2d2  (311 ) N: 200 (+ 0) (P:  3.21%) (Q: -0.18384) (U: 0.01390) (Q+U: -0.16994) (V: -0.5581) 
e2e7  (330 ) N: 437 (+ 1) (P:  0.81%) (Q:  0.76966) (U: 0.00160) (Q+U:  0.77127) (V: -0.2049) 
```

### DIFF
--- a/src/mcts/params.cc
+++ b/src/mcts/params.cc
@@ -109,24 +109,27 @@ const OptionId SearchParams::kSmartPruningFactorId{
     "pruning is deactivated."};
 const OptionId SearchParams::kFpuStrategyId{
     "fpu-strategy", "FpuStrategy",
-    "How is an eval of unvisited node determined. \"reduction\" subtracts "
-    "--fpu-reduction value from the parent eval. \"absolute\" sets eval of "
-    "unvisited nodes to the value specified in --fpu-value."};
-// TODO(crem) Make FPU in "reduction" mode use fpu-value too. For now it's kept
-// for backwards compatibility.
-const OptionId SearchParams::kFpuReductionId{
-    "fpu-reduction", "FpuReduction",
-    "\"First Play Urgency\" reduction (used when FPU strategy is "
-    "\"reduction\"). Normally when a move has no visits, "
-    "it's eval is assumed to be equal to parent's eval. With non-zero FPU "
-    "reduction, eval of unvisited move is decreased by that value, "
-    "discouraging visits of unvisited moves, and saving those visits for "
-    "(hopefully) more promising moves."};
+    "How is an eval of unvisited node determined. \"First Play Urgency\" "
+    "changes search behavior to visit unvisited nodes earlier or later by "
+    "using a placeholder eval before checking the network. The value specified "
+    "with --fpu-value results in \"reduction\" subtracting that value from the "
+    "parent eval while \"absolute\" directly uses that value."};
 const OptionId SearchParams::kFpuValueId{
     "fpu-value", "FpuValue",
-    "\"First Play Urgency\" value. When FPU strategy is \"absolute\", value of "
-    "unvisited node is assumed to be equal to this value, and does not depend "
-    "on parent eval."};
+    "\"First Play Urgency\" value used to adjust unvisited node eval based on "
+    "--fpu-strategy."};
+const OptionId SearchParams::kFpuStrategyAtRootId{
+    "fpu-strategy-at-root", "FpuStrategyAtRoot",
+    "How is an eval of unvisited root children determined. Just like "
+    "--fpu-strategy except only at the root level and adjusts unvisited root "
+    "children eval with --fpu-value-at-root. In addition to matching the "
+    "strategies from --fpu-strategy, this can be \"same\" to disable the "
+    "special root behavior."};
+const OptionId SearchParams::kFpuValueAtRootId{
+    "fpu-value-at-root", "FpuValueAtRoot",
+    "\"First Play Urgency\" value used to adjust unvisited root children eval "
+    "based on --fpu-strategy-at-root. Has no effect if --fpu-strategy-at-root "
+    "is \"same\"."};
 const OptionId SearchParams::kCacheHistoryLengthId{
     "cache-history-length", "CacheHistoryLength",
     "Length of history, in half-moves, to include into the cache key. When "
@@ -197,8 +200,10 @@ void SearchParams::Populate(OptionsParser* options) {
   options->Add<FloatOption>(kSmartPruningFactorId, 0.0f, 10.0f) = 1.33f;
   std::vector<std::string> fpu_strategy = {"reduction", "absolute"};
   options->Add<ChoiceOption>(kFpuStrategyId, fpu_strategy) = "reduction";
-  options->Add<FloatOption>(kFpuReductionId, -100.0f, 100.0f) = 1.2f;
-  options->Add<FloatOption>(kFpuValueId, -1.0f, 1.0f) = -1.0f;
+  options->Add<FloatOption>(kFpuValueId, -100.0f, 100.0f) = 1.2f;
+  fpu_strategy.push_back("same");
+  options->Add<ChoiceOption>(kFpuStrategyAtRootId, fpu_strategy) = "absolute";
+  options->Add<FloatOption>(kFpuValueAtRootId, -100.0f, 100.0f) = 1.0f;
   options->Add<IntOption>(kCacheHistoryLengthId, 0, 7) = 0;
   options->Add<FloatOption>(kPolicySoftmaxTempId, 0.1f, 10.0f) = 2.2f;
   options->Add<IntOption>(kMaxCollisionEventsId, 1, 1024) = 32;
@@ -225,8 +230,12 @@ SearchParams::SearchParams(const OptionsDict& options)
       kSmartPruningFactor(options.Get<float>(kSmartPruningFactorId.GetId())),
       kFpuAbsolute(options.Get<std::string>(kFpuStrategyId.GetId()) ==
                    "absolute"),
-      kFpuReduction(options.Get<float>(kFpuReductionId.GetId())),
       kFpuValue(options.Get<float>(kFpuValueId.GetId())),
+      kFpuAbsoluteAtRoot(
+          options.Get<std::string>(kFpuStrategyAtRootId.GetId()) == "absolute"),
+      kFpuReductionAtRoot(options.Get<std::string>(
+                              kFpuStrategyAtRootId.GetId()) == "reduction"),
+      kFpuValueAtRoot(options.Get<float>(kFpuValueAtRootId.GetId())),
       kCacheHistoryLength(options.Get<int>(kCacheHistoryLengthId.GetId())),
       kPolicySoftmaxTemp(options.Get<float>(kPolicySoftmaxTempId.GetId())),
       kMaxCollisionEvents(options.Get<int>(kMaxCollisionEventsId.GetId())),
--- a/src/mcts/params.cc
+++ b/src/mcts/params.cc
@@ -202,7 +202,7 @@ void SearchParams::Populate(OptionsParser* options) {
   options->Add<ChoiceOption>(kFpuStrategyId, fpu_strategy) = "reduction";
   options->Add<FloatOption>(kFpuValueId, -100.0f, 100.0f) = 1.2f;
   fpu_strategy.push_back("same");
-  options->Add<ChoiceOption>(kFpuStrategyAtRootId, fpu_strategy) = "absolute";
+  options->Add<ChoiceOption>(kFpuStrategyAtRootId, fpu_strategy) = "same";
   options->Add<FloatOption>(kFpuValueAtRootId, -100.0f, 100.0f) = 1.0f;
   options->Add<IntOption>(kCacheHistoryLengthId, 0, 7) = 0;
   options->Add<FloatOption>(kPolicySoftmaxTempId, 0.1f, 10.0f) = 2.2f;
@@ -232,10 +232,13 @@ SearchParams::SearchParams(const OptionsDict& options)
                    "absolute"),
       kFpuValue(options.Get<float>(kFpuValueId.GetId())),
       kFpuAbsoluteAtRoot(
+          (options.Get<std::string>(kFpuStrategyAtRootId.GetId()) == "same" &&
+           kFpuAbsolute) ||
           options.Get<std::string>(kFpuStrategyAtRootId.GetId()) == "absolute"),
-      kFpuReductionAtRoot(options.Get<std::string>(
-                              kFpuStrategyAtRootId.GetId()) == "reduction"),
-      kFpuValueAtRoot(options.Get<float>(kFpuValueAtRootId.GetId())),
+      kFpuValueAtRoot(options.Get<std::string>(kFpuStrategyAtRootId.GetId()) ==
+                              "same"
+                          ? kFpuValue
+                          : options.Get<float>(kFpuValueAtRootId.GetId())),
       kCacheHistoryLength(options.Get<int>(kCacheHistoryLengthId.GetId())),
       kPolicySoftmaxTemp(options.Get<float>(kPolicySoftmaxTempId.GetId())),
       kMaxCollisionEvents(options.Get<int>(kMaxCollisionEventsId.GetId())),
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -189,13 +189,17 @@ int64_t Search::GetTimeToDeadline() const {
 
 namespace {
 inline float GetFpu(const SearchParams& params, Node* node, bool is_root_node) {
+  // Use root FPU behavior unless it's "same"
+  if (is_root_node) {
+    if (params.GetFpuAbsoluteAtRoot()) return params.GetFpuValueAtRoot();
+    if (params.GetFpuReductionAtRoot())
+      return -node->GetQ() -
+             params.GetFpuValueAtRoot() * std::sqrt(node->GetVisitedPolicy());
+  }
   return params.GetFpuAbsolute()
              ? params.GetFpuValue()
-             : ((is_root_node && params.GetNoise()) ||
-                !params.GetFpuReduction())
-                   ? -node->GetQ()
-                   : -node->GetQ() - params.GetFpuReduction() *
-                                         std::sqrt(node->GetVisitedPolicy());
+             : -node->GetQ() -
+                   params.GetFpuValue() * std::sqrt(node->GetVisitedPolicy());
 }
 
 inline float ComputeCpuct(const SearchParams& params, uint32_t N) {
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -189,17 +189,10 @@ int64_t Search::GetTimeToDeadline() const {
 
 namespace {
 inline float GetFpu(const SearchParams& params, Node* node, bool is_root_node) {
-  // Use root FPU behavior unless it's "same"
-  if (is_root_node) {
-    if (params.GetFpuAbsoluteAtRoot()) return params.GetFpuValueAtRoot();
-    if (params.GetFpuReductionAtRoot())
-      return -node->GetQ() -
-             params.GetFpuValueAtRoot() * std::sqrt(node->GetVisitedPolicy());
-  }
-  return params.GetFpuAbsolute()
-             ? params.GetFpuValue()
-             : -node->GetQ() -
-                   params.GetFpuValue() * std::sqrt(node->GetVisitedPolicy());
+  const auto value = params.GetFpuValue(is_root_node);
+  return params.GetFpuAbsolute(is_root_node)
+             ? value
+             : -node->GetQ() - value * std::sqrt(node->GetVisitedPolicy());
 }
 
 inline float ComputeCpuct(const SearchParams& params, uint32_t N) {
--- a/src/mcts/params.h
+++ b/src/mcts/params.h
@@ -78,11 +78,8 @@ class SearchParams {
     return options_.Get<bool>(kLogLiveStatsId.GetId());
   }
   float GetSmartPruningFactor() const { return kSmartPruningFactor; }
-  bool GetFpuAbsolute() const { return kFpuAbsolute; }
-  float GetFpuValue() const { return kFpuValue; }
-  bool GetFpuAbsoluteAtRoot() const { return kFpuAbsoluteAtRoot; }
-  bool GetFpuReductionAtRoot() const { return kFpuReductionAtRoot; }
-  float GetFpuValueAtRoot() const { return kFpuValueAtRoot; }
+  bool GetFpuAbsolute(bool at_root) const { return at_root ? kFpuAbsoluteAtRoot : kFpuAbsolute; }
+  float GetFpuValue(bool at_root) const { return at_root ? kFpuValueAtRoot : kFpuValue; }
   int GetCacheHistoryLength() const { return kCacheHistoryLength; }
   float GetPolicySoftmaxTemp() const { return kPolicySoftmaxTemp; }
   int GetMaxCollisionEvents() const { return kMaxCollisionEvents; }
@@ -149,7 +146,6 @@ class SearchParams {
   const bool kFpuAbsolute;
   const float kFpuValue;
   const bool kFpuAbsoluteAtRoot;
-  const bool kFpuReductionAtRoot;
   const float kFpuValueAtRoot;
   const int kCacheHistoryLength;
   const float kPolicySoftmaxTemp;
--- a/src/mcts/params.h
+++ b/src/mcts/params.h
@@ -79,8 +79,10 @@ class SearchParams {
   }
   float GetSmartPruningFactor() const { return kSmartPruningFactor; }
   bool GetFpuAbsolute() const { return kFpuAbsolute; }
-  float GetFpuReduction() const { return kFpuReduction; }
   float GetFpuValue() const { return kFpuValue; }
+  bool GetFpuAbsoluteAtRoot() const { return kFpuAbsoluteAtRoot; }
+  bool GetFpuReductionAtRoot() const { return kFpuReductionAtRoot; }
+  float GetFpuValueAtRoot() const { return kFpuValueAtRoot; }
   int GetCacheHistoryLength() const { return kCacheHistoryLength; }
   float GetPolicySoftmaxTemp() const { return kPolicySoftmaxTemp; }
   int GetMaxCollisionEvents() const { return kMaxCollisionEvents; }
@@ -116,8 +118,9 @@ class SearchParams {
   static const OptionId kLogLiveStatsId;
   static const OptionId kSmartPruningFactorId;
   static const OptionId kFpuStrategyId;
-  static const OptionId kFpuReductionId;
   static const OptionId kFpuValueId;
+  static const OptionId kFpuStrategyAtRootId;
+  static const OptionId kFpuValueAtRootId;
   static const OptionId kCacheHistoryLengthId;
   static const OptionId kPolicySoftmaxTempId;
   static const OptionId kMaxCollisionEventsId;
@@ -144,8 +147,10 @@ class SearchParams {
   const bool kNoise;
   const float kSmartPruningFactor;
   const bool kFpuAbsolute;
-  const float kFpuReduction;
   const float kFpuValue;
+  const bool kFpuAbsoluteAtRoot;
+  const bool kFpuReductionAtRoot;
+  const float kFpuValueAtRoot;
   const int kCacheHistoryLength;
   const float kPolicySoftmaxTemp;
   const int kMaxCollisionEvents;
--- a/src/selfplay/tournament.cc
+++ b/src/selfplay/tournament.cc
@@ -99,8 +99,6 @@ void SelfPlayTournament::PopulateOptions(OptionsParser* options) {
   defaults->Set<float>(SearchParams::kTemperatureId.GetId(), 1.0f);
   defaults->Set<bool>(SearchParams::kNoiseId.GetId(), true);
   defaults->Set<float>(SearchParams::kFpuValueId.GetId(), 0.0f);
-  defaults->Set<std::string>(SearchParams::kFpuStrategyAtRootId.GetId(),
-                             "same");
   defaults->Set<std::string>(SearchParams::kHistoryFillId.GetId(), "no");
   defaults->Set<std::string>(NetworkFactory::kBackendId.GetId(),
                              "multiplexing");
--- a/src/selfplay/tournament.cc
+++ b/src/selfplay/tournament.cc
@@ -98,7 +98,9 @@ void SelfPlayTournament::PopulateOptions(OptionsParser* options) {
   defaults->Set<float>(SearchParams::kSmartPruningFactorId.GetId(), 0.0f);
   defaults->Set<float>(SearchParams::kTemperatureId.GetId(), 1.0f);
   defaults->Set<bool>(SearchParams::kNoiseId.GetId(), true);
-  defaults->Set<float>(SearchParams::kFpuReductionId.GetId(), 0.0f);
+  defaults->Set<float>(SearchParams::kFpuValueId.GetId(), 0.0f);
+  defaults->Set<std::string>(SearchParams::kFpuStrategyAtRootId.GetId(),
+                             "same");
   defaults->Set<std::string>(SearchParams::kHistoryFillId.GetId(), "no");
   defaults->Set<std::string>(NetworkFactory::kBackendId.GetId(),
                              "multiplexing");


mooskagh:
That's intended only for test, not for merge, right?

sergiovieri:
Could you elaborate a bit what you mean by "wider initially like a0"?

Mardak:
This could be for merge if it's less desirable to implement virtual loss and batching more like AlphaZero on TPUs. Root FPU=1 makes it more explicit matching Matthew's self-play 800 visit expectation to visit every root children rather than leaving it to chance of asynchronous parallelism.

@sergiovieri Matthew from DeepMind says `in very early stages we fill batches with nodes that we otherwise wouldn't have searched` so in the very first batch, it'll happen to pick the first 16 highest prior moves to evaluate as the move pending evaluation assumes it is bad/loss/checkmate, so the next pick/simulation will skip over some moves to fill up the batch. Depending on the hardware and latency and threading, even more moves can get picked that a sequential MCTS would not have.

Specifically for the stalemate situation, AZ would probably have looked at the top 16 prior moves, so the 99.90%, 0.05% and half of the 0.00% prior moves. With unvisited Q=loss and extremely close to 0% prior leading to U~=0, a sequential search would stick with the stalemate move because it thinks all other moves are losing; however, because AZ happens to include 15 other moves, it realizes Q=loss is wrong and Q=near-win. Root FPU=1 here makes all root children get visited allowing search to then realize every move other than the 99.90% prior is winning.

oscardssmith:
This also seems good for training. Getting at least 1 visit will partially prevent self reinforcing policy sharpening. (policy shaping tree, shaping policy)

alreadydone:
Instead of setting FPU=1, AZ might have done something similar to LZ Go, which is just to set the evaluation of expanding nodes (i.e. node whose evaluations/policy aren't ready yet) to a very low value to avoid selecting them (but instead select their siblings):
https://github.com/leela-zero/leela-zero/blob/master/src/UCTNode.cpp#L268

Mardak:
Yup, AZ's batching behavior is very similar to FPU=1 for Chess because their eager batching happens to grab most/all of the ~35 children whereas for Go, ~250 children probably aren't all visited.

Mardak:
I updated PR realizing we don't want to break t40 with a new client that didn't support the existing self-play.

Edited into original comment: Added `fpu-strategy=root-absolute` to do absolute at root and relative otherwise. Self-play maintains current previous default `fpu-strategy=relative` and still sets `fpu-reduction=0`, so that's unaffected by the code simplification of not checking noise (but would prevent wanting to use reduction with noise except at root -- do we want to keep that?). To turn on new behavior for training, server would set `fpu-strategy=root-absolute fpu-value=1` (with `fpu-reduction=0` still set on the client).

Mardak:
Ah, I guess we probably want to also support FPU=-1 for self-play similar to t40 except at root. So many configurations… ;)

Eddh:
(with `fpu-value=1 fpu-reduction=0` still set on the client). 
If we have fpu-value=1 at root why not activate fpu-reduction with optimal tuning for non-root nodes ?

patrickDBL:
Does this change gain significant Elo over the current search?

Mardak:
@mooskagh Updated approach to add `--early-root-widening` flag which changes FPU behavior at root only for search (leaving verbose, temperature, other `GetFPU` callers unaffected).

> Does this change gain significant Elo over the current search?

I ran selfplay tournament with 50105 with mostly training run 2 params with `--player2.no-early-root-widening`, and it did gain some Elo but not significantly.

```tournamentstatus P1: +301 -261 =266 Win: 52.42% Elo: 16.80 LOS: 95.42% P1-W: +162 -118 =135 P1-B: +139 -143 =131 ```

The training data is likely a bit improved, and here's some interesting games where the new flag allowed player1 to find very low prior moves that player2 didn't expect resulting in quick resigns. The training data would likely result in training these priors towards 80% instead of 0% that they had been. Yes, Dirichlet noise helps find these too, so the change here just makes it a bit more consistent.

As usual with training data, the below are just some games that end quickly with wild swings but the outcome isn't the only interesting result. Even for already won or already lost positions, the network learns from MCTS providing better search probabilities to generalize.

---

White thought it was winning (so black thought it was losing) and didn't expect black to move the rook for a discovery check to win a free rook. White resigns without another move.
https://lichess.org/irjPpfMD#60
```
position startpos moves g1f3 e7e6 d2d4 g8f6 c2c4 b7b6 g2g3 c8b7 f1g2 c7c5 d4d5 e6d5 f3h4 g7g6 b1c3 f8g7 e1g1 e8g8 c1g5 b8a6 c3d5 b7d5 g2d5 a8b8 d1d2 a6b4 d5f3 b6b5 g5f4 b5c4 f4b8 d8b8 a2a3 b4a6 a1c1 d7d5 f3d5 f8d8 e2e4 f6d5 e4d5 a6c7 c1c4 d8d5 d2c2 c7e6 b2b4 e6d4 c2d1 c5b4 a3b4 b8b5 c4c8 g7f8 f1e1 g8g7 g1g2 b5b7 c8c4

30… d5c5 (P:  0.93%) (Q: -0.27458)
```
irjppfmd

---

White thought it was slightly behind and definitely didn't expect a 0.02% prior rook sacrifice to win a queen. White resigns without another move.
https://lichess.org/4aD347R1#52
```
position startpos moves g1f3 g8f6 d2d4 e7e6 b2b3 f8e7 e2e3 d7d5 c1b2 c7c5 d4c5 e7c5 a2a3 e8g8 b1d2 a7a6 c2c4 c5e7 c4d5 f6d5 f1e2 b8c6 e1g1 e7f6 b2f6 d8f6 d1c1 d5c3 e2d3 f8d8 c1c2 h7h6 b3b4 c3d5 d3h7 g8h8 h7e4 c8d7 d2b3 c6e5 b3c5 d7b5 f1c1 a8c8 f3e5 f6e5 e4d5 d8d5 a3a4 b5c6 c2c3

26… d5d1 (P:  0.02%) (Q:  0.10799)
```
4ad347r1

---

Both sides were thinking the game was drawn in this equal-material QRPP endgame, but white searching wide finds the lowest prior 0.01% winning move where black resigns a couple moves later.
https://lichess.org/NcYepjgb#99
```
position startpos moves g1f3 e7e6 d2d4 f8e7 e2e3 g8f6 c2c4 e8g8 f1d3 d7d5 e1g1 c7c5 c4d5 e6d5 d4c5 e7c5 b1c3 b8c6 a2a3 a7a5 b2b3 d5d4 c3e4 f6e4 d3e4 d4e3 c1e3 c5e3 f2e3 d8e7 d1d3 h7h6 f3d4 c8d7 e4d5 c6d4 e3d4 d7c6 d5c6 b7c6 f1c1 e7f6 c1c4 f8e8 a1f1 f6d6 d3f3 d6a3 f3f7 g8h8 c4a4 a3d6 h2h3 e8f8 f7c4 f8f1 c4f1 a8f8 f1c4 d6g3 a4a1 f8f2 c4c6 g3e3 c6d5 f2f8 g1h1 e3d3 h1h2 d3d2 d5e5 f8f2 e5e8 h8h7 e8e4 h7g8 b3b4 d2b4 e4d5 g8h7 a1a5 b4b8 h2g1 b8f4 a5a1 f2d2 g1h1 d2d4 d5b3 d4b4 b3c2 h7g8 a1d1 b4b8 d1a1 f4d4 a1d1 d4b2

50 d1d8 (P:  0.01%) (Q:  0.01082)
```
ncyepjgb

---

White thought it was losing initially before searching wide to find this 0.09% rook sacrifice to win a queen. Black resigns without another move.
https://lichess.org/XP1bn6Pf#55
```
position startpos moves e2e4 e7e5 b1c3 g8f6 g1h3 f8c5 f1c4 e8g8 e1g1 c7c6 d2d3 b7b5 c4b3 a7a5 a2a3 d7d6 h3g5 c5b6 c3e2 h7h6 g5f3 f8e8 c2c4 b8a6 c4b5 c6b5 c1e3 b6e3 f2e3 a8a7 a1c1 d8b6 d3d4 f6e4 f3d2 e4d2 d1d2 b5b4 b3d5 c8e6 d5e6 f7e6 d2c2 a7c7 c2g6 e8e7 c1e1 b6c6 d4e5 d6e5 e1d1 c7c8 e2g3 c6c2

28 d1d8 (P:  0.09%) (Q: -0.11440)
```
xp1bn6pf

**Comment by Mardak** [comment]
To be clear, the positions in the previous comment, the resigning side played a blundering move because it didn't expect the tactic. The idea is that training improves the priors for the tactical move so that the network learns to avoid putting itself in a bad situation.

Overall behavior is that the network searches deep as best as it knows according to the network, and search wide at root to find the network's blindspots. Root is special because that's what is used to train future move probabilities.

**Comment by Mardak** [comment]
One potential open question is how early/urgent should visiting children be with this change? Right now I've set it to 1.0f but U for already visited winning positions would result in Q+U greater than that of very low prior unvisited moves. Making the value 2.0f would make it more likely but still not guaranteed, e.g., the stalemate capture promote to queen position has Q+U almost 4.

```
e7f8q (P: 99.90%) (Q:  0.99146) (D:  0.000) (U: 2.99718) (Q+U:  3.98864)
```

**Comment by Eddh** [comment]
The only thing I don't like about this is that it doesn't really transfer to other games like Go. In go exploring every move at root would be bad, a balance would need to be found, which would probably be some kind of FPU= running parent Value. But in Chess I don't see any issues.

**Comment by Mardak** [comment]
Indeed, visiting all for Go isn't the behavior we noticed in #748. It's just that as a simplification of the AZ's behavior of visiting 26 root children out of 33 for that position is pretty close to visiting all without trying to emulate the TPU behavior more precisely. I suppose a closer approximation would be to allow some number of root children to always be visited, and that could then be more directly reused for Go.. But then unclear if this behavior is desired for Go anyway…

**Comment by jkormu** [comment]
This pr makes the search wider in the root which could affect badly with endgame temp as any move with visits could be picked by temp. On the other hand, searching wider in the root should help finding good moves with low prior.
To test which one matters more, I ran a match with training like settings with endgame temp 0.45 between lc0 master and lc0 pr750.
The shared settings:
```
arg="--backend=cudnn-fp16" arg="--weights=$PWD/weights_run2_32930.pb.gz" arg='--smart-pruning-factor=0.0' arg='--threads=1' arg='--minibatch-size=32' arg='--cpuct=2.5' arg='--cpuct-factor=0.00' arg='--fpu-reduction=0.00' arg='--fpu-value=-1.00' arg='--policy-softmax-temp=1.00' arg='--max-collision-events=1' arg='--max-collision-visits=1' arg="--temperature=1.1" arg="--temp-endgame=0.45" arg="--temp-cutoff-move=16" arg="--temp-visit-offset=-0.25" arg="--noise"
```
In addition to these settings, master used `arg='--fpu-strategy=absolute'` and lc0_pr750 `arg='--fpu-strategy=root-absolute'`.
Results with 2 ply repeated opening book show versions being pretty much equal or pr750 marginally worse:

```
   # PLAYER                 :  RATING  ERROR  POINTS  PLAYED   (%)  CFS(%)    W     D    L  D(%)
   1 lc0_master_20190221    :       0      4  1707.5    3372  50.6      84  966  1483  923  44.0
   2 lc0_pr750_20190224     :      -4      4  1664.5    3372  49.4     ---  923  1483  966  44.0
```

Then to test if this is caused by endgame temp I ran a match with `--temp-endgame=0.0` for both engines and rest of the settings as above:

```
   # PLAYER                 :  RATING  ERROR  POINTS  PLAYED   (%)  CFS(%)    W     D    L  D(%)
   1 lc0_pr750_20190224     :       9      5  1280.5    2500  51.2      97  614  1333  553  53.3
   2 lc0_master_20190221    :       0      5  1219.5    2500  48.8     ---  553  1333  614  53.3
```

It indeed seems that pr750 might not play that well together with endgame temp but the effect is not big.

**Comment by Mardak** [comment]
Thanks for running the numbers with 32930. Just making sure, the latter numbers without endgame temperature for more match-like settings over 2500 games, a very mature network is able to gain 9 Elo when visiting root children earlier?

There's many ways to adjust temperature depending on how the first set of games were lost. If they were purely from 1-visit moves getting picked, then a lower temperature or more negative visit-offset could be used. If they're from playing down the bad lines due to network overvaluing positions that would have been hidden by search not visiting low priors, then it's actually a beneficial outcome for the network to lessen the value of those positions for future networks.

**Comment by jkormu** [comment]
Yes 9+/-4 elo. Also my next question was if the temp could be counter acted with visit offset -1 since presumably most of the root blunders get no more than one visit. I have a match running with offset -0.999 (For some reason -1 is not possible).

**Comment by oscardssmith** [comment]
@jkormu, are you sure you aren't testing one version against itself? 9 elo seems way too small

**Comment by jkormu** [comment]
@oscardssmith Double checked and it seems to be master vs pr750. What kind of Elo gains you are expecting with this pr in training like setting? Please note that in the above tests master and pr750 had identical setting (including all noise settings) with only difference in --fpu-strategy. The test did not state that dropping noise from 0.45 to 0.0 gains 9 Elo. In first match both used 0.45 endgame temp and in latter both used 0.0.

**Comment by oscardssmith** [comment]
It's not so much that I have specific expectations as I get suspicious when I see multiple tests within 10 elo, as I've been burned by this before.

**Comment by jkormu** [comment]
Results for  temp `--temp-visit-offset=-0.999` and `--temp-endgame=0.45` show that offset of -1 can be used to counteract the negative effects of wider root search together with temp: 
```
   # PLAYER                 :  RATING  ERROR  POINTS  PLAYED   (%)  CFS(%)     W     D     L  D(%)
   1 lc0_pr750_20190224     :      10      3  2465.5    4800  51.4     100  1360  2211  1229  46.1
   2 lc0_master_20190221    :       0      3  2334.5    4800  48.6     ---  1229  2211  1360  46.1
```

Test was with 800 nodes, repeated 2 move book and both engines shared following settings:
`arg="--backend=cudnn-fp16" arg="--weights=$PWD/weights_run2_32930.pb.gz" arg='--smart-pruning-factor=0.0' arg='--threads=1' arg='--minibatch-size=32' arg='--cpuct=2.5' arg='--cpuct-factor=0.00' arg='--fpu-reduction=0.00' arg='--fpu-value=-1.00' arg='--policy-softmax-temp=1.00' arg='--max-collision-events=1' arg='--max-collision-visits=1' arg="--temperature=1.1" arg="--temp-endgame=0.45" arg="--temp-cutoff-move=16" arg="--temp-visit-offset=-0.999" arg="--noise"`

The only setting that was different was `--fpu-strategy': pr750 used here `root-absolute` and master `absolute`.

**Comment by Mardak** [comment]
Here's another 95+ LOS/CFS result from a more recent t50 network (50180) but this time with accelerated selfplay settings (lower visits, higher resign, temperature cutoff [all games were different])
```
tournamentstatus P1: +281 -238 =369 Win: 52.42% Elo: 16.84 LOS: 97.05% P1-W: +153 -115 =175 P1-B: +128 -123 =194

--player2.no-early-root-widening --fpu-strategy=absolute --visits=1000 --cpuct=2.5 --resign-percentage=20 --temperature=1.2 --temp-endgame=0.45 --temp-cutoff-move=16 --temp-visit-offset=-.25 --temp-value-cutoff=1 --minimum-kldgain-per-node=0.000008 --resign-wdlstyle=true --no-share-trees
```

**Comment by Mardak** [comment]
@mooskagh Why is this marked "not for merge?" Are there other plans to implement a more AlphaZero-like search, which happens to improve training of tactics? It seems possible that lc0 has tried to work around tactical deficiency by additionally diverging from AlphaZero with 2.2 policy softmax temperature. So is there an explicit decision to not behave more like AlphaZero?

**Comment by mooskagh** [comment]
It didn't have that behaviour switchable off in params, that's why I decided it was just for testing.

File: src/mcts/params.cc [hunk 1]
```diff
@@ -104,6 +104,12 @@ const OptionId SearchParams::kSmartPruningFactorId{
     "promising moves from being considered even earlier. Values less than 1 "
     "causes hopeless moves to still have some attention. When set to 0, smart "
     "pruning is deactivated."};
+const OptionId SearchParams::kEarlyRootWideningId{
```
mooskagh: Maybe have pair of params `FpuStrategyAtRoot` (with values `absolute`, `relative` and `same` for same is not in root), and `FpuValueAtRoot`

Mardak:
@mooskagh Updated PR with your suggestion of FpuStrategyAtRoot and FpuValueAtRoot; and at the same time converted FpuReduction to FpuValue. I made "same" ignore FpuValueAtRoot and documented help as such to avoid complexities of thinking "same" might mean.. "if strategy is reduction, reduce eval by --fpu-value for most nodes except reduce root children by --fpu-value-at-root instead."

This maintains the default selfplay behavior of using fpu reduction 0, and for plain defaults, keeps the default fpu reduction 1.2 and adds fpu root absolute 1.

src/mcts/params.cc
Mardak: Updated PR as suggested

src/mcts/search.cc
```diff
@@ -189,13 +189,17 @@ int64_t Search::GetTimeToDeadline() const {
...
 namespace {
 inline float GetFpu(const SearchParams& params, Node* node, bool is_root_node) {
+  // Use root FPU behavior unless it's "same"
+  if (is_root_node) {
+    if (params.GetFpuAbsoluteAtRoot()) return params.GetFpuValueAtRoot();
+    if (params.GetFpuReductionAtRoot())
+      return -node->GetQ() -
+             params.GetFpuValueAtRoot() * std::sqrt(node->GetVisitedPolicy());
+  }
   return params.GetFpuAbsolute()
```
Mardak: Not sure if you wanted this kept as a single return statement (for inline?) -- if so, the logic could be:
```c
return is_root_node && (absoluteRoot || reductionRoot)
  ? absoluteRoot ? valueRoot : -Q - valueRoot*visited
  : absolute
    ? value
    : -Q - value*visited
```

Mardak: Alternatively, if we make the param getters be absoluteRoot and sameRoot (instead of adding reductionRoot):
```c
return is_root_node && !sameRoot
  ? absoluteRoot ? valueRoot : -Q - valueRoot*visited
  ? absolute ? value : -Q - value*visited
```

mooskagh: Maybe parse everything in `Params` class and make `GetFpuAbsolute` and `GetFpuValue` have parameter there, and leave here something like:
```cpp
  const auto value = params.GetFpuValue(is_root_node);
  return params.GetFpuAbsolute(is_root_node) ? value : (-Q - value * std::sqrt(node->GetVisitedPolicy()));
```

Mardak: Interesting. I believe this would make it easier for "same" to allow differing values at root. Should that be allowed? Or keep the PR's behavior of "same" == don't treat root differently?

```
strategy = reduction
value = 1.2
strategy-at-root = same (i.e., reduction)
value-at-root = 0

or

strategy = absolute
value = -1
strategy-at-root = same (i.e., absolute)
value-at-root = 1
```

mooskagh: I was thinking that if value is `same` then `Params` initialized both root's strategy and value from ones of non-root.
Otherwise is user wants to specify `--fpu-value=1.23` and not have special behaviour to root, they willl also have to specify `--fpu-root-value=1.23`.

src/mcts/params.cc
```diff
@@ -197,8 +200,10 @@ void SearchParams::Populate(OptionsParser* options) {
   options->Add<FloatOption>(kSmartPruningFactorId, 0.0f, 10.0f) = 1.33f;
   std::vector<std::string> fpu_strategy = {"reduction", "absolute"};
   options->Add<ChoiceOption>(kFpuStrategyId, fpu_strategy) = "reduction";
-  options->Add<FloatOption>(kFpuReductionId, -100.0f, 100.0f) = 1.2f;
-  options->Add<FloatOption>(kFpuValueId, -1.0f, 1.0f) = -1.0f;
+  options->Add<FloatOption>(kFpuValueId, -100.0f, 100.0f) = 1.2f;
+  fpu_strategy.push_back("same");
+  options->Add<ChoiceOption>(kFpuStrategyAtRootId, fpu_strategy) = "absolute";
```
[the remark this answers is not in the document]
Mardak: Switched to "same" (and updated selfplay to not need to override)

src/mcts/search.cc
Mardak: Okay, I made the Params constructor check for "same" to initialize to the other values.

Mardak:
@Tilps The server should probably set `--fpu-value=-1` in addition to the existing `--fpu-strategy=absolute` before many people use a version with this.

Tilps:
--fpu-value=-1.0 added to training parameters.

ASilver:
@Mardak So are you saying that in your preliminary tests, FPUR at 0.5 in training conditions actually increases exploration _and_ strength?

PR #754, opened 2019-02-24
SendMovesStats on every info update.

killerducky:
Use this magic combination of UCI parameters to print out our raw values periodically in addition to the UCI standard ones. I'd like to use this mode for my lc0_analyzer.py script, I think others could find it useful too.

This seems intuitive enough to avoid yet another UCI param, because if they have VerboseStats on and a large MultiPv value they probably want to see as much information as possible.

### DIFF
--- a/src/mcts/params.cc
+++ b/src/mcts/params.cc
@@ -96,6 +96,9 @@ const OptionId SearchParams::kNoiseId{
 const OptionId SearchParams::kVerboseStatsId{
     "verbose-move-stats", "VerboseMoveStats",
     "Display Q, V, N, U and P values of every move candidate after each move."};
+const OptionId SearchParams::kLogLiveStatsId{
+    "log-live-stats", "LogLiveStats",
+    "Do VerboseMoveStats on every info update."};
 const OptionId SearchParams::kSmartPruningFactorId{
     "smart-pruning-factor", "SmartPruningFactor",
     "Do not spend time on the moves which cannot become bestmove given the "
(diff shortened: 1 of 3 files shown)


killerducky:
Here's what I plan to use this for:
https://github.com/killerducky/lc0_analyzer/wiki/Lc0LiveAnalyzer

oscardssmith:
I don't really like this as, at some point multi-pv will likely change how nodes at root are distributed

mooskagh:
What exactly it is for?
If we need to show `N` for separate moves, I think it's better to have a separate command line flag which switches `SendUciInfo()` to output `N` (and with some effor other `common_info`) per edge (like we do with `Q` already) rather than the same for all nodes.

I think it would also look better in chess GUIs.

But if we need also to output `P`, `U+V` etc, then it's harder, yes.

killerducky:
I need at least `P` and for each move to have it's own `N`. We could change `N` to be per child instead of total when MultiPv>1, it makes some sense but also is a bit magical/surprising to the user. But it's hard to output `P` unless we put `string P` at the very end of the line.

But I'm using this to understand better how Lc0 search works, so I like seeing all the other numbers too. Sounds like people don't like the magic combo of options, instead we can do a new hidden option.

`MoreVerboseStats`? Having trouble thinking of better names...

mooskagh:
`StatsWhileThinking` ?

killerducky:
`LiveStats` ?

mooskagh:
`LiveStats` is fine I think. Maybe `LogLiveStats`?

killerducky:
Changed code to `LogLiveStats` version.

PR #755, opened 2019-02-24
Add auto kld thresholding for use in time control situations.

Tilps:
Yay 4 more parameters to tune... heh.

The exponent and multiplier and interval ratio come from my attempt to do some baseline tuning by reversing what threshold is needed to get an average of x nodes.  

Using interval ratio instead of a constant interval should help the algorithm be more stable regardless of time control.  I doubt there is much need to tweak this parameter - could maybe make it a bit smaller I guess.
The multiplier and exponent are likely net specific - as a net gets smarter the multiplier probably drops, not sure if we should expect the exponent to vary much net to net.  I tuned these using latest T40.  I haven't validated these constants for large nodes per move (and the exponent is probably pretty sensitive), but they have been working pretty well as a fit so far in the range of 800 to 6400 nodes per move.

The main thing that needs tuning is the auto-kldgain-move-fraction - I vaguely think it probably should be something in the range of 1/slowmover to 0.5/slowmover - and maybe slow mover should go way up.

So if someone has time, a clop of move-fraction and slowmover together could be good.

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -314,12 +314,31 @@ NNCacheLock Search::GetCachedNNEval(Node* node) const {
 }
 
 void Search::UpdateKLDGain() {
-  if (params_.GetMinimumKLDGainPerNode() <= 0) return;
+  if (params_.GetMinimumKLDGainPerNode() <= 0 &&
+      params_.GetAutoKLDGainMoveFraction() <= 0)
+    return;
 
   SharedMutex::Lock nodes_lock(nodes_mutex_);
   Mutex::Lock lock(counters_mutex_);
-  if (total_playouts_ + initial_visits_ >=
-      prev_dist_visits_total_ + params_.GetKLDGainAverageInterval()) {
+  int interval = params_.GetKLDGainAverageInterval();
+  double threshold = params_.GetMinimumKLDGainPerNode();
+  // Automode overrides above.
+  if (params_.GetAutoKLDGainMoveFraction() > 0) {
+    if (remaining_playouts_ == std::numeric_limits<int>::max()) {
+      // Need a remaining playouts estimate before we can think about correct
+      // threshold.
+      return;
+    }
+    int64_t expected_total_playouts = total_playouts_ + remaining_playouts_;
+    double target_playouts =
+        expected_total_playouts * params_.GetAutoKLDGainMoveFraction();
+    interval = static_cast<int>(target_playouts *
+                                params_.GetAutoKLDGainIntervalRatio());
+    threshold = params_.GetAutoKLDGainMultiplier() *
+                std::pow(expected_total_playouts + initial_visits_,
+                         params_.GetAutoKLDGainExponent());
+  }
+  if (total_playouts_ + initial_visits_ >= prev_dist_visits_total_ + interval) {
     std::vector<uint32_t> new_visits;
     for (auto edge : root_node_->Edges()) {
       new_visits.push_back(edge.GetN());
@@ -327,19 +346,19 @@ void Search::UpdateKLDGain() {
     if (prev_dist_.size() != 0) {
       double sum1 = 0.0;
       double sum2 = 0.0;
-      for (int i = 0; i < new_visits.size(); i++) {
+      for (size_t i = 0; i < new_visits.size(); i++) {
         sum1 += prev_dist_[i];
         sum2 += new_visits[i];
       }
       double kldgain = 0.0;
-      for (int i = 0; i < new_visits.size(); i++) {
+      for (size_t i = 0; i < new_visits.size(); i++) {
         double o_p = prev_dist_[i] / sum1;
         double n_p = new_visits[i] / sum2;
         if (prev_dist_[i] != 0) {
           kldgain += o_p * log(o_p / n_p);
         }
       }
-      if (kldgain / (sum2 - sum1) < params_.GetMinimumKLDGainPerNode()) {
+      if (kldgain / (sum2 - sum1) < threshold) {
         kldgain_too_small_ = true;
       }
     }
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -335,7 +335,7 @@ void Search::UpdateKLDGain() {
     interval = static_cast<int>(target_playouts *
                                 params_.GetAutoKLDGainIntervalRatio());
     threshold = params_.GetAutoKLDGainMultiplier() *
-                std::pow(expected_total_playouts + initial_visits_,
+                std::pow(target_playouts + initial_visits_,
                          params_.GetAutoKLDGainExponent());
   }
   if (total_playouts_ + initial_visits_ >= prev_dist_visits_total_ + interval) {
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -331,7 +331,7 @@ void Search::UpdateKLDGain() {
     }
     int64_t expected_total_playouts = total_playouts_ + remaining_playouts_;
     double target_playouts =
-        expected_total_playouts * params_.GetAutoKLDGainMoveFraction();
+        expected_total_playouts * params_.GetAutoKLDGainMoveFraction() / limits_.slowmover;
     interval = static_cast<int>(target_playouts *
                                 params_.GetAutoKLDGainIntervalRatio());
     threshold = params_.GetAutoKLDGainMultiplier() *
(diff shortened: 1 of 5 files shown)


Tilps:
Added curve adjustments based on expanding to fit the 800 to 15000 nodes range.

Tilps:
Current fit seems okay having added a data point at 28k nodes.  Will run one last data point tomorrow somewhere around 100k nodes.  I don't have the patience to go any higher.

Tilps:
105k nodes fit seems close enough with the current formula.  Took longer than I expected to get 10k positions done - but done it is.

Tilps:
I've got some small time control, small sample size positive results for 1.3 slowmover and 0.7 AutoKLDGainMoveRatio.  So I think clop should be able to get some good results.

zz4032:
Did a tuning with Network ID 32930 (in selfplay) and both parameters as "LinearParameters" and bounds [0;3]:
```
Slowmover = 1.47 (default: 1.0)
AutoKLDGainMoveFraction = 1.760 (default: 0.0)
```
plot
slowmover
kld-gain-movefraction

Tilps:
Heh I did not expect such a high value for MoveFraction, greater than 1 was not expected at all...

I guess it must be mostly working on the early outs gaining banked up time, rather than the extra depth in specific moves.
I'll do some elo testing.

jjoshua2:
Smart pruning already was too low by default at 17k nodes so something that
simulates increasing that should gain elo at low nodes up to 100k or 200k
per move.

wakamex:
took 12k games (16s on 2060 vs. 2s SFdev), but finally got clop to converge to slowmover=1.00273 and AutoKLDGainMoveFraction=0.401881 in the interval [1,10] and [0.05,1]

seems pretty insensitive to the exact numbers, with only 3 elo over sm=1.35 frac=0.49
image

looking at the plot, even less sensitive to the fraction, though strongly prefers slowmover<1.5
image

will run a quick strength test vs. default and edit in here

Tilps:
I'm not having much luck playing against SF with either of these tunes or my own random guess.  Nothing statistically significant yet, but both of the tunes seemed possibly weaker.

Tilps:
Seems it might only ever be a win in selfplay and not against SF.  Happy if anyone finds evidence otherwise.

oscardssmith:
Is there a way to clop vs sf?

zz4032:
Test match with my own tuned settings:
```
   1 lc0          :     9.0   16.4    86.0     814      75.8
   2 lc0-tuned    :     0.0   ----     ---     814      75.8
```
Never had success with Slowmover tunings to be honest...

Tilps:
I tried slowmover 5 and move ratio 0.2 for fun - interestingly it seems not significantly weaker (maybe slightly stronger but well under error bars - only 200 games played so far vs SF)  in my testing so far despite such a high slowmover.
Given this I think maybe I need to pay closer attention to how the endgame interacts with slowmover and move ratio.  When time gets low the slowmover is no longer actually fully applied.  I think move ratio needs to correspondingly rise in such low time scenarios to avoid blundering since it having a lower fraction is predicated on the fact that slow mover is high.  I think that should be a straightforward elo positive relative to current logic.

wakamex:
re-ran clop properly on slowmover only (keeping fraction at 0.5) and ended up with... 1.3!

yes you can clop vs. sf, just set
```
opponents = [
    'cmd=E:\lc0\sf190208 option.Threads=4 tc=16'
]
```

jjoshua2:
Can someone test with smart pruning at 1.8? And do a clop of kld then I
guess?

oscardssmith:
@jjoshua2 kld might reduce the need for smart pruning, although not eliminate it.

Tilps:
Added a change which should help with endgame blundering due to slow mover not being applied to its full extent due to there not being enough remaining time.
New expected range for move ratio is something near 1.
I ran some testing with 1.0 and slowmover 5 and it seemed pretty much equal to defaults when playing against SF.  It might be easier to clop now, since the parameters aren't quite so intertwined as they were before.

Tilps:
Also merged master.

src/mcts/search.cc
```diff
@@ -317,32 +317,51 @@ NNCacheLock Search::GetCachedNNEval(Node* node) const {
...
-  if (params_.GetMinimumKLDGainPerNode() <= 0) return;
+  if (params_.GetMinimumKLDGainPerNode() <= 0 &&
+      params_.GetAutoKLDGainMoveFraction() <= 0)
+    return;
 
   SharedMutex::Lock nodes_lock(nodes_mutex_);
   Mutex::Lock lock(counters_mutex_);
-  if (total_playouts_ + initial_visits_ >=
-      prev_dist_visits_total_ + params_.GetKLDGainAverageInterval()) {
+  int interval = params_.GetKLDGainAverageInterval();
```
wakamex: so `KLDGainAverageInterval` is only used if `AutoKLDGainMoveFraction` is not >0, in which case `interval` gets overwritten by `target_playouts * AutoKLDGainIntervalRatio`... so clopping AverageInterval with MoveFraction > 0 would be meaningless?

wakamex: so non-auto mode is used for training, and auto mode is for tc? with no overlap between parameters? we should only be tuning auto parameters for tc then

Tilps: Auto mode is for TC and non-auto for training - that is correct.  Trying to use both will just use auto.

wakamex:
on new defaults, did a quick clop (3s, ~1knpm) on slowmover, fraction, and pruning, which gave surprisingly good convergence, though it's still drifting after 1800 games, it's now at sm=2.55 frac =1.20 prun=1.78

no real difference vs. SF with that clop at that tc after 200 games (cfs 
```
   # PLAYER                :  RATING  ERROR  POINTS  PLAYED   (%)  CFS(%)    W    D    L  D(%)
   1 sf190208              :     0.0   ----   286.0     400  71.5     100  239   94   67  23.5
   2 lc0pr755clop-41168    :  -163.9   47.0    57.5     200  28.8      55   35   45  120  22.5
   3 lc0pr755def-41168     :  -168.4   49.1    56.5     200  28.3     ---   32   49  119  24.5
```

rerunning clop with longer tc (16s, ~5knpm) and tighter limits excluding values where the current clop didn't win at all (sm 0.5-5 frac 0.5-2 prun 0.5-3)

jjoshua2:
Haha I told you 1.8 at these shorter tc! My guess is the current is good
for tcec time controls still though. Maybe 1.5 for in between.

wakamex:
proper clop gave me sm1.0 frac1.14 prun1.57 to little effect:

16s tc:
```   # PLAYER                :  RATING  ERROR  POINTS  PLAYED   (%)  CFS(%)    W    D    L  D(%)
   1 sf190208              :     0.0   ----    64.0     104  61.5      99   48   32   24  30.8
   2 lc0pr755clop-41168    :   -82.1   64.3    47.5     106  44.8      56   28   39   39  36.8
   3 lc0pr755def-41168     :   -86.7   65.0    46.5     106  43.9     ---   26   41   39  38.7```

8+0.2s tc shows does worse, so doesn't seem stable across tc's
```   # PLAYER                :  RATING  ERROR  POINTS  PLAYED   (%)  CFS(%)    W    D    L  D(%)
   1 lc0pr755def-41168     :    22.3   62.3    42.5      80  53.1      62   25   35   20  43.8
   2 lc0pr755clop-41168    :     9.2   62.1    40.0      78  51.3      61   27   26   25  33.3
   3 sf190208              :     0.0   ----    75.5     158  47.8     ---   45   61   52  38.6```

Tilps:
Huge error bars - but yeah it seems SF is good at exploiting issues which this change does not address at all - or something.

jjoshua2:
That might mean instead of focusing on spending more time in these
positions it might be necessary to decrease fpu reduction or increase
policy flattening. Probably some combination of all three, but using that
extra time to do more width instead of depth.

Naphthalin:
As kldgain mainly aims at the nodes in search, this is probably obsolete in the new smooth time manager by crem merged in v0.25.0; this PR can probably be closed.

Tilps:
I don't think this is obsolete because of the time manager in any way, kld instead of smart pruning are just different strategies that are hidden in the inner parts of whatever time manager we use.

However we tried plenty of times and got no where with kld being better than smart pruning alone, so it can be closed.

PR #756, opened 2019-02-24
Update position.h

D0g1nG1thub:


borg323:
Thanks for your suggestion, but this file hasn't been modified so far this year so it is too early to update the copyright.

PR #766, opened 2019-02-26
Update syzygy.cc

D0g1nG1thub:


borg323:
This file has an already updated copyright year. You are updating the original author's copyright, but this should only happen when we merge code from the upstream repository.

PR #771, opened 2019-02-28
Update TempVisitOffset help text and allowed values.

Mardak:
r?@mooskagh Cleans up the text and makes the range symmetrical.

@jkormu wanted to test with an offset of -1 but couldn't https://github.com/LeelaChessZero/lc0/pull/750#issuecomment-468019066, and #technical had some confusion about whether the offset should be negative or not.


Mardak:
Here's the search code using offset:
https://github.com/LeelaChessZero/lc0/blob/9d374646c527e5575179d131c992eb9b2ddc27dc/src/mcts/search.cc#L590-L605

jhorthos:
Mardak can you put a compiled version in Appveyor? (or is it already there? i can't remember what the link would be).

Tilps:
Behavior of lower_bound seems safe in the presence of equal values, so should be fine.

PR #772, opened 2019-02-28
Use the best move eval instead of highest eval for temperature cutoff.

Mardak:
r?@mooskagh The current implementation didn't seem like it supports AlphaZero's `by softmax sam- pling with a temperature of 10.0 among moves for which the value was no more than 1% away from the best move for the first 30 plies`

In particular, a random single visit with high eval towards the end of limited node search could prevent selecting moves that got more visits.

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -599,8 +599,10 @@ EdgeAndNode Search::GetBestChildWithTemperature(Node* parent,
             root_limit.end()) {
       continue;
     }
-    if (edge.GetN() + offset > max_n) max_n = edge.GetN() + offset;
-    if (edge.GetQ(fpu) > max_eval) max_eval = edge.GetQ(fpu);
+    if (edge.GetN() + offset > max_n) {
+      max_n = edge.GetN() + offset;
+      max_eval = edge.GetQ(fpu);
+    }
   }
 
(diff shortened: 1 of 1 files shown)


Mardak:
Looking back at #568 I suppose it does say "best move" although I would think MCTS' "best" is most visited instead of highest eval. ??

src/mcts/search.cc
```diff
@@ -595,8 +595,10 @@ EdgeAndNode Search::GetBestChildWithTemperature(Node* parent,
             root_limit.end()) {
       continue;
     }
-    if (edge.GetN() + offset > max_n) max_n = edge.GetN() + offset;
-    if (edge.GetQ(fpu) > max_eval) max_eval = edge.GetQ(fpu);
+    if (edge.GetN() + offset > max_n) {
+      max_n = edge.GetN() + offset;
+      if (edge.GetQ(fpu) > max_eval) max_eval = edge.GetQ(fpu);
```
Mardak: We probably don't want the conditional assignment here? If the best/most-visited move's Q drops, it might not be able to replace the max_eval.

mooskagh: Probably no need for condition indeed.

Mardak: Removed

PR #776, opened 2019-03-03
Track number of edges, nodes, etc.

killerducky:
Use them for RAM limit, and print them in verbose stats.

* Still logs the approximation of nodes, but uses an exact edges*edge_size + nodes*node_size for the limit.
* Included a few more stats as well for outputting along with verbose move stats as "info string"
* For now I didn't modify the "real" info. One candidate is to change NPS somehow, e.g. num_nodes_/time.
* Cache hits and prefetch into cache confuse things, especially prefetch isn't tracked to see if they are actually used later or not.
* Most of the code doing this already holds nodes_mutex_. But I did have to add two short sections of taking that lock.
* Overhead of counting tree size was ~20ms on my machine for ~2M node tree.

* I tested a mate in one position, old method stopped very early. This method correctly allows many more playouts since hitting terminals does not create more nodes in memory.

### DIFF
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -169,16 +169,16 @@ SearchLimits EngineController::PopulateSearchLimits(
                                               *params.movetime - move_overhead);
   }
   if (params.nodes) limits.visits = *params.nodes;
-  const int ram_limit = options_.Get<int>(kRamLimitMbId.GetId());
+  const int64_t ram_limit = options_.Get<int>(kRamLimitMbId.GetId());
   if (ram_limit) {
     const auto cache_size =
         options_.Get<int>(kNNCacheSizeId.GetId()) * kAvgCacheItemSize;
-    int64_t limit = (ram_limit * 1000000LL - cache_size) / kAvgNodeSize;
+    limits.ram_limit = ram_limit * 1000000LL - cache_size;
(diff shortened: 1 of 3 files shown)


killerducky:
Some more stats. It's a little slower than I hoped.
```
UpdateTreeSize took 301ms. edges:73263164 nodes:1807524
UpdateTreeSize took 282ms. edges:73278966 nodes:1807930
UpdateTreeSize took 303ms. edges:73282542 nodes:1808025
UpdateTreeSize took 306ms. edges:73290457 nodes:1808237
UpdateTreeSize took 292ms. edges:73293454 nodes:1808326

UpdateTreeSize took 2660ms. edges:658011775 nodes:16319913
UpdateTreeSize took 2632ms. edges:658019470 nodes:16320100
UpdateTreeSize took 2622ms. edges:658030740 nodes:16320402
UpdateTreeSize took 2647ms. edges:658037605 nodes:16320574
UpdateTreeSize took 2628ms. edges:658044866 nodes:16320768
```

For this test I set RamLimitMb, let it hit the limit, then run `go` again.

killerducky:
I plan to do a version that counts nodes/edges in the constructors/destructors. But probably not for awhile, closing this for now.

PR #778, opened 2019-03-04
Quick patch for negative gamma

Ttl:
Still wrong, but at least it's consistently wrong. Fixes at least some of the T40 problems.

PR #779, opened 2019-03-04
Fold batch norm to weights and biases

Ttl:
Handles negative and zero gammas. This is a proper fix for #778.


src/neural/cuda/network_cudnn.cc
```diff
@@ -207,22 +207,22 @@ class CudnnNetwork : public Network {
...
         has_se_ = true;
       }
-      processConvBlock(weights.residual[i].conv1, true, 3);
-      processConvBlock(weights.residual[i].conv2, true, 3);
+      weights.residual[i].conv1.FoldBN(3);
+      weights.residual[i].conv2.FoldBN(3);
     }
     if (conv_policy_) {
-      processConvBlock(weights.policy1, true, 1);
+      weights.policy1.FoldBN(3);
```
Tilps: Was the old code wrong here?  It still had 1 rather than 3...

src/neural/cuda/network_cudnn.cc
```diff
@@ -660,11 +660,11 @@ class CudnnNetwork : public Network {
...
 
-    // Compute reciprocal of std-dev from the variances (so that it can be
-    // just multiplied).
+    // Variance to gamma.
     std::vector<float>& stddev = block.bn_stddivs;
-    for (auto&& w : stddev) {
-      w = 1.0f / std::sqrt(w + epsilon);
+    for (auto i = 0; i < block.bn_stddivs.size(); i++) {
+      block.bn_gammas[i] *= 1.0f / std::sqrt(block.bn_stddivs[i] + epsilon);
+      block.bn_stddivs[i] = 1.0f;
```
Tilps: Is this function still used at all?  I don't think it is - and thus we can just delete it.
Especially since I am very confused about whether the foldBNLayer code path is correct in the presence of gamma folding above.

borg323:
Running `./lc0 -w net41306 --backend=check --backend-opts=mode=histo,cudnn,blas`
```
Absolute error histogram for a batch of 65
      |                                                                                         |
      |                 #                                                                       |
      |                 #                                                                       |
      |                 ##                                                                      |
 0.15 +                 ##                                                                      +
      |                ###                                                                      |
      |                ###                                                                      |
      |                ####                                                                     |
      |                ####                                                                     |
  0.1 +                ####                                                                     +
      |               #####                                                                     |
      |               ######                                                                    |
      |               ######                                                                    |
      |              #######                                                                    |
 0.05 +              ########                                                                   +
      |              ########                                                                   |
      |             ##########                                                                  |
      |            #############                                                                |
      |# #######################################################                                |
      +----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+
   -inf  -15  -14  -13  -12  -11  -10   -9   -8   -7   -6   -5   -4   -3   -2   -1    0    1  +inf 
```
In contrast, with `./lc0 -w net41306 --backend=check --backend-opts=mode=histo,cudnn,opencl`
```
Absolute error histogram for a batch of 78
      |                                                                                         |
  0.2 +                                        ##                                               +
      |                                        ##                                               |
      |                                        ##                                               |
      |                                        ##                                               |
      |                                        ##                                               |
 0.15 +                                        ###                                              +
      |                                        ###                                              |
      |                                        ###                                              |
      |                                       ####                                              |
      |                                       ####                                              |
  0.1 +                                       #####                                             +
      |                                       #####                                             |
      |                                       #####                                             |
      |                                       ######                                            |
      |                                       ######                                            |
 0.05 +                                      #######                                            +
      |                                      ########                                           |
      |                                      #########                                          |
      |                                     ############                                        |
      |                               ##################################################        |
      +----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+
   -inf  -15  -14  -13  -12  -11  -10   -9   -8   -7   -6   -5   -4   -3   -2   -1    0    1  +inf 
```
Unfortunately, there are large differences between opencl and cudnn also with old nets.
`./lc0 -w net11248 --backend=check --backend-opts=mode=histo,cudnn,opencl`
```
Absolute error histogram for a batch of 58
      |                                                                                         |
      |                                                           #                             |
      |                                                           #                             |
 0.25 +                                                           #                             +
      |                                                           #                             |
      |                                                           #                             |
      |                                                           #                             |
      |                                                           ##                            |
  0.2 +                                                          ###                            +
      |                                                          ###                            |
      |                                                          ###                            |
      |                                                          ###                            |
      |                                                          ###                            |
 0.15 +                                                          ###                            +
      |                                                          ####                           |
      |                                                          ####                           |
      |                                                          ####                           |
      |                                                          ####                           |
  0.1 +                                                          ####                           +
      |                                                          ####                           |
      |                                                          ####                           |
      |                                                          #####                          |
      |                                                          #####                          |
 0.05 +                                                          #####                          +
      |                                                          ######                         |
      |                                                         ########                        |
      |                                                         ##########                      |
      |                                                       ##########################        |
      +----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+
   -inf  -15  -14  -13  -12  -11  -10   -9   -8   -7   -6   -5   -4   -3   -2   -1    0    1  +inf 
```

src/neural/cuda/network_cudnn.cc
ankan-ban: Yes, the old code looks wrong as we create the convolution as 3x3 when conv_policy_ is enabled: https://github.com/LeelaChessZero/lc0/blob/master/src/neural/cuda/network_cudnn.cc#L325

Tilps: As discussed on discord, while it was 'wrong' there is no change in behavior between 1 and 3 in practice so long as the actual filter size is 3.

Ttl:
All backends except for Tensorflow should be working now. I disabled it at the moment since it wouldn't give correct output.

Mardak:
As a sanity check for mac/opencl, looks like this PR behaves the same way as #778 in my 7 piece value test:
screen shot 2019-03-05 at 7 50 46 am

gsobala:
I confirm that the OpenCL backend is working correctly on macos on both SE and non-SE nets when compared to blas using backend=check

src/neural/opencl/clsource/convolve1.opencl
```diff
@@ -107,11 +106,10 @@ __kernel void merge_bn(
...
   }
-  if (means) {
-    const float mean = vload_net_t(o, means);
-    const float scale_stddiv = vload_net_t(o, stddivs);
-    
-    sum = scale_stddiv * (sum - mean);
+  if (biases) {
+    const float bias = vload_net_t(o, biases);
+
+    sum = sum + bias;
```
Tilps: This changes whether relu is defined from being based on existance of means, to existance of biases.
As far as I can tell this has no effect since all convole1's in the net post past and present always have relu - but it could be worth someone double checking.

src/neural/cuda/network_cudnn.cc
```diff
@@ -201,28 +201,10 @@ class CudnnNetwork : public Network {
...
-      processConvBlock(weights.residual[i].conv2, true, 3);
-    }
-    if (conv_policy_) {
-      processConvBlock(weights.policy1, true, 1);
-      // weights.policy doesn't have batch norm with convolutional policy
-      // so no need to call processConvBlock for it.
-    } else {
-      processConvBlock(weights.policy, true, 1);
+    // 0. Check for SE.
+    if (weights.residual[0].has_se) {
```
Tilps: I assume no one will ever try a net with 0 residual layers...

oscardssmith:
Time for version bump now?

Issue #780, opened 2019-03-05
Entropy-guided exploration for better tactical capabilities


kk2796:
I originally posted this idea in [Google Forum](https://groups.google.com/d/msg/lczero/BWj7cOT0uSc/Jb8nDC5xAwAJ) and was encouraged to echo it here.   That thread lays out an intuitive case that using entropy to guide exploration can yield highly efficient search strategies that cannot be realized with "success likelihood" and "visit counts" alone.  For the skeptics, please feel free to peruse the original thread (or to see something a bit more concrete, skim down in this thread to ***COOLEST POINT OF ENTIRE POST IMO*** )

I'm not going to spend any more time arguing for this change; I think the strongest case for it arises from exploring how it could/should work.  On that note - I've taken a bit of time to explore the math of this idea, and I'd like to lay out a plausible implementation.  I do not claim this is an optimal implementation.

First, it's necessary to define entropy in the context of PUCT search.  This is the context where the search is in state "s" and must choose an action "a" using probabilities that align with the various U(s,a) values of all actions.  The action chosen will accord to a playout to the node associated with that action.  There is already a good amount of information that's calculated and pulled together to compute U(s,a) for action a:

- Q(s,a) - the average of all value-head evaluations from prior visits to the node at a and its descendants
- cPuct - multiplier to increase the impact of exploration 
- P(s,a) - the a-priori policy-head probability of action a
- N(s,b) - the number of times this state has been reached.
- N(s,a) - the number of times action a has already been chosen (always <= N(s,b)).

To this mix, I introduce the term E(s,a).  E(s,a) is the entropy associated with the subtree rooted at the node of action a.  The entropy of a subtree is the entropy of the distribution of probabilities of a playout from the root of that subtree reaching any particular leaf node (either an unvisited node or terminal "mate" node).  
An algorithm for computing E(s,a) for subtree T: 
1) Enumerate every possible leaf node that can be reached with a playout from the root node of T.
2) For each playout, computing the step-wise probability of each discrete action (that is, comparing magnitude of U(s',a') of each step to the sum of U(s',*) of all that step's sibling actions).  
3) compute the product of all the playout's step-wise probabilities (from root to leaf) to attain the final probability of the playout
4) repeat 2-3 for all possible/legal playouts enumerated in (1)
5) compute entropy from this final set of probabilities, using information-theory formula for entropy of N probabilities:
[...]

PR #791, opened 2019-03-09
Value head standard deviation exploration term.

ddobbelaere:
This adds a value head standard deviation exploration term for (hopefully) better tactical capabilities (as proposed in https://github.com/LeelaChessZero/lc0/issues/780#issuecomment-470843033).

The main idea is that the uncertainty term of the UCT search is adapted to (new term in bold):

U(s,a) = cPuct ⋅ P(s,a) ⋅ sqrt(N(s,b)) ⋅ 1/(1+N(s,a)) + **stddev-factor ⋅ S(s,a)**,

with S(s,a) the standard deviation of the raw net value Qs over all visited nodes in the subtree (including multivisits). The backup update step is very efficient (incremental formula even for multivisits) and uses a variant of https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Welford's_online_algorithm (extended to the multivisit case).

The goal of this PR is in the first place to investigate the effect on tactical capabilities (as a function of the still unknown stddev-factor). If you have other ideas to include the S(s,a) term in the equation, please share your ideas!

Possible follow-ups:

- Investigate performance for fixed node tests on a standard tactics test suite and compare with master (stddev-factor = 0). Note that the node count at which the best move is found is also an interesting metric (it might be that this PR more quickly finds the best move).
- CLOP the unknown stddev-factor.
- Investigate other ways to include S(s,a) in the U(s,a) term.
- Investigate match play performance for an optimal value of stddev-factor.

### DIFF
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -242,7 +242,11 @@ void Node::CancelScoreUpdate(int multivisit) {
 }
 
 void Node::FinalizeScoreUpdate(float v, float d, int multivisit) {
-  // Recompute Q.
+  // Update the sum of squared differences.
+  sum_squared_diff_q_ +=
+      (v - q_) * (v - q_) * multivisit * n_ / (n_ + multivisit);
+
+  // Recompute Q and D.
   q_ += multivisit * (v - q_) / (n_ + multivisit);
   d_ += multivisit * (d - d_) / (n_ + multivisit);
 
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -159,7 +159,7 @@ class Node {
   float GetQStdDev() const {
     // Note that we use an unbiased standard deviation estimator (sqrt(n_-1)
     // denominator).
-    return n_ > 0 ? sqrt(sum_squared_diff_q_ / (n_ - 1)) : 0.0f;
+    return n_ > 1 ? sqrt(sum_squared_diff_q_ / (n_ - 1)) : 0.0f;
   }
 
   // Returns whether the node is known to be draw/lose/win.
@@ -364,11 +364,11 @@ class EdgeAndNode {
     return edge_ ? edge_->GetMove(flip) : Move();
   }
 
-  // Returns U = numerator * p / N + stddev_multiplier * GetQStdDev().
+  // Returns U = numerator * p / N + stddev_factor * GetQStdDev().
   // Passed numerator is expected to be equal to (cpuct * sqrt(N[parent])).
-  float GetU(float numerator, float stddev_multiplier) const {
+  float GetU(float numerator, float stddev_factor) const {
     return numerator * GetP() / (1 + GetNStarted()) +
-           stddev_multiplier * GetQStdDev();
+           stddev_factor * GetQStdDev();
   }
 
   int GetVisitsToReachU(float target_score, float numerator,
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -155,6 +155,10 @@ class Node {
   // for terminal nodes.
   float GetQ() const { return q_; }
   float GetD() const { return d_; }
+  // Returns standard deviation of value head in subtree.
+  float GetQStdDev() const {
+    return n_ > 0 ? sqrt(sum_squared_diff_q_ / n_) : 0.0f;
+  }
 
   // Returns whether the node is known to be draw/lose/win.
   bool IsTerminal() const { return is_terminal_; }
@@ -267,6 +271,9 @@ class Node {
   // of the player who "just" moved to reach this position, rather than from the
   // perspective of the player-to-move for the position.
   float q_ = 0.0f;
+  // Sum of squared differences of the value head of the neural network over all
+  // visited nodes in subtree w.r.t. the average q_.
+  float sum_squared_diff_q_ = 0.0f;
   // Averaged draw probability. Works similarly to Q, except that D is not
   // flipped depending on the side to move.
   float d_ = 0.0f;
@@ -340,6 +347,7 @@ class EdgeAndNode {
   float GetD() const {
     return (node_ && node_->GetN() > 0) ? node_->GetD() : 0.0f;
   }
+  float GetQStdDev() const { return node_ ? node_->GetQStdDev() : 0.0f; }
   // N-related getters, from Node (if exists).
   uint32_t GetN() const { return node_ ? node_->GetN() : 0; }
   int GetNStarted() const { return node_ ? node_->GetNStarted() : 0; }
@@ -354,10 +362,11 @@ class EdgeAndNode {
     return edge_ ? edge_->GetMove(flip) : Move();
   }
 
-  // Returns U = numerator * p / N.
+  // Returns U = numerator * p / N + stddev_factor * GetQStdDev().
   // Passed numerator is expected to be equal to (cpuct * sqrt(N[parent])).
-  float GetU(float numerator) const {
-    return numerator * GetP() / (1 + GetNStarted());
+  float GetU(float numerator, float stddev_factor) const {
+    return numerator * GetP() / (1 + GetNStarted()) +
+           stddev_factor * GetQStdDev();
   }
 
   int GetVisitsToReachU(float target_score, float numerator,
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -315,7 +315,7 @@ class Node {
 
 // A basic sanity check. This must be adjusted when Node members are adjusted.
 #if defined(__i386__) || (defined(__arm__) && !defined(__aarch64__))
-static_assert(sizeof(Node) == 52, "Unexpected size of Node for 32bit compile");
+static_assert(sizeof(Node) == 56, "Unexpected size of Node for 32bit compile");
 #else
 static_assert(sizeof(Node) == 80, "Unexpected size of Node");
 #endif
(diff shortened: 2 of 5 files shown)


jkormu:
Did some initial tests and seems like the reasonable values for stddev-factor quite a bit below 1.0.
[Here](https://script.google.com/macros/s/AKfycbzA7OyZd_j_ZZ98HWDpkbTRmFMiivnbDrPIvJEKk47SgMQPI574/exec?file_name=pr791-stddev-factors.html) are search trees visualized for different search tree factors (be warned that the plot is 19MB download).

For example here is tree shape for stddev-factor=1.0:
image

And here for stdev-factor=0.0:
image

Is it expected that value 1.0 has so drastic effect? And what is the expected range of reasonable values?

wakamex:
I got a quick positive result with 0.1, manually chosen after 0.5 was way worse.

seems promising to tune, though number of parameters is so high now, they seem almost unmanageable to tune all in combination. some improved tuning system would be fantastic.

```
tc 8+0.2s on 2060
   # PLAYER            :  RATING  ERROR  POINTS  PLAYED   (%)  CFS(%)    W    D    L  D(%)
   1 lc0pr791-41445    :    26.1   21.0    79.0     138  57.2      99   45   68   25  49.3
   2 lc0-41445         :   -26.1   21.0    59.0     138  42.8     ---   25   68   45  49.3
```
in combination with some other manual settings, includes FPU PR since I used the circleCI build, but it's turned off. moveOverhead accidentally left at 200ms, but doubt that had any impact.
image

kk2796:
I wonder if this use-case for standard deviation and variance might be better ways to work with upper bound altogether.   Could we use standard error, SE(s,a), as a replacement term for U(s,a)?  This would be independent of the objective of the motivation behind entropy - but might be a big enough improvement that a need to factor in entropy becomes moot.     

Instead of using standard deviation, we use standard error to establish confidence intervals that shrink as visit counts increase.  There are a few things that have to come together simultaneously for this to make sense.  

1.  First (and maybe this applies irrespective of any other changes) we should clean up how we choose Q of unvisited actions.  If the value head (I'll call V(s)) represents "where I think this game will end with perfect play going forward." then wouldn't it make sense to estimate the Q(s,a) should be as V(s) for the action a of the top-policy move?  And for all other actions a', Q(s,a') should be somewhere between -1 and V(s).  Exactly where Q(s,a') falls between -1 and V(s) could be proportional to where policy(a') falls between 0 and policy(a).  
So if value-head was 0.2, and policy-head showed 3 legal moves, a, b, c as P(a)=0.5, P(b)=0.4, and P(c)=0.1 then initial Q(s,a) would be 0.2.  For Q(s,b), because policy(b) is 80% of policy(a), we would set Q(s,b) to be 80% of the distance between -1 and 0.2, which is 0.96 up from -1, which would be Q(s,b) = -0.04.  And by similar logic, initial estimate of Q(s,c) would be -1 + .24 = -0.76.  


2. For initial upper bound confidence interval of estimate unvisited action , we can assume a normal probability curve between Q(s,*) and +1 for each action's initial Q per above.  Then, our initial guess is treated as the first data point for Q and subsequent visits as additional data points, yielding standard deviation and other truly meaningful statistics.  We then choose actions based on their current average of Q plus the standard error of that Q (which will shrink with increasing visits to good-nodes, driving exploration of less-visited nodes).

oscardssmith:
Yeah, I think standard error is probably what we want instead of standard deviation.

ddobbelaere:
I am currently running a very extensive tactics suite kindly provided by jhortos for stddev-factor = 0.1 and stddev-factor = 0 (master) and the results are impressive!

As I don't have a fast CPU, (one of the best T35 nets) net 36092 is used (128x10b SE). Parameters of lc0 are `--smart-pruning-factor=0 --minibatch-size=16` (all others default). Each problem is being investigated over 10000 nodes. We call a problem *solved* if the best move over those 10000 moves is the correct move. The *number of nodes needed* is defined as the number of nodes after which the best move stays the correct move till the end (10000 nodes), which means the search doesn't change its opinion in the meanwhile. Here are the preliminary results:

```
stddev-factor = 0; solved 101/208 problems; 1487 nodes needed on average
stddev-factor = 0.1; solved 114/208 problems (101 common with 0); 1476 nodes needed on average (1161 over commonly solved ones)
````

Not only are there more problems solved, but the average number of nodes needed decreases from 1487 to 1161 over the commonly solved problems (stddev-factor = 0.1 finds all solutions that master (stddev-factor = 0) finds).

I'll update the list after it has run overnight.

I suspect that stddev-factor = 0.1 is still far from optimal (I only took this value because @wakamex obtained good results with it). Maybe we can rerun the tests with a better (clopped) value.

**Comment by kk2796** [comment]
@ddobbelaere no worries - my newer ideas are definitely in need of a bit more refinement (and better/clearer justification of them).  Meanwhile, nice work in laying the foundation for collecting empirical data and starting to get some actual results.  Seeing unrefined/unoptimized approaches demonstrate improvements of this magnitude is really exciting.

**Comment by wakamex** [comment]
std 0.05 looks even more promising at my short 8+0.2s tc (guessing about 7k nodes)

```   # PLAYER                   :  RATING  ERROR  POINTS  PLAYED   (%)  CFS(%)    W    D    L  D(%)
   1 lc0-32930                :     0.0   ----   944.0    1500  62.9      99  746  396  358  26.4
   2 lc0pr791std0.05-41454    :   -41.5   34.2   132.5     300  44.2      73   89   87  124  29.0
   3 lc0pr788-41454           :   -57.2   34.0   126.0     300  42.0      58   73  106  121  35.3
   4 lc0-41454                :   -62.0   32.9   124.0     300  41.3      93   82   84  134  28.0
   5 lc0pr791std0.1-41454     :   -98.1   36.5   109.5     300  36.5     100   76   67  157  22.3
   6 lc0-50420                :  -230.8   40.0    64.0     300  21.3     ---   38   52  210  17.3

White advantage = 36.27 +/- 7.89
Draw rate (equal opponents) = 28.59 % +/- 1.23
```

the number of parameters to tune seems prohibitively large for clop. my last 2 parameter clops took around 5k games to converge over about 2 days.

quicker convergence for more parameters seems possible with bayesian optimisation:
[deepmind paper](https://arxiv.org/abs/1812.06855)
discussion:
[leela-zero](https://github.com/leela-zero/leela-zero/issues/2085)
[fishcooking](https://groups.google.com/forum/#!topic/fishcooking/dzBJZCBzRY4)

Gilad-B:
Is it possible for a change like this to increase Elo with a (relatively) tactically weak net, but not be as good with the tactically strongest nets?
Maybe Elo should be tested with the tactically best nets (which currently seem to come from Test10, afaik).

kk2796:
@ddobbelaere 

I suspect a key reason you aren't getting more "bang for buck" (and maybe the reason you're getting meager improvements when stddev is reduced by factors like 0.1 and 0.05) is that standard deviation may not be the right statistic to use.  I suspect stderr is much more fitting to this scenario.   

**stderr = (stddev / sqrt(N(s,a)))**

Coding-wise, I think all that's needed is a to add stderr changes alongside (or in place of) existing stddev changes.

In node.h:

add at line 162:
// Returns standard error of value head in subtree.
  float GetQStdErr() const {
    return n_ > 0 ? sqrt(sum_squared_diff_q_ / n_) / sqrt(n_) : 0.0f;
  }

add at line 351:
float GetQStdErr() const { return node_ ? node_->GetQStdErr() : 0.0f; }

modify lines 365-370 to:
  // Returns U = numerator * p / N + stddev_factor * GetQStdErr().
  // Passed numerator is expected to be equal to (cpuct * sqrt(N[parent])).
  float GetU(float numerator, float stddev_factor) const {
    return numerator * GetP() / (1 + GetNStarted()) +
           stderr_factor * GetQStdErr();
  }

In params.cc 

Add at line 67:
const OptionId SearchParams::kStdErrFactorId{
    "stderr-factor", "StdErrFactor",
    "Multiplier for the standard error of Q in the uncertainty \"UCT "
    "search\" term.  "};

Add at line 95:
   options->Add<FloatOption>(kStdErrFactorId, -1000.0f, 1000.0f) = 0.0f;

Add at line 235:
    kStdDevFactor(options.Get<float>(kStdDevFactorId.GetId())),

In parms.h

Add at line 52:
float GetStdErrFactor() const { return kStdErrFactor; }

Add at line 110:
  static const OptionId kStdErrFactorId;

Add at line 148: 
 const float kStdErrFactor;


And similar changes in search.cc to swap stderr functionality in place of existing stddev functionality...

ddobbelaere:
Good idea, I created another PR with your idea to avoid any confusion keep the code clean.

ddobbelaere:
Updated the PR with an unbiased standard deviation estimator and dropped the cpuct multiplier from the standard deviation term (note that this effect on optimal parameters, e.g. above results can be reproduced by multiplying stddev-factor by approx. 3).

killerducky:

ddobbelaere:
Thanks for pointing out. It was sloppy of me to just put a TODO above the method call but not further think about it.

In this specific case I don't think it matters though for positive stddev-factor, as then the number of visits to reach the second best edge are underestimated (as the extra uncertainty term of the best child is not taken into account, but it is taken into account for the second best child). This means, again for positive stddev-factor, a shortcut taken by choosing the cache child is always justified.

Naphthalin:
This is relevant to the discussion collected in #1231 about changing the PUCT formula.

ddobbelaere:
Closing this, we're end 2020 now. FWIW, feel free to steal/borrow ideas/code from it :)

PR #796, opened 2019-03-13
Value head standard error exploration term.

ddobbelaere:
This adds a value head standard error exploration term for (hopefully) better tactical capabilities, as proposed by @kk2796 in https://github.com/LeelaChessZero/lc0/pull/791#issuecomment-472197225.

The main idea is that the uncertainty term of the UCT search is adapted to (new term in bold):

U(s,a) = cPuct ⋅ P(s,a) ⋅ sqrt(N(s,b)) ⋅ 1/(1+N(s,a)) + **stderr-factor ⋅ SE(s,a)**,

with SE(s,a) = S(s,a)/sqrt(N(s,a)) the standard error of the raw net value Qs (and terminal Qs) of all playouts in the subtree (including multivisits) and S(s,a) the standard deviation. The backup update step is very efficient (incremental formula even for multivisits) and uses a variant of https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Welford's_online_algorithm (extended to the multivisit case).

The goal of this PR is in the first place to investigate the effect on tactical capabilities (as a function of the still unknown stderr-factor). If you have other ideas to include the SE(s,a) term in the equation, please share your ideas!

Possible follow-ups:

- Investigate performance for fixed node tests on a standard tactics test suite and compare with master (stderr-factor = 0). Note that the node count at which the best move is found is also an interesting metric (it might be that this PR more quickly finds the best move).
- CLOP the unknown stderr-factor.
- Investigate other ways to include SE(s,a) in the U(s,a) term.
- Investigate match play performance for an optimal value of stderr-factor.

### DIFF
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -242,7 +242,11 @@ void Node::CancelScoreUpdate(int multivisit) {
 }
 
 void Node::FinalizeScoreUpdate(float v, float d, int multivisit) {
-  // Recompute Q.
+  // Update the sum of squared differences.
+  sum_squared_diff_q_ +=
+      (v - q_) * (v - q_) * multivisit * n_ / (n_ + multivisit);
+
+  // Recompute Q and D.
   q_ += multivisit * (v - q_) / (n_ + multivisit);
   d_ += multivisit * (d - d_) / (n_ + multivisit);
 
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -155,6 +155,10 @@ class Node {
   // for terminal nodes.
   float GetQ() const { return q_; }
   float GetD() const { return d_; }
+  // Returns standard deviation of value head in subtree.
+  float GetQStdDev() const {
+    return n_ > 0 ? sqrt(sum_squared_diff_q_ / n_) : 0.0f;
+  }
 
   // Returns whether the node is known to be draw/lose/win.
   bool IsTerminal() const { return is_terminal_; }
@@ -267,6 +271,9 @@ class Node {
   // of the player who "just" moved to reach this position, rather than from the
   // perspective of the player-to-move for the position.
   float q_ = 0.0f;
+  // Sum of squared differences of the value head of the neural network over all
+  // visited nodes in subtree w.r.t. the average q_.
+  float sum_squared_diff_q_ = 0.0f;
   // Averaged draw probability. Works similarly to Q, except that D is not
   // flipped depending on the side to move.
   float d_ = 0.0f;
@@ -340,6 +347,7 @@ class EdgeAndNode {
   float GetD() const {
     return (node_ && node_->GetN() > 0) ? node_->GetD() : 0.0f;
   }
+  float GetQStdDev() const { return node_ ? node_->GetQStdDev() : 0.0f; }
   // N-related getters, from Node (if exists).
   uint32_t GetN() const { return node_ ? node_->GetN() : 0; }
   int GetNStarted() const { return node_ ? node_->GetNStarted() : 0; }
@@ -354,10 +362,11 @@ class EdgeAndNode {
     return edge_ ? edge_->GetMove(flip) : Move();
   }
 
-  // Returns U = numerator * p / N.
+  // Returns U = numerator * p / N + stddev_factor * GetQStdDev().
   // Passed numerator is expected to be equal to (cpuct * sqrt(N[parent])).
-  float GetU(float numerator) const {
-    return numerator * GetP() / (1 + GetNStarted());
+  float GetU(float numerator, float stddev_factor) const {
+    return numerator * GetP() / (1 + GetNStarted()) +
+           stddev_factor * GetQStdDev();
   }
 
   int GetVisitsToReachU(float target_score, float numerator,
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -157,7 +157,9 @@ class Node {
   float GetD() const { return d_; }
   // Returns standard error of value head in subtree.
   float GetQStdErr() const {
-    return n_ > 0 ? sqrt(sum_squared_diff_q_) / n_ : 0.0f;
+    // Note that we use an unbiased standard deviation estimator (sqrt(n_-1)
+    // denominator).
+    return n_ > 1 ? sqrt(sum_squared_diff_q_ / (n_ - 1) / n_) : 0.0f;
   }
 
   // Returns whether the node is known to be draw/lose/win.
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -362,11 +362,11 @@ class EdgeAndNode {
     return edge_ ? edge_->GetMove(flip) : Move();
   }
 
-  // Returns U = numerator * p / N + stderr_multiplier * GetQStdErr().
+  // Returns U = numerator * p / N + stderr_factor * GetQStdErr().
   // Passed numerator is expected to be equal to (cpuct * sqrt(N[parent])).
-  float GetU(float numerator, float stderr_multiplier) const {
+  float GetU(float numerator, float stderr_factor) const {
     return numerator * GetP() / (1 + GetNStarted()) +
-           stderr_multiplier * GetQStdErr();
+           stderr_factor * GetQStdErr();
   }
 
   int GetVisitsToReachU(float target_score, float numerator,
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -315,7 +315,7 @@ class Node {
 
 // A basic sanity check. This must be adjusted when Node members are adjusted.
 #if defined(__i386__) || (defined(__arm__) && !defined(__aarch64__))
-static_assert(sizeof(Node) == 52, "Unexpected size of Node for 32bit compile");
+static_assert(sizeof(Node) == 56, "Unexpected size of Node for 32bit compile");
 #else
 static_assert(sizeof(Node) == 80, "Unexpected size of Node");
 #endif
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -362,11 +362,11 @@ class EdgeAndNode {
     return edge_ ? edge_->GetMove(flip) : Move();
   }
 
-  // Returns U = numerator * p / N + stddev_factor * GetQStdDev().
+  // Returns U = numerator * p / N + stddev_multiplier * GetQStdDev().
   // Passed numerator is expected to be equal to (cpuct * sqrt(N[parent])).
-  float GetU(float numerator, float stddev_factor) const {
+  float GetU(float numerator, float stddev_multiplier) const {
     return numerator * GetP() / (1 + GetNStarted()) +
-           stddev_factor * GetQStdDev();
+           stddev_multiplier * GetQStdDev();
   }
 
   int GetVisitsToReachU(float target_score, float numerator,
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -155,9 +155,9 @@ class Node {
   // for terminal nodes.
   float GetQ() const { return q_; }
   float GetD() const { return d_; }
-  // Returns standard deviation of value head in subtree.
-  float GetQStdDev() const {
-    return n_ > 0 ? sqrt(sum_squared_diff_q_ / n_) : 0.0f;
+  // Returns standard error of value head in subtree.
+  float GetQStdErr() const {
+    return n_ > 0 ? sqrt(sum_squared_diff_q_) / n_ : 0.0f;
   }
 
   // Returns whether the node is known to be draw/lose/win.
@@ -347,7 +347,7 @@ class EdgeAndNode {
   float GetD() const {
     return (node_ && node_->GetN() > 0) ? node_->GetD() : 0.0f;
   }
-  float GetQStdDev() const { return node_ ? node_->GetQStdDev() : 0.0f; }
+  float GetQStdErr() const { return node_ ? node_->GetQStdErr() : 0.0f; }
   // N-related getters, from Node (if exists).
   uint32_t GetN() const { return node_ ? node_->GetN() : 0; }
   int GetNStarted() const { return node_ ? node_->GetNStarted() : 0; }
@@ -362,11 +362,11 @@ class EdgeAndNode {
     return edge_ ? edge_->GetMove(flip) : Move();
   }
 
-  // Returns U = numerator * p / N + stddev_multiplier * GetQStdDev().
+  // Returns U = numerator * p / N + stderr_multiplier * GetQStdErr().
   // Passed numerator is expected to be equal to (cpuct * sqrt(N[parent])).
-  float GetU(float numerator, float stddev_multiplier) const {
+  float GetU(float numerator, float stderr_multiplier) const {
     return numerator * GetP() / (1 + GetNStarted()) +
-           stddev_multiplier * GetQStdDev();
+           stderr_multiplier * GetQStdErr();
   }
 
   int GetVisitsToReachU(float target_score, float numerator,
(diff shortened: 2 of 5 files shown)


ASilver:
Sounds interesting, nice one. Just a comment: while testing tactics is obviously attractive, I would start by making a plain Elo (strength) test. Better tactics at the cost of better play won't be of much help after all. One can improve tactics just by raising the cpuct values beyond their optimal, but the overall play then deteriorates.

jhorthos:
Will run some tests on jhorthos tactics suite at varying stderr-factor settings

kk2796:
mobile, so my opinion briefly

tactical position testing : elo testing :: petri dish testing of antibiotic : clinical trial

jhorthos:
ddobbelaere - I missed that you had done some of this testing on the tactics suite - I guess I am duplicating some of your work... but I don't have currently implemented the test for how quickly the solution is found. Should add that I suppose.

oscardssmith:
@jhorthos just to be clear, is your clop for tactics or for elo?

ddobbelaere:
Thanks for you efforts. I ran the suite via python-chess and parsed the info strings (based on lc0_analyzer.py).

I defined the number of nodes needed to solve a problem as the number at which the best move is correct and doesn't change until the end (10000 nodes in my case). So let's say a is the correct move and the info string's best moves are b, a, c, c, d, a, a, a with reported number of nodes n1, n2, n3, n4, n5, n6, n7, n8, then n6 is the number of nodes needed in this case.

jhorthos:
ah nice method - i haven't used info strings directly - i just get the moves directly inside python chess. yes that is what I meant by game-play CLOP. Elo.

jhorthos:
What does "including multivisits" mean? When a node is visited N times, its Q is added N times to the list of values used to compute stderr?

ddobbelaere:
The average Q(s,a) of a subtree, and its standard error SE(s,a), is updated with the Q of each *leaf* node that is arrived at after an MCTS *playout* from root to leaf node (maximizing Q+U at each step). There are two such possible leaf nodes:

- Non-terminal leaf nodes are said to be *expanded* and need a NN query to get its Q and policy values.
- Terminal leaf nodes (TB hit, checkmate, stalemate, draw by insufficient material, draw by repetition) have a known Q, so they do not need a NN query.

Because we use a highly parallellized version of MCTS (with minibatches in each search thread, but also with multiple search threads), we do playouts before the results of the NN queries are available to us. This can lead to so-called *collisions* (we arrive at the same unexpanded node multiple times), of which we allow a limited amount. Once the node is expanded (the NN eval has taken place) and we update Q(s,a) and SE(s,a), we count the Q of this node as having been visited multiple times, hence the name *multivisits*.

jhorthos:
I wonder if you would mind compiling a version without Cpuct in the second term? Or maybe that makes no sense.

Ah I understand the multivisit now - you just mean that only terminal nodes and expanded (NN tested) nodes even have a Q.

ddobbelaere:
Done, your CLOP is probably telling us that there is an optimal value for **cPuct * stderr-factor** (the previous multiplicative factor) that doesn't depend on cPuct itself, such that it doesn't make a lot of sense to include cPuct in the factor.

I initially thought it was a good idea to have cPuct there as I thought some scaling with node counts of stderr term would maybe be beneficial. This initial thought can be called questionable now (why do we want the stderr term to scale the same way as the original term, it actually makes no sense).

Leaving out cPuct will have the extra advantage that changing cPuct won't have any adverse effects if stderr-factor is different from 0. Earlier an increase of cPuct would have to be accompanied by a decrease of stderr-factor, which is kind of strange and awkward.

jhorthos:
Have been thinking more. I think stderr is the right indicator rather than stddev because I think you want it to be sensitive to the number of tested nodes. As the local tree expands, if the variance measure stays constant I imagine you want to gradually discount it, as stderr does.

One question to be sure I understand. "raw net value Qs (and terminal Qs) of all playouts in the subtree" means the subtree from the parent the unexpanded node in question correct? All these terms take getting used to. So there is an as-yet unevaluated node X for which we are calculating U, and the "subtree" above is the tree with the **parent** of node X as its root.

kk2796:
If we want to really clean this up, the supplied constant should be "confidence-interval", e.g. "%95", and derived "stderr multiplier", which is actually statistical "t",  is derived from that using t-table:
http://sphweb.bumc.bu.edu/otlt/MPH-Modules/BS/BS704_Confidence_Intervals/SysBP_example-T_table.jpg

The nice thing there is that while a confidence interval specification would be constant, the t-value drops with N just like stderr itself drops with N.  This has a nice dual-edge facet - impact of stderr slowly drops with high number of visits (so works "intuitively" with extremeley high visit-count) but t-factor ensures lots of focus is directed at low-N / high-stddev nodes (which may have huge margins of error in their initial estimates; thus be worth exploring more even if initial Q averages are relatively poor)

Going off the deep end a bit - we should **definitely** decouple this from cPuct as - to a large extent - I feel like this is sort of doing the job the original U factor was trying to do... just, "the right way".  Won't know until CLOP results come in, but I suspect that optimizations will point toward increasing the contribution of the t-test/stderr factor (so, larger confidence-interval params) while decreasing the contribution of original U(s,a) factor (so cPuct trends toward 0).

kk2796:
If this does prove successful, it'd be ironic that the U factor was so-named as a reference to "Upper Confidence Bound"... effectively we'd be replacing it with the statistically correct definition of the upper confidence boundary.

jhorthos:
U now means the upper uncertainty bound. :-)

kk2796:
And as for the question, "but what would we do with policy then? " Well, for 1, it still serves a purpose in training the network (it's a meaningful artifact of the search process that the NN is well-served to be able to predict).   But for 2 - there's a MUCH more intuitively obvious use for that too.  I described this once elsewhere.  Get rid of the silly/nonsensical "first move urgency factor" and just the use current-node's P(s,*) array to set initial Q value of each of it's children.  Each child gets an initial Q value between -1 (expected value of worst possible move) and current-value-head (expected value of optimal-play move).  The actual value from that range for any given child should be proportional to where that child's P(s,a) value lies between 0 and the large policy of the entire P(s,*) array.  So the child with biggest P(s,a) gets initial Q = 100% of current-value-head.  Child with P(s,a) near 0% gets Q value very close to -1.

kk2796:
I'm having too much fun with this... but can't let go of the notion of the deepmind guys kicking around ideas for how to design PUCT formula.  

DM1: I like using average Q for exploitation, but it needs to have some sort of exploration term... 
DM2: Yeah, something that somehow captures 'how big might this actually be', you know, above what our current best guess is.
DM1: Right!  We'll call it an 'upper bound'!  The less certain we are about how good the move is, the bigger this 'upper bound' would be on its value.  
DM2: We can use policy head!  We hadn't use it yet, so in this case....
(there's a light tap on door... older guy sticks his head in)
Stan: Hey, guys - overheard your conversation about stats - need any ideas?
DM1 [muttering to DM2] "Ah Christ, it's Stan, the statistician." [Aloud] "Hey Stan!  Hah - thanks, but no, we're not doing "statistics" here.  Times are changing. We're building NN's.  Machine learning stuff.. a whole new math.  But we'll let you know if we need help finding an average or something."
Stan: Okay... you sure?  Cuz I thought I heard you saying..." 
DM2, [interrupting] Yep, thanks Stan, but this is visionary stuff they don't teach in college - no time to explain it to you, but maybe you can read up on it when we publish it in Science."  [muttering to DM1] "I thought Stan retired like a year ago!
(Stan departs, muttering something about "... dang yungins...")
DM2: Okay - where were we?  Oh, right, Upper bound.  Okay, so, my idea is, start with a policy value.  It's a bit small, so let's bump it up with a mulitplier, some sort of constant to our PUCT formula..."
DM3: Perfect - then we just use a scale so it drives exploration less and less the more a node is visited. 
 So let's multiply by 1/N.  
DM1: Oh - wait - what about N=0 though?  How about, 1/(N+1).  
DM2: Yeah, that's much better! And we can use it to force all nodes to be visited eventually, by multiplying it also by the total parent-node visits.  Scaled down of course, to something like log(N) of parent or sqrt(n of parent).  
DM1: We are so amazingly smart, But there's only one way to figure out the best solution.
DM2: Ask Stan?
(raucous laughter fills the room)
DM1: Good one, good one!  No, for real, let's fire up 10000 TPUs for a few hours and sort this out with the infallible method of "trial and error"

ddobbelaere:
@kk2796 I really like your suggestion to replace the fixed stderr-factor by the statistical _t_ corresponding to a given (one-sided) fixed confidence interval, which changes as a function of the number of nodes N(s,a). A few critical remarks that I would like to bring up however are:

- The notion of confidence interval corresponding to _t_ is only mathematically correct if the following two assumptions of the t-distribution are satisfied: statistical *independence* of the Qs and the *normality* (Gaussian) assumption. It can already be seen that in some cases (namely multivisits), the independence assumption is blatantly violated. Moreover, the true distribution of Q might be approximately Gaussian (we don't know actually), but from the condition that it lies between -1 and 1, it is already apparent that the a perfect fit to a Gaussian distribution is impossible.
- If we ever try this out (and I think we should, but maybe in a separate follow-up PR), we should resort to approximations, as calculating the exact inverse of the t-distribution CDF seems computationally too expensive.

Looking again at the equations, and at the definition of t in https://en.wikipedia.org/wiki/Student%27s_t-distribution#How_the_t-distribution_arises, I noticed that we are now using the biased sample variance (with denominator n) instead of the unbiased sample variance (with denominator n-1) that is also present in the definition of t. This leads to the observation that right now, an opposite scaling behavior w.r.t. t is happening at low nodes, as can be seen in the following figures.

scaling_factors_85
scaling_factors_95

The varying t is shown as a function of node count, along with a comparison with fixed stderr-factor, both for an unbiased sample variance and biased sample variance (the one we are using now).

In the limit of N(s,a) going to infinity, all the curves converge to the same constant.

**In conclusion: it seems very reasonable to change the biased sample variance to the unbiased version in this PR, but let's allow @jhorthos to share his findings of the current state of the PR (biased variance) first.**

kk2796:
@ddobbelaere excellent assessment!  Fully on board with waiting for findings of current idea before exploring improvements.

A thought that I'd sort of been keepinng "at bay" that ties in to your comment about the lack of independence of Qs.  Bearing in mind that any given node has a truly correct value of -1, 0, or 1; and that (in theory - unless search is fundamentally flawed) a large enough visits to any given node will should result in Qs that converge to the "truly correct" value (where large-enough might be astronomically large), it stands to reason that more recent visits should be considered "more likely to be more accurate" than early-on visits.  To that end - independent of everything else already under consideration... I wonder if Q values (along with their contributions to variance) should be weighted according to visit count.  Something like: 
n=1: newest node is 1/1= 100% of Q
n=2: newest node is 2/3 = 66% of Q - oldest is 33%
n=3: newest node is 3/6 = 50% of Q - oldest is .1666%
...
n=100: newest node is 100/5050 = 5% of Q - oldest is .0002%

oscardssmith:
I think that is probably a bad idea, as it potentially stops Q from converging due to weird cyclical effects.

kk2796:
@oscardssmith  Perhaps weighting according to N is a bit excessive - while the N=100 move is probably a better estimate than N=5; it does seem a bit over-the-top to say it's worth 20x as much.  Perhaps a more subdued idea would work better - like weighting each Q value according to the depth of the node it came from.  
But I do think this (and couple other ideas, such as using P(s,a) to seed the n=0 value of Q(s,a)) will be much more compelling grounds to explore *after* we see some successful outcomes for the stderr idea.

ddobbelaere:
Thanks to @jhorthos for the extensive testing. Here are his CLOP results he sent me a while ago along with his comments:

> Ok here is the CLOP report. First, the "max" value recorded occasionally:

games | cpuct | stderr-factor
----|----|----
2455 | 2.93 | 0.82
2707 |2.95 |0.49
3000| 3.09 |0.09
3083 |2.95| 0.47 _(very unstable in stderr)_
3200| 2.92 |0.65
3500 |3.19| 0.05
3784 |3.02| 0.54
4199| 3.05 |0.42
4276 |3.04| 0.43
4774| 3.09|  0.23
4817|  3.11|  0.22

> Now the weighted image.

clop

> ranges cpuct 1-4 and stderr-factor -1 to 3
so I would be inclined to use an eyeball line through that cloud, which goes from about 2.5 cpuct 1.7 sef to 3.0 cpuct 0.0 sef.
Not very satisfactory, since it suggest above cpuct 3.0 that stderr-factor should be negative.
But almost 5,000 games is enough to cause me to stop. Usually the result with 2 factors is a lot clearer by then.

kk2796:
That is definitely messy!  My initial inclination is to take this as grounds to move immediately to testing the unbiased mean (denominator = n-1) to see if that yields a direct improvement just by virtue of expressing the right statistical measure.  

However, I've got one concern that's independent of the distinction of biased vs unbiased.   Recalling my prior comment, I did predict a negative correlation:  

> Won't know until CLOP results come in, but I suspect that optimizations will point toward increasing the contribution of the t-test/stderr factor (so, larger confidence-interval params) while decreasing the contribution of original U(s,a) factor (so cPuct trends toward 0).

While this came through, it certainly did not come through very cleanly.  I now wonder if the main reason we saw such a noisy mess was because the U term is the only means of policy P(s,a) to influence search.  Leaving everything else aside, I think it's safe to say that P(s,a) is valuable information and so any change that forces it out of the picture, no matter what other benefits that change may bring, is going to suffer the consequences of throwing out perfectly useful information.   

So if my theory holds and the new SE term "wants" to be a better/cleaner exploration term than U, and U must be reduced via shrinking cpuct to make room for SE, there's an unavoidable catch-22: increasing SE's contribution will diminish the usage of any information contained in P(s,a) in the overall PUCT search.   The analogy that occurs to me is that P(s,a) is like the "baby" preventing us from dumping out enough of the (murky) U-factor "bathwater" to make room for the (clean/fresh) SE-factor "water".  If that analogy is apt, then the only way we'd truly be able to detect the benefit of SE is to give the P(s,a) some other way to participate (beyond it's cpuct-weighted contribution to the U factor).   

The idea I expressed [above](https://github.com/LeelaChessZero/lc0/pull/796#issuecomment-473674025) was to use P(s,a) to set a "best guess" at initial value of unexplored node, instead of relying on "first-move-priority" logic.  Though that was originally meant as a follow-up idea to play with  (contingent on success of this SE pull request), I now suspect that there maybe a mutual dependency: that the SE factor can only work effectively if P(s,a) is somehow expressed outside of U factor.

killerducky:

kk2796:
@ddobbelaere FYI - I am coding the "init-Q" changes and will handle as a separate pull request.  While I see the two ideas as related, there's no logical reason the change can't be made independent from your stderr changes (and merged down the road). 

BTW, this will be my first "toe dip" into actual lc0 coding, so I may reach out for assistance.

ddobbelaere:
@killerducky Thanks for pointing out. It was sloppy of me to just put a TODO above the method call but not further think about it.

In this specific case I don't think it matters though **for positive stderr-factor**, as then the number of visits to reach the second best edge are underestimated (as the extra uncertainty term of the best child is not taken into account, but it is taken into account for the second best child). This means, again for positive stddev-factor, a shortcut taken by choosing the cache child is always justified.

Note that the behavior for negative stderr-factor is incorrect however (the number of visits can be overestimated), if this would make any sense anyway (but the CLOP considered them, so it might have influenced its convergence and estimated best values).

@kk2796 Sure, go ahead, glad to help.

Naphthalin:
This is relevant to the discussion collected in #1231.

ddobbelaere:
Closing this, we're end 2020 now. FWIW, feel free to steal/borrow ideas/code from it :)

PR #809, opened 2019-03-20
Only short-circuit search with a non-0 SmartPruningFactor.

Mardak:
r?@mooskagh Similar to https://github.com/LeelaChessZero/lc0/commit/82b3f208d33fac38c768f2ad2bd4d45c30599fa9 which fixed `go infinite searchmoves a2a4`… this commit fixes it for `go nodes 100 searchmoves a2a4` where in both cases, the intent of searchmoves is to have search focus on the desired move, so the `possible_moves` short-circuit behavior is undesired.

Before:
```
go nodes 100 searchmoves a2a4
info string a2a4  (207 ) N:       0 (+ 1) (P:  2.42%) (Q:  0.08519) (D:  0.000) (U: 0.03634) (Q+U:  0.12152) (V:  -.----) 
bestmove a2a4 ponder g8f6
```

After:
```
go nodes 100 searchmoves a2a4
info string a2a4  (207 ) N:     108 (+ 9) (P:  2.42%) (Q: -0.05578) (D:  0.000) (U: 0.00642) (Q+U: -0.04936) (V: -0.0646) 
bestmove a2a4 ponder h7h5
```

(One of my use cases is `go nodes 2 searchmoves <move>` to get the network eval V for a specific move from a desired position.)

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -982,7 +982,8 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend(
       second_best_edge.Reset();
     }
 
-    if (is_root_node && possible_moves <= 1 && !search_->limits_.infinite) {
+    if (is_root_node && possible_moves <= 1 && !search_->limits_.infinite &&
+        params_.GetSmartPruningFactor()) {
       // If there is only one move theoretically possible within remaining time,
       // output it.
       Mutex::Lock counters_lock(search_->counters_mutex_);


mooskagh:
Note that in this case it will still have smart pruning.
So if you do `go nodes 1000000 searchmoves e2e4 d2d4` and after 600000 nodes it decides that d2d4 is not going to win, it won't allocate visits there at all.

While it's expected that `go infinite` will be infinite, suppressing smart pruning (partially) in case of `go nodes` is not that clear..

oscardssmith:
would it make sense to add an exact flag that determines whether exactly or about x nodes are used?

Mardak:
Would you prefer the check look for searchmoves having only 1 entry instead of not-empty? `go nodes … searchmoves <one move>` most likely intends to search more than 1 visit.

killerducky:
I think it's easier to leave the code alone, and if users don't want `go nodes N` to stop early, they can set `SmartPruningFactor=0`.

Mardak:
`SmartPruningFactor=0` still causes a single searchmove to short circuit search as it appears like there's only 1 possible move.

mooskagh:
Hm, maybe we should not do that if SmartPruningFactor=0, as a special case. Factor = 0 -> no smart pruning. :)
Want to stop early with factor=0? Set it to 0.0000001.
Not sure if good idea.

Mardak:
> Hm, maybe we should not do that if SmartPruningFactor=0, as a special case. Factor = 0 -> no smart pruning. :)

Updated PR as suggested

hans-ekbrand:
If the user request a node based search limit, then the non-expert user expects Leela to actually evaluate that number of nodes. `SmartPruningFactor` is a feature for time-management, but it shouldn't interfere in this case, that's my strong opinion. The problem is really that `SmartPruningFactor` defaults to a non-zero value, which from a user perspective is problematic, since the user might not know of all the default values, the user might not even be aware that `SmartPruningFactor` exists or understand how it works. A good program just works even for non-expert level users.

Mardak:
Potentially obviated by #884 which removes the `possible_moves` optimization.

fohristiwhirl:
Yes please. I think examining 1 move in particular for a limited number of nodes is a thing some analysts want to do.

Mardak:
I believe @dkappe was looking for a way to do single move analysis

PR #810, opened 2019-03-21
Add stop command to selfplay interactive mode to allow for graceful exit.

Tilps:
Doesn't stop any games in progress as opposed to quit which terminates the app.

If we want to really aggressively go after short game bias we would also need to ensure each thread plays the same number of games, but at least with this in place we can look at migrating the client to use interactive mode to reduce short game bias.  Every thread same number of games could be another flag in future.

### DIFF
--- a/src/selfplay/loop.cc
+++ b/src/selfplay/loop.cc
@@ -85,6 +85,11 @@ void SelfPlayLoop::CmdStart() {
       std::make_unique<std::thread>([this]() { tournament_->RunBlocking(); });
 }
 
+void SelfPlayLoop::CmdStop() {
+  tournament_->Stop();
+  tournament_->Wait();
+}
+
 void SelfPlayLoop::SendGameInfo(const GameInfo& info) {
   std::vector<std::string> responses;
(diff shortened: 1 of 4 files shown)


src/selfplay/loop.cc
```diff
@@ -85,6 +85,11 @@ void SelfPlayLoop::CmdStart() {
       std::make_unique<std::thread>([this]() { tournament_->RunBlocking(); });
 }
 
+void SelfPlayLoop::CmdStop() {
+  tournament_->Stop();
+  tournament_->Wait();
```
Tilps: This wait is here so you can issue stop/quit as a pair - and when the app has exited its done.

Tilps:
Actually I don't think we need an option to make each thread do the same number of games - it doesn't change the bias level.

src/selfplay/loop.cc
mooskagh: "Nicer" solution would be stop+isready+quit, but selfplay is not real UCI anyway.

Issue #753, opened 2019-02-23
Take reused nodes into account when budgeting time.


mooskagh:
Currently we budget move time for Leela to think.
However, (I think) we really want to aim for particular number of nodes at the moment of making the move. 
Particularly, if tree was mostly reused, we need to think less, and if no tree was reused, we need to think more.

A particualr case of this issue is first move (where no tree is usually reused). There is a PR which tries to solve this for the first move https://github.com/LeelaChessZero/lc0/pull/645, but I really think it should be solved in more generic way.

Particualry, totally unexpected move (no tree reuse) should trigger the same behaviour as first move, and partially unexpected move should also increase thinking time.

mooskagh:
Actually, probably idea of "budgeting nodes not time" is not entierly correct.
It can be either nodes or time, but it should take reused nodes (or reused time, e.g. reused nodes from last move converted back to seconds) into account.

Naphthalin:
Sounds like this was implemented with #1195, so issue can probably be closed.

PR #813, opened 2019-03-24
Add EdgeDiscardFactor

MelleKoning:
As #812 but now excluding CP, so only having the addition of EdgeDiscardFactor.

Copying text from #812 :

Looking at issues #686 and #753 the issue of how best to use thinking time seems a paramount issue.

One thought here is to discard moves from root in case they are lagging too far behind the best move being considered. When discarding those moves earlier in the process this will save time for future upcoming moves, especially in time limited games where added time is given for each move.

The default setting to discard moves is currently at half the number of nodes, meaning that given the remaining budgeted time of the move you can discard other moves that can never catch up with the best move.

However, discarding even sharper will save time for future moves. This will lead to more thinking time when best_move and second_best are close together in future moves of the game risking the current move of course...

Tested with factor 1.25f, so increasing the discard factor with 25%, including root-strategy absolute and CP enabled gave the following result in 14 test games with CuteChess.
Time settings: 5 seconds the game + 3 sec a move
Results: 5 wins, 3 losses, 6 draws.
Positive result but too low number of games to give confidence.

Might need a lot of CLOP testing to figure out what factor works best.

### DIFF
--- a/src/mcts/params.cc
+++ b/src/mcts/params.cc
@@ -178,6 +178,11 @@ const OptionId SearchParams::kKLDGainAverageInterval{
     "kldgain-average-interval", "KLDGainAverageInterval",
     "Used to decide how frequently to evaluate the average KLDGainPerNode to "
     "check the MinimumKLDGainPerNode, if specified."};
+const OptionId SearchParams::kEdgeDiscardFactor{
+	"edge-discard-factor", "EdgeDiscardFactor",
+	"Factor used for discarding thinking time for moves that can not reach bestmoves nodes. Factor of 1.00 is default "
+	"higher factor will sooner discard other edges thereby saving time for future moves." };
+
 
 void SearchParams::Populate(OptionsParser* options) {
   // Here the uci optimized defaults" are set.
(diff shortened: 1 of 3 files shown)


killerducky:
Is this different from SmartPruningFactor?

MelleKoning:
Well, now you mention it maybe it is not different at all. 



Looking at it, Only difference is where the calculation takes place but in essence Looks the same indeed... smartpruning establishes a different calculated number of playouts. Ehm, do you know why default of smartpruning is set to 1.33 instead of just one?

oscardssmith:
it tested stronger.

MelleKoning:
Ok, with a factor of 2.0 for EdgeDiscardFactor (will just overlap with SmartPruningFactor) + CP I got even these scores against default:

```
Score of lc0.Net.41666 vs lc0pr812df20.Net.41666: 2 - 11 - 37 [0.410]
Elo difference: -63.23 +/- 47.74

50 of 50 games finished.

```
For a match with same net versus same net with same openings reversed that seems pretty decisive.

jjoshua2:
Some quick tests show smart pruning is strongest around 2.0 as well. If all
you are doing is increasing smart pruning maybe it's better to do that
directly? At longer tc the optimal pruning is less though.

MelleKoning:
Thanks for that information @jjoshua2 and @killerducky simply was not aware of that. Looking at the code now, understanding what remaining_playouts means indeed believe this EdgeDiscardFactor is just the same as SmartPruning parameter. So I will just do that directly and remove these 812 and 813 PR's shortly.
Thanks,

Issue #814, opened 2019-03-26
Need gtest for best_child_cached_ code


killerducky:
This is a gotcha for people changing the PUCT formula. It would be nice to have a gtest for this logic, that would detect when someone changes the main PUCT formula but doesn't realize it breaks this caching logic. I think a majority of those PRs have this bug, like #746 

Here are a few code snippets of how the caching works:

`Node* possible_shortcut_child = node->GetCachedBestChild();`

```
      int estimated_visits_to_change_best =
          best_edge.GetVisitsToReachU(second_best, puct_mult, fpu);
      // Only cache for n-2 steps as the estimate created by GetVisitsToReachU
      // has potential rounding errors and some conservative logic that can push
      // it up to 2 away from the real value.
      node->UpdateBestChild(best_edge,
                            std::max(0, estimated_visits_to_change_best - 2));
```

```
  int GetVisitsToReachU(float target_score, float numerator,
                        float default_q) const {
    const auto q = GetQ(default_q);
    if (q >= target_score) return std::numeric_limits<int>::max();
    const auto n1 = GetNStarted() + 1;
    return std::max(
        1.0f,
        std::min(std::floor(GetP() * numerator / (target_score - q) - n1) + 1,
                 1e9f));
  }
```

killerducky:
I quickly scanned some PRs and tagged those that I guess might have a problem.

killerducky:
See https://github.com/LeelaChessZero/lc0/pull/619 for original PR for this code.

killerducky:
Tilps just pointed out Node::FinalizeScoreUpdate() invalidates best_child_cached_. That's why the code can assume Q does not change because it's only valid during batch gathering.

Naphthalin:
Having this kind of test would indeed be helpful; currently, the first sign usually is a serious nps drop. Any news on this?

PR #816, opened 2019-03-27
Policy scaled fpu strategy

oscardssmith:
They apparently gain 500 elo at 800 nodes.

### DIFF
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -160,6 +160,13 @@ float Edge::GetP() const {
   return ret;
 }
 
+float Edge::GetQInit() const { return qInit_; }
+
+void Edge::SetQInit(float val) {
+  assert(-1.0f <= val && val <= 1.0f);
+  qInit_ = val;
+}
+
 std::string Edge::DebugString() const {
   std::ostringstream oss;
   oss << "Move: " << move_.as_string() << " p_: " << p_ << " GetP: " << GetP();
@@ -192,6 +199,7 @@ void Node::CreateEdges(const MoveList& moves) {
   assert(!edges_);
   assert(!child_);
   edges_ = EdgeList(moves);
+
 }
 
 Node::ConstIterator Node::Edges() const { return {edges_, &child_}; }
@@ -270,8 +278,9 @@ void Node::UpdateBestChild(const Iterator& best_edge, int visits_allowed) {
 
 Node::NodeRange Node::ChildNodes() const { return child_.get(); }
 
-void Node::ReleaseChildren() { gNodeGc.AddToGcQueue(std::move(child_)); }
 
+
+void Node::ReleaseChildren() { gNodeGc.AddToGcQueue(std::move(child_)); }
 void Node::ReleaseChildrenExceptOne(Node* node_to_save) {
   // Stores node which will have to survive (or nullptr if it's not found).
   std::unique_ptr<Node> saved_node;
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -86,6 +86,12 @@ class Edge {
   float GetP() const;
   void SetP(float val);
 
+  // Returns or sets value of Move policy prior returned from the neural net
+  // (but can be changed by adding Dirichlet noise). Must be in [0,1].
+  float GetQInit() const;
+  void SetQInit(float val);
+
+
   // Debug information about the edge.
   std::string DebugString() const;
 
@@ -100,6 +106,7 @@ class Edge {
   // Probability that this move will be made, from the policy head of the neural
   // network; compressed to a 16 bit format (5 bits exp, 11 bits significand).
   uint16_t p_ = 0;
+  float qInit_ = 0.0;
 
   friend class EdgeList;
 };
@@ -137,6 +144,7 @@ class Node {
 
   // Creates edges from a movelist. There has to be no edges before that.
   void CreateEdges(const MoveList& moves);
+
 
   // Gets parent node.
   Node* GetParent() const { return parent_; }
@@ -335,7 +343,16 @@ class EdgeAndNode {
 
   // Proxy functions for easier access to node/edge.
   float GetQ(float default_q) const {
-    return (node_ && node_->GetN() > 0) ? node_->GetQ() : default_q;
+	  
+  //  return (node_ && node_->GetN() > 0)
+       //        ? node_->GetQ()
+           //    : (node_->GetOwnEdge()) ? node_->GetOwnEdge()->GetQInit()
+                        //               : default_q;
+
+	      return (node_ && node_->GetN() > 0)
+             ? node_->GetQ()
+             : edge_->GetQInit();
+
   }
   float GetD() const {
     return (node_ && node_->GetN() > 0) ? node_->GetD() : 0.0f;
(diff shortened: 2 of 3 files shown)


simlu:
Looks like the actual change is very small. Lots of formatting noise in this one...

oscardssmith:
yeah

kk2796:
A little backstory - After making this change, I unit-tested it with a corrupted test environment.  My Arena GUI was sneaking an older net into my tournament settings, leading me to test my changes with latest net against original binary that - unbeknownst to me - was using a very old net.  This led to scenario where my initial test cases blew my socks off (showing ELO gains of +500 or more).  My enthusiasm overwhelmed my better judgement, and I slammed the accelerator down to get the changes into the hands of others for verification.  That rush to publish left a wide wake of woefully hackish code... not only poorly formatted, but the modifications themselves were "POC" level hacks I'd not normally let see the light of day. 

If nobody else beats me to the punch, within next 24-48 hours, I'll clean this up.  It still is exactly the change I'd like to see combined with some other ideas; however, the "right" way to do it is probably to introduce this as an 3rd alternate for FPU strategy.

mooskagh:
I'd like to get more confirmations that that's actually gaining some Elo.

If it is, it still have other issues to resolve (other than formatting and removing support of current FPU handling).
Namely, it grows Edge size twofolds, which is ~+120 bytes per node.

Instead, we can e.g. store maxP in visited_policy_ (which is only used in FPU-strategy=reduction) and adjust Q formula to use it.

Also, please provide better name and description for the PR.

Also, check whether this change suffers from a problem described in #814 (probably not, as it affects Q but not U).

kk2796:
Thanks @mooskagh .  I think a brief/fitting title for the PR would be "Policy-scaled FPU strategy", but open to suggestions.  That said, I don't see any buttons/links to change the title though - perhaps original author @oscardssmith would need to make that change?  Or an admin?

kk2796:
@mooskagh 
Thanks, I see what you mean about the wasted space.   

Currently coding/unit-testing:
1. Storing maxP_ at node level (as separate variable from visited_policy_)
2. Dynamically computing edges' qInit values from maxP (rather than storing per qInit per edge)  
3. Adding "policy-scaled" as new (non-default) option for FPU selection.

This should eliminate the real space-waster, qInit-per-edge, and thus almost complete recoup the ~120 bytes-per-node overhead of the original change.  I'm really not a fan of overloading the visited_policy_ variable for the meager savings of 2 additional bytes per node.  However, if memory real estate truly is that pricey, I would be willing to merge maxP_ and visited_policy_ together, but would aim to rename the variable to something that reflected its dual-usage character.  

Or - dummy C++ question here - is there a way to avoid allocating space for whichever of maxP_ or visited_policy_ is not used?  

I should also add - I don't have the GPUs to test extensively (neither deep node count nor broad game-count), but the rudimentary testing (performed after rebuilding my test environment from scratch and taking extreme care to validate all UCI parameters of the tests) I've done is indicative that 
1. My changes have no effect (good or bad) on fixed-node-count, self-play ELO using default parameters 
2. The changes do seem to give a slight edge when combined with lower CPUCT (1.5), vs unmodified code using default CPUCT (3.0).

It's not much, but I do think it's enough to merit some broader testing once I've got the PR updated with aforementioned changes.  Especially in lieu of the fact that I'm cautiously optimistic that this change and the change in #796 will be mutually beneficial... so really, "don't break anything" is all I really hoped out of this PR in the first place.

zz4032:
I tuned CPuct for current dev and PR816 at TC with ~900 nodes/move.
Tuned CPuct for master is 3.02, for PR816 only slightly lower with 2.85.
Test match:
```
   # PLAYER                  :  RATING  ERROR  LOS(%)   GAMES  DRAWS(%)
   1 lc0_PR816_CPuct_2.85    :    10.0    8.2    99.2    2000      72.0
   2 lc0_CPuct_3.02          :     0.0   ----     ---    2000      72.0

White ELO advantage = 15.6
Draw rate (equal opponents) = 72.3 %
```

kk2796:
While doing some unit testing, I decided to look a little closer at how close the PR816 change was to meeting the objective: setting fpu to a value that works as a decent estimate for the actual Q value of the unvisited node.    The following is a copy/paste from a Discord write-up I did, discussing what I found:

> 
> I've been slowly plugging away at PR816 - the idea of setting FPU of each edge to a value between -1 and parent-Q, interpolated according to where that edge's policy lies between 0% and the max-policy of all edges.  A key assumption was that this interpolation approach would provide a rough, but plausible, initial estimate of the Q value of each unvisited node.   I decided to check that assumption by forcing a depth-first tree walk of 10,000 nodes from starting position, recording all policy&value-head results, then comparing interpolated-fpu values of each edge to its node's Q (value-head result).
> Interpolated-fpu vs node-Q-value
> 
> https://cdn.discordapp.com/attachments/425419999096733706/562821410897133608/unknown.png
> 
> The x-axis is the interpolated "guess" at Q-of-node; y-axis is actual Q-of-node
> The "pretty" part (linear and very close guess) is comprised almost entirely of the "best-policy" edges.  It basically means that most of the time, the Q-value of the best policy is very very close to the Q-value of its parent.  Everything else is "what policy-head thinks of moves that aren't the best".
> There's a temptation to say, "this shows how good Leela's policy head is at focusing on just the best moves ... and treating everything else as, basically, garbage."  But I think this chart demonstrates a deep problem.  While Leela's policy head is very adept at picking the best move, it is very poor at differentiating between "mediocre" and "abysmally bad" moves.  Consider (this is with net 41714) position reached after "d2d4 b7b5".
> In that position, as white, there are 2 terrible moves... but pretty much everything else is strong.  c1h6 is obviously the worst possible move (loses bishop).  And it has worst policy of 0.88%.  Next worse is moving king out with e1d2: 1.07%.
> b1c3 is certainly not "great" - but compared to those turds, it's a completely different realm of "not great". 
>  It's at least plausible that it could amount to a follow-up worth exploring - not at the front of the list... but worth a look eventually. Yet policy of b1c3 is 1.25% - a hair over e1d1.
> I guess my overarching point, and concern driven by that graph, is that this looks very unhealthy: how can Leela be expected to explore well when her policy head virtually fails to give any differential between abysmal moves like e1d2 / c1h6 and mediocre moves like b1c3.  Apologies for wall-of-text. I'll recap all this in the PR as well, but wanted to get my thoughts here for others to consider.

kk2796:
More on-topic with this PR:

1) Initially this PR was kindly opened by @oscardssmith, who did the legwork of applying my code changes to his personal fork of lc0.
2) Those code-changes were ugly - really just POC hacks to test an idea (that I got way overhyped about :-) ).  
3) I've now developed and am nearly done unit-testing a refactored version of the code.  I've learned enough about github work flow to have created my own personal fork, "kk2796/lc0" and I have been using a new branch on that fork for the refactored code.   

I'm now seeking "best path forward" advice.  Without moving to a completely new PR, is there a simple way for me to take the new version of my code changes (committed to a new branch of kk2796/lc0 fork) and swap it in for the version of the code that oscardssmith based this current PR on?  Or would it be cleaner to open a new PR, point to this one, and presumably close this one?

oscardssmith:
I think close and new will be easier.

PR #817, opened 2019-03-27
Play move with the highest lower confidence bound

Ttl:
Same as https://github.com/leela-zero/leela-zero/pull/2290. Some features still missing from the LZ PR: ~minimum visit ratio~ and avoiding pruning of the best LCB move in time management.

```
800 node test with network 41724 (MinibatchSize=32 and 5 piece TB)
Score of lc0_lcb vs lc0_master: 85 - 62 - 253  [0.529] 400
Elo difference: 20.00 +/- 20.61, LOS: 97.11 %, DrawRatio: 63.2 %
```

### DIFF
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -243,12 +243,11 @@ void Node::CancelScoreUpdate(int multivisit) {
 
 void Node::FinalizeScoreUpdate(float v, float d, int multivisit) {
   // Recompute Q.
-  float old_delta = (v - q_);
+  // Need to divide by 4 to scale to 0..1 range probability.
+  q_squared_diff_ +=
+      0.25f * (v - q_) * (v - q_) * multivisit * n_ / (n_ + multivisit);
   q_ += multivisit * (v - q_) / (n_ + multivisit);
   d_ += multivisit * (d - d_) / (n_ + multivisit);
-  float new_delta = (v - q_);
-  // Need to divide by 4 to scale to 0..1 range probability.
-  q_squared_diff_ += 0.25f * old_delta * new_delta;
 
   // If first visit, update parent's sum of policies visited at least once.
   if (n_ == 0 && parent_ != nullptr) {
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -243,8 +243,12 @@ void Node::CancelScoreUpdate(int multivisit) {
 
 void Node::FinalizeScoreUpdate(float v, float d, int multivisit) {
   // Recompute Q.
+  float old_delta = (v - q_);
   q_ += multivisit * (v - q_) / (n_ + multivisit);
   d_ += multivisit * (d - d_) / (n_ + multivisit);
+  float new_delta = (v - q_);
+  // Need to divide by 4 to scale to 0..1 range probability.
+  q_squared_diff_ += 0.25f * old_delta * new_delta;
 
   // If first visit, update parent's sum of policies visited at least once.
   if (n_ == 0 && parent_ != nullptr) {
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -38,6 +38,7 @@
 #include "neural/encoder.h"
 #include "neural/writer.h"
 #include "utils/mutex.h"
+#include "utils/stats.h"
 
 namespace lczero {
 
@@ -155,6 +156,7 @@ class Node {
   // for terminal nodes.
   float GetQ() const { return q_; }
   float GetD() const { return d_; }
+  float GetSquaredDiff() const { return q_squared_diff_; }
 
   // Returns whether the node is known to be draw/lose/win.
   bool IsTerminal() const { return is_terminal_; }
@@ -270,6 +272,8 @@ class Node {
   // Averaged draw probability. Works similarly to Q, except that D is not
   // flipped depending on the side to move.
   float d_ = 0.0f;
+  // Sum of squared differences to mean. Used for calculating variance.
+  float q_squared_diff_ = 0.0f;
   // Sum of policy priors which have had at least one playout.
   float visited_policy_ = 0.0f;
   // How many completed visits this node had.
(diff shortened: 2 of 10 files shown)


killerducky:
I think you'll need something similar to this in meson.build:

`deps += [ dependency('boost', modules : ['thread', 'system' ], static : true) ]`

Ttl:
Oh, boost isn't used anywhere. Somehow it works on my machine. Probably no point in adding it just for this one function call.

Ttl:
I clopped ci_alpha and min_ratio parameters at 800 nodes with T50 network against master branch:

ci_alpha_clop_max

ci_alpha_clop_plot

X-axis: `GammaParameter cialpha 1e-7 1e-2`
Y-axis: `LinearParameter minimum-lcb-n-ratio 0.05 0.95`

src/mcts/node.cc
```diff
@@ -243,8 +243,12 @@ void Node::CancelScoreUpdate(int multivisit) {
 
 void Node::FinalizeScoreUpdate(float v, float d, int multivisit) {
   // Recompute Q.
+  float old_delta = (v - q_);
   q_ += multivisit * (v - q_) / (n_ + multivisit);
   d_ += multivisit * (d - d_) / (n_ + multivisit);
+  float new_delta = (v - q_);
+  // Need to divide by 4 to scale to 0..1 range probability.
+  q_squared_diff_ += 0.25f * old_delta * new_delta;
```
ddobbelaere: This formula is incorrect, the increment has to be multiplied by `multivisit`. Compare this with the one used in #791, which is exact, even in the multivisit case.

Now the variance is underestimated, which might explain the rather low clopped values for ci_alpha.

src/mcts/node.h
```diff
@@ -340,6 +344,27 @@ class EdgeAndNode {
...
+    if (!node_) {
+      return -1e6f;
+    }
+    auto visits = node_->GetN();
+    if (visits < 2) {
+      // Return large negative value if not enough visits.
+      return -1e6f + visits;
+    }
+
+    auto stddev = std::sqrt(GetVariance(1.0f) / visits);
```
ddobbelaere: Technically this is the standard error (standard deviation divided by sqrt(N)). Maybe `stderr` is a better variable name.

Or is this the standard deviation of something in another context?

src/mcts/node.cc
Ttl: I didn't want to deal with multivisits yet. #791 is not correct either as it gives different answer depending on the order of the samples (Check for example [0, 1] vs [1, 0]).

`ci_alpha` is very similar to what I clopped in LZ and lack of multivisits probably doesn't affect it. The reason is that the evaluations aren't really normally distributed. See discussion in LZ PR.

src/mcts/node.h
Ttl: I was thinking of the estimated distribution for the mean which has this standard deviation. Standard error seems to be correct too.

src/mcts/node.cc
ddobbelaere: You are already dealing with multivisits (you cannot circumvent them), as you divide the variance by N-1 (which includes multivisits).

I insist #791 is correct and can give you the derivation if you're interested.

Ttl: You are correct about the multivisits and I should add support for it.

Check the case above and keep in mind that `n_` is initialized to zero and `q_` is mean of the previous samples, not mean of all samples. With multivisit = 1 it should match Welford's algorithm, but it doesn't.

Ttl: Okay, forget the previous messages. I checked it more carefully and it seems to give correct answer for multivisit = 1. I trust it works for other multivisit values too.

ddobbelaere: Okay, here's the derivation anyway, as reference. 
IMG_20190328_203618293

src/mcts/node.h
ddobbelaere: Okay, that seems logical, I didn't think about it that way :).

Ttl:
Some testing at high nodes, nothing conclusive. TC 30+0.5s. Ryzen 1700X and RTX2080. SF with 16 threads.

```
   # PLAYER              :  RATING  ERROR  POINTS  PLAYED   (%)  CFS(%)
   1 lc0_41724_master    :    24.7   15.8   274.0     512    54      67
   2 lc0_41724_lcb       :    17.8   15.4   269.0     512    53      96
   3 stockfish_dev       :     0.0    9.8   481.0    1024    47     ---
```

borg323:
Quick patch to fix appveyor builds: https://github.com/borg323/lc0/commit/bad4503f41c0d391ed17ef7c418bbc633c4c4110

zz4032:
TC 4.8s/game+0.02s/move (~900 nodes/move)
```
   1 lc0_PR817    :    47.4   12.5   100.0    1000      60.7
   2 lc0          :     0.0   ----     ---    1000      60.7

White ELO advantage = 56.5
Draw rate (equal opponents) = 64.3 %
```
TC 17.04s/game+0.071s/move (~6000 nodes/move)
```
   1 lc0          :     0.0   ----    60.0    1000      69.0
   2 lc0_PR817    :    -1.5   11.2     ---    1000      69.0

White ELO advantage = 68.1
Draw rate (equal opponents) = 73.9 %
```

zz4032:
Tried retuning this PR and noticed that at lower nodes/move I'm getting a weighted plot similar to yours:
stc_8 9E-4_0 30


But with increased time control (~12000 nodes/move) the optimum area shifts and split's in two:
ltc_4 2E-6_0 51

Clearly a reason why it doesn't perform as expected at LTC.

Naphthalin:
As far as I know, this is still relevant, and not solved by time manager or so. Might be a candidate for @kiudee tuning it once updated, and merged if useful.

Naphthalin:
Closing this for a combination of several reasons:
1) implementation requires boost
2) too many search conflicts
3) #1543 contains a more recent LCB implementation
4) implementation of #1734 will allow a much easier reimplementation, compared to updating this branch.

PR #819, opened 2019-03-28
Group _pext_u64() calls to improve code readability.

benini:
Reduce the number of:
#if defined(NO_PEXT)

Similar to Stockfish:
https://github.com/official-stockfish/Stockfish/blob/master/src/bitboard.h#L88

### DIFF
--- a/src/chess/board.cc
+++ b/src/chess/board.cc
@@ -201,6 +201,17 @@ struct MagicParams {
   // Number of bits to shift.
   uint8_t shift_bits_;
 #endif
+
+  uint64_t Index(BitBoard occupancy) const {
+#if defined(NO_PEXT)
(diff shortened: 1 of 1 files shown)


mooskagh:
Good idea, but probably even better to move all such code to something like `utils/bits.h`, full of inline functions.

Naphthalin:
As this is a minor code style change and can still be merged, a merge decision should be met quickly.

mooskagh:
Yeah my suggestion was to move the function into `utils/bits.h`

benini:
Of course you know the organization of the lc0 code better than I do, but in my opinion the current file is the best location for the code. If _pext_u64 is not available, the code does a different thing (magic numbers) and cannot be used in other contexts, but only for the generation of rook and bishop moves.

PR #821, opened 2019-03-28
Additional reporting details for selfplay to support T50+

Tilps:


src/selfplay/game.cc
```diff
@@ -171,6 +176,15 @@ std::vector<Move> SelfPlayGame::GetMoves() const {
 }
 
 float SelfPlayGame::GetWorstEvalForWinnerOrDraw() const {
+  if (options_[0].uci_options->Get<bool>(kResignWDLStyleId.GetId())) {
```
mooskagh: Is it theoretically possible that player1 and player2 have different kResignWDLStyle values set? Then `options_[0]` won't work.

Tilps: Theoretically - yes. But not using the same resign style for both sides seems pretty silly.

src/selfplay/game.cc
```diff
@@ -117,15 +119,18 @@ void SelfPlayGame::Play(int white_threads, int black_threads, bool training,
     eval = (eval + 1) / 2;
     if (eval < min_eval_[idx]) min_eval_[idx] = eval;
     const int move_number = tree_[0]->GetPositionHistory().GetLength() / 2 + 1;
+    auto best_w = (best_eval.first + 1.0f - best_eval.second) / 2.0f;
```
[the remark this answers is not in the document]
Tilps: Feels a bit like overkill.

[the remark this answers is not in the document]
mooskagh: For me it doesn't, but ok. :)

src/selfplay/game.cc
mooskagh: If it's <=3 lines of code change to add it, would be nice to support it to avoid future surprises.

Tilps: Actually its not.  Its not the final position that matters for this function, its every single position so far. So its a bunch more work to support them having different values.

PR #700, opened 2019-01-26
Basic certainty propagation

Videodr0me:
Basic certainty propagation.

Please enable by setting the option "certainty-propagation" to "true" (default is disabled = false). 
Also it is highly recommended to also set "two-fold-draw-scoring" to "true" (default is normal threefold draw scoring = false). The two options are independent of each other, so you can also try leela with two-fold draw scoring without turning on certainty propagation.

This is PR#2 of the sliced PR 487 changes. In comparison to 487 it reduces code complexity by leaving out:

- Negabound search function
- One-ply-look ahead
- Informative tournament stats (see PR#1) 
- certainty-depth-parameter (as there is no look-ahead)

It retains a small elo plus, does display mate, displayes correct scores in tb positions (instead of leelas eval), and features optional 2-fold draw scoring. On the plus side:

+ CPU overhead is marginal, so no need to increase threads on very fast multiple GPUs systems. Basically  CP is a parameter-less on/off option.

As there are ongoing efforts to speed-up the move generator and include a transpositon/caching scheme that includes terminal/certain nodes and efficiently handles in game transpositions, its best to first get this basic version into master, and only later add the look-ahead-search. For more info check PR#487

Issue #775, opened 2019-03-02
Noise doesn't work with tree reuse


Ttl:
Originally reported by aart on discord a while ago. It seems that when root node is made from tree reuse dirichlet noise is not applied.

Works if  `TrimTreeAtHead` is called. Self-play calls it so this doesn't apply to training.

Logs: https://pastebin.com/hC7KBc7u

The first move has noised policy as it's not generated by tree-reuse, but the second move has identical policy with and without noise.

Mardak:
If this issue becomes more general to fix up tree-reuse issues, a separate thing to address is from https://github.com/LeelaChessZero/lc0/pull/822#issuecomment-484213160 where a move reusing trees with a terminal child could make the new head terminal and prevent its children from getting visits.

Naphthalin:
The question probably is whether there is any use case for tree reuse + dirichlet noise at the same time where NN cache can't replace the tree reuse; as we use noise only in training afaik, this doesn't seem to be the case (?)

mooskagh:
Probably it's a good idea to throw and error in this case (dirichlet noise + tree reuse), but it's not clear whether it's actually a good idea.
In any case, I think it makes sense to close this issue as unfeasible.

PR #822, opened 2019-03-29
Also make parents Terminal if any move is a win or all moves are loss or draw.

Mardak:
r?@mooskagh Not as feature-full as #700, but this behaves more like TB terminals. Convert a node when search notices children have terminal behavior so future search can stop sooner. Slight refactoring of NodeToProcess to as TerminalHit behaved the same as Extension(1).

Example:
Screen Shot 2019-03-29 at 2 27 33 PM

```
./lc0 -w 11258 --verbose-move-stats --fpu-strategy-at-root=absolute --minibatch-size=1 --smart-pruning-factor=0
position fen 3k4/p4Q2/3N3p/1p1B4/8/1P2P3/2R2PPP/6K1 w - - 1 31 moves f7a7

# just loading the position
go nodes 1    
h6h5 N:       0 (+ 0) (P: 47.68%) (Q:  1.00000) (V:  -.----)
b5b4 N:       0 (+ 0) (P: 52.32%) (Q:  1.00000) (V:  -.----)

# root fpu=1 visit both; nothing special
go nodes 2
h6h5 N:       1 (+ 0) (P: 47.68%) (Q: -0.97491) (V: -0.9749)
b5b4 N:       1 (+ 0) (P: 52.32%) (Q: -0.97261) (V: -0.9726)

# b5b4 finds c2c8#
go nodes 3
h6h5 N:       1 (+ 0) (P: 47.68%) (Q: -0.97491) (V: -0.9749)
b5b4 N:       2 (+ 0) (P: 52.32%) (Q: -0.98630) (V: -0.9726)

# b5b4 c2c8# converts b5b4 into a loss
go nodes 5
h6h5 N:       2 (+ 0) (P: 47.68%) (Q: -0.98746) (V: -0.9749)
b5b4 N:       3 (+ 0) (P: 52.32%) (Q: -1.00000) (V: -1.0000) (T)

# h6h5 c2c8# converts h6h5 into a loss
go nodes 7
h6h5 N:       3 (+ 0) (P: 47.68%) (Q: -1.00000) (V: -1.0000) (T)
b5b4 N:       4 (+ 0) (P: 52.32%) (Q: -1.00000) (V: -1.0000) (T)
```

Then looking at one move before:
Screen Shot 2019-03-29 at 1 50 47 PM

```
position fen 3k4/p4Q2/3N3p/1p1B4/8/1P2P3/2R2PPP/6K1 w - - 1 31

# just normal out of order finding checkmate moves
go nodes 10
c2c6 N:       0 (+ 0) (P:  1.94%) (Q:  1.00000) (V:  -.----)
f7c7 N:       0 (+ 0) (P:  1.97%) (Q:  1.00000) (V:  -.----)
d6c8 N:       0 (+ 0) (P:  1.98%) (Q:  1.00000) (V:  -.----)
d5e4 N:       0 (+ 0) (P:  2.06%) (Q:  1.00000) (V:  -.----)
d5b7 N:       1 (+ 0) (P:  2.21%) (Q:  0.97211) (V:  0.9721)
d5c6 N:       1 (+ 0) (P:  2.16%) (Q:  0.97785) (V:  0.9778)
c2c7 N:       1 (+ 0) (P:  2.24%) (Q:  0.97764) (V:  0.9776)
d6b7 N:       1 (+ 0) (P:  3.93%) (Q:  1.00000) (V:  1.0000) (T)
f7e8 N:       2 (+ 0) (P:  4.16%) (Q:  1.00000) (V:  1.0000) (T)
c2c8 N:       3 (+ 0) (P:  7.86%) (Q:  1.00000) (V:  1.0000) (T)

# just normal search visits
go nodes 400
d6c8 N:       8 (+ 0) (P:  1.98%) (Q:  0.98583) (V:  0.9752)
d5e4 N:       8 (+ 0) (P:  2.06%) (Q:  0.98539) (V:  0.9736)
f7c7 N:       9 (+ 0) (P:  1.97%) (Q:  1.00000) (V:  1.0000) (T)
d5b7 N:       9 (+ 0) (P:  2.21%) (Q:  0.98568) (V:  0.9721)
d5c6 N:       9 (+ 0) (P:  2.16%) (Q:  0.99107) (V:  0.9778)
c2c7 N:       9 (+ 0) (P:  2.24%) (Q:  0.98915) (V:  0.9776)
d5e6 N:      11 (+ 0) (P:  2.53%) (Q:  0.99073) (V:  0.9777)
d6b7 N:      18 (+ 0) (P:  3.93%) (Q:  1.00000) (V:  1.0000) (T)
f7e8 N:      20 (+ 0) (P:  4.16%) (Q:  1.00000) (V:  1.0000) (T)
c2c8 N:      37 (+ 0) (P:  7.86%) (Q:  1.00000) (V:  1.0000) (T)

# f7a7 (from first example) has enough visits that all its moves are losses, so convert to win
go nodes 500
f7a7 N:       9 (+ 0) (P:  1.52%) (Q:  1.00000) (V:  1.0000) (T)
f7f6 N:      10 (+ 0) (P:  1.74%) (Q:  1.00000) (V:  1.0000) (T)
f7c7 N:      11 (+ 0) (P:  1.97%) (Q:  1.00000) (V:  1.0000) (T)
d5b7 N:      11 (+ 0) (P:  2.21%) (Q:  0.98829) (V:  0.9721)
d5c6 N:      11 (+ 0) (P:  2.16%) (Q:  0.99269) (V:  0.9778)
c2c7 N:      12 (+ 0) (P:  2.24%) (Q:  0.99186) (V:  0.9776)
d5e6 N:      13 (+ 0) (P:  2.53%) (Q:  0.99216) (V:  0.9777)
d6b7 N:      23 (+ 0) (P:  3.93%) (Q:  1.00000) (V:  1.0000) (T)
f7e8 N:      24 (+ 0) (P:  4.16%) (Q:  1.00000) (V:  1.0000) (T)
c2c8 N:      46 (+ 0) (P:  7.86%) (Q:  1.00000) (V:  1.0000) (T)

# lots of moves converting to win
go nodes 600
f7a7 N:      10 (+ 0) (P:  1.52%) (Q:  1.00000) (V:  1.0000) (T)
f7f6 N:      12 (+ 0) (P:  1.74%) (Q:  1.00000) (V:  1.0000) (T)
b3b4 N:      13 (+ 0) (P:  1.87%) (Q:  1.00000) (V:  1.0000) (T)
f7c7 N:      13 (+ 0) (P:  1.97%) (Q:  1.00000) (V:  1.0000) (T)
d5c6 N:      15 (+ 0) (P:  2.16%) (Q:  1.00000) (V:  1.0000) (T)
d5b7 N:      15 (+ 0) (P:  2.21%) (Q:  1.00000) (V:  1.0000) (T)
c2c7 N:      15 (+ 0) (P:  2.24%) (Q:  1.00000) (V:  1.0000) (T)
d5e6 N:      17 (+ 0) (P:  2.53%) (Q:  1.00000) (V:  1.0000) (T)
d6b7 N:      27 (+ 0) (P:  3.93%) (Q:  1.00000) (V:  1.0000) (T)
f7e8 N:      28 (+ 0) (P:  4.16%) (Q:  1.00000) (V:  1.0000) (T)
c2c8 N:      54 (+ 0) (P:  7.86%) (Q:  1.00000) (V:  1.0000) (T)
```

### DIFF
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -393,6 +393,9 @@ void NodeTree::MakeMove(Move move) {
   for (auto& n : current_head_->Edges()) {
     if (n.GetMove() == move) {
       new_head = n.GetOrSpawnNode(current_head_);
+      // Ensure head is not terminal, so search can extend or visit children of
+      // "terminal" positions, e.g., WDL hits, converted terminals, 3-fold draw.
+      if (new_head->IsTerminal()) new_head->MakeNotTerminal();
       break;
     }
   }
@@ -442,11 +445,6 @@ bool NodeTree::ResetToPosition(const std::string& starting_fen,
   // retain old n_ and q_ (etc) data, even though its old children were
   // previously trimmed; we need to reset current_head_ in that case.
   if (!seen_old_head) TrimTreeAtHead();
-
-  // Make sure the head is not terminal, so search can extend or visit children
-  // of "terminal" positions, e.g., WDL hits, converted terminals, 3-fold draw.
-  if (current_head_->IsTerminal()) current_head_->MakeNotTerminal();
-
   return seen_old_head;
 }
 
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -241,12 +241,13 @@ void Node::MakeNotTerminal() {
       const auto n = child.GetN();
       if (n > 0) {
         n_ += n;
-        q_ -= child.GetQ(0.0f) * n;
+        // Flip Q for opponent.
+        q_ += -child.GetQ(0.0f) * n;
         d_ += child.GetD() * n;
       }
     }
 
-    // Recompute with current eval (instead of network's) and children's eval
+    // Recompute with current eval (instead of network's) and children's eval.
     q_ /= n_;
     d_ /= n_;
   }
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -230,6 +230,17 @@ void Node::MakeTerminal(GameResult result) {
   }
 }
 
+void Node::MakeNotTerminal() {
+  is_terminal_ = false;
+  n_ = 0;
+
+  // If we have edges, we've been extended (1 visit), so include children too.
+  if (edges_) {
+    n_++;
+    for (const auto& child : Edges()) n_ += child.GetN();
+  }
+}
+
 bool Node::TryStartScoreUpdate() {
   if (n_ == 0 && n_in_flight_ > 0) return false;
   ++n_in_flight_;
@@ -430,9 +441,11 @@ bool NodeTree::ResetToPosition(const std::string& starting_fen,
   // previously searched position, which means that the current_head_ might
   // retain old n_ and q_ (etc) data, even though its old children were
   // previously trimmed; we need to reset current_head_ in that case.
-  // Also, if the current_head_ is terminal, reset that as well to allow forced
-  // analysis of WDL hits, or possibly 3 fold or 50 move "draws", etc.
-  if (!seen_old_head || current_head_->IsTerminal()) TrimTreeAtHead();
+  if (!seen_old_head) TrimTreeAtHead();
+
+  // Make sure the head is not terminal, so search can extend or visit children
+  // of "terminal" positions, e.g., WDL hits, converted terminals, 3-fold draw.
+  if (current_head_->IsTerminal()) current_head_->MakeNotTerminal();
 
   return seen_old_head;
 }
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -237,7 +237,18 @@ void Node::MakeNotTerminal() {
   // If we have edges, we've been extended (1 visit), so include children too.
   if (edges_) {
     n_++;
-    for (const auto& child : Edges()) n_ += child.GetN();
+    for (const auto& child : Edges()) {
+      const auto n = child.GetN();
+      if (n > 0) {
+        n_ += n;
+        q_ -= child.GetQ(0.0f) * n;
+        d_ += child.GetD() * n;
+      }
+    }
+
+    // Recompute with current eval (instead of network's) and children's eval
+    q_ /= n_;
+    d_ /= n_;
   }
 }
 
--- a/src/mcts/params.cc
+++ b/src/mcts/params.cc
@@ -152,6 +152,13 @@ const OptionId SearchParams::kOutOfOrderEvalId{
     "in the cache or is terminal, evaluate it right away without sending the "
     "batch to the NN. When off, this may only happen with the very first node "
     "of a batch; when on, this can happen with any node."};
+const OptionId SearchParams::kStickyEndgamesId{
+    "sticky-endgames", "StickyEndgames",
+    "When an end of game position is found during search, allow the eval of "
+    "the previous move's position to stick to something more accurate. For "
+    "example, if at least one move results in checkmate, then the position "
+    "should stick as checkmated. Similarly, if all moves are drawn or "
+    "checkmated, the position should stick as drawn or checkmate."};
 const OptionId SearchParams::kSyzygyFastPlayId{
     "syzygy-fast-play", "SyzygyFastPlay",
     "With DTZ tablebase files, only allow the network pick from winning moves "
@@ -209,6 +216,7 @@ void SearchParams::Populate(OptionsParser* options) {
   options->Add<IntOption>(kMaxCollisionEventsId, 1, 1024) = 32;
   options->Add<IntOption>(kMaxCollisionVisitsId, 1, 1000000) = 9999;
   options->Add<BoolOption>(kOutOfOrderEvalId) = true;
+  options->Add<BoolOption>(kStickyEndgamesId) = false;
   options->Add<BoolOption>(kSyzygyFastPlayId) = true;
   options->Add<IntOption>(kMultiPvId, 1, 500) = 1;
   std::vector<std::string> score_type = {"centipawn", "win_percentage", "Q"};
@@ -244,6 +252,7 @@ SearchParams::SearchParams(const OptionsDict& options)
       kMaxCollisionEvents(options.Get<int>(kMaxCollisionEventsId.GetId())),
       kMaxCollisionVisits(options.Get<int>(kMaxCollisionVisitsId.GetId())),
       kOutOfOrderEval(options.Get<bool>(kOutOfOrderEvalId.GetId())),
+      kStickyEndgames(options.Get<bool>(kStickyEndgamesId.GetId())),
       kSyzygyFastPlay(options.Get<bool>(kSyzygyFastPlayId.GetId())),
       kHistoryFill(
           EncodeHistoryFill(options.Get<std::string>(kHistoryFillId.GetId()))),
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -1301,10 +1301,9 @@ void SearchWorker::DoBackupUpdateSingleNode(
 
     // A non-winning terminal move needs all other moves to have the same value.
     if (can_convert && v != 1.0f) {
-      auto edges = p->Edges();
-      can_convert = std::all_of(edges.begin(), edges.end(), [&](const auto& e) {
-        return e.IsTerminal() && e.GetQ(0.0f) == v;
-      });
+      for (const auto& edge : p->Edges()) {
+        can_convert = can_convert && edge.IsTerminal() && edge.GetQ(0.0f) == v;
+      }
     }
 
     // Convert the parent to a terminal loss if at least one move is winning or
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -1279,7 +1279,8 @@ void SearchWorker::DoBackupUpdateSingleNode(
   }
 
   // For the first visit to a terminal, maybe convert ancestors to terminal too.
-  auto can_convert = node->IsTerminal() && !node->GetN();
+  auto can_convert =
+      params_.GetStickyEndgames() && node->IsTerminal() && !node->GetN();
 
   // Backup V value up to a root. After 1 visit, V = Q.
   float v = node_to_process.v;
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -906,12 +906,8 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend(
       return NodeToProcess::Collision(node, depth, collision_limit);
     }
     // Either terminal or unexamined leaf node -- the end of this playout.
-    if (!node->HasChildren()) {
-      if (node->IsTerminal()) {
-        return NodeToProcess::TerminalHit(node, depth, 1);
-      } else {
-        return NodeToProcess::Extension(node, depth);
-      }
+    if (node->IsTerminal() || !node->HasChildren()) {
+      return NodeToProcess::Visit(node, depth);
     }
     Node* possible_shortcut_child = node->GetCachedBestChild();
     if (possible_shortcut_child) {
@@ -1282,18 +1278,48 @@ void SearchWorker::DoBackupUpdateSingleNode(
     return;
   }
 
+  // For the first visit to a terminal, maybe convert ancestors to terminal too.
+  auto can_convert = node->IsTerminal() && !node->GetN();
+
   // Backup V value up to a root. After 1 visit, V = Q.
   float v = node_to_process.v;
   float d = node_to_process.d;
-  for (Node* n = node; n != search_->root_node_->GetParent();
-       n = n->GetParent()) {
+  for (Node *n = node, *p; n != search_->root_node_->GetParent(); n = p) {
+    p = n->GetParent();
+
+    // Current node might have become terminal from some other descendant, so
+    // backup the rest of the way with more accurate values.
+    if (n->IsTerminal()) {
+      v = n->GetQ();
+      d = n->GetD();
+    }
     n->FinalizeScoreUpdate(v, d, node_to_process.multivisit);
+
+    // Convert parents to terminals except the root or those already converted.
+    can_convert = can_convert && p != search_->root_node_ && !p->IsTerminal();
+
+    // A non-winning terminal move needs all other moves to have the same value.
+    if (can_convert && v != 1.0f) {
+      auto edges = p->Edges();
+      can_convert = std::all_of(edges.begin(), edges.end(), [&](const auto& e) {
+        return e.IsTerminal() && e.GetQ(0.0f) == v;
+      });
+    }
+
+    // Convert the parent to a terminal loss if at least one move is winning or
+    // to a terminal win or draw if all moves are loss or draw respectively.
+    if (can_convert) {
+      p->MakeTerminal(v == 1.0f ? GameResult::BLACK_WON
+                                : v == -1.0f ? GameResult::WHITE_WON
+                                             : GameResult::DRAW);
+    }
+
     // Q will be flipped for opponent.
     v = -v;
 
     // Update the stats.
     // Best move.
-    if (n->GetParent() == search_->root_node_ &&
+    if (p == search_->root_node_ &&
         search_->current_best_edge_.GetN() <= n->GetN()) {
       search_->current_best_edge_ =
           search_->GetBestChildNoTemperature(search_->root_node_);
--- a/src/mcts/search.h
+++ b/src/mcts/search.h
@@ -271,12 +271,8 @@ class SearchWorker {
                                    int collision_count) {
       return NodeToProcess(node, depth, true, collision_count);
     }
-    static NodeToProcess Extension(Node* node, uint16_t depth) {
+    static NodeToProcess Visit(Node* node, uint16_t depth) {
       return NodeToProcess(node, depth, false, 1);
-    }
-    static NodeToProcess TerminalHit(Node* node, uint16_t depth,
-                                     int visit_count) {
-      return NodeToProcess(node, depth, false, visit_count);
     }
 
    private:
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -162,6 +162,8 @@ class Node {
 
   // Makes the node terminal and sets it's score.
   void MakeTerminal(GameResult result);
+  // Makes the node not terminal and updates its visits.
+  void MakeNotTerminal();
 
   // If this node is not in the process of being expanded by another thread
   // (which can happen only if n==0 and n-in-flight==1), mark the node as
--- a/src/mcts/params.h
+++ b/src/mcts/params.h
@@ -85,6 +85,7 @@ class SearchParams {
   int GetMaxCollisionEvents() const { return kMaxCollisionEvents; }
   int GetMaxCollisionVisitsId() const { return kMaxCollisionVisits; }
   bool GetOutOfOrderEval() const { return kOutOfOrderEval; }
+  bool GetStickyEndgames() const { return kStickyEndgames; }
   bool GetSyzygyFastPlay() const { return kSyzygyFastPlay; }
   int GetMultiPv() const { return options_.Get<int>(kMultiPvId.GetId()); }
   std::string GetScoreType() const {
@@ -123,6 +124,7 @@ class SearchParams {
   static const OptionId kMaxCollisionEventsId;
   static const OptionId kMaxCollisionVisitsId;
   static const OptionId kOutOfOrderEvalId;
+  static const OptionId kStickyEndgamesId;
   static const OptionId kSyzygyFastPlayId;
   static const OptionId kMultiPvId;
   static const OptionId kScoreTypeId;
@@ -152,6 +154,7 @@ class SearchParams {
   const int kMaxCollisionEvents;
   const int kMaxCollisionVisits;
   const bool kOutOfOrderEval;
+  const bool kStickyEndgames;
   const bool kSyzygyFastPlay;
   const FillEmptyHistory kHistoryFill;
   const int kMiniBatchSize;


Mardak:
The behavior on search with this change:

1) convert to loss when any child is checkmate prevents visiting other children, so this saves all future visits to these children
2) nodes converted to loss are less likely to be visited by search favoring Q>-1 moves
3) convert to win when all children are checkmated, so this node is more likely to be visited with Q=1 instead of other moves that might require NN lookup

No direct changes to root behavior to allow searching moves as normal except some of its children will have more accurate Q to direct search.

Mardak:
Here's the position that inspired me to write up this PR.

Screen Shot 2019-03-29 at 10 04 53 AM

All runs miss this clearance / defense annihilation move Qxh7 (11258 0.26%, 22202 0.31%, 32930 0.27%, 41750 0.29%, 50783 0.17%) https://lichess.org/VGNIwD67#70

Looking at the move before the screenshot to try to avoid blundering with Qxa3:
```
./lc0 -w 11258 --verbose-move-stats --smart-pruning-factor=0
position startpos moves g1f3 e7e6 c2c4 g8f6 d2d4 f8b4 c1d2 b4e7 a2a3 d7d5 e2e3 e8g8 b1c3 b8d7 d1c2 b7b6 c4d5 e6d5 f1e2 c8b7 f3e5 a7a6 e1g1 c7c5 d2e1 a8c8 c2f5 d8c7 f2f4 c8d8 a1c1 f8e8 e1h4 b6b5 e2d3 d7f8 f5h3 f6e4 c3e4 d5e4 d3b1 c5c4 f4f5 e7h4 h3h4 f7f6 e5g4 f8d7 g4f2 d7b6 f2e4 b7e4 b1e4 e8e7 c1e1 d8e8 f1f4 g8h8 e4f3 c4c3 b2c3 c7c3 h2h3 e7e3 e1e3 c3e3 g1h2 b6c4 f3d5

# without PR
go nodes 800
c4d2 N:      47 (+ 0) (P: 13.63%) (Q:  0.00268) (V:  0.0637)
e3a3 N:     580 (+ 0) (P: 14.29%) (Q:  0.23623) (V:  0.3569)

go nodes 1600
h7h6 N:     483 (+ 1) (P:  3.92%) (Q:  0.18075) (V:  0.0814)
e3a3 N:     625 (+ 0) (P: 14.29%) (Q:  0.14722) (V:  0.3569)

# with PR
go nodes 800
c4d2 N:      48 (+ 0) (P: 13.63%) (Q:  0.00272) (V:  0.0637) 
e3a3 N:     564 (+ 0) (P: 14.29%) (Q: -1.00000)(V: -1.0000) (T) 

go nodes 1600
e3a3 N:     564 (+ 0) (P: 14.29%) (Q: -1.00000) (V: -1.0000) (T) 
h7h6 N:     568 (+ 0) (P:  3.92%) (Q:  0.17601) (V:  0.0814) 
```

Where with the PR, the checkmate move makes the forced king move a loss which then makes Qxh7 become a win and the blundering Qxa3 a loss. With Qxa3 appearing as Q: -1, additional visits avoid it and quickly find an alternate h6 move.

Mardak:
I modified the position/fen of the previous comment as a TB sanity check for a 10-piece position:

Screen Shot 2019-03-29 at 6 42 58 PM

```
position fen 7k/6pp/5p2/1p6/2q2R1Q/1B6/7K/8 w - - 0 36

go nodes 1600000
…
h4g3 N:   14626 (+ 1) (P:  2.27%) (Q:  0.98233) (V:  0.9885)
h4g4 N:   15603 (+ 3) (P:  2.35%) (Q:  0.98313) (V:  0.9657)
f4c4 N:  414512 (+ 0) (P: 18.75%) (Q:  1.00000) (V:  1.0000) (T)
b3c4 N: 1277607 (+ 0) (P: 33.73%) (Q:  1.00000) (V:  1.0000) (T)
```

Where after capturing c4 queen with Rxc4 or Bxc4, eventually all possible black moves are found to be losses because some followup white move is found to be a TB win. Verbose move stats showing (T) for a 10 piece position, neat! 😄

Mardak:
Here's a comparison with 1 million nodes before/after PR for a test position:
Screen Shot 2019-03-29 at 8 18 15 PM

```
./lc0 -w 50783 --verbose-move-stats --smart-pruning-factor=0
position fen 7k/6pp/2p2p2/1R6/2r2R1Q/1B6/7K/8 b - - 0 36
go nodes 1000000

# without PR
info depth 7 seldepth 18 time 276485 nodes 1000095 score cp -6885 hashfull 1000 nps 3617 tbhits 1410 pv g7g5 h4h5 c4e4 h5h6 e4e7 h6h3 h7h5 h2g1 e7g7 g1g2 g7a7
c4c3 N:   42000 (+ 5) (P:  5.92%) (Q: -0.99767) (V:  -.----)
c4a4 N:   42258 (+ 3) (P:  5.33%) (Q: -0.99606) (V:  -.----)
c6c5 N:   42299 (+ 5) (P:  5.45%) (Q: -0.99634) (V:  -.----)
c4c2 N:   42655 (+ 8) (P:  5.10%) (Q: -0.99534) (V:  -.----)
c4c1 N:   43181 (+ 3) (P:  5.85%) (Q: -0.99708) (V:  -.----)
c4c5 N:   47580 (+ 3) (P:  6.12%) (Q: -0.99633) (V:  -.----)
c4d4 N:   48964 (+ 3) (P:  5.87%) (Q: -0.99537) (V:  -.----)
f6f5 N:   49652 (+ 6) (P:  6.09%) (Q: -0.99567) (V: -0.9950)
h8g8 N:   50759 (+15) (P:  6.11%) (Q: -0.99543) (V:  -.----)
c6b5 N:   51637 (+ 9) (P:  5.58%) (Q: -0.99409) (V:  -.----)
c4e4 N:   52993 (+ 7) (P:  6.24%) (Q: -0.99515) (V:  -.----)
c4b4 N:   56625 (+ 6) (P:  5.54%) (Q: -0.99297) (V:  -.----)
c4f4 N:   68245 (+ 4) (P:  5.26%) (Q: -0.99071) (V: -0.9991)
h7h5 N:   70686 (+17) (P:  6.45%) (Q: -0.99226) (V:  -.----)
g7g6 N:   76761 (+68) (P:  6.05%) (Q: -0.99088) (V: -0.9943)
h7h6 N:   79466 (+13) (P:  6.72%) (Q: -0.99153) (V: -0.9928)
g7g5 N:  134333 (+ 0) (P:  6.30%) (Q: -0.98741) (V: -0.9940)

# with PR
info depth 4 seldepth 18 time 181733 nodes 1000288 score cp -7273 hashfull 1000 nps 5504 tbhits 756 pv g7g5 h4h5 c4e4 h5h6 e4e7 h6h3 e7e2 h2g1 e2g2 g1h1
c4f4 N:   38452 (+ 0) (P:  5.26%) (Q: -1.00000) (V: -1.0000) (T)
c4a4 N:   38976 (+ 0) (P:  5.33%) (Q: -1.00000) (V: -1.0000) (T)
c6c5 N:   39857 (+ 0) (P:  5.45%) (Q: -1.00000) (V: -1.0000) (T)
c4c1 N:   42779 (+ 0) (P:  5.85%) (Q: -1.00000) (V: -1.0000) (T)
c4d4 N:   42879 (+ 0) (P:  5.87%) (Q: -1.00000) (V: -1.0000) (T)
c4c3 N:   43259 (+ 0) (P:  5.92%) (Q: -1.00000) (V: -1.0000) (T)
g7g6 N:   44184 (+ 0) (P:  6.05%) (Q: -1.00000) (V: -1.0000) (T)
f6f5 N:   44485 (+ 0) (P:  6.09%) (Q: -1.00000) (V: -1.0000) (T)
c4c5 N:   44731 (+ 0) (P:  6.12%) (Q: -1.00000) (V: -1.0000) (T)
c4e4 N:   45634 (+ 0) (P:  6.24%) (Q: -1.00000) (V: -1.0000) (T)
c4c2 N:   46594 (+ 0) (P:  5.10%) (Q: -0.99702) (V:  -.----)
h7h5 N:   47173 (+ 0) (P:  6.45%) (Q: -1.00000) (V: -1.0000) (T)
c6b5 N:   61614 (+18) (P:  5.58%) (Q: -0.99496) (V:  -.----)
h8g8 N:   63343 (+ 5) (P:  6.11%) (Q: -0.99560) (V:  -.----)
c4b4 N:   64565 (+ 5) (P:  5.54%) (Q: -0.99444) (V:  -.----)
h7h6 N:  110478 (+17) (P:  6.72%) (Q: -0.99171) (V: -0.9939)
g7g5 N:  181284 (+13) (P:  6.30%) (Q: -0.98886) (V:  -.----)
```

In this case, the search differences with PR:
- g7g5 pv has 35% more visits (46951 more visits)
- 1m visits took 34% less time (95 fewer seconds)
- final nps 52% higher (1887 extra nps)


And testing from startpos but just 100k nodes:
```
# without PR
info depth 12 seldepth 39 time 46118 nodes 100002 score cp 44 hashfull 302 nps 2168 tbhits 0 pv g1f3 d7d5 d2d4 e7e6 g2g3 c7c5 f1g2 g8f6 e1g1 c5d4 f3d4 e6e5 d4b3 c8e6 c2c4 b8c6 c4d5 f6d5 b1d2 d5f6 d2e4 f6e4 g2e4 e6h3 f1e1 f8b4 c1d2 b4d2 d1d2 d8d2 b3d2

# with PR
info depth 12 seldepth 39 time 46122 nodes 100275 score cp 44 hashfull 302 nps 2174 tbhits 0 pv g1f3 d7d5 d2d4 e7e6 g2g3 c7c5 f1g2 g8f6 e1g1 c5d4 f3d4 e6e5 d4b3 c8e6 c2c4 b8c6 c4d5 f6d5 b1d2 d5f6 d2e4 f6e4 g2e4 e6h3 f1e1 f8b4 c1d2 b4d2 d1d2 d8d2 b3d2
```
Looks like no noticeable difference in speed.

Mardak:
@Ttl One interesting aspect I noticed with nodes becoming terminal losses is the visits remain so if what search had been thinking is the best move leads to the opponent having a terminal win, search needs enough visits for a 2nd-best move to exceed and be picked even though we know for sure the most visited move is bad. Maybe we should pick move to play based on lower confidence bound like https://github.com/leela-zero/leela-zero/pull/2290 ?

E.g., from https://github.com/LeelaChessZero/lc0/pull/822#issuecomment-478164951
```
go nodes 800
c4d2 N:      48 (+ 0) (P: 13.63%) (Q:  0.00272) (V:  0.0637) 
e3a3 N:     564 (+ 0) (P: 14.29%) (Q: -1.00000) (V: -1.0000) (T) 

go nodes 1600
e3a3 N:     564 (+ 0) (P: 14.29%) (Q: -1.00000) (V: -1.0000) (T) 
h7h6 N:     568 (+ 0) (P:  3.92%) (Q:  0.17601) (V:  0.0814) ```
It would be unfortunate if e3a3 was picked somewhere between 800 visits and 1590 visits even though we know we shouldn't play it. But even then, the behavior is still no worse than the existing search picking this bad move.

Also somewhat related is these terminal loss moves probably shouldn't get any more visits, but we end up "wasting" CPU cycles adjusting the "uncertainty" for these moves.

Both of these aren't really specific to this issue though, so probably be their own PRs.

**Comment by ddobbelaere** [comment]
Nice PR. I like the concept that you conclude that a node is terminal during _playout_ instead of during backup, which is very neat. The code is also very non-obtrusive it has to be said.

I think you can even go one step further and also infer certain draws. In the logic, just add a `should_become_draw path` based on `Q == 0 && node->IsTerminal()`. This way it might also help with three fold reps, perpetuals, stalemate etc.

**Comment by Mardak** [comment]
>  think you can even go one step further and also infer certain draws. 

Shhh 😉 I was trying to keep this PR small. Yes, it should just be adding a similar to should_become_draw declaration, accumulation, and conditional termination. Although I wasn't really sure how often all possible moves show up as certain drawn…

**Comment by Videodr0me** [comment]
Hi there. I opted against doing it in picking nodes because generally there are only few certain nodes and the code in backpropagatin only adds overhead if and only if the node to process is actually certain (or has any bounds, as i also determine bounds). So it has no overhead for all other (the vast majority) of paths explored in node picking. Also (not currently done in PR487 or PR700) in backprop you can cease to do any checks if there has been no change in certainty state when going up a level.

That being said, you could easily move your code to backprop with almost no changes. Your code is node-centric which makes it fit well with current structure of backprop and picknode. I also contemplated that but decided against doing it. The main reasons are in how i envision the future of search: 
(1)  transpositions 
(2)  look-ahead search
In both cases require a more edge centric view as nodes can be reached via different edges and edges will need to hold path depended information. Also with the look-ahead we do no longer have to create nodes that we know to be certain (as the edges already contain that information). This was under the paradigm: reduce the memory footprint as much as possible.

While this probably was not the best decision in terms of "short-term" mergability, i still believe that it will be unavoidable in the long-term if we ever want to handle transpositions and/or lookahead search efficiently. 

If you want to retain as much of the functionality of PR487 or PR700 as possibly. You could just move the certainty status to the node structure. Then the code changes collapse/simplify greatly in the getters, the only thing you have to ditch is the look-ahead or PR487.

Some other remarks. You may have a problem if a terminal node gets to be root if opponent plays into mate. So you might have to unterminal root when initializing search. Also if you want to retain a previously calculated mating sequence, you need to change the trim at head behaviour, as currently you trim if the head is terminal, basically throwing away the search result leading to the terminal condition. Then you also could check for certain draws and bounds (this is very effective for discovering draws). 

I am back from holidays and will update PR487and PR700 to a new PR called "Advanced Search". It will be more optimized (e.g. edge centric backprop) and over time I will add some additional properties:
1) All Nodes along the PV will be expanded A0 style by filling non-complete batches. This improves on FPU_at_root options by extending that concept to all nodes along the PV. It comes almost at no cost as it replaces the inefficient prefetch into cache. 
2) Reduced trolling by preferring terminal mates over proven mating sequences over TB wins and more efficient use of dtz information 
3) distance to mate measure

**Comment by Videodr0me** [comment]
> 
> @Ttl One interesting aspect I noticed with nodes becoming terminal losses is the visits remain so if what search had been thinking is the best move leads to the opponent having a terminal win, search needs enough visits for a 2nd-best move to exceed and be picked even though we know for sure the most visited move is bad. Maybe we should pick move to play based on lower confidence bound like [leela-zero/leela-zero#2290](https://github.com/leela-zero/leela-zero/pull/2290) ?
> 
No need, you can do what PR700 does and adjust the tuple that is used for GetBestChildrenNoTemperature. That solves this issue.

> > Also somewhat related is these terminal loss moves probably shouldn't get any more visits, but we end up "wasting" CPU cycles adjusting the "uncertainty" for these moves.
>
This can only be done safely at root. As terminal losses need to be hit (even repeatedly) to inform search about the inferiority of the path. If policy guided UCT hits these nodes depite having a Q of -1 the line must be really bad and these hits help to start exploring alternatives early. At root you can safely skip these as there is no information left. I tried various schemes (in the tree, not at root), not visiting loosing childs at all, decreasing visit probability with N, decreasing visit probability with depth. The result was that not visiting at all looses elo, results of the more "refined" schemes (N, depth) were inconclusive. So i would strongly advise to only skip loosing moves at root.

**Comment by roy7** [comment]
@Mardak Something I toyed with long ago was, if I decided a child needed pruned (in this case with Chess, a terminal node that is a loss) I'd remove the search and eval data from that move all the way back up the tree to overcome the "averaging" problem of MCTS. This was separate from using max_lcb to choose moves, the idea being instead of needing the 2nd move to get enough visits to outnumber the 1st move, I'd lower the # of visits to the 1st move if I pruned a big branch in the tree. (Lowering the average win rate much faster, and promoiting 2nd move faster than waiting it out.)

Just food for thought if you ever want to try it.

**Comment by Videodr0me** [comment]
> 
> 
> @Mardak Something I toyed with long ago was, if I decided a child needed pruned (in this case with Chess, a terminal node that is a loss) I'd remove the search and eval data from that move all the way back up the tree to overcome the "averaging" problem of MCTS. This was separate from using max_lcb to choose moves, the idea being instead of needing the 2nd move to get enough visits to outnumber the 1st move, I'd lower the # of visits to the 1st move if I pruned a big branch in the tree. (Lowering the average win rate much faster, and promoting 2nd move faster than waiting it out.)
> 
> Just food for thought if you ever want to try it.
Hi, I solved this issue by doing this:
(1) if a subbranch is made certain, which sometimes means a huge swing in eval, I adjust the scores in the tree as if all previous moves to this subbranch already yielded the certain result. This leads to leela changing its mind much more quickly without having to wait for the averaging operator to catch up.
(2) luckily you can do that basically for free, as you can do it on the same backprop that backpropagates the result anyways. Basically  you just adjust the weight of the result and through the linear tree algebra you get a consistent tree just by using the normal backprop mechanism - no additional traversals needed, no extra nodes visited. 
You can see how i did that in PR700 or PR487 by checking the section in backprop where i compare the previous q to the current q and adjust the weights: https://github.com/LeelaChessZero/lc0/blob/d6cfd84e5fc7ca2ee32da90713788b056c3be3fe/src/mcts/search.cc#L1429

**Comment by Mardak** [comment]
Thanks for the suggestions. Terminating parents from `DoBackupUpdateSingleNode` does seem to be slightly more optimal:
1) avoids wasting 1 visit on the forward pass
2) avoids accumulating terminating condition across children in the common case

The downside is that it's slightly more complex requiring a new children loop in each step of backup, but the extra loop should only trigger when a terminal is added to the tree. This conditional behavior aspect would allow backup to do other more complex changes without impacting the forward common case.

**Comment by Videodr0me** [comment]
> Thanks for the suggestions. Terminating parents from `DoBackupUpdateSingleNode` does seem to be slightly more optimal:
> 
>     1. avoids wasting at least 1 visit on the forward pass
> 
>     2. more accurate Q for `PickNodeToExtend` 1 batch earlier
> 
>     3. avoids accumulating terminating condition across children in the common case
> 
> 
> The downside is that it's slightly more complex requiring a new children loop in each step of backup, but the extra loop should only trigger when a parent has a new terminal child. This conditional behavior aspect would allow backup to do other more complex changes without impacting the forward common case.

Exactly. Even if the original node is terminal you do not need to check all childs on every depth. As soon as you get to a node that does not get to be made terminal you can stop checking all childs from that node onward until backprop is finished. Thats the most optimal way and also avoids some other issues with doing it in picknode. If you move it there and also make nodes a terminal draw if all childs are terminal draws then you have the simplest version i can think of.

And if you succeed in getting that merged, we can then add features more features over time. Just for reference here is my findings in how much elo each part of PR487 gains:
(1) Bare bones CP, like this PR is not measurable. But - and thats the nice thing which attracted me to all this - it is at least >=0 for theoretical reasons and needs no tuning.
(2) Bounds checking and score adjusting is about +10 Elo
(3) one-ply-lookahead and 2-fold-draw in conjunction is also about+10Elo
(4) The other features of 487 and 700 do not impact elo, but reduce trolling, display mate, prefer mates over TB hits, play immediatly if mate is proven. These are more "cosmetic" in nature.

**Comment by zz4032** [comment]
TC 15.12s/game + 0.063s/move (~4000nodes/move)
lc0_PR822 with `option.Terminator=true`.
```
   1 lc0          :     0.0   ----    94.9    1000      71.3
   2 lc0_PR822    :    -9.1   10.9     ---    1000      71.3

White ELO advantage = 70.7
Draw rate (equal opponents) = 77.6 %
```

**Comment by Mardak** [comment]
Added a commit to do Terminator from DoBackupUpdateSingleNode instead of PickNodeToExtend. Here's an example of the more efficient visit usage using the first position from the first comment https://github.com/LeelaChessZero/lc0/pull/822#issue-265873448:

```
position fen 3k4/p4Q2/3N3p/1p1B4/8/1P2P3/2R2PPP/6K1 w - - 1 31 moves f7a7

# old PickNodeToExtend behavior
h6h5 N:       1 (+ 1) (P: 47.68%) (Q: -0.97491) (V: -0.9749)
b5b4 N:       2 (+ 0) (P: 52.32%) (Q: -0.98630) (V: -0.9726)

h6h5 N:       2 (+ 0) (P: 47.68%) (Q: -0.98746) (V: -0.9749)
b5b4 N:       3 (+ 0) (P: 52.32%) (Q: -1.00000) (V: -1.0000) (T)

h6h5 N:       3 (+ 0) (P: 47.68%) (Q: -1.00000) (V: -1.0000) (T)
b5b4 N:       4 (+ 0) (P: 52.32%) (Q: -1.00000) (V: -1.0000) (T)

# new DoBackupUpdateSingleNode behavior
h6h5 N:       1 (+ 0) (P: 47.68%) (Q: -0.97491) (V: -0.9749)
b5b4 N:       2 (+ 0) (P: 52.32%) (Q: -1.00000) (V: -1.0000) (T)

h6h5 N:       2 (+ 0) (P: 47.68%) (Q: -1.00000) (V: -1.0000) (T)
b5b4 N:       3 (+ 0) (P: 52.32%) (Q: -1.00000) (V: -1.0000) (T)
```
Instead of taking 7 visits to identify both moves as terminal, the new behavior finds it 2 visits earlier.

Also, the new commit adds in draw checks. Here's a dummy test position:
Screen Shot 2019-03-31 at 6 07 36 PM

```
position fen qk1K4/8/8/8/8/8/8/Q7 w - - 1 1

# progression --no-terminator
a1a8 N:       1 (+ 0) (P:  5.96%) (Q:  0.04149) (V:  0.0415)
a1a8 N:       2 (+ 0) (P:  5.96%) (Q:  0.02074) (V:  0.0415)
a1a8 N:       3 (+ 0) (P:  5.96%) (Q:  0.01383) (V:  0.0415)

# progression --terminator
a1a8 N:       1 (+ 0) (P:  5.96%) (Q:  0.04149) (V:  0.0415)
a1a8 N:       2 (+ 0) (P:  5.96%) (Q:  0.00000) (V:  0.0000) (T)
```

**Comment by Mardak** [comment]
@Ttl Oh nice! I didn't notice your #817 earlier. Here's the output with PRs combined:

```
position startpos moves g1f3 e7e6 c2c4 g8f6 d2d4 f8b4 c1d2 b4e7 a2a3 d7d5 e2e3 e8g8 b1c3 b8d7 d1c2 b7b6 c4d5 e6d5 f1e2 c8b7 f3e5 a7a6 e1g1 c7c5 d2e1 a8c8 c2f5 d8c7 f2f4 c8d8 a1c1 f8e8 e1h4 b6b5 e2d3 d7f8 f5h3 f6e4 c3e4 d5e4 d3b1 c5c4 f4f5 e7h4 h3h4 f7f6 e5g4 f8d7 g4f2 d7b6 f2e4 b7e4 b1e4 e8e7 c1e1 d8e8 f1f4 g8h8 e4f3 c4c3 b2c3 c7c3 h2h3 e7e3 e1e3 c3e3 g1h2 b6c4 f3d5

go nodes 700
h7h6 N:      15 (+ 0) (P:  3.92%) (Q:  0.09498) (LCB:  0.43525) (V:  0.0814)
e8d8 N:      12 (+ 0) (P:  2.80%) (Q:  0.11855) (LCB:  0.49849) (V:  0.0983)
e3a3 N:     542 (+ 1) (P: 14.29%) (Q:  0.28268) (LCB:  0.61505) (V:  0.3569)
bestmove e3a3 ponder d5f7

go nodes 1000
e3a3 N:     561 (+ 0) (P: 14.29%) (Q: -1.00000) (LCB:  0.00000) (V: -1.0000) (T)
…
h7h6 N:      74 (+ 1) (P:  3.92%) (Q:  0.13845) (LCB:  0.51435) (V:  0.0814)
e8d8 N:      93 (+ 0) (P:  2.80%) (Q:  0.15288) (LCB:  0.53667) (V:  0.0983)
bestmove e3a3 ponder d5f7

go nodes 1300
e3a3 N:     561 (+ 0) (P: 14.29%) (Q: -1.00000) (LCB:  0.00000) (V: -1.0000) (T)
…
e8d8 N:     136 (+ 1) (P:  2.80%) (Q:  0.15390) (LCB:  0.54272) (V:  0.0983)
h7h6 N:     269 (+ 0) (P:  3.92%) (Q:  0.16087) (LCB:  0.54902) (V:  0.0814)
bestmove h7h6 ponder f4f3
```
Looks like your PR definitely helps avoid picking known losing moves! 😄

**Comment by Mardak** [comment]
I ran 1000 selfplay with 800 visits to see what's the behavior of parents getting terminated with this PR.

- Generally, at 800 visits, the terminated nodes are far from root and the Q value before termination wasn't too far off.
- A position that has a checkmate move for the opponent would likely have a Q around -0.65. MCTS probably would have preferred not visiting that parent anyway, so terminating it to Q: -1 makes it even less likely to visit.
- Similarly positions that have all opponent moves as losses (most likely a position with a single forced move) already had a Q around 0.85, so MCTS would have favored the move anyway.
- For the rare positions that could terminate as draws from this data, all of them were positions that had a single forced opponent draw move, and the Q was likely to be around 0.05.

Here's some stats of terminated parents:
- 46337 loss terminals
  - 0 with 1 edge
  - 13 with 2 edges
  - 13 with 3 edges
  - 130 with 4-9 edges
  - 46181 with 10+ edges
- 5860 won terminals
  - 5522 with 1 edge
  - 328 with 2 edges
  - 10 with 3 edges
  - 0 with 4+ edges
- 497 draw terminals
  - 497 with 1 edge
  - 0 with 2+ edges

So there are definitely a lot more parents that could be terminated as loss because one of generally 10+ possible moves are opponent with checkmate. There were a small portion of positions where the opponent had multiple moves leading to checkmate allowing a parent terminated as won, and overall roughly 90% fewer won than loss terminated parents. Similarly roughly 90% fewer draw than won terminated parents.

q terminated

```
tournamentstatus final
P1: +443 -430 =127
Win: 50.65%
Elo:  4.52
LOS: 67.00%
P1-W: +236 -201 =63
P1-B: +207 -229 =64
```

**Comment by Videodr0me** [comment]
I checked the code (briefly) and everything looks ok now. I really think you cannot do a simpler version of this. So lets keep our fingers crossed that this gets commited, so we have a starting point to maybe build on top of this. 

Your stats match mine, and are the reason why i push for two-fold draw scoring. Certain draws will then increase substantially in endgame positions (and the lookahead will increase those even further at no gpu cost - just cpu).

Last two things I would suggest is to call it "certainty-propagation" or maybe even more accurately "terminal-propagation". An also you should maybe not trim the head completely - because currently as soon as a terminal becomes root (start of the mating sequence) all previous computation is lost because the tree gets deleted. This might lead to strange sudden death behavior. You could just change the call to trimathead to only occur if the terminal node has no children.

**Comment by Mardak** [comment]
There's a set of issues related to trimming the head or not that would complicate this PR, so not sure if it's worth delaying getting things merged vs cleaning it up in its own PR. Similarly naming discussions often lead to time wasted bikeshedding, so I usually just let the reviewer decide if we even need to have a option/param.

**Comment by gonzalezjo** [comment]
Probably a stupid question: What are the functional differences between this and PR800 now?

File: src/mcts/params.cc [hunk 1]
```diff
@@ -87,6 +87,8 @@ const OptionId SearchParams::kTemperatureVisitOffsetId{
     "Adjusts visits by this value when picking a move with a temperature. If a "
     "negative offset reduces visits for a particular move below zero, that "
     "move is not picked. If no moves can be picked, no temperature is used."};
+const OptionId SearchParams::kTerminatorId{
+    "terminator", "Terminator", "I'll be back!"};
```
mooskagh: I'd write a better help for it. :) And after the help is written and it's clear what that option actually does, possibly it will also make sense to rename the option too.

src/mcts/node.cc
```diff
@@ -230,6 +230,17 @@ void Node::MakeTerminal(GameResult result) {
   }
 }
 
+void Node::AllowBranching() {
```
mooskagh: `MakeNotTerminal` ? `UnmakeTerminal` ? `UndoMakeTerminal`?

src/mcts/search.h
```diff
@@ -271,13 +271,9 @@ class SearchWorker {
                                    int collision_count) {
       return NodeToProcess(node, depth, true, collision_count);
     }
-    static NodeToProcess Extension(Node* node, uint16_t depth) {
+    static NodeToProcess Single(Node* node, uint16_t depth) {
```
mooskagh: It feels that ther should be a better name than `Single` but I cannot come up with it.
`TerminalHit`, `Collision`, and `Extension` hinted what will happen with this node, while `Single` just means that's no multivisit.. Maybe `Visit`? `LeafVisit`? `LeafEval`? `Eval`? `Backprop`? `Process`? :)

src/mcts/search.cc
```diff
@@ -1281,18 +1277,44 @@ void SearchWorker::DoBackupUpdateSingleNode(
     return;
   }
 
+  // For the first visit to a terminal, maybe convert ancestors to terminal too.
+  auto can_convert = params_.GetTerminator() && node->IsTerminal() && !node->GetN();
```
mooskagh: Line longer than 80 characters, run `clang-format -style=google` (or try to fix that manually as we still don't have merge check for that)

src/mcts/search.cc
```diff
@@ -1281,18 +1277,44 @@ void SearchWorker::DoBackupUpdateSingleNode(
...
+      v = n->GetQ();
+      d = n->GetD();
+    }
     n->FinalizeScoreUpdate(v, d, node_to_process.multivisit);
+
+    // Convert parents to terminals except the root or those already converted.
+    can_convert = can_convert && p != search_->root_node_ && !p->IsTerminal();
+
+    // A non-winning terminal move needs all other moves to have the same value.
+    if (can_convert && v != 1.0f)
```
mooskagh: Braces around body of `if` because it's multiline.

src/mcts/search.cc
```diff
@@ -1281,18 +1277,44 @@ void SearchWorker::DoBackupUpdateSingleNode(
...
+      for (const auto& edge : p->Edges())
```
mooskagh: Braces around body of `for` because it's on a separate line.
Consider this (although I'm not sure that it's more readable):
```cpp
auto edges = p->Edges();
can_convert = std::all_of(edges.begin(), edges.end(), [&](const Edge& e) {
  return e.IsTerminal() && e.GetQ(0.0f) == v;
});
```

src/mcts/search.cc
```diff
@@ -1281,18 +1277,44 @@ void SearchWorker::DoBackupUpdateSingleNode(
...
+        can_convert = can_convert && edge.IsTerminal() && edge.GetQ(0.0f) == v;
+
+    // Convert the parent to a terminal loss if at least one move is winning or
+    // to a terminal win or draw if all moves are loss or draw respectively.
+    if (can_convert)
+      p->MakeTerminal(v == 1.0f ? GameResult::BLACK_WON
```
mooskagh: Braces around multiline `if` body.

src/mcts/params.cc
Mardak: What do we want this to be called? Was waiting on your feedback before putting in actual values.

Mardak: Also, I wasn't sure if we wanted this as an actual param or just for testing the PR.

src/mcts/node.cc
Mardak: Renamed to `MakeNotTerminal`

src/mcts/search.h
Mardak: Renamed to `Visit` (as Leaf implies no children but "terminals" might have children in the tree).

src/mcts/search.cc
Mardak: Added newline for now (as depending on params decision, this might change again).

src/mcts/search.cc
Mardak: Added braces.

src/mcts/search.cc
Mardak: Switched to `all_of` but with `const auto &` instead of `const Edge&` as it's actually a `EdgeAndNode`.

src/mcts/search.cc
Mardak: Added braces.

src/mcts/params.cc
mooskagh: I thought we were going to merge it (with default off at first at least for training).
Maybe running speed test on random backend makes sense to be sure it doesn't slow things down, but from the code it doesn't seem that it should.

As `Certainty Propagation` is reserved for a more global changes, something like...
`TerminalBubble` (so that "terminals bubble up from their nodes to parent")
`TerminalSpill`  (so that "terminals spill down from their nodes to parent"?
`StickyTerminals` (just sounds fine)
`TerminalPropagation` (or `PropagateTerminals`) (to make it similar to `Certainty` but not really the same)
Something like that?

Mardak:
I updated the first commit addressing review comments, so the only remaining item is if we want to keep the second commit adding the testing `--terminator` option https://github.com/LeelaChessZero/lc0/pull/822/commits/0a2b234a40d4fa8d69e5739d5b773e081ac2ca3e or if this should just be the default behavior.

If we do want the param, we'll need a name, and I'll add a description that matches the name. I believe generally if these are exposed as params, they should be more "user friendly" than "terminal," etc.

src/mcts/params.cc
Mardak: `StickyEndgames` ?

mooskagh: Fine. :)
Although "endgame" in chess means something different, terminal can also happen in opening and midgame. `StickyGameEndging`?.. meh :)
`StickyOutcome`... `StickyEnd` :) Maybe `StickyNodes`? Not relevant, but sounds nice.

But I'm fine with endgames too!

Mardak:
I updated the first commit addressing all but the param review comments. The second commit I added `--sticky-endgames` param with description:

https://github.com/LeelaChessZero/lc0/blob/eb0ea5f74e6c79e5eb2e63fdae730afd7970feb0/src/mcts/params.cc#L155-L161

Is this what you were thinking of random backend?
```
./lc0 --backend=random --smart-pruning-factor=0 --no-sticky-endgames
go nodes 10000000
info depth 7 seldepth 15 time 69522 nodes 10000097 score cp 0 hashfull 1000 nps 143840 tbhits 0 pv d2d3 g7g6 f2f4 a7a6 d1d2 c7c5 d2c3 d7d5 a2a4

./lc0 --backend=random --smart-pruning-factor=0 --sticky-endgames
go nodes 10000000
info depth 7 seldepth 15 time 69522 nodes 10000012 score cp 0 hashfull 1000 nps 143839 tbhits 0 pv e2e3 c7c6 e1e2 c6c5 h2h3 e7e5 b2b4 d8f6 h3h4 f6f3
```

mooskagh:
My suggestion with std::all_of broke the build, probably easier to revert to handwritten loop.
Otherwise looks good!

killerducky:
@Mardak do you know why this test showed an Elo regression?
https://github.com/LeelaChessZero/lc0/pull/822#issuecomment-478357950

Is the code slower? Or was there a problem that has been fixed since then?

Mardak:
The old code was a bit inefficient in the common case affecting all node picking vs the current code only doing extra work when a new terminal was visited. There is also a fix in the current code to retain existing tree if the game played into a "terminal" position where the old code would have required re-searching already visited nodes.

src/mcts/search.cc
```diff
@@ -1282,18 +1278,48 @@ void SearchWorker::DoBackupUpdateSingleNode(
...
+    }
     n->FinalizeScoreUpdate(v, d, node_to_process.multivisit);
+
+    // Convert parents to terminals except the root or those already converted.
+    can_convert = can_convert && p != search_->root_node_ && !p->IsTerminal();
+
+    // A non-winning terminal move needs all other moves to have the same value.
+    if (can_convert && v != 1.0f) {
+      for (const auto& edge : p->Edges()) {
+        can_convert = can_convert && edge.IsTerminal() && edge.GetQ(0.0f) == v;
```
Tilps: This logic can be extended so that if all nodes are terminal, but they don't all agree on v  - then parent can be made terminal draw. But theoretically assuming imperfect opponent play maybe it could be better not to do so.

Videodr0me: Yes if all childs are terminal (or certain in my parlance) the maximum should be returned, so if all moves are (certain) losses and one is a (certain) draw, we can safely consider the position a certain draw. The perfect play assumption is not really relevant, as all opponents of leela either share the same codebase (different NNs) or are much superior in propagating minimax scores (basically all a/b engines). I would suggest to implement this - its simple and should maybe only add one line of code.

Videodr0me:
I still find the name of the option a little misleading, as mates in the opening or middle gamephase are also affected. I would still suggest calling it "terminal-propagation". Also in its current simple form it can be an always on feature so maybe we don't need an option after all.

Videodr0me:
I saw that you avoid trimming the head now and recalculate N. But i vaguely remember that I also had to change makemove, as otherwise there are problems in self-play and treesharing - did you test that?

src/mcts/search.cc
Mardak: This could be extended in a separate PR to improve the logic.

Mardak:
> i vaguely remember that I also had to change makemove, as otherwise there are problems in self-play and treesharing - did you test that?

It does look like the root can end up terminal if selfplay is run with `--reuse-tree` but the default is off, and it seems likely it will stay that way for training unless other things get fixed first, e.g., applying noise to existing nodes. I'll add a note to #775 if it becomes a more general issue of fixing tree-reuse issues.

Videodr0me:
> It does look like the root can end up terminal if selfplay is run with `--reuse-tree` but the default is off, and it seems likely it will stay that way for training unless other things get fixed first, e.g., applying noise to existing nodes. I'll add a note to #775 if it becomes a more general issue of fixing tree-reuse issues.

Things should not break if people use different options. The change required to fix this is basically identical to what you already did in Resettoposition, you just have to add the same lines of code to makemove. You can check PR700 for reference.

Mardak:
But how does it break? The outcome of the games are the same.

Mardak:
> The change required to fix this is basically identical to what you already did in Resettoposition, you just have to add the same lines of code to makemove.

Copy/pasting code is not a good solution in general as it often leads to bugs and code maintenance issues down the line that need additional review cycles or mental overhead causing.

I can file a followup PR with a better fix than just duplicating code.

Videodr0me:
> But how does it break? The outcome of the games are the same.

There might be several issues, games might not be played to their conclusion according to the rules of chess, terminating early before mate or a real drawn terminal is actually on the board and i think there might even be some edge cases where it can crash (but this might or might not apply to your current implementation).

Videodr0me:
> Copy/pasting code is not a good solution in general as it often leads to bugs and code maintenance issues down the line that need additional review cycles or mental overhead causing.

Obviously it was not meant as suggesting to cut/paste the code verbatim. It was just meant as a pointer on what could to be done in makemove to avoid potential problems - naturally it needs to be adapted. The aim was to raise awareness of potential issues - so we can get this into master without breaking anything.

Videodr0me:
> Probably a stupid question: What are the functional differences between this and PR800 now?

PR800 is bases on PR700 (basic certainty propagation). The differences to this PR are:

- PR700/800 display mate scores, this PR does not.
- PR700/800 prefer terminal wins over certain wins, this PR does not differentiate between the two. This leads to situations where this PR will postpone a mate-in-one just before the 50 move counter runs out.
- PR700/800 differentiates between mates proven with and without TB hits, this PR does not. This leads to situations where this PR might go for a long TB win instead of a shorter direct mate
- PR700/800 establishes proper minimax scores, this PR does not. So cases where all moves, except one drawing move are losing, are not propagated upward making this PR slightly less effective.
- PR700/800 establish certain bounds. For example if one side can "force" a draw but has alternatives, the position is scored as a lower bound draw. This is very effective in properly scoring repeatitions and is more effective than this PR.
- PR700/800 have an option to enable "two-fold" draw scoring, which increases the number of certain nodes and cuts down on tree combinatorics also making it more effective than this PR.
- PR700/800 store information in the edges, while this PR overloads the terminal bool in nodes. This makes this PR fit the current codebase more elegantly. 

But all of this is inconsequential, as many of the strength gaining features can be added to this PR later once its in master over time. You can think of PR487/700/800 as a proof-of-concept of what is possible if you push this concept to its limits. In doing that i imposed  - maybe to much - of my vision on how the codebase should evolve, and generally underestimated that the collaborative approach is better served with implementing features in small increments.And this is what this PR is about.

src/mcts/node.cc
```diff
@@ -230,6 +230,17 @@ void Node::MakeTerminal(GameResult result) {
   }
 }
 
+void Node::MakeNotTerminal() {
+  is_terminal_ = false;
```
Tilps: I think we should re-calculate Q and D here as well. 
Also are we confident that n_in_flight/best_cached_child/visited_policy are all correct?

Mardak: Added Q and D recomputation in https://github.com/LeelaChessZero/lc0/pull/822/commits/802e1ff48faec449d3e93af17e6caa5945ac2f91

The other values are updated in the usual FinalizeScoreUpdate calls from backup.

Tilps: Seems fine.

Tilps:
> r?@mooskagh or @Tilps Could you additionally review the added commit [f39a874](https://github.com/LeelaChessZero/lc0/commit/f39a874c47b831f10ecce06962bec9df3c1025b1) that moves the original terminal trimming head behavior into `MakeMove` as a conditional `MakeNotTerminal()`? This proactively fixes selfplay with tree reuse if TBs were supported in selfplay too.

I wondered whether MakeNotTerminal should be kept in case MakeMove isn't called - but I guess its fine since the making of terminals never applies to the root - so even if we start from fen which is late game and able to be proven terminal, the root itself doesn't get marked and so does not need to be reset.
So I think this is fine.

src/mcts/node.cc
```diff
@@ -237,7 +237,18 @@ void Node::MakeNotTerminal() {
   // If we have edges, we've been extended (1 visit), so include children too.
   if (edges_) {
     n_++;
-    for (const auto& child : Edges()) n_ += child.GetN();
+    for (const auto& child : Edges()) {
+      const auto n = child.GetN();
+      if (n > 0) {
+        n_ += n;
+        q_ -= child.GetQ(0.0f) * n;
```
mooskagh: Would be nice to add a comment why it is `-=` here instead of `+=`.

src/mcts/node.cc
```diff
@@ -237,7 +237,18 @@ void Node::MakeNotTerminal() {
...
+        d_ += child.GetD() * n;
+      }
+    }
+
+    // Recompute with current eval (instead of network's) and children's eval
```
mooskagh: Period at the end of a sentence.

src/mcts/node.cc
Mardak: Would `+= -child.GetQ()` instead be more immediately clear without a comment?

Mardak: Switched to `+=` and added comment:
https://github.com/LeelaChessZero/lc0/blob/b9b5e2fe8ceb9b4aa3abebad8f037547f3e9905c/src/mcts/node.cc#L244-L245

src/mcts/node.cc
Mardak: Added.

PR #824, opened 2019-03-30
get number of visits per line in multi-pv

aartdappel:


mooskagh:
There is #826 with the parameter for the same, so closing this one.

PR #832, opened 2019-04-19
Get Forsyth-Edwards Notation : GetFen()

MelleKoning:
Requested by jjosh on Discord, addition of a GetFen() to create a fen from current position.


MelleKoning:
The test reveals the GetFen() for black positions does not function properly. Does anybody know what is different in the ChessBoard when it's black to move and thus why GetFen() gives strange results?
Please see position_test.cc

MelleKoning:
added a command to get the current position from the engine.

When no search is done yet, the latest known fen (startpos) is returned.
After search, the fen of current position is returned.

Example:


```
position startpos moves e2e4 e7e5
getfen
rnbqkbnr/pppppppp/8/8/8/8/PPPPPPPP/RNBQKBNR w KQkq - 0 0
go nodes 2
bestmove g1f3 ponder g8f6
getfen
rnbqkbnr/pppp1ppp/8/4p3/4P3/8/PPPP1PPP/RNBQKBNR w KQkq - 0 2
```

MelleKoning:
From discord: author of Nibbler, a GUI for Lc0, wrote a fen parser.

Plea for merge: It can be tricky to get the fen writer reader right to also include enpassent, maybe it helps GUI writers in general to not have to reinvent the wheel. Having a fen reader will therefore be beneficial for lc0 gui writers. It just saves them for having to write that on their own when the engine alteady supports it via uci commands.

src/chess/position.cc
```diff
@@ -137,4 +142,87 @@ uint64_t PositionHistory::HashLast(int positions) const {
   return HashCat(hash, Last().GetNoCaptureNoPawnPly());
 }
 
+// PrintFen outputs a FEN notation of the board. 
+// based on https://en.wikipedia.org/wiki/Forsyth%E2%80%93Edwards_Notation#Examples
+string Position::GetFen() const  {
+	string result;
+	string enpassant;
```
mooskagh: It's better to declare variables at the point of the first use rather than in the beginning of the function.

src/chess/position.cc
```diff
@@ -137,4 +142,87 @@ uint64_t PositionHistory::HashLast(int positions) const {
...
+	ChessBoard board = GetBoard();
+	bool fenflipped = false;
```
mooskagh: With `GetWhiteBoard()`, no need to flip it and have this variable.

src/chess/position.cc
```diff
@@ -137,4 +142,87 @@ uint64_t PositionHistory::HashLast(int positions) const {
...
+	if (board.flipped()) {
+		board.Mirror();
+		fenflipped = true;
+	}
+	int emptycounter = 0;
+	for (int i = 7; i >= 0; --i) {
+		for (int j = 0; j < 8; ++j) {
+			if (emptycounter > 0 &&
+				(
+				    board.our_king().get(i,j) ||
```
mooskagh: I think this logic should be moved to another function. Other a free-standing one or member function of `ChessBoard` class. Something like:

```cpp
char GetPieceAt(const ChessBoard& board, int row, int col) {
  BoardSquare sq(row, col);
  if (!(ours() | theirs()).get(sq)) return '\0';
  if (board.our_king() == sq) return 'K';
  if (board.their_king() == sq) return 'k';
  char res = 'N';
  if (board.bishops().get(sq)) res = 'B';
  else if (board.queens().get(sq)) res = 'Q';
  else if (board.rooks().get(sq)) res = 'R';
  return flipped() ? std::tolower(res) : res;
}

File: src/chess/position.cc [hunk 4]
```diff
@@ -137,4 +142,87 @@ uint64_t PositionHistory::HashLast(int positions) const {
...
+				else if (board.queens().get(i, j)) {
+					c = 'q';
+				}
+				else if (board.rooks().get(i, j)) {
+					c = 'r';
+				}
+				else {
+					c = 'n';
+				}
+				if (board.ours().get(i, j)) c = std::toupper(c); // capitals are for Black
```
mooskagh: From reading the code, it seems capitals are for white.

src/chess/position.cc
```diff
@@ -137,4 +142,87 @@ uint64_t PositionHistory::HashLast(int positions) const {
...
+				emptycounter++;
+			}			
+		}
+		if (emptycounter > 0) result += std::to_string(emptycounter);
+		if (i > 0) result += "/";
+		emptycounter = 0;
+	}
+	string castlings_no_fenflip = board.castlings().as_string();
+	if (fenflipped) {
+		board.Mirror();
```
mooskagh: `const ChessBoard& side_to_move_board = GetBoard();`

src/chess/position.cc
```diff
@@ -137,4 +142,87 @@ uint64_t PositionHistory::HashLast(int positions) const {
...
+	}
+	enpassant = "-";
+	for (auto sq : board.en_passant()) {
+		// Our internal representation stores en_passant 2 rows away
+		// from the actual sq.
+		if (sq.row() == 0) {
```
mooskagh: Actually as rank is always `3` for black en passants and `6` for white, no need for side_to_move_board at all.

src/chess/position.cc
```diff
@@ -137,4 +142,87 @@ uint64_t PositionHistory::HashLast(int positions) const {
...
+			enpassant = ((BoardSquare)(sq.as_int() + 16)).as_string();
```
mooskagh: A bit hacky, but saves some typing. :) And no need for loop.
```
auto sq = *board.en_pasant().begin();
result += std::string(1, 'a' + sq.col()) + IsBlackToMove() ? "3" : "6";
```

src/chess/position.cc
```diff
@@ -137,4 +142,87 @@ uint64_t PositionHistory::HashLast(int positions) const {
...
+		}
+		if (emptycounter > 0) result += std::to_string(emptycounter);
+		if (i > 0) result += "/";
+		emptycounter = 0;
+	}
+	string castlings_no_fenflip = board.castlings().as_string();
+	if (fenflipped) {
+		board.Mirror();
+	}
+	enpassant = "-";
```
mooskagh:
```cpp
If (board.en_passant().empty()) {
 result += "-";
} else {
...
```

src/chess/position.cc
```diff
@@ -137,4 +142,87 @@ uint64_t PositionHistory::HashLast(int positions) const {
...
+		// from the actual sq.
+		if (sq.row() == 0) {
+			enpassant = ((BoardSquare)(sq.as_int() + 16)).as_string();
+		}
+		else {
+			enpassant = ((BoardSquare)(sq.as_int() - 16)).as_string();
+		}
+	}
+	result += IsBlackToMove() ? " b" : " w";
+	result += " " + castlings_no_fenflip;
```
mooskagh: Can call `board.castlings().as_string()` here without any temporary variables.

src/chess/uciloop.cc
```diff
@@ -153,13 +154,18 @@ bool UciLoop::DispatchCommand(
...
-        StrSplitAtWhitespace(GetOrEmpty(params, "moves"));
-    CmdPosition(GetOrEmpty(params, "fen"), moves);
-  } else if (command == "go") {
+	  if (ContainsKey(params, "fen") == ContainsKey(params, "startpos")) {
+		  throw Exception("Position requires either fen or startpos");
+	  }
+	  const std::vector<std::string> moves =
+		  StrSplitAtWhitespace(GetOrEmpty(params, "moves"));
+	  CmdPosition(GetOrEmpty(params, "fen"), moves);
+  } else if (command == "get") {
```
mooskagh: There's no uci command "get" (I've just checked stockfish doesn't have it either).
If stockfish supports it in some way, it's better to be consistent.
If not, maybe we don't need it either.

In any case, I don't think that adding non-standard commands in `UciLoop` is a good idea.
Instead, we should probably add `virtual bool CmdNonStandardCommand(const std::string& command);` which uci loop should call in such cases, and `EngineLoop::CmdNonStandardCommand` should just have `if (command == "blah") DoThings(); `

src/engine.cc
```diff
@@ -306,6 +306,12 @@ void EngineController::SetPosition(const std::string& fen,
   search_.reset();
 }
 
+std::string EngineController::GetPosition() {
+	if (tree_ != nullptr) {
+		return tree_->HeadPosition().GetFen();
+	}
+	return "need a 'go nodes..' before returning fen position.";
```
mooskagh: Throw exception (from `utils/exception.h`) instead.

src/engine.h
```diff
@@ -82,12 +82,15 @@ class EngineController {
       const GoParams& params,
       std::chrono::steady_clock::time_point start_time);
 
+  std::string  GetPosition();
```
mooskagh: Better `GetCurrentPositionFen()`.

src/chess/position_test.cc
```diff
@@ -17,12 +17,39 @@
...
 
 namespace lczero {
 
+TEST(Position, SetFenGetFen) {
+		std::vector<Position> positions;
+		ChessBoard board;
+		std::vector<std::string> fen_splitted;
+		std::vector<std::string> source_fens = {
+			"r3k2r/p1ppqpb1/bn2pnp1/3PN3/1p2P3/2N2Q1p/PPPBBPPP/R3K2R w KQkq - 1 1" ,
+			"r3k2r/p1ppqpb1/bn2pnp1/3PN3/1p2P3/2N2Q1p/PPPBBPPP/R3K2R w KQkq e3 1 1" , // has en_passent space e3
```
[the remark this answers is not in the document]
MelleKoning: Installed clang-format and ran it.

src/chess/position.cc
```diff
@@ -27,9 +27,14 @@
 
 #include "chess/position.h"
 #include <cassert>
+#include <cstring>
+#include <cstdlib>
+#include <cctype>
 
 namespace lczero {
 
+using std::string;
```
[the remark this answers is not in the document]
MelleKoning: Removed

src/chess/position.cc
```diff
@@ -137,4 +142,87 @@ uint64_t PositionHistory::HashLast(int positions) const {
   return HashCat(hash, Last().GetNoCaptureNoPawnPly());
 }
 
+// PrintFen outputs a FEN notation of the board. 
+// based on https://en.wikipedia.org/wiki/Forsyth%E2%80%93Edwards_Notation#Examples
+string Position::GetFen() const  {
+	string result;
+	string enpassant;
+	ChessBoard board = GetBoard();
```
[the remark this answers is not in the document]
MelleKoning: Thanks, introduced GetWhiteBoard();

src/chess/position.cc
MelleKoning: Moved enpassant to where it is used

src/chess/position.cc
MelleKoning: Removed fenflipped

src/chess/position.cc
MelleKoning: Yes, just initializing as empty with dash(-) and overwriting when there is an en_passent space

src/chess/position.cc
MelleKoning: fixed

src/chess/position.cc
MelleKoning: Adjusted the comment

MelleKoning: Adjusted the comment

src/chess/position.cc
MelleKoning: I don't think it is stored at rank three and six but on 0 and 7; when I asked about en_passent Tilps in discord pointed me to the code here which I simply re-used here: https://github.com/Tilps/lc0/blob/rescore_tb/src/selfplay/loop.cc#L282

src/chess/position.cc
MelleKoning: Have changed the for loop to a hacky line but at least kept using the provided as_string()

src/chess/uciloop.cc
MelleKoning: There is no uci command indeed. Now that lc0 is champion we could set our own standards.

Just kidding ;) 
Introduced a CmdNonStandardUciCommand method and moved code up to EngineLoop. However, there is a command-parser in UciLoop which I can't circumvent easily, so "get","fen" is still there...

src/engine.cc
MelleKoning: Now using throw exception

src/engine.h
MelleKoning: Renamed.

src/chess/uciloop.cc
```diff
@@ -208,7 +203,7 @@ bool UciLoop::DispatchCommand(
   } else if (command == "quit") {
     return false;
   } else {
-    throw Exception("Unknown command: " + command);
+    return CmdNonStandardUciCommand(command, params);
```
mooskagh: My idea was that `CmdNonStandardUciCommand` returns `true` if it managed to process command and `false` if not, so here it would still be:
```cpp
else if (!CmdNonStandardUciCommand(command, params)) {
  throw Exception("Unknown command: " + command);
}
```

And in `UciLoop::CmdNonStandardUciCommand` just `return false`

src/engine.cc
```diff
@@ -453,20 +456,42 @@ void EngineLoop::CmdPosition(const std::string& position,
...
 void EngineLoop::CmdStop() { engine_.Stop(); }
 
+bool ContainsKey(const std::unordered_map<std::string, std::string>& params,
+	const std::string& key) {
+	return params.find(key) != params.end();
+}
+
+bool EngineLoop::CmdNonStandardUciCommand(const std::string & command, const std::unordered_map<std::string, std::string>& params)
+{
+	if (command == "get") {
```
mooskagh: Why not just a `getfen`? Is it so in some other engine?
Also, maybe just pass unparsed command to `CmdNonStandardUciCommand` (not command and params), so that it can parse it as it wishes. That way we can even compare with `"get fen"` (because currently it would work with `get asd asd fen sdjksd sdfkls ldkjfds` but won't with `get asd fen sdjkfl sdjklsdjk`, which is not very nice).

MelleKoning: First I started out with 'getfen' but I did not like it as it is not English, and does not seem to match the other command/parameter pairs. I future we might want to support to get other stuff. 'get fen' is just a starting point. For analyses purposes we could extend further with other non-uci commands. Not sure what though, yet but for example 'get epd' if chess960 would be supported. It's not really part of the scope and when something new is introduced we can change this later.

src/chess/uciloop.cc
MelleKoning: Found a way to move out 'get fen' from the uci commands. Having the sole 'line string' loses the parsing of command and tokens, but I guess it's good for now.

src/chess/position.cc
```diff
@@ -142,87 +140,73 @@ uint64_t PositionHistory::HashLast(int positions) const {
...
+  for (int i = 7; i >= 0; --i) {
+    for (int j = 0; j < 8; ++j) {
+      if (emptycounter > 0 &&
+          (board.our_king().get(i, j) || board.their_king().get(i, j) ||
+           board.pawns().get(i, j) || board.ours().get(i, j) ||
+           board.theirs().get(i, j))) {
+        result += std::to_string(emptycounter);
+        emptycounter = 0;
+      }
+      char piece = GetPieceAt(board, i, j);
```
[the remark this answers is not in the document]
MelleKoning: Yes, that makes the algorithm much better. Adjusted the code.

src/chess/position.cc
```diff
@@ -142,87 +140,73 @@ uint64_t PositionHistory::HashLast(int positions) const {
...
+      if (piece != '.') {
```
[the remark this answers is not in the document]
MelleKoning: Done

src/chess/position.cc
```diff
@@ -142,87 +140,73 @@ uint64_t PositionHistory::HashLast(int positions) const {
...
-	result += " " + std::to_string(GetNoCaptureNoPawnPly());
-	result += " " + std::to_string(ply_count_); 
-	return result;
+// PrintFen outputs a FEN notation of the board.
+// based on
+// https://en.wikipedia.org/wiki/Forsyth%E2%80%93Edwards_Notation#Examples
+std::string Position::GetFen() const {
+  std::string result;
+  const ChessBoard& board = GetWhiteBoard();
+  int emptycounter = 0;
```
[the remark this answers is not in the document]
MelleKoning: Altered.

src/chess/position.cc
```diff
@@ -142,87 +140,73 @@ uint64_t PositionHistory::HashLast(int positions) const {
...
+    if (emptycounter > 0) result += std::to_string(emptycounter);
+    if (i > 0) result += "/";
+    emptycounter = 0;
+  }
+  std::string enpassant = "-";
+  if (!board.en_passant().empty()) {
+    auto sq = *board.en_passant().begin();
+    // Our internal representation stores en_passant 2 rows away
+    // from the actual sq.
+    enpassant = ((BoardSquare)(sq.as_int() + (IsBlackToMove() ? +16 : -16)))
```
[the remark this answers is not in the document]
MelleKoning: Yes, good suggestion! Had to switch the 2 and 5, black can take on row 2, white on row 5 :)

src/chess/position.cc
```diff
@@ -142,87 +140,73 @@ uint64_t PositionHistory::HashLast(int positions) const {
...
+  }
+  result += IsBlackToMove() ? " b" : " w";
+  result += " " + board.castlings().as_string();
+  result += " " + enpassant;
+  result += " " + std::to_string(GetNoCaptureNoPawnPly());
+  result += " " + std::to_string(ply_count_);
+  return result;
 }
 
+char Position::GetPieceAt(const ChessBoard& board, int row, int col) const {
```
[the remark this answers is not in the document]
MelleKoning: Adjusted.

src/chess/position.cc
```diff
@@ -142,87 +140,73 @@ uint64_t PositionHistory::HashLast(int positions) const {
...
+      c = 'B';
+    } else if (board.queens().get(row, col)) {
+      c = 'Q';
+    } else if (board.rooks().get(row, col)) {
+      c = 'R';
+    } else {
+      c = 'N';
+    }
+    if (board.theirs().get(row, col)) {
+      c = std::tolower(c);  // capitals are for White
```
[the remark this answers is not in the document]
MelleKoning: Fixed.

src/chess/position.h
```diff
@@ -65,6 +67,8 @@ class Position {
   const ChessBoard& GetBoard() const { return us_board_; }
   // Gets board from the point of view of opponent.
   const ChessBoard& GetThemBoard() const { return them_board_; }
+  // Gets board from the point of view of the white player.
+  const ChessBoard& GetWhiteBoard() const { return us_board_.flipped() ? them_board_ : us_board_; };
```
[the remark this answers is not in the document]
MelleKoning: Have clang-format installed as a plug-in to visual studio. Must say that the default behavior is not always ideal, that is, I find the lines a little bit too short. For example: I do not use the clang-format on the position_test.cc file because it breaks up the fen-strings!

src/chess/uciloop.h
```diff
@@ -76,9 +76,13 @@ class UciLoop {
                            const std::vector<std::string>& /*moves*/) {
     throw Exception("Not supported");
   }
-  virtual void CmdGetFen() {
-	  throw Exception("Not supported");
+    
+  // can be overridden in engine
```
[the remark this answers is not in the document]
MelleKoning: Adjusted.

src/chess/uciloop.cc
MelleKoning: Had to re-wire a few things to ensure the {"get","fen"} was moved out of the uciloop.cc file. Now indeed the typed in line will be forwarded to the engine class without command/token parsing.

src/chess/position.cc
```diff
@@ -142,87 +140,73 @@ uint64_t PositionHistory::HashLast(int positions) const {
...
+    }
+  }
+  return (c == '?' ? '.' : c);  // '?' means '.' no piece found
```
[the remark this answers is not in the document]
MelleKoning: Used the '?' during creation to see if I got all different pieces covered. Now that it works indeed no longer needed; changed to initialize with zero-char now.

src/chess/uciloop.cc
```diff
@@ -131,8 +133,13 @@ void UciLoop::RunLoop() {
     LOGFILE << ">> " << line;
     try {
       auto command = ParseCommand(line);
-      // Ignore empty line.
-      if (command.first.empty()) continue;
+      // Unknown UCI command, check non standard.
+      if (command.first.empty()) {
+        if (!CmdNonStandardUciCommand(line)) {
```
mooskagh: Does it work this way?.. We should get there only when command is empty, but in this case CmdNonStandardUciCommand() will have nothing to match against.

I suggest we either call `CmdNonStandardUciCommand()` from within DispatchCommand() (and in this case we need to add "get" command into `kKnownCommands` like it's done for non-standard "xyzzy" command), or

call "TryHandleNonStandardCommand()" before calling ParseCommand() (and skip parsing if it's successful). That doesn't look very nice though.

src/chess/uciloop.h
```diff
@@ -76,6 +76,11 @@ class UciLoop {
                            const std::vector<std::string>& /*moves*/) {
     throw Exception("Not supported");
   }
+  // To support non-UCI commands this can be overridden in the engine class.
+  virtual bool CmdNonStandardUciCommand(const std::string& command) {
+	  throw Exception("Unknown command: " + command);
```
mooskagh: I think default behaviour should be `return false;` rather than throwing exception?

src/engine.cc
```diff
@@ -306,6 +309,12 @@ void EngineController::SetPosition(const std::string& fen,
   search_.reset();
 }
 
+std::string EngineController::GetCurrentPositionFen() {
+  if (tree_ != nullptr) {
+    return tree_->HeadPosition().GetFen();
+  }
+  throw Exception("Need a 'go nodes..' before returning fen position.");
```
mooskagh: Actually we need `position startpos|fen` I think, and not go nodes.
I think error message like "No game in progress", "Game not started" etc would be more appropriate.

src/engine.cc
```diff
@@ -446,14 +455,33 @@ void EngineLoop::CmdUciNewGame() { engine_.NewGame(); }
...
+  std::string fen = engine_.GetCurrentPositionFen();
+  return SendResponse(fen);
+}
 void EngineLoop::CmdGo(const GoParams& params) { engine_.Go(params); }
 
 void EngineLoop::CmdPonderHit() { engine_.PonderHit(); }
 
 void EngineLoop::CmdStop() { engine_.Stop(); }
 
+bool ContainsKey(const std::unordered_map<std::string, std::string>& params,
```
mooskagh: Instead of duplicating the function, maybe it makes sense to create `utils/misc.h` and move it there? (also use it in uciloop.cc). Something like:

```cpp
template<class T, U>
bool ContainsKey(const T& container, const U& key) {
   return container.find(key) != container.end();
}

File: src/chess/position.cc [hunk 31]
```diff
@@ -27,9 +27,37 @@
...
 #include <cassert>
+#include <cctype>
+#include <cstdlib>
+#include <cstring>
 
 namespace lczero {
 
+// GetPieceAt returns the piece found at row, col on board or the null-char '\0'
+// in case no piece there.
+char GetPieceAt(const ChessBoard& board, int row, int col) {
```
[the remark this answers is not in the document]
MelleKoning: Sorry, I don't get it. I moved it close to top of the file, still inside namespace lczero already.. I can't put the function outside (on top of) namespace lczero as it has a reference to ChessBoard. 
FYI: I'm not all too familiar with C++ 11 so please enlighten me :)

src/chess/uciloop.h
MelleKoning: Done

src/engine.cc
MelleKoning: Hm, the way the fen is looked up, the position should already have been processed through the engine. For some reason this set of commands does not work:

> position fen 8/2p5/3p4/KP5r/1R3p1k/8/4P1P1/8 w - - 1 1
> get fen
> error Need a 'go nodes..' before returning fen position.

but this does:

 
> position fen 8/2p5/3p4/KP5r/1R3p1k/8/4P1P1/8 w - - 1 1
> go nodes 1
> ...
> get fen
> 8/2p5/3p4/KP5r/1R3p1k/8/4P1P1/8 w - - 1 0

src/engine.cc
MelleKoning: Good catch; it was dead code. Is now removed.

src/chess/position.cc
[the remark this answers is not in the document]
MelleKoning: Ok, found out what anonymous namespace is and moved it there

Naphthalin:
What is the current stance of this PR? If this isn't implemented yet but useful, we should decide on merging it after an update.

MelleKoning:
What update would be needed to enable merge?

Naphthalin:
it shows 3 merge conflicts, that's what I referred to as "update". But I think before doing the work, the decision whether this will be merged should be made.

src/engine.h
```diff
@@ -131,10 +134,12 @@ class EngineLoop : public UciLoop {
   void CmdUciNewGame() override;
   void CmdPosition(const std::string& position,
                    const std::vector<std::string>& moves) override;
+  void CmdGetFen();
```
mooskagh, line 137: This is not needed anymore.

mooskagh:
This was/is almost ready, I believe it never got back after a round of comments.
After rebasing and checking that it works fine with FRC, I think it's good to be merged.

liamlim:
What is needed for this PR to get merged? Is some testing needed? Maybe I could help but I think it would be good to know from maintainers what is the state of this PR. It was reviewed many times and now it's rebased on current master

Naphthalin:
We are currently in the process of collecting which PRs to merge into `v0.26.0`, so right now PRs are getting selected with having that in mind.

Naphthalin:
@liamlim it seems to me that there are a few comments by @mooskagh which weren't addressed yet; other than that I don't see anything holding this PR back.

liamlim:
What exactly is necessary to test about FRC for this PR to address comments by mooskagh? I think that the only think that makes FRC different from regular chess are castling rights and these work just fine in my opinion.

MelleKoning:
This PR is long due. It works any updates could

liamlim:
I think that all comments in this PR has been addressed and it works just fine. Is there any additional work needed for this to get merged?

borg323:
A new review is needed, since I did extensive modifications.

Naphthalin:
@mooskagh any chance you can take a more thorough look here? The inner function `Position::GetFEN()` is likely needed for transposition related things, especially for updating the time manager to account for transpositions near root.

src/chess/uciloop.cc
```diff
@@ -205,9 +211,8 @@ bool UciLoop::DispatchCommand(
     SendResponse("Nothing happens.");
   } else if (command == "quit") {
     return false;
-  } else {
-    throw Exception("Unknown command: " + command);
   }
+
```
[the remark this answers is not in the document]
mooskagh: Ok, after one more thought... It's becoming too complicated, so I take back my suggestion from 2 years ago, sorry to change my mind.

Now I believe it's cleaner just to add yet another command (`fen`) to `kKnownCommands` and add `CmdFen()` to `UciLoop`.

src/engine.cc
```diff
@@ -173,6 +173,27 @@ void EngineController::SetPosition(const std::string& fen,
   search_.reset();
 }
 
+std::string EngineController::GetCurrentPositionFen() {
+  if (tree_ != nullptr) {
+    return tree_->HeadPosition().GetFen();
+  } else if (current_position_) {
```
mooskagh, line 179: I think it would be cleaner to split this function into two. How about that:

1. Make `current_position_` non `std::optional` (initialize it to `ChessBoard::kStartposFen, {}` on construction in in new_game`.
2. Remove `else` branch in `if (current_position_)` in `EngineController::Go`.
3. Have function `Position FoldPosition(const CurrentPosition& pos)` which would do what this branch does (no need to put it in a class). (or some better name than `FoldPosition` but it's good to have a name that implies that it's not a O(1) operation... `BuildPosition()`? `ApplyPositionMoves()`?)
4. Call `FoldPosition(current_position_).GetFen()` instead of `GetCurrentPositionFen()`.
5. (optional) rename `CurrentPosition` (struct name, not the variable) to something better. :)

src/chess/position.cc
```diff
@@ -128,4 +158,39 @@ uint64_t PositionHistory::HashLast(int positions) const {
   return HashCat(hash, Last().GetRule50Ply());
 }
 
+// PrintFen outputs a FEN notation of the board.
+// based on
+// https://en.wikipedia.org/wiki/Forsyth%E2%80%93Edwards_Notation#Examples
+std::string Position::GetFen() const {
```
mooskagh, line 164: This function doesn't really access (or has to access) any private member variables, consider making it a free-standing function rather than a member. <https://www.drdobbs.com/cpp/how-non-member-functions-improve-encapsu/184401197>, <https://flylib.com/books/en/2.887.1.37/1/>

src/engine.h
mooskagh: This is needed again...
Sorry for making to do and undo extra work, I've just took a look what it would take to add this "NonStandardUciCommand" function cleanly and it really doesn't seem to be that clean.

mooskagh (changes requested):

src/chess/board.h
```diff
@@ -136,6 +136,23 @@ class ChessBoard {
 
     void Mirror() { data_ = ((data_ & 0b11) << 2) + ((data_ & 0b1100) >> 2); }
 
+    std::string as_string() const {
+      if (data_ == 0) return "-";
+      std::string result;
+      if (queenside_rook() == FILE_A && kingside_rook() == FILE_H) {
```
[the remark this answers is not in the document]
borg323: Seems to work fine in a few tests (mostly lichess and cutechess). It can be changed, but is it worth it? E.g. start position 508 would be `rbqnknbr/pppppppp/8/8/8/8/PPPPPPPP/RBQNKNBR w KQkq - 0 1` if we also check the king's position rather than `rbqnknbr/pppppppp/8/8/8/8/PPPPPPPP/RBQNKNBR w AHah - 0 1` and getting this right is not trivial. (Also putting the later form in the lichess board editor loses black's castling rights).
So I'd rather leave it like that until we get actual issues reported.

src/engine.cc
borg323: Note that now the position is never taken from the tree - this is implied in the above and makes in practice, as a `position` followed by `fen` would give the fen for the tree head with the previous version instead of the just set position.

borg323:
With the last revision I changed the command to a simple `fen`, if anyone objects now is the time to speak up.

MelleKoning:
Nice work borg!

mooskagh (approved):

PR #841, opened 2019-04-29
Recalibrate centipawn formula

QxC4eva:
It is widely agreed that Leela's eval in centipawns is inflated and a headache to viewers in tournaments such as TCEC and CCC. It has led some to believe that Leela "is just too optimistic" or "overestimates positions", or that high evals "is just a characteristic of neural networks". Furthermore, it has made the TCEC and CCC adjudication rules less reliable. All of this can be resolved by a slight tweak to the conversion formula.

The process that led to the current formula `290.680623072 * tan(1.548090806 * Q)` is as follows:
- Evals of Stockfish and Leela (Apr/15/2018) were obtained by removing pieces from the starting position
- Based on correlation analysis, `300 * tan(1.5 * Q)` was found to be a close predictor of SF's eval
- It was agreed to use tan() over logit due to its flatter tails and tendency to give more disparity in evals for winrates above 90%
- The formula was then modified to `290.680623072 * tan(1.548090806 * Q)` in order to reach exactly 12800cp when Q=1 (100% winrate)
- Note that "1.548090806" was calculated from `atan(12800 / 290.680623072)`

This method has a few problems:
- A very outdated version of Leela (test10 at its infancy) was used
- Analysis was done on a low sample size of 15 positions
- Leela would not have come across the start position with pieces removed in her training
- The formula was calibrated to positions that are rarely, if at all, seen in ordinary games

It is likely some of these issues have led to the iconic inflation of Leela's eval. For better calibration, analysis should be done on more "realistic" positions with the latest versions of both engines. To do this, I took the games from the recent TCEC 15 Premier Division and extracted all positions where the evals of Leela and Bluefish (a live spectator engine @ 88C/176T Stockfish 11-Dev no contempt) were available. [The data can be found here.](https://pastebin.com/raw/5PG1aEMv) Weighting was applied to ensure that each game has the same influence regardless of its length, and positions where either engine has over 8x the eval of the other were removed. Stockfish isn't the only traditional engine out there so to avoid specific engine bias, the evals of Bluefish were first compared to that of Komodo, Houdini, and itself:
Bluefish Graph
The overall trend slope was found to be 0.69894. This was multiplied to all of Bluefish's evals to give what I'll call the "BF-K-H eval". I then ran a script to test Leela's conversion formula `k * tan(atan(12800 / k) * Q)` for all values of k between 0.000000001 and 400.000000000. It was found that when k = 111.714640912, the relationship between BF-K-H eval and Leela's eval becomes a 1:1 ratio. I therefore propose this as the new formula:

`cp = 111.714640912 * tan(1.5620688421 * Q)`

Below is a scatter plot to show its differences to the old formula. The lack of negative evals was due to Leela having no white losses or black wins, except against Ethereal (game 42) but that's when Bluefish decided to go offline...
Leela Graph

And here's a comparison of the old and new formulae:
Formula Graph
With this change, Leela's eval will shrink by a factor of ~2.6 in most cases. Mate scores will remain at +128.00 or -128.00. +1.10 will become a 75% win percentage, +1.50 will be 80%, and +3.30 will be 90%.

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -127,6 +127,8 @@ void Search::SendUciInfo() REQUIRES(nodes_mutex_) {
     auto& uci_info = uci_infos.back();
     if (score_type == "centipawn") {
       uci_info.score = 111.714640912 * tan(1.5620688421 * edge.GetQ(default_q));
+    } else if (score_type == "centipawn_2018") {
+      uci_info.score = 290.680623072 * tan(1.548090806 * edge.GetQ(default_q));
     } else if (score_type == "win_percentage") {
       uci_info.score = edge.GetQ(default_q) * 5000 + 5000;
     } else if (score_type == "Q") {
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -126,7 +126,7 @@ void Search::SendUciInfo() REQUIRES(nodes_mutex_) {
     uci_infos.emplace_back(common_info);
     auto& uci_info = uci_infos.back();
     if (score_type == "centipawn") {
-      uci_info.score = 290.680623072 * tan(1.548090806 * edge.GetQ(default_q));
+      uci_info.score = 111.714640912 * tan(1.5620688421 * edge.GetQ(default_q));
     } else if (score_type == "win_percentage") {
       uci_info.score = edge.GetQ(default_q) * 5000 + 5000;
     } else if (score_type == "Q") {
--- a/src/mcts/params.cc
+++ b/src/mcts/params.cc
@@ -219,7 +219,8 @@ void SearchParams::Populate(OptionsParser* options) {
   options->Add<BoolOption>(kStickyEndgamesId) = false;
   options->Add<BoolOption>(kSyzygyFastPlayId) = true;
   options->Add<IntOption>(kMultiPvId, 1, 500) = 1;
-  std::vector<std::string> score_type = {"centipawn", "win_percentage", "Q"};
+  std::vector<std::string> score_type = {"centipawn", "centipawn_2018",
+                                         "win_percentage", "Q"};
   options->Add<ChoiceOption>(kScoreTypeId, score_type) = "centipawn";
   std::vector<std::string> history_fill_opt{"no", "fen_only", "always"};
   options->Add<ChoiceOption>(kHistoryFillId, history_fill_opt) = "fen_only";


mooskagh:
I like the idea and graphs are beautiful. :)
I'd like to hear opinions of people who came up with original numbers though.
@fli @syjytg @killerducky 

See also #728 #25 https://github.com/glinscott/leela-chess/pull/497  https://github.com/glinscott/leela-chess/pull/207

As we have switchable eval mode, it may also make sense to keep current as a separate mode for now.

borg323:
To keep the current formula as a separate mode, you can do something like the following:
```
diff --git a/src/mcts/params.cc b/src/mcts/params.cc
index d2cace8..4ec7e0f 100644
--- a/src/mcts/params.cc
+++ b/src/mcts/params.cc
@@ -219,7 +219,8 @@ void SearchParams::Populate(OptionsParser* options) {
   options->Add<BoolOption>(kStickyEndgamesId) = false;
   options->Add<BoolOption>(kSyzygyFastPlayId) = true;
   options->Add<IntOption>(kMultiPvId, 1, 500) = 1;
-  std::vector<std::string> score_type = {"centipawn", "win_percentage", "Q"};
+  std::vector<std::string> score_type = {"centipawn", "centipawn_2018",
+                                         "win_percentage", "Q"};
   options->Add<ChoiceOption>(kScoreTypeId, score_type) = "centipawn";
   std::vector<std::string> history_fill_opt{"no", "fen_only", "always"};
index c690855..5db6817 100644
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -126,6 +126,8 @@ void Search::SendUciInfo() REQUIRES(nodes_mutex_) {
     if (score_type == "centipawn") {
+      uci_info.score = 111.714640912 * tan(1.5620688421 * edge.GetQ(default_q));
+    } else if (score_type == "centipawn_2018") {
       uci_info.score = 290.680623072 * tan(1.548090806 * edge.GetQ(default_q));
     } else if (score_type == "win_percentage") {
       uci_info.score = edge.GetQ(default_q) * 5000 + 5000;
```

QxC4eva:
An extra mode for the old formula sounds like a great idea. I'll try your suggestion @borg323 

See also glinscott/leela-chess#322 as to how the original numbers were decided. The bulk of the discussion behind it was in the Discord [starting around here.](https://discordapp.com/channels/425419482568196106/425419999096733706/434911034017447946)

gonzalezjo:
I wonder if this might actually gain Elo by making Leela more likely to accept a draw adjudication in endgames she'd otherwise struggle with ;)


Either way, this is an awesome PR--both in terms of addressing a major flaw, and in terms of being written up and presented really well! Hope to see this merged soon.

src/mcts/search.cc
```diff
@@ -126,7 +126,9 @@ void Search::SendUciInfo() REQUIRES(nodes_mutex_) {
     uci_infos.emplace_back(common_info);
     auto& uci_info = uci_infos.back();
     if (score_type == "centipawn") {
-      uci_info.score = 290.680623072 * tan(1.548090806 * edge.GetQ(default_q));
+      uci_info.score = 111.714640912 * tan(1.5620688421 * edge.GetQ(default_q));
+	} else if (score_type == "centipawn_2018") {
```
borg323: Your editor used tabs here

QxC4eva: Yeah sorry it should be fixed now X)

FinalCut64:
Maybe a quick sanity check to see if this change actually helps in playing strenght or at least is elo neutral?
I mean, I love features and tweaks, and this one looks nice but the main objective here is to improve the level of play.
I don't have a gpu otherwise I would gladly test this.

gonzalezjo:
It’s strictly cosmetic and has no impact on playing strength.

oscardssmith:
it could change strength with adjudication, but I agree it can be merged without worrying about it.

agentspock:
Hi, there i am converting the Cp back to a Q using the formula : Qpercent = atan(cp/290.680623072)/1.548090806;   I am getting valules, lie for cp = -0.15, the Q is -0.00033. That does translate to -0.03% chance of winning? Another move now shows, cp=0.67  and my calculated Q = 0.00148 so is that 0.148% chance of winning?  And another cp = 17.3, Q = 0.0383, so is that 3.83% chance of winning? Any guidance would be appreciated.

PR #842, opened 2019-05-02
update download links in README

borg323:
Some help requests originated from people downloading the 0.19 release branch, following the instructions. The updated links will get the latest code from the release/0.21 branch in every case.

Note that I couldn't find a github api to get the latest release zip and tar, so the links point to the head of the release/0.21 branch (like the `git clone` command).

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,11 +17,11 @@ Versioning follows the Semantic Versioning guidelines, with major, minor and pat
 Download using git:
 
 ```
-git clone -b release/0.19 --recurse-submodules https://github.com/LeelaChessZero/lc0.git
+git clone -b release/0.21 --recurse-submodules https://github.com/LeelaChessZero/lc0.git
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -17,11 +17,11 @@ Versioning follows the Semantic Versioning guidelines, with major, minor and pat
 Download using git:
 
 ```
-git clone -b release/0.19 --recurse-submodules https://github.com/LeelaChessZero/lc0.git
+git clone -b release/0.21 --recurse-submodules https://github.com/LeelaChessZero/lc0.git
```
mooskagh: Maybe add a banner with large bold red flashing all-caps font:
```
REPLACE 0.21 WITH THE LATEST RELEASE VERSION
```
?

borg323: Adding this for future reference (since the PR is already merged): This will only address the clone command, the archive download links are not showing the version number - If I don't find something a set that always gets the latest release, the phrasing should be revised.

PR #843, opened 2019-05-02
print Lc0 in red

borg323:
Adds minimal WIN32 handling of ANSI CSI escape codes to StderrLogMessage so that the Lc0 logo is shown in red.


src/main.cc
```diff
@@ -35,9 +35,9 @@
 
 int main(int argc, const char** argv) {
   LOGFILE << "Lc0 started.";
-  CERR << "       _";
+  CERR << "\033[1;31m       _";
```
mooskagh: Maybe add `utils/esc_codes.h` with something like:
```cpp
class EscCodes {
   EscCodes() = delete; // To disallow instantiation.
   static constexpr const char * kRed = "\033[1;31m";
};
```

And then use it as `CERR << EscCodes::kRed << ...`
?

It you think it's overengineering for a single use case, feel free to leave it as it is.

src/utils/logging.cc
```diff
@@ -30,11 +30,55 @@
...
+#include <windows.h>
+#endif
+
 namespace lczero {
 
 namespace {
 size_t kBufferSizeLines = 200;
 const char* kStderrFilename = "<stderr>";
+
+#ifdef _WIN32
```
mooskagh: It's not nice to have platform-specific code under ifdefs, especially given it's not that related to the core purpose of logging.
Maybe add this function to `filesystem.h` (and corresponding `.win32.cc`/ `.posix.cc`)?
It's not that filesystem related but adding one more set of files is probably an overkill. Or maybe `filesystem.*` may be renamed to something like `os.*`, but it doesn't look too good either.

src/utils/logging.cc
```diff
@@ -80,8 +124,23 @@ StderrLogMessage::StderrLogMessage(const char* file, int line)
...
+    if (len == std::string::npos) break;
+    s = s.substr(len);
+    len = 2;
+    while (s[len] && !(s[len] >= 0x40 && s[len] <= 0x7E)) len++;
+    std::cerr.flush();
+    std::cerr << HandleCSI(s.substr(0, ++len));
+    s = s.substr(len);
+  };
+  std::cerr << std::endl;
+  log_ << s2;
```
mooskagh: I'm not sure it's a good idea to strip anything (including esc codes from the log).

src/utils/logging.cc
```diff
@@ -80,8 +124,23 @@ StderrLogMessage::StderrLogMessage(const char* file, int line)
     : log_(file, line) {}
 
 StderrLogMessage::~StderrLogMessage() {
-  std::cerr << str() << std::endl;
-  log_ << str();
+  std::string s = str();
+  std::string s2 = "";
+  size_t len;
```
mooskagh: This should be factored out to a separate function.
Also this feels a bit wastful to do for every single log message (especially given that lots of string copying is going on, and we don't yet use std::string_view as we don't use C++17), but objectively there's no difference.

src/main.cc
mooskagh: Ok, maybe it's overengineering, but maybe just defining those as constants one line above is good enough.

borg323:
The new version has less platform dependent code, skipping ANSI escape codes on older windows that don't support them (there is an easy way to detect their availability and enable them). Also there is no log filtering, the codes will pass through.

src/utils/esc_codes.h
```diff
@@ -0,0 +1,68 @@
...
+  static const char* RedBg() { return flag_ ? "\033[41m" : ""; }
+  static const char* GreenBg() { return flag_ ? "\033[42m" : ""; }
+  static const char* YellowBg() { return flag_ ? "\033[43m" : ""; }
+  static const char* BlueBg() { return flag_ ? "\033[44m" : ""; }
+  static const char* MagendaBg() { return flag_ ? "\033[45m" : ""; }
+  static const char* CyanBg() { return flag_ ? "\033[46m" : ""; }
+  static const char* WhiteBg() { return flag_ ? "\033[47m" : ""; }
+
+ private:
+  static bool flag_;
```
mooskagh: Something like `enabled_`?
And as there's data to store anyway, why not to make it a usual class (e.g. singleton if init() is doing something just once)

Both suggestions are optional, it's fine to merge it as it is.

src/utils/esc_codes.cc
```diff
@@ -0,0 +1,49 @@
...
+bool EscCodes::flag_;
+
+void EscCodes::Init() {
+#ifdef _WIN32
+  HANDLE h = GetStdHandle(STD_OUTPUT_HANDLE);
+  DWORD mode;
+  GetConsoleMode(h, &mode);
+  flag_ = SetConsoleMode(h, mode | ENABLE_VIRTUAL_TERMINAL_PROCESSING);
+#else
+  flag_ = true;
```
mooskagh: Well, for unix it would be nice to look into `$TERM` environment variable too, but I don't know what to look there for.

PR #846, opened 2019-05-07
Enable Exponential Decay Option

AlexisOlson:
This piggybacks off of the linear decay option and does not require any new parameters.


AlexisOlson:
This PR does not change any existing behavior but allows for an exponential noise decay if a negative integer is passed in instead of a positive integer. That integer determines the decay rate. This image shows what the decay functions looks like for parameter -50 versus +50 with an initial temp of 3.

Decay

mooskagh:
Is there a need for it? If yes, I'd suggest to have `--tempdecay-mode` enum switch for that (like FPU strategy, for example).

But I thought that most complaints were that linear decay is too fast, not that it's too slow, and this change makes it decay even faster.

Temperature less than 0.3..0.5 is almost the same as 0, it's very improbable that it will have any effect, so there's no reason to have that long tail too. (also there is a danger of float overflow if not computed carefully enough)

AlexisOlson:
The speed of the decay is primarily governed by the moves to 0 (or to 1/10th). Given any exponential curve, there exist lines that are both steeper and shallower. With a linear decay, if you want to drop fast for the first couple moves, then temp is gone entirely rather quickly. Exponential decay allows you to drop faster for earlier moves and slower for later moves.

To me, the green curve seems more likely to be a desired decay than any of the straight lines.

Comparison

mooskagh:
What I'm saying is the function should be concave (convex upwards), not convex downwards.

For temperature, only small range (say, 0.4 -- 1.5) makes sense.
More than 1.5 behaves like equiprobable random, less than 0.4 is almost the same as 0, in almost all cases just best move will be picked.

So properties of the temperature function should be like following:
1. It should stay close for initial value (which is usually 1) at first and not decay fast.
2. Later it should decay fast, and when it reached low value (e.g. 0.4), it's better to make it go towards zero faster.

Exponential decay is convex (downwards), so it makes the function to stay longer outside of reasonable parameters, and only have "reasonable" parameters for short period of time.

mooskagh:
Something like `(1-move/tempdecay_moves)^(1/factor)` is what I suggest.
Currently factor is 1, but we could make it parameter and set to 2 or 3.

https://www.google.com/search?ei=2erRXLyRNoKHk74P_MmPgAQ&q=%281-x%2F10%29%5E%281%2F3%29&oq=%281-x%2F10%29%5E%281%2F3%29&gs_l=psy-ab.3..0i8i30l10.78413.80741..81118...3.0..0.226.553.1j2j1......0....1..gws-wiz.......0i71j0i7i30j0i8i7i30j0i7i5i30.HB4LQVOzsUs

AlexisOlson:
Interesting. A concave down function like you suggest is counter-intuitive to me if the goal is to generate opening diversity without adversely affecting strength too much. I'll have to think about that.

If you add a factor `f` as you suggest, then if you allow it to be a float > 0, then you could get either concave up with `f < 1` or concave down with `f > 1`.

Factor

jhorthos:
mooskagh I have an extensive set of tests that show temp in any range has a significant effect on Elo - it is non-linear so less of an issue around 0.1 or 0.2, but 0.4 has a pretty hefty bad effect. I think most of the data is on some tab on sheet3.

AlexisOlson:
So would it be better if I abandon this PR and write one with a `tempdecay-factor` (my `f` above) instead which will allow custom concavity?

I think `0.1 < f < 10` with `f=1` (linear) default would be reasonable.

jhorthos:
Both. :-) Personally I like convex.

jhorthos:
Oh another thing is that "1.5 behaves like equiprobable random" is not true in theory or in fact. temp 1.1, 2.0, 3.0, and 4.0 all have quite different properties in my openings (albeit with linear temp decay to it isn't simple).

Naphthalin:
I think this PR is obsolete now due to #1212 being merged, giving enough opportunities to manipulate the temp decay. @AlexisOlson can you check that?

AlexisOlson:
@Naphthalin It's certainly not exactly equivalent but should be simple and close enough.

PR #850, opened 2019-05-13
verify backend-opts syntax correctness

borg323:
This is based on an idea by @mooskagh - at least I hope so, it was a few months ago...


src/utils/optionsdict.cc
```diff
@@ -322,4 +322,14 @@ void OptionsDict::AddSubdictFromString(const std::string& str) {
   parser.ParseMain(this);
 }
 
+std::string OptionsDict::FindNotUsed() const {
```
mooskagh: I'd make a function like `CheckAllOptionsRead()`, which throws exception itself, but probably matter of taste.

src/neural/factory.cc
```diff
@@ -119,7 +119,12 @@ std::unique_ptr<Network> NetworkFactory::LoadNetwork(
   OptionsDict network_options(&options);
   network_options.AddSubdictFromString(backend_options);
 
-  return NetworkFactory::Get()->Create(backend, weights, network_options);
+  auto ptr = NetworkFactory::Get()->Create(backend, weights, network_options);
+
+  auto x = network_options.FindNotUsed();
+  if (x.size()) throw Exception("Unknown backend option \"" + x + "\"");
```
mooskagh: Why not `(!x.empty())`? Seems more aligned with intent.

src/utils/optionsdict.cc
mooskagh: ...And would make this function get `(std::string& path_from_parent)` (for `--backend-opts=a(cpu=4)` it would be "a", so that it could complain about `a.cpu` being unknown).

src/utils/optionsdict.h
```diff
@@ -95,18 +95,26 @@ class OptionsDict : TypeDict<bool>,
...
+  // Returns the first option in the dict that has not been read to
+  // find syntax errors in options added using AddSubdictFromString
+  std::string FindNotUsed() const;
+
   bool HasSubdict(const std::string& name) const;
 
  private:
   const OptionsDict* parent_ = nullptr;
   std::map<std::string, OptionsDict> subdicts_;
+  mutable std::map<std::string, bool> used_;
```
mooskagh: Maybe instead modify TypeDict?
```
template <typename T>
class TypeDict {
 protected:
  struct V {
      const T& Get() const { is_used_ = true; return value_; }
      void Set(const T& v) { is_used_ = false; value_ = v; }
      bool IsSet() const { return is_used_; }
    private:
      mutable bool is_used_ = false;
      T value_;
  };
  std::unordered_map<std::string, V> dict_;
};
```

src/utils/optionsdict.cc
mooskagh: ...So it would be nice to also check subdicts, and it would also work if value is accessed through some child..
Cases like that will fail though:
`gpu=1,(gpu=2),(gpu=3)`, if gpu=1 is not used. But probably that's good.

src/neural/factory.cc
borg323: First thing that came to mind and never revised - will change it.

src/utils/optionsdict.cc
borg323: Hmm, I didn't consider the subdict case - need more testing. The reason for not throwing the exception directly was to have a backend-opts specific error message.

mooskagh: Maybe throw another exception (but still derived from `lczero::Exception`) and rethrow converted to more readable exception in the calling code?
Maybe it's overengineering though.

src/utils/optionsdict.h
borg323: Is the added complexity worth it?

src/neural/factory.cc
```diff
@@ -119,7 +119,11 @@ std::unique_ptr<Network> NetworkFactory::LoadNetwork(
   OptionsDict network_options(&options);
   network_options.AddSubdictFromString(backend_options);
 
-  return NetworkFactory::Get()->Create(backend, weights, network_options);
+  auto ptr = NetworkFactory::Get()->Create(backend, weights, network_options);
+
```
mooskagh: That it totally matter of taste, but I wouldn't add empty lines here. :)

src/utils/optionsdict.cc
```diff
@@ -322,4 +322,17 @@ void OptionsDict::AddSubdictFromString(const std::string& str) {
   parser.ParseMain(this);
 }
 
+void OptionsDict::CheckAllOptionsRead(
+    const std::string& path_from_parent) const {
+  std::string s = path_from_parent.empty() ? "" : path_from_parent + '.';
```
mooskagh: If you would add `'.'` at the function call at line 334, that check would not be needed at all.

src/utils/optionsdict.cc
```diff
@@ -322,4 +322,17 @@ void OptionsDict::AddSubdictFromString(const std::string& str) {
...
+  for (auto const& option : used_) {
+    if (!option.second) {
+      throw Exception("Unknown option: " + s + option.first);
+    }
+  }
+  for (auto const& dict : subdicts_)
+    dict.second.CheckAllOptionsRead(s + dict.first);
```
mooskagh: Braces around the body of the loop as it's on a separate line.

src/utils/optionsdict.h
mooskagh: It looks cleaner that way..
(no need to worry about dicts_ and used_ being in sync).
Also if there is a type mismatch (e.g. expected `gpu=2` but passed `gpu=2.0` instead), current code won't notice it.

mooskagh: Also unsed empty subdictionary is not detected (with either approach), probably not a problem.

src/utils/optionsdict.cc
borg323: I would need to add the dot in factory.cc as well. Moreover, the idea is to be able to call this with an empty string, I'm not convinced `backend.option` is the best format for this (although this is the way I have it now).

src/utils/optionsdict.h
borg323: Type mismatches were already caught, but now the error messages are more informative. Any suggestions for a cleaner way to do this?

src/utils/optionsdict.cc
```diff
@@ -322,4 +322,32 @@ void OptionsDict::AddSubdictFromString(const std::string& str) {
   parser.ParseMain(this);
 }
 
+void OptionsDict::CheckAllOptionsRead(
+    const std::string& path_from_parent) const {
+  std::string s = path_from_parent.empty() ? "" : path_from_parent + '.';
+  for (auto const& option : TypeDict<bool>::dict_) {
```
mooskagh: Maybe add a function to TypeDict to avoid some code duplication?
Something like:

```cpp
void EnsureNoUnusedOptions(const std""string& prefix, const std::string& type_name) {
  for (auto const& option: dict_) {
     // ... ("Unknown " + type_name + " option: " + prefix + option.first);
  }
}
```

And then here just:
```
   TypeDict<int>::EnsureNoUnusedOptions("integer", s);
   // ...
```
?

src/utils/optionsdict.h
```diff
@@ -95,6 +104,10 @@ class OptionsDict : TypeDict<bool>,
   // is still in scope, when the parent pointer is used
   void AddSubdictFromString(const std::string& str);
 
+  // Throws an exception for the first option in the dict that has not been read
+  // to find syntax errors in options added using AddSubdictFromString
```
mooskagh: Period character at the end of the sentence. :)

PR #852, opened 2019-05-16
SE layer fix when not using fused kernel

ankan-ban:
- only SE layer needs transposed weights.
- need to store non-transposed weights too in case we have to fall back.


Tilps:
I'm feeling confused - the PR title says this is a fix for when not using fused kernel - but I see 0 logical changes outside of the kUseFusedSELayerpath.

ankan-ban:
The issue was that we were always transposing weights for the FC layers when kUseFusedSELayer is true (in load weights function). If a matching filter size isn't found, the fused implementation can return failure and we fall back to non fused path (even when kUseFusedSELayer is set), and the non fused path requires non transposed weights.

Tilps:
Ahh - could alternatively kUseFusedSELayer not be a constant but instead calculated by whether we actually support fusing - so there is no fallback?
I assume there is negligable performance impact of this change as is? If so then maybe it doesn't matter.

ankan-ban:
There should be no performance impact (only one extra copy done at loading time).

kUseFusedSELayer flag was supposed to be only a compile time for debugging.

PR #853, opened 2019-05-17
Q value difference pruning

MelleKoning:
This is a test-pull request on trying to improve smart pruning by using the known Q value of the best root-move.

The idea here is that if the best root-move is clearly better than all the other root-moves being considered to just make that move quicker and thus save some time for future moves.

Why?
- Suppose your opponent goes for a queen exchange and takes your queen. The logical move to do is to take back your opponents queen and don't think about that all too much. This way you save some time for future moves.

Additional information:
Within the search there already exists smart pruning as well as a cut-off when the best found move has such a large number of playouts that no other root move can catch up. That is still in place. The only addition in this PR is that there will be a slightly faster cutoff in case the number of playouts has reached a certain threshhold AND there is a slight Q difference with all the other root moves.

The current values for this earlier cut-off have been tested on very short time controls and very low node counts. It's a known fact that on low node counts, higher values for smart pruning are very effective, so this PR should also be tested with Longer Time Controls (LTC) with higher node counts to see if this improves strength, if at all. 

A test is running currently for 100 games at +1.5 sec but as I do not have a good GPU would like to see if others can test, just to check if this improves strength or not.

Thoughts on strength:
- The cutoff early as shown here has a higher risk of playing a _worse_ move and this should be weighted against the advantage for saving time for those cases where more thinking is useful _in future_. 
- I believe that eventually the TimeControls should be directed by the net outputs instead of trying to handcraft within the search like in this PR.

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -879,7 +879,7 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend(
 
   // Fetch the current best root node visits for possible smart pruning.
   const int64_t best_node_n = search_->current_best_edge_.GetN();
-
+  
   // True on first iteration, false as we dive deeper.
   bool is_root_node = true;
   uint16_t depth = 0;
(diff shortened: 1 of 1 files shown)


MelleKoning:
TM also mentioned in issue #869 
In this PR the smart pruning is just made a bit 'sharper' based on Q difference between the best move and the remaining calculated Q's of the current evaluation/position.

Naphthalin:
This is an interesting approach, and might work well together with our smart-pruning logic in general. However, crem's new "smooth nodes per move" time manager might already fix this as well, as these are the situations with nearly full tree reuse.

PR #854, opened 2019-05-17
Estimated moves testing

AlexisOlson:
Reshape estimated moves curve.

This tweak is intended to help performance with sudden death TC and is only for testing purposes at this time.


AlexisOlson:
Preliminary testing shows this not to be a good tweak. I do not recommend further testing until I've analyzed my results and made further changes.

PR #855, opened 2019-05-18
Add 320 and 352 channel support for fused SE layer

ankan-ban:
On RTX 2060:

```
352x16 network
baseline: Benchmark final time 9.84818s calculating 11340.2 nodes per second.
opt: Benchmark final time 9.94462s calculating 11666.1 nodes per second.

320x24 network:
baseline: Benchmark final time 7.73397s calculating 9637.74 nodes per second.
opt: Benchmark final time 7.67333s calculating 10030.6 nodes per second.
```


jhorthos:
What should the SE ratio be for the 320x24 network?

PR #858, opened 2019-05-23
 use Eigen as a BLAS alternative 

borg323:
This is a port of <https://github.com/leela-zero/leela-zero/pull/1692>.


oscardssmith:
How does it compare performance-wise?

borg323:
I did a quick test with a T35 net and it is about 8-10% quicker than MKL on my i5. More tests are of course welcome.

lealgo:
Builds and runs on Android:

```
HWHWI:/data/local/tmp $ ./lc0 benchmark -w /sdcard/lc0/36089 --threads=8 --max-prefetch=4 --minibatch-size=128                                                                                              
       _
|   _ | |
|_ |_ |_| v0.22.0-dev built May 23 2019
Loading weights file from: /sdcard/lc0/36089
Creating backend [blas]...
Using Eigen version 3.3.7
BLAS max batch size is 256.
Benchmark time 268ms, 2 nodes, 7 nps, move e2e4
Benchmark time 474ms, 3 nodes, 6 nps, move e2e4
Benchmark time 679ms, 5 nodes, 7 nps, move e2e4
Benchmark time 883ms, 9 nodes, 10 nps, move e2e4
Benchmark time 1089ms, 14 nodes, 12 nps, move e2e4
Benchmark time 1438ms, 24 nodes, 16 nps, move e2e4
Benchmark time 1882ms, 34 nodes, 18 nps, move e2e4
Benchmark time 2101ms, 40 nodes, 19 nps, move e2e4
Benchmark time 2363ms, 58 nodes, 24 nps, move e2e4
Benchmark time 2705ms, 80 nodes, 29 nps, move e2e4
Benchmark time 3054ms, 102 nodes, 33 nps, move e2e4
Benchmark time 3365ms, 138 nodes, 41 nps, move e2e4
Benchmark time 3935ms, 182 nodes, 46 nps, move e2e4
Benchmark time 4103ms, 203 nodes, 49 nps, move e2e4
Benchmark time 4491ms, 247 nodes, 54 nps, move e2e4
Benchmark time 4973ms, 272 nodes, 54 nps, move e2e4
Benchmark time 5580ms, 332 nodes, 59 nps, move e2e4
Benchmark time 6490ms, 420 nodes, 64 nps, move e2e4
Benchmark time 7038ms, 454 nodes, 64 nps, move e2e4
Benchmark time 7754ms, 526 nodes, 67 nps, move e2e4
Benchmark time 7818ms, 526 nodes, 67 nps, move e2e4
bestmove e2e4
Benchmark final time 8.85549s calculating 71.1423 nodes per second.

HWHWI:/data/local/tmp $ file lc0
lc0: ELF shared object, 64-bit LSB arm64, dynamic (/system/bin/linker64), for Android 28, built by NDK r19c (5345600), not stripped
```

lealgo:
Android build :D

[lc0-eigen-aarch64.zip](https://github.com/LeelaChessZero/lc0/files/3217283/lc0-eigen-aarch64.zip)

src/neural/blas/network_blas.cc
```diff
@@ -379,6 +384,11 @@ BlasNetwork::BlasNetwork(const WeightsFile& file, const OptionsDict& options)
                                                        pol_channels, channels);
   }
 
+#ifdef USE_EIGEN
+  std::cerr << "Using Eigen version " << EIGEN_WORLD_VERSION << "."
```
mooskagh: Use `CERR` from logging.h instead so that it also gets into log?

src/neural/blas/convolution1.cc
```diff
@@ -44,7 +56,7 @@ void Convolution1::Forward(const size_t batch_size, const size_t input_channels,
 
     const float* batch_input = input + i * kSquares * input_channels;
     float* batch_output = output + i * kSquares * output_channels;
-
+#ifndef USE_EIGEN
```
mooskagh: Could BLAS and Eigen in theory co-exist?
Then if would be better to have templated functions (e.g. `<bool is_eigen>`) which would be shown up as different backends.

See for example:
cudnn vs cudnn-fp16
https://github.com/LeelaChessZero/lc0/blob/master/src/neural/cuda/network_cudnn.cc#L805
https://github.com/LeelaChessZero/lc0/pull/5#issuecomment-393979584

tensorflow vs tensorflow-cpu
https://github.com/LeelaChessZero/lc0/blob/master/src/neural/network_tf.cc#L342

May be not that straighforward though as different header files are needed (so #ifdefs will be needed in any case), and calling non-included functions in `if(false) {}` won't compile.
Calling them from non-instanciated function template specializations will work though:

```cpp
template<>
DoStuff<false /* is eigen*/>() {
  known_functions();
}

template<>
DoStuff<true /* is eigen*/>() {
  unknown_functions();  // Is fine
}

#if defined(HAVE_EIGEN)
REGISTER_NETWORK("eigen", MakeBlasNetwork<true>, 90)
#endif
#if defined(HAVE_BLAS)
REGISTER_NETWORK("blas", MakeBlasNetwork<false>, 90)
#endif
```

Not sure if easy enough to be bothered with, but that way it would be nicer.

Also there is this brute-force way to allow co-existance:
```cpp
template<bool is_eigen> MyFunction() {
#ifdef HAVE_EIGEN
  if (is_eigen) {
    // Do eigen stuff.
  }
#endif
#ifdef HAVE_BLAS
  if (!is_eigen) {
    // Do blas stuff.
  }
#endif
}

#if defined(HAVE_EIGEN)
REGISTER_NETWORK("eigen", MakeBlasNetwork<true>, 90)
#endif
#if defined(HAVE_BLAS)
REGISTER_NETWORK("blas", MakeBlasNetwork<false>, 90)
#endif
```

src/neural/blas/network_blas.cc
borg323: Should I do this for the other instances in the file as well, or save that for a separate pr?

mooskagh: Either is fine, probably better to keep changes separate.

src/neural/blas/convolution1.cc
borg323: I don't think there is a reason for this: eigen is only suitable for the build machine, so blas is the way to go for redistributable cpu only binaries. The only use case I can see is for benchmarking between eigen and mkl on a specific machine - is there any other use case?

PR #859, opened 2019-05-24
Create readme_for_dummies

TesseractA:
This file intends to translate implied instructions to actual instructions. This is intended to reduce the people that need help, and answer some questions about how the engine operates.

This file is only a start. It could also link to a Youtube/Video tutorial as well.


TesseractA:
I could make significant improvements to this file, however it seems unnecessary if you can link a Youtube/Twitch video that can do the exact same thing.

PR #865, opened 2019-05-28
Add pushd command in build.sh

TesseractA:
Hmm I'm not entirely sure this is right. I'm rather new at doing bash commands.

This was intended to stack directories and enable one to do a remote execution of this shell script into build.sh.

Almost all credit to borg.


borg323:
There have been a few help requests on discord from people trying to run build.sh from other directories.
This change has been tested on both linux and mac.

PR #868, opened 2019-05-30
Embed the weights file into lc0 executable

lealgo:
The main motivation here is to use it in some platforms like Android that may have limited configuration options. It could also simplify packaging. One could imagine a situation where is desirable to deliver just one binary that contains a preset configuration & network weights file.

The proposed solution is very simple and should work across platforms. It consists of the following: during the build process the compressed weights file is converted into source files (cpp+h) as `const unsigned char[]`. The sources are conditionally included from `loader.cc` where instead of a file, the array is gunzipped and fed to the existing protobuf parse. The mechanism of auto-discover and load from file is unaffected, and the embedded weights are loaded only when those first are not specified.

The shortcoming of this idea is the increased size of the executable. For small network sizes this is no big deal even on mobile devices. And maybe even with the full 20b networks we can keep the executable size below or about the same size as other "portable" apps.

I'm really open to criticism on this idea and the proposed implementation.


oscardssmith:
I think this makes a lot of sense and default bundling the strongest tested net (so a 10b for the CPU and Android build) would probably make set up way easier for less technically inclined users.

borg323:
A quick comment: I think the `embedder.py` should move into an existing directory, either `third_party` if completely unchanged or else to `scripts` (and the modifications documented in a comment). Opinions?

meson.build
```diff
@@ -137,7 +137,23 @@ else
...
+embedder = find_program('embed/embedder.py', required: false)
+if get_option('embed') and embedder.found()
+  # Rule to convert binary files to source + header.
+  embedder_gen = generator(embedder,
+    output : ['@BASENAME@.cpp', '@BASENAME@.h'],
+    arguments : ['@INPUT@', '@OUTPUT0@', '@OUTPUT1@']
+  )
+  # Add network weights file if found
+  weights_file = 'embed/embedded_weights'
+  if run_command('test', '-f', weights_file).returncode() == 0
```
borg323: This doesn't look like it will work on windows. [checkdir.py](https://github.com/LeelaChessZero/lc0/blob/master/checkdir.py) was implemented to check the existence of directories in a corss-platform way, so a similar approach should work here as well.

meson_options.txt
```diff
@@ -127,3 +127,8 @@ option('protobuf-3-6-0',
        type: 'boolean',
        value: false,
        description: 'Use the protobuf 3.6.0 subproject')
+
+option('embed',
+       type: 'boolean',
```
borg323: What about making this a `string` option with the file to embed?

src/neural/factory.cc
```diff
@@ -59,7 +62,11 @@ NetworkFactory::Register::Register(const std::string& name, FactoryFunc factory,
 }
 
 void NetworkFactory::PopulateOptions(OptionsParser* options) {
+#ifdef EMBEDDED_WEIGHTS
+  options->Add<StringOption>(NetworkFactory::kWeightsId) = kEmbedded;
```
borg323: Other options to consider instead of changing the default:
- use the embedded net if autodiscovery fails.
- add the embedded net to the search list (with a timestamp same with the executable) so that newer nets are preferred.

Alternatively (or simultaneously) have a separate (bool) option to use the builtin net.

lealgo: I like the simplicity of you first bullet. If auto-discovery fails then load the embedded. That way the default is kept and also `factory.cc` is left untouched. We lose only one small thing, the ability to specify the embedded weights by config, forcing the user to hide the weights from the auto-discover.

meson_options.txt
lealgo: I thought of that first, but then I realized that you compile against one exact header and const name from the real source (`loader.cc`). Then the build process must supply such header and symbol. The original data file name could be changed though, having meson pass the header name to the embedder.py as a new second parameter. That way the builder calls with `-Dembed_weights=/my/weights` and meson calls embedder.py with the real filename and the fixed header name for the source (for weights). Later maybe another `-Dembed_config=/my/config`

src/neural/factory.cc
borg323: I wasn't going for simple, just more intuitive user interface. I expect the `factory.cc` changes should stay to be able to force using the embedded net (or go with an additional option for this).

scripts/embedder.py
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
lealgo: I like it. More control on the header is good. I'll try something later. Thanks!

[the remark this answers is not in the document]
lealgo: Sorry, I've only had time to try the hex representation. Cool because it's more compact than decimal BUT now as it's a string we have offended the compiler:

`warning: string literal of length 6998518 exceeds maximum length 65536 that C++ compilers are required to support [-Woverlength-strings]`

Not sexy anymore. The array doesn't have that limitation, and after all it lives as a fat source only for a brief lapse until it's compiled to object. 

By the way @mooskagh your int to hex had issues. I'm not a python guy but `bytes()` is [0..255] and it wasn't needed, worked with just `rb'\x%02x' % x`.

PR #884, opened 2019-06-10
Factor out time management code

mooskagh:
Not sure it made code better, not worse, but the intention was the following:

Factor out time management code from search/engine to a separate directory in order to:
1. Be able to write better time manager (e.g. carry nps from one search to another, change time budget as the search goes, etc).
2. Move complexity out of Search and Engine classes.
3. Have reusable part of the code which can be used for search v2.

It should be mostly no-op, but there are some differences, e.g.
- remaining time and nodes are stored per thread, not globally, that means that threads may sometimes disagree on that.
- `lc0 benchmark` doesn't support smart pruning now.

I'll probably add tests as the change is large and messay.


mooskagh:
Instead of passing limits to a search and search itself watching for it, this change makes it work the opposite:

1. "Stopper" is passed to the search.
2. Search calls stopper after every iteration, passing needed stats.
3. Stopper returns whether search should be stopped (and also passes "hints": estimated nodes remaining, and estimated time remaining)

oscardssmith:
This change seems great. Getting unnecessary complication out of search is highly appreciated.

borg323:
`lc0 benchmark` not supporting smart pruning is probably an improvement.

gonzalezjo:
> Not sure it made code better, not worse

Seems like it's better overall.

src/mcts/timemgr/stoppers.cc
```diff
@@ -0,0 +1,235 @@
...
+}
+
+int64_t TimeLimitStopper::GetTimeLimitMs() const { return time_limit_ms_; }
+
+///////////////////////////
+// DepthStopper
+///////////////////////////
+bool DepthStopper::ShouldStop(const IterationStats& stats, TimeManagerHints*) {
+  if (stats.average_depth >= depth_) {
+    LOGFILE << "Stopped search: Reached depth.";
```
borg323: return true;

mooskagh: Done.

src/mcts/stoppers/stoppers.cc
```diff
@@ -0,0 +1,235 @@
+/*
+  This file is part of Leela Chess Zero.
+  Copyright (C) 2018-2019 The LCZero Authors
```
borg323: Is this intentional? All other files in this directory mention only 2019.

PR #885, opened 2019-06-10
Set probabilities for illegal moves in a loop

DanielUranga:
Set probabilities for illegal moves in a loop instead of using std::memset.
`std::memset(result.probabilities, -1, sizeof(result.probabilities));` is setting all bytes in `result.probabilites` as `0xffffffff`.

### DIFF
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -339,7 +339,9 @@ V4TrainingData Node::GetV4TrainingData(GameResult game_result,
   // Prevent garbage/invalid training data from being uploaded to server.
   if (total_n <= 0.0f) throw Exception("Search generated invalid data!");
   // Set illegal moves to have -1 probability.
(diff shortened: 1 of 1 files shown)


src/mcts/node.cc
```diff
@@ -339,7 +339,9 @@ V4TrainingData Node::GetV4TrainingData(GameResult game_result,
   // Prevent garbage/invalid training data from being uploaded to server.
   if (total_n <= 0.0f) throw Exception("Search generated invalid data!");
   // Set illegal moves to have -1 probability.
-  std::memset(result.probabilities, -1, sizeof(result.probabilities));
+  for (auto& probability : result.probabilities) {
```
mooskagh: `std::fill(begin(result.probabilities), end(result.probabilities), -1);` ?

PR #887, opened 2019-06-13
Check for null parent in search backup.

Mardak:
r?@mooskagh or @Tilps Fix a crash for FEN with no move history and no moves.

before
```
position fen 8/8/8/8/8/8/8/8 w - - 0 1
go nodes 1
Segmentation fault: 11

position fen 8/8/8/8/8/8/8/8 w - - 0 1 moves a1a1
go nodes 1
bestmove a1a1
```

after
```
position fen 8/8/8/8/8/8/8/8 w - - 0 1
go nodes 1
bestmove a1a1

position fen 8/8/8/8/8/8/8/8 w - - 0 1 moves a1a1
go nodes 1
bestmove a1a1
```

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -1301,6 +1301,9 @@ void SearchWorker::DoBackupUpdateSingleNode(
     }
     n->FinalizeScoreUpdate(v, d, node_to_process.multivisit);
 
+    // Nothing left to do without ancestors to update.
+    if (!p) break;
+
     // Convert parents to terminals except the root or those already converted.
     can_convert = can_convert && p != search_->root_node_ && !p->IsTerminal();
 


src/mcts/search.cc
```diff
@@ -1302,7 +1302,8 @@ void SearchWorker::DoBackupUpdateSingleNode(
     n->FinalizeScoreUpdate(v, d, node_to_process.multivisit);
 
     // Convert parents to terminals except the root or those already converted.
-    can_convert = can_convert && p != search_->root_node_ && !p->IsTerminal();
+    can_convert =
```
mooskagh: Probably cleaner to add `if (!p) break;` in a line above.
Although I'm a bit lost how it happened that the loop condition is not enough anymore..

Mardak: Before #822, there were no uses of the parent within the loop.

PR #889, opened 2019-06-13
Prefer terminal wins (and avoid loss) when picking best children.

### LINKED ISSUE #837: Search or move selection should prefer (shorter) mates
Not implemented in #822 but exists in #700, I quickly put together a buggy implementation to explore what's there to consider, and what is actually meaningful to implement.

Screen Shot 2019-04-22 at 10 26 44 AM
```
position fen 7k/6pp/5p2/8/5R1Q/8/7K/8 b - - 0 1 moves g7g6 h4f6 h8g8

f4e4 N:    9282 (P:  3.87%) (Q:  1.00000) (U: 0.05047) (Q+U:  1.05047) (V:  1.0000) (T101) 
h2h3 N:   15911 (P:  6.63%) (Q:  1.00000) (U: 0.05047) (Q+U:  1.05047) (V:  1.0000) (T101) 
f6e6 N:   44797 (P: 36.96%) (Q:  0.99995) (U: 0.10002) (Q+U:  1.09997) (V:  0.9984) 
f6e7 N:  108277 (P:  1.32%) (Q:  1.00000) (U: 0.00147) (Q+U:  1.00147) (V:  1.0000) (T3) 
f6d8 N:  121259 (P:  1.48%) (Q:  1.00000) (U: 0.00147) (Q+U:  1.00147) (V:  1.0000) (T3) 
f6f7 N:  252327 (P:  2.03%) (Q:  1.00000) (U: 0.00097) (Q+U:  1.00097) (V:  1.0000) (T2) 
f6f8 N:  519160 (P:  2.03%) (Q:  1.00000) (U: 0.00047) (Q+U:  1.00047) (V:  1.0000) (T1) 
```

Here I embedded distance in the node's Q value by multiplying Q with the distance, e.g., q_ == 1.0f for a move is checkmate and q_ == -2.0f is checkmated in 2 (ply), and exposed `GetQ` as bounded [-1, 1] and added a `GetDistance` for use in Backup and PickNode and VerboseStats. The actual implementation will probably be different when optimizing for performance of the desired feature.

The more interesting question is what do we actually want?

If we have search focus on moves that have a mate/terminal win, depending on how much that is biased, it can prevent finding an even shorter mate. E.g., above h2h3 has the highest prior (using 50783) of the terminals and preferring it could have prevented search from realizing the lower prior move f6f8 is checkmate.

On the flip side, allowing search to explore more even with a known mate could result in search still directing too many visits to moves with high priors, e.g.,
```
position fen 7k/6pp/5p2/8/5R1Q/8/7K/8 b - - 0 1 moves g7g6 h4f6 h8g8 f6e6 g8g7

e6g6 N:     132 (P:  1.12%) (Q:  1.00000) (U: 0.06639) (Q+U:  1.06639) (V:  1.0000) (T100) 
e6f7 N:     175 (P:  1.50%) (Q:  1.00000) (U: 0.06749) (Q+U:  1.06749) (V:  1.0000) (T102) 
f4f7 N:     184 (P:  1.62%) (Q:  1.00000) (U: 0.06904) (Q+U:  1.06904) (V:  1.0000) (T105) 
e6f6 N:     541 (P:  1.33%) (Q:  1.00000) (U: 0.01932) (Q+U:  1.01932) (V:  1.0000) (T5) 
e6e7 N:    6863 (P:  1.34%) (Q:  1.00000) (U: 0.00154) (Q+U:  1.00154) (V:  1.0000) (T2) 
e6d6 N:   16134 (P: 38.51%) (Q:  1.00000) (U: 0.01885) (Q+U:  1.01885) (V:  1.0000) (T4) 
```

Alternatively, if picking shorter mates is done only for selecting the best move instead of search, this then wouldn't help improve training data (over what would happen naturally with `--stick-endgames` bumping the eval to +1/0/-1). If this is actually preferred for "cosmetic" play style, then the implementation would probably be quite different focusing on `GetBestChildrenNoTemperature` doing a "one-time" dynamic traversal for shorter mate instead of continuously doing it during search.

I'm sure @Videodr0me has additional thoughts. 😄

Mardak:
r?@mooskagh Fix #837. Favor shorter checkmates then longer mates then shorter tablebase wins then … then longer losses then shorter losses then lastly checkmated. Adds a helper `struct TerminalScore` to sort children. Also reports `score mate` instead of `score cp` for terminal wins and losses.

~Not quite #837 but if there's only one mate found, it's quite likely to be the shortest mate as search didn't need to explore more to find deeper mates. But very much dependent on the network if it prefers some moves over others. E.g.,~

Screen Shot 2019-06-13 at 3 20 19 PM
```
./lc0 -w 42100 --verbose-move-stats --smart-pruning-factor=0 --no-syzygy-fast-play --minibatch-size=1
position fen 8/8/8/3Q4/k7/8/2K5/8 w - - 0 1
```

before:
```
go nodes 100
nodes 100 score cp 4532 hashfull 0 nps 90 tbhits 2 pv d5d8 a4b5 d8d7 b5c4 d7e7 c4d5 e7d8
…
c2c3 N:       0 P:  2.00% Q:  0.28611 
…
d5b7 N:       0 P:  2.34% Q:  0.28611 
…
d5d8 N:      59 P: 20.96% Q:  0.99688 
bestmove d5d8 ponder a4b5


go nodes 1000
nodes 1000 score cp 4780 hashfull 3 nps 242 tbhits 2 pv d5d8 a4b5 d8d7 b5c5 d7e7 c5c4 e7d8
…
c2c3 N:      21 P:  2.00% Q:  1.00000 (T) 
…
d5b7 N:      24 P:  2.34% Q:  0.99717 
…
d5d8 N:     221 P: 20.96% Q:  0.99713 
bestmove d5d8 ponder a4b5


go nodes 10000
nodes 10000 score cp 4692 hashfull 12 nps 737 tbhits 2 pv d5d8 a4b5 d8b8 b5a6 b8c8 a6b5 c8d8 b5a6
…
c2c3 N:     227 P:  2.00% Q:  1.00000 (T) 
…
d5b7 N:     266 P:  2.34% Q:  1.00000 (T) 
…
d5d8 N:    2191 P: 20.96% Q:  0.99705 
bestmove d5d8 ponder a4b5
```
after:
```

go nodes 100
nodes 100 score cp 4542 hashfull 0 nps 69 tbhits 2 pv d5d8 a4b5 d8d7 b5c4 d7e7 c4d5 e7d8
…
c2c3 N:       0 P:  2.00% Q:  0.28608 
…
d5b7 N:       0 P:  2.34% Q:  0.28608 
…
d5d8 N:      59 P: 20.96% Q:  0.99689 
bestmove d5d8 ponder a4b5


go nodes 1000
nodes 1000 score mate 2 hashfull 3 nps 204 tbhits 2 pv c2c3 a4a3 d5a8
…
c2c3 N:      21 P:  2.00% Q:  1.00000 (T) 
…
d5b7 N:      24 P:  2.34% Q:  0.99716 
…
d5d8 N:     221 P: 20.96% Q:  0.99712 
bestmove c2c3 ponder a4a3


go nodes 10000
nodes 10000 score mate 2 hashfull 12 nps 663 tbhits 2 pv c2c3 a4a3 d5a8
…
c2c3 N:     227 P:  2.00% Q:  1.00000 (T) 
…
d5b7 N:     266 P:  2.34% Q:  1.00000 (T) 
…
d5d8 N:    2191 P: 20.96% Q:  0.99706 
bestmove c2c3 ponder a4a3
```

### DIFF
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -234,7 +234,9 @@ void Node::MakeNotTerminal() {
   is_terminal_ = false;
   n_ = 0;
 
-  // If we have edges, we've been extended (1 visit), so include children too.
+  // If a child exists on a terminal, this node was previously extended (1
+  // visit), so include children for recalculation too. Otherwise, treat this
+  // node as unvisited to populate from NN computations (including policy data).
   if (child_) {
     n_++;
     for (const auto& child : Edges()) {
@@ -388,6 +390,57 @@ V4TrainingData Node::GetV4TrainingData(GameResult game_result,
 /////////////////////////////////////////////////////////////////////////
 // EdgeAndNode
 /////////////////////////////////////////////////////////////////////////
+
+namespace {
+// Placeholder plies to differentiate actual game end and tablebase.
+const int kTablebaseTerminalPlies = 999;
+
+// Recursively count terminal edges to reach true terminal.
+int CountPliesToTrueTerminal(Node* terminal) {
+  auto ret = 0;
+
+  // Follow "terminals" until a true terminal with no children.
+  if (terminal->HasChildren()) {
+    const auto follow_q = -terminal->GetQ();
+    std::vector<int> plies;
+    for (auto edge : terminal->Edges()) {
+      if (edge.IsTerminal() && edge.GetQ(0.0f) == follow_q) {
+        plies.push_back(CountPliesToTrueTerminal(edge.node()));
+      }
+    }
+
+    // No terminal children, so we have a tablebase position.
+    if (plies.empty()) {
+      ret = kTablebaseTerminalPlies;
+    } else if (follow_q == 1.0f) {
+      // Minimize plies for winning moves.
+      ret = *std::min_element(std::begin(plies), std::end(plies));
+    } else if (follow_q == -1.0f) {
+      // Maximize plies for losing moves.
+      ret = *std::max_element(std::begin(plies), std::end(plies));
+    }
+  }
+
+  // Include the current move in the ply count.
+  return ret + 1;
+}
+}  // namespace
+
+int EdgeAndNode::GetMovesTillCheckmate() const {
+  auto Q = GetQ(0.0f);
+  if (!IsTerminal() || !Q) return 0;
+  return std::copysign((CountPliesToTrueTerminal(node_) + 1) / 2, Q);
+}
+
+int EdgeAndNode::GetTerminalSortingKey() const {
+  auto moves = GetMovesTillCheckmate();
+  if (moves) {
+    // Give shortest winning moves the highest score then tablebase wins while
+    // also avoiding shortest losing moves.
+    moves = std::copysign(kTablebaseTerminalPlies * 2, moves) - moves;
+  }
+  return moves;
+}
 
 std::string EdgeAndNode::DebugString() const {
   if (!edge_) return "(no edge)";
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -235,7 +235,7 @@ void Node::MakeNotTerminal() {
   n_ = 0;
 
   // If we have edges, we've been extended (1 visit), so include children too.
-  if (edges_) {
+  if (child_) {
     n_++;
     for (const auto& child : Edges()) {
       const auto n = child.GetN();
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -126,7 +126,10 @@ void Search::SendUciInfo() REQUIRES(nodes_mutex_) {
     ++multipv;
     uci_infos.emplace_back(common_info);
     auto& uci_info = uci_infos.back();
-    if (score_type == "centipawn") {
+    const auto mate = scored.GetMate();
+    if (mate) {
+      uci_info.mate = mate;
+    } else if (score_type == "centipawn") {
       uci_info.score = 295 * edge.GetQ(default_q) /
                        (1 - 0.976953126 * std::pow(edge.GetQ(default_q), 14));
     } else if (score_type == "centipawn_2018") {
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -102,7 +102,8 @@ void ApplyDirichletNoise(Node* node, float eps, double alpha) {
 }  // namespace
 
 void Search::SendUciInfo() REQUIRES(nodes_mutex_) {
-  auto edges = GetBestChildrenNoTemperature(root_node_, params_.GetMultiPv());
+  auto edges =
+      GetBestChildrenNoTemperature(root_node_, params_.GetMultiPv(), true);
   const auto score_type = params_.GetScoreType();
 
   std::vector<ThinkingInfo> uci_infos;
@@ -535,7 +536,7 @@ void Search::EnsureBestMoveKnown() REQUIRES(nodes_mutex_)
 
   final_bestmove_ = temperature
                         ? GetBestChildWithTemperature(root_node_, temperature)
-                        : GetBestChildNoTemperature(root_node_);
+                        : GetBestChildNoTemperature(root_node_, true);
 
   if (final_bestmove_.HasNode() && final_bestmove_.node()->HasChildren()) {
     final_pondermove_ = GetBestChildNoTemperature(final_bestmove_.node());
@@ -543,8 +544,8 @@ void Search::EnsureBestMoveKnown() REQUIRES(nodes_mutex_)
 }
 
 // Returns @count children with most visits.
-std::vector<EdgeAndNode> Search::GetBestChildrenNoTemperature(Node* parent,
-                                                              int count) const {
+std::vector<EdgeAndNode> Search::GetBestChildrenNoTemperature(
+    Node* parent, int count, bool calcTerm) const {
   MoveList root_limit;
   if (parent == root_node_) {
     PopulateRootMoveLimit(&root_limit);
@@ -563,8 +564,8 @@ std::vector<EdgeAndNode> Search::GetBestChildrenNoTemperature(Node* parent,
             root_limit.end()) {
       continue;
     }
-    edges.emplace_back(edge.GetTerminalSortingKey(), edge.GetN(), edge.GetQ(0),
-                       edge.GetP(), edge);
+    edges.emplace_back(calcTerm ? edge.GetTerminalSortingKey() : 0, edge.GetN(),
+                       edge.GetQ(0), edge.GetP(), edge);
   }
   const auto middle = (static_cast<int>(edges.size()) > count)
                           ? edges.begin() + count
@@ -578,8 +579,9 @@ std::vector<EdgeAndNode> Search::GetBestChildrenNoTemperature(Node* parent,
 }
 
 // Returns a child with most visits.
-EdgeAndNode Search::GetBestChildNoTemperature(Node* parent) const {
-  auto res = GetBestChildrenNoTemperature(parent, 1);
+EdgeAndNode Search::GetBestChildNoTemperature(Node* parent,
+                                              bool calcTerm) const {
+  auto res = GetBestChildrenNoTemperature(parent, 1, calcTerm);
   return res.empty() ? EdgeAndNode() : res.front();
 }
 
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -121,12 +121,11 @@ void Search::SendUciInfo() REQUIRES(nodes_mutex_) {
 
   int multipv = 0;
   const auto default_q = -root_node_->GetQ();
-  for (const auto& scored : edges) {
-    const auto& edge = scored.edge;
+  for (const auto& edge : edges) {
     ++multipv;
     uci_infos.emplace_back(common_info);
     auto& uci_info = uci_infos.back();
-    const auto mate = scored.GetMate();
+    const auto mate = edge.GetMovesTillCheckmate();
     if (mate) {
       uci_info.mate = mate;
     } else if (score_type == "centipawn") {
@@ -544,8 +543,8 @@ void Search::EnsureBestMoveKnown() REQUIRES(nodes_mutex_)
 }
 
 // Returns @count children with most visits.
-std::vector<Search::TerminalScore> Search::GetBestChildrenNoTemperature(
-    Node* parent, int count) const {
+std::vector<EdgeAndNode> Search::GetBestChildrenNoTemperature(Node* parent,
+                                                              int count) const {
   MoveList root_limit;
   if (parent == root_node_) {
     PopulateRootMoveLimit(&root_limit);
@@ -556,7 +555,7 @@ std::vector<Search::TerminalScore> Search::GetBestChildrenNoTemperature(
   // * If two nodes have equal number:
   //   * If that number is 0, the one with larger prior wins.
   //   * If that number is larger than 0, the one with larger eval wins.
-  using El = std::tuple<TerminalScore, uint64_t, float, float>;
+  using El = std::tuple<int, uint64_t, float, float, EdgeAndNode>;
   std::vector<El> edges;
   for (auto edge : parent->Edges()) {
     if (parent == root_node_ && !root_limit.empty() &&
@@ -564,24 +563,24 @@ std::vector<Search::TerminalScore> Search::GetBestChildrenNoTemperature(
             root_limit.end()) {
       continue;
     }
-    edges.emplace_back(TerminalScore(edge), edge.GetN(), edge.GetQ(0),
-                       edge.GetP());
+    edges.emplace_back(edge.GetTerminalSortingKey(), edge.GetN(), edge.GetQ(0),
+                       edge.GetP(), edge);
   }
   const auto middle = (static_cast<int>(edges.size()) > count)
                           ? edges.begin() + count
                           : edges.end();
   std::partial_sort(edges.begin(), middle, edges.end(), std::greater<El>());
 
-  std::vector<TerminalScore> res;
+  std::vector<EdgeAndNode> res;
   std::transform(edges.begin(), middle, std::back_inserter(res),
-                 [](const El& x) { return std::get<0>(x); });
+                 [](const El& x) { return std::get<4>(x); });
   return res;
 }
 
 // Returns a child with most visits.
 EdgeAndNode Search::GetBestChildNoTemperature(Node* parent) const {
   auto res = GetBestChildrenNoTemperature(parent, 1);
-  return res.empty() ? EdgeAndNode() : res.front().edge;
+  return res.empty() ? EdgeAndNode() : res.front();
 }
 
 // Returns a child chosen according to weighted-by-temperature visit count.
@@ -740,57 +739,6 @@ Search::~Search() {
   Abort();
   Wait();
   LOGFILE << "Search destroyed.";
-}
-
-//////////////////////////////////////////////////////////////////////////////
-// TerminalScore
-//////////////////////////////////////////////////////////////////////////////
-
-Search::TerminalScore::TerminalScore(EdgeAndNode edge) : edge(edge) {
-  if (edge.IsTerminal()) {
-    score_ = edge.GetQ(0.0f);
-    if (score_) {
-      plies_ = CalcPlies(edge.node());
-    }
-  }
-}
-
-bool Search::TerminalScore::operator<(const TerminalScore& other) const {
-  // Prefer the better outcome.
-  if (score_ != other.score_) {
-    return score_ < other.score_;
-  }
-  // Prefer shorter wins and longer losses.
-  return score_ * (other.plies_ - plies_) < 0;
-}
-
-int Search::TerminalScore::CalcPlies(Node* terminal) {
-  auto ret = 0;
-
-  // Follow "terminals" until a true terminal with no children.
-  if (terminal->HasChildren()) {
-    const auto follow_q = -terminal->GetQ();
-    std::vector<int> plies;
-    for (auto edge : terminal->Edges()) {
-      if (edge.IsTerminal() && edge.GetQ(0.0f) == follow_q) {
-        plies.push_back(CalcPlies(edge.node()));
-      }
-    }
-
-    // No terminal children, so we have a tablebase position.
-    if (plies.empty()) {
-      ret = 999;
-    } else if (follow_q == 1.0f) {
-      // Minimize plies for winning moves.
-      ret = *std::min_element(std::begin(plies), std::end(plies));
-    } else if (follow_q == -1.0f) {
-      // Maximize plies for losing moves.
-      ret = *std::max_element(std::begin(plies), std::end(plies));
-    }
-  }
-
-  // Include the current move in the ply count.
-  return ret + 1;
 }
 
 //////////////////////////////////////////////////////////////////////////////
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -121,7 +121,8 @@ void Search::SendUciInfo() REQUIRES(nodes_mutex_) {
 
   int multipv = 0;
   const auto default_q = -root_node_->GetQ();
-  for (const auto& edge : edges) {
+  for (const auto& scored : edges) {
+    const auto& edge = scored.edge;
     ++multipv;
     uci_infos.emplace_back(common_info);
     auto& uci_info = uci_infos.back();
@@ -540,18 +541,19 @@ void Search::EnsureBestMoveKnown() REQUIRES(nodes_mutex_)
 }
 
 // Returns @count children with most visits.
-std::vector<EdgeAndNode> Search::GetBestChildrenNoTemperature(Node* parent,
-                                                              int count) const {
+std::vector<Search::TerminalScore> Search::GetBestChildrenNoTemperature(
+    Node* parent, int count) const {
   MoveList root_limit;
   if (parent == root_node_) {
     PopulateRootMoveLimit(&root_limit);
   }
   // Best child is selected using the following criteria:
+  // * Best terminal score (favor wins, avoid losses).
   // * Largest number of playouts.
   // * If two nodes have equal number:
   //   * If that number is 0, the one with larger prior wins.
   //   * If that number is larger than 0, the one with larger eval wins.
-  using El = std::tuple<uint64_t, float, float, EdgeAndNode>;
+  using El = std::tuple<TerminalScore, uint64_t, float, float>;
   std::vector<El> edges;
   for (auto edge : parent->Edges()) {
     if (parent == root_node_ && !root_limit.empty() &&
@@ -559,23 +561,24 @@ std::vector<EdgeAndNode> Search::GetBestChildrenNoTemperature(Node* parent,
             root_limit.end()) {
       continue;
     }
-    edges.emplace_back(edge.GetN(), edge.GetQ(0), edge.GetP(), edge);
+    edges.emplace_back(TerminalScore(edge), edge.GetN(), edge.GetQ(0),
+                       edge.GetP());
   }
   const auto middle = (static_cast<int>(edges.size()) > count)
                           ? edges.begin() + count
                           : edges.end();
   std::partial_sort(edges.begin(), middle, edges.end(), std::greater<El>());
 
-  std::vector<EdgeAndNode> res;
+  std::vector<TerminalScore> res;
   std::transform(edges.begin(), middle, std::back_inserter(res),
-                 [](const El& x) { return std::get<3>(x); });
+                 [](const El& x) { return std::get<0>(x); });
   return res;
 }
 
 // Returns a child with most visits.
 EdgeAndNode Search::GetBestChildNoTemperature(Node* parent) const {
   auto res = GetBestChildrenNoTemperature(parent, 1);
-  return res.empty() ? EdgeAndNode() : res.front();
+  return res.empty() ? EdgeAndNode() : res.front().edge;
 }
 
 // Returns a child chosen according to weighted-by-temperature visit count.
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -914,7 +914,7 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend(
       return NodeToProcess::Collision(node, depth, collision_limit);
     }
     // Either terminal or unexamined leaf node -- the end of this playout.
-    if (node->IsTerminal() || !node->HasChildren()) {
+    if (node->IsTerminal() || !node->GetN()) {
       return NodeToProcess::Visit(node, depth);
     }
     Node* possible_shortcut_child = node->GetCachedBestChild();
@@ -1067,7 +1067,6 @@ void SearchWorker::ExtendNode(Node* node) {
           node->MakeTerminal(GameResult::DRAW);
         }
         search_->tb_hits_.fetch_add(1, std::memory_order_acq_rel);
-        return;
       }
     }
   }
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -743,6 +743,57 @@ Search::~Search() {
 }
 
 //////////////////////////////////////////////////////////////////////////////
+// TerminalScore
+//////////////////////////////////////////////////////////////////////////////
+
+Search::TerminalScore::TerminalScore(EdgeAndNode edge) : edge(edge) {
+  if (edge.IsTerminal()) {
+    score_ = edge.GetQ(0.0f);
+    if (score_) {
+      plies_ = CalcPlies(edge.node());
+    }
+  }
+}
+
+bool Search::TerminalScore::operator<(const TerminalScore& other) const {
+  // Prefer the better outcome.
+  if (score_ != other.score_) {
+    return score_ < other.score_;
+  }
+  // Prefer shorter wins and longer losses.
+  return score_ * (other.plies_ - plies_) < 0;
+}
+
+int Search::TerminalScore::CalcPlies(Node* terminal) {
+  auto ret = 0;
+
+  // Follow "terminals" until a true terminal with no children.
+  if (terminal->HasChildren()) {
+    const auto follow_q = -terminal->GetQ();
+    std::vector<int> plies;
+    for (auto edge : terminal->Edges()) {
+      if (edge.IsTerminal() && edge.GetQ(0.0f) == follow_q) {
+        plies.push_back(CalcPlies(edge.node()));
+      }
+    }
+
+    // No terminal children, so we have a tablebase position.
+    if (plies.empty()) {
+      ret = 999;
+    } else if (follow_q == 1.0f) {
+      // Minimize plies for winning moves.
+      ret = *std::min_element(std::begin(plies), std::end(plies));
+    } else if (follow_q == -1.0f) {
+      // Maximize plies for losing moves.
+      ret = *std::max_element(std::begin(plies), std::end(plies));
+    }
+  }
+
+  // Include the current move in the ply count.
+  return ret + 1;
+}
+
+//////////////////////////////////////////////////////////////////////////////
 // SearchWorker
 //////////////////////////////////////////////////////////////////////////////
 
--- a/src/mcts/search.h
+++ b/src/mcts/search.h
@@ -115,6 +115,7 @@ class Search {
     bool operator<(const TerminalScore& other) const {
       return score_ < other.score_;
     }
+    int GetMate() const { return std::copysign((plies_ + 1) / 2, score_); }
 
    private:
     float score_ = 0.0f;
--- a/src/mcts/search.h
+++ b/src/mcts/search.h
@@ -100,13 +100,12 @@ class Search {
   // Computes the best move, maybe with temperature (according to the settings).
   void EnsureBestMoveKnown();
 
-  class TerminalScore;
   // Returns a child with most visits, with or without temperature.
   // NoTemperature is safe to use on non-extended nodes, while WithTemperature
   // accepts only nodes with at least 1 visited child.
   EdgeAndNode GetBestChildNoTemperature(Node* parent) const;
-  std::vector<TerminalScore> GetBestChildrenNoTemperature(Node* parent,
-                                                          int count) const;
+  std::vector<EdgeAndNode> GetBestChildrenNoTemperature(Node* parent,
+                                                        int count) const;
   EdgeAndNode GetBestChildWithTemperature(Node* parent,
                                           float temperature) const;
 
@@ -195,22 +194,6 @@ class Search {
   const SearchParams params_;
 
   friend class SearchWorker;
-};
-
-// Helper to score and sort terminals for preferred moves.
-class Search::TerminalScore {
- public:
-  EdgeAndNode edge;
-
-  TerminalScore(EdgeAndNode edge);
-  bool operator<(const TerminalScore& other) const;
-  int GetMate() const { return std::copysign((plies_ + 1) / 2, score_); }
-
- private:
-  float score_ = 0.0f;
-  int plies_ = 0;
-
-  static int CalcPlies(Node* terminal);
 };
 
 // Single thread worker of the search engine.
--- a/src/mcts/search.h
+++ b/src/mcts/search.h
@@ -97,64 +97,10 @@ class Search {
   const SearchParams& GetParams() const { return params_; }
 
  private:
-  // Helper to score and sort terminals for preferred moves.
-  struct TerminalScore {
-    EdgeAndNode edge;
-
-    TerminalScore(EdgeAndNode edge) : edge(edge) {
-      if (edge.IsTerminal()) {
-        score_ = edge.GetQ(0.0f);
-
-        // Prefer shorter wins and longer losses.
-        if (score_ != 0.0f) {
-          plies_ = CalcPlies(edge.node());
-          score_ *= 1 - 0.5f * plies_ / TB_PLIES;
-        }
-      }
-    }
-    bool operator<(const TerminalScore& other) const {
-      return score_ < other.score_;
-    }
-    int GetMate() const { return std::copysign((plies_ + 1) / 2, score_); }
-
-   private:
-    float score_ = 0.0f;
-    int plies_ = 0;
-
-    static const int TB_PLIES = 1000;
-    static int CalcPlies(Node* terminal) {
-      auto ret = 0;
-
-      // Follow "terminals" until a true terminal with no children.
-      if (terminal->HasChildren()) {
-        const auto follow_q = -terminal->GetQ();
-        std::vector<int> plies;
-        for (auto edge : terminal->Edges()) {
-          if (edge.IsTerminal() && edge.GetQ(0.0f) == follow_q) {
-            plies.push_back(CalcPlies(edge.node()));
-          }
-        }
-
-        // No terminal children, so we have a tablebase position.
-        if (plies.empty()) {
-          ret = TB_PLIES;
-        } else if (follow_q == 1.0f) {
-          // Minimize plies for winning moves.
-          ret = *std::min_element(std::begin(plies), std::end(plies));
-        } else if (follow_q == -1.0f) {
-          // Maximize plies for losing moves.
-          ret = *std::max_element(std::begin(plies), std::end(plies));
-        }
-      }
-
-      // Include the current move in the ply count.
-      return ret + 1;
-    }
-  };
-
   // Computes the best move, maybe with temperature (according to the settings).
   void EnsureBestMoveKnown();
 
+  class TerminalScore;
   // Returns a child with most visits, with or without temperature.
   // NoTemperature is safe to use on non-extended nodes, while WithTemperature
   // accepts only nodes with at least 1 visited child.
@@ -249,6 +195,22 @@ class Search {
   const SearchParams params_;
 
   friend class SearchWorker;
+};
+
+// Helper to score and sort terminals for preferred moves.
+class Search::TerminalScore {
+ public:
+  EdgeAndNode edge;
+
+  TerminalScore(EdgeAndNode edge);
+  bool operator<(const TerminalScore& other) const;
+  int GetMate() const { return std::copysign((plies_ + 1) / 2, score_); }
+
+ private:
+  float score_ = 0.0f;
+  int plies_ = 0;
+
+  static int CalcPlies(Node* terminal);
 };
 
 // Single thread worker of the search engine.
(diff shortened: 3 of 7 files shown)


Videodr0me:
I would extend your scheme slightly to prefer terminal winning children without children over those with children. This would eliminate the extreme cases where leela would not play a mate in one postponing it just before the 50 move rule runs out. 

You would just need to change the bool to an int and just change the formular slightly to give a higher value to those terminal wins without children.

Also you could think about doing the same thing to avoid proven loosing terminals.

Mardak:
Refactored into a `struct EdgeScore` that knows how how to compare so we don't need to have a dummy edge comparison. Allows the struct to adjust score down the line, e.g., for shorter mate distances.

Mardak:
With the latest update, it'll avoid known losses, e.g., example from https://github.com/LeelaChessZero/lc0/pull/822#issuecomment-478164951
Screen Shot 2019-03-29 at 10 04 53 AM
```
position startpos moves g1f3 e7e6 c2c4 g8f6 d2d4 f8b4 c1d2 b4e7 a2a3 d7d5 e2e3 e8g8 b1c3 b8d7 d1c2 b7b6 c4d5 e6d5 f1e2 c8b7 f3e5 a7a6 e1g1 c7c5 d2e1 a8c8 c2f5 d8c7 f2f4 c8d8 a1c1 f8e8 e1h4 b6b5 e2d3 d7f8 f5h3 f6e4 c3e4 d5e4 d3b1 c5c4 f4f5 e7h4 h3h4 f7f6 e5g4 f8d7 g4f2 d7b6 f2e4 b7e4 b1e4 e8e7 c1e1 d8e8 f1f4 g8h8 e4f3 c4c3 b2c3 c7c3 h2h3 e7e3 e1e3 c3e3 g1h2 b6c4 f3d5

info … nodes 820 score cp 82 hashfull 5 nps 108 tbhits 0 pv e3a3 h4h5 e8d8 f4h4 h7h6 d5e6 a3d6 h2h1 d6g3 h4g4
info … nodes 821 score cp 7 hashfull 5 nps 113 tbhits 0 pv c4d2 d5e6 e8d8 h4f2 e3f2
…
c4d2 N:      46 P: 13.63% Q:  0.02581 
e3a3 N:     634 P: 14.29% Q: -1.00000 (T) 
bestmove c4d2 ponder d5e6
```

src/mcts/search.cc
```diff
@@ -540,42 +541,43 @@ void Search::EnsureBestMoveKnown() REQUIRES(nodes_mutex_)
...
+  using El = std::tuple<EdgeScore, uint64_t, float, float>;
   std::vector<El> edges;
   for (auto edge : parent->Edges()) {
     if (parent == root_node_ && !root_limit.empty() &&
         std::find(root_limit.begin(), root_limit.end(), edge.GetMove()) ==
             root_limit.end()) {
       continue;
     }
-    edges.emplace_back(edge.GetN(), edge.GetQ(0), edge.GetP(), edge);
+    edges.emplace_back(EdgeScore(edge), edge.GetN(), edge.GetQ(0), edge.GetP());
```
Mardak: I suppose arguably, with the `EdgeScore::operator<` there doesn't need to be a tuple anymore…

Mardak: Although unless we cache the results of GetN/Q/P in EdgeScore, pulling them out once in the tuple might be slightly faster ??

Videodr0me:
> 
> 
> Refactored into a `struct EdgeScore` that knows how how to compare so we don't need to have a dummy edge comparison. Allows the struct to adjust score down the line, e.g., for shorter mate distances.

I can see that you do not want to convolute populating the tuple with a large expression containing conditionals for the terminal cases. But, I am not sure if the new struct  EdgeScore makes sense. It does not contain the full edge score - which would be the full tuple (or an aggregate that computed from the tuple that reflects the ordering) and not just the first entry (the terminal winning/loosing conditions). Also the terminal properties are not even an attribute of an Edge but of a Node (so maybe its more an EdgeNodeTerminalScore or something). If EdgesScore would replace the full tuple, i could see the merits. One could then use the ">" or "<" operator for comparisons and sorting and how the score is computed would be shielded from getbestmove. 

But thats just cosmetics - i have no strong opinion on this. Its good that its now avoiding loosing moves as well. Now you just need to factor in preferring winning terminals without children over those with children - maybe just by multiplying the score by 2 if there are no children, this would also avoid losing by a mate in one in favour of longer mates.

MelleKoning:
@Videodr0me @Mardak We can see where this is going :) 
In this PR there is not yet priority for a mate win over a TB win as both get scored the same value.

However when we think back to the additions of CertaintyState / CertaintyResult that were added to the Edge as in #700 we can see EdgeScore be extended to ensure a mate-win gets priority over a TB win again. 
Having priority for mates over TB wins might shorten the game when both are in sight in a position.
https://github.com/LeelaChessZero/lc0/pull/700/files#diff-2b6cf6f06881eb37c1ebee83cfd70353R87

Videodr0me:
> 
> 
> @Videodr0me @Mardak We can see where this is going :)
> In this PR there is not yet priority for a mate win over a TB win as both get scored the same value.
> 
> However when we think back to the additions of CertaintyState / CertaintyResult that were added to the Edge as in #700 we can see EdgeScore be extended to ensure a mate-win gets priority over a TB win again.
> Having priority for mates over TB wins might shorten the game when both are in sight in a position.
> https://github.com/LeelaChessZero/lc0/pull/700/files#diff-2b6cf6f06881eb37c1ebee83cfd70353R87

Yes, in PR700 I also prefered normal mates over proven wins that include a TB position in the proven subtree. This would require to have an additonal status bit "based_on_propagatedt_bhit". I believe that the terminal bool already takes one byte of memory (according to standards AFAIK) and thus could easily hold 8 status bits instead of just one without increasing the memory footprint. But its a very minor issue, leela should win regardless - only in sudden death situations it might prevent a rare loss on time. So maybe we shouldn't repeat my mistake to make this too "perfect" with lots of code changes, and just take it step-by-step.

Mardak:
Here's some quick speed tests with random backend:
```
./lc0 --backend=random --backend-opts=uniform=true --verbose-move-stats --smart-pruning-factor=0
go nodes 1000000

master (b73cb5f) median 9972 ms

info depth 7 seldepth 13 time 9998 nodes 1000400 score cp 0 hashfull 1000 nps 100060 tbhits 0 pv g1f3 a7a5 h2h4 b7b5 b1a3 a8a7 a1b1 b8a6 b1a1 b5b4
info depth 7 seldepth 13 time 9972 nodes 1000035 score cp 0 hashfull 1000 nps 100284 tbhits 0 pv b1c3 c7c6 h2h3 a7a5 a1b1 a8a7 b1a1 b8a6 a1b1 a5a4
info depth 7 seldepth 13 time 10022 nodes 1000001 score cp 0 hashfull 1000 nps 99780 tbhits 0 pv g1f3 c7c6 h2h3 a7a5 b1a3 a8a7 a1b1 b8a6 b1a1 a5a4
info depth 7 seldepth 13 time 9949 nodes 1000045 score cp 0 hashfull 1000 nps 100517 tbhits 0 pv g1f3 b7b5 h2h3 a7a5 b1a3 a8a7 a1b1 b8a6 b1a1 b5b4
info depth 7 seldepth 13 time 9897 nodes 1000259 score cp 0 hashfull 1000 nps 101066 tbhits 0 pv g1f3 b8c6 h2h4 g8h6 b1a3 a8b8 a1b1 b8a8 b1a1 h6g4


with PR median 9881 ms

info depth 7 seldepth 13 time 9850 nodes 1000255 score cp 0 hashfull 1000 nps 101548 tbhits 0 pv a2a3 b8a6 a1a2 a8b8 b1c3 b8a8 g1f3 a8b8 h1g1 b8a8
info depth 7 seldepth 13 time 9881 nodes 1000166 score cp 0 hashfull 1000 nps 101221 tbhits 0 pv b1c3 b8a6 a1b1 a8b8 b1a1 b8a8 a1b1 a8b8 b1a1 b8a8
info depth 7 seldepth 13 time 9839 nodes 1000120 score cp 0 hashfull 1000 nps 101648 tbhits 0 pv b1c3 b8a6 a1b1 a8b8 b1a1 b8a8 a1b1 a8b8 b1a1 b8a8
info depth 7 seldepth 13 time 9965 nodes 1000480 score cp 0 hashfull 1000 nps 100399 tbhits 0 pv c2c4 b8a6 b1a3 a8b8 a1b1 b8a8 b1a1 a8b8 a1b1 b8a8
info depth 7 seldepth 13 time 10037 nodes 1000429 score cp 0 hashfull 1000 nps 99674 tbhits 0 pv b1c3 b8a6 a1b1 a8b8 b1a1 b8a8 a1b1 a8b8 b1a1 b8a8
```

Mardak:
Here's a position from CCC 9 Quarterfinals Game 588 where Stoofvlees expected Rh7# and on the previous move, Lc0 v0.21.2-42482 announced +128.00 because there was some followup move that led to mate, but when searching for move 59, the network priors heavily favored a non-terminal and dropped its score to +52.93 to eventually mate in 22 moves.
https://www.chess.com/computer-chess-championship#event=ccc-9-the-gauntlet-quarterfinals&game=588

Screen Shot 2019-06-30 at 10 20 03 AM

```
position startpos moves d2d4 d7d5 g1f3 g8f6 e2e3 e7e6 f1d3 b7b6 b2b3 f8d6 c2c4 c8b7 c4d5 e6d5 c1a3 c7c5 b1c3 b8d7 a1c1 e8g8 e1g1 f8e8 a3b2 f6e4 c3e2 d7f6 e2g3 a7a5 d3b5 e8e6 g3f5 d6f8 f3e5 e4d6 b5d3 a5a4 f5d6 f8d6 d3f5 e6e8 b3a4 b7a6 f1e1 c5c4 b2c3 a6c8 f5c2 f6e4 c2e4 d5e4 e5c4 d6h2 g1h2 d8h4 h2g1 e8e6 c4e5 e6h6 f2f3 h6h5 d1b3 c8e6 d4d5 e6c8 d5d6 c8e6 b3b6 h7h6 b6c6 h4h1 g1f2 h1h4 f2e2 a8a4 c1c2 e4f3 g2f3 e6a2 d6d7 a2b3 c2b2 h5e5 b2b3 a4a2 e2d3 g8h7 b3b8 a2a6 d7d8Q a6c6 d8h4 f7f6 c3e5 c6e6 h4h5 e6c6 e5f4 c6c5 h5c5 g7g5 d3c4 h7g6 c4b3 h6h5 b3a2 h5h4 a2a3 h4h3 a3a2 h3h2 a2b1 h2h1Q e1h1 g6f7 c5d5 f7e7
go nodes 50000

…
d5d6 N:     529 (P:  0.90%) (Q:  1.00000) (T)  #3
…
h1h7 N:     617 (P:  1.05%) (Q:  1.00000) (T)  #1
…
b8d8 N:     703 (P:  1.20%) (Q:  1.00000) (T)  #2
b8b7 N:     711 (P:  1.21%) (Q:  1.00000) (T)  #2
…
b8c8 N:     725 (P:  1.24%) (Q:  1.00000) (T)  #2
…
b8e8 N:     829 (P:  1.41%) (Q:  1.00000) (T)  #3
f4d6 N:    1442 (P:  2.46%) (Q:  1.00000) (T)  #2
d5c4 N:   15051 (P: 29.15%) (Q:  0.99715)      #4
bestmove f4d6 ponder e7d7
```

So this PR would have at least mated in 2 moves instead of 22 moves but wouldn't play the mate in 1 -- to be fixed later.

MelleKoning:
To fix build you probably need same change as in #893

Mardak:
Yup with #893 merged, the appveyor build would be green: https://ci.appveyor.com/project/LeelaChessZero/lc0/builds/25652739

Mardak:
Just to follow up the previous position https://github.com/LeelaChessZero/lc0/pull/889#issuecomment-507054097 but with the latest from this PR:

The --verbose-move-stats output is pretty similar to what was previously shared except here's the --multipv=10 output:
```
score mate 1 hashfull 90 nps 898 tbhits 0 multipv 1 pv h1h7
score mate 2 hashfull 90 nps 898 tbhits 0 multipv 2 pv b8b7 e7e8 h1h8 
score mate 3 hashfull 90 nps 898 tbhits 0 multipv 3 pv f4d6 e7d7 d6c5 d7c7 d5b7
score mate 3 hashfull 90 nps 898 tbhits 0 multipv 4 pv d5d6 e7f7 d6d7 f7g6 b8g8
score mate 4 hashfull 90 nps 898 tbhits 0 multipv 5 pv b8c8 f6f5 c8c7 e7f6 c7f7 f6g6 d5f5 
score mate 4 hashfull 90 nps 898 tbhits 0 multipv 6 pv b8e8 e7e8 h1h8 e8e7 h8h7 e7e8 d5a8 
score mate 4 hashfull 90 nps 898 tbhits 0 multipv 7 pv b8d8 f6f5 d8d7 e7f6 d7f7 f6g6 d5f5 
score cp 5565 hashfull 90 nps 898 tbhits 0 multipv 8 pv d5c4 f6f5 c4b3 e7f6 b3a4 f6e7 a4b5 e7f6 b5a6 f6g7
score cp 9369 hashfull 90 nps 898 tbhits 0 multipv 9 pv b8b6 e7f8 d5d8 f8g7 d8b8 g7f7 b8a7 f7g6 a7a6 g6f5
score cp 3893 hashfull 90 nps 898 tbhits 0 multipv 10 pv d5b5 e7f7 b5a6 f7g7 a6a7 g7g6 a7a8 g6f7 b8c8 f7e6
```

Notably the mate in 1 is multipv 1 even though it only has ~600 visits out of 50k. Also, the reported mates can be higher because --sticky-endgames stops searching other possibly shorter winning descendant moves, so the network's priors do affect how accurate the reported mates are.

Mardak:
Here's a position from CCC 9 Quarterfinals where --syzygy-fast-play would have sacced the queen with 117. Qe2+ forcing 117… Kxe2 resulting in a tablebase win for white and a Bishop+Knight endgame.

Screen Shot 2019-07-02 at 1 13 13 AM
https://lichess.org/analysis/standard/3N4/8/2B5/5K2/8/8/Q7/4k3_w_-_-_0_117

Here's the top 5 moves after 10k visits:
```
score mate 3 hashfull 18 nps 637 tbhits 5 multipv 1 pv f5e4 e1f1 e4e3 f1g1 a2g2
score mate 3 hashfull 18 nps 637 tbhits 5 multipv 2 pv c6f3 e1f1 a2e2 f1g1 e2g2
score mate 4 hashfull 18 nps 637 tbhits 5 multipv 3 pv f5f4 e1d1 f4e3 d1c1 e3d3 c1d1 a2b1
score mate 5 hashfull 18 nps 637 tbhits 5 multipv 4 pv c6a4 e1f1 a4b5 f1g1 b5a6 g1h1 a6b7 h1g1 a2g2
score mate 501 hashfull 18 nps 637 tbhits 5 multipv 5 pv a2e2 e1e2 f5e4
```

Again, network priors and tablebase positions result in an upper bound for the reported mate, but at least at root, search has an opportunity to find a shorter mate.

Here's after playing white Ke4:
```
score mate -2 hashfull 18 nps 192500 tbhits 2 multipv 1 pv e1f1 e4e3 f1g1 a2g2
score mate -2 hashfull 18 nps 192500 tbhits 2 multipv 2 pv e1d1 e4d3 d1c1 a2a1
```

Then after black Kf1:
```
score mate 2 hashfull 26 nps 1383 tbhits 7 multipv 1 pv e4e3 f1g1 a2g2
score mate 2 hashfull 26 nps 1383 tbhits 7 multipv 2 pv e4f3 f1e1 a2e2
score mate 2 hashfull 26 nps 1383 tbhits 7 multipv 3 pv e4d3 f1e1 a2e2
score mate 3 hashfull 26 nps 1383 tbhits 7 multipv 4 pv c6a4 f1g1 e4f3 g1f1 a2b1
score mate 3 hashfull 26 nps 1383 tbhits 7 multipv 5 pv c6b5 f1g1 e4f3 g1h1 a2g2
```

Then after white Ke3:
```
score mate -1 hashfull 27 nps 128000 tbhits 2 multipv 1 pv f1g1 a2g2
score mate -1 hashfull 27 nps 128000 tbhits 2 multipv 2 pv f1e1 a2a1
```

Then after black Kg1:
```
score mate 1 hashfull 29 nps 3283 tbhits 6 multipv 1 pv a2g2
score mate 1 hashfull 29 nps 3283 tbhits 6 multipv 2 pv a2f2
score mate 2 hashfull 29 nps 3283 tbhits 6 multipv 3 pv c6d5 g1f1 a2b1
score mate 2 hashfull 29 nps 3283 tbhits 6 multipv 4 pv c6a8 g1f1 a2b1
score mate 2 hashfull 29 nps 3283 tbhits 6 multipv 5 pv e3f3 g1f1 a2a1
```

Mardak:
Here's some speed tests of master and latest PR:

```
./lc0 --backend=random --backend-opts=uniform=true --verbose-move-stats --smart-pruning-factor=0
go nodes 1000000


master (b73cb5f) median 9861 ms

time 9806 nodes 1000304 score cp 0 hashfull 1000 nps 102009 tbhits 0 pv g1f3 f7f5 h2h3 d7d5 b1a3 b8d7 a1b1 a8b8 b1a1 f5f4
time 9866 nodes 1000360 score cp 0 hashfull 1000 nps 101394 tbhits 0 pv d2d3 h7h5 c2c4 f7f5 b1d2 b8a6 a1b1 a8b8 c4c5
time 9792 nodes 1000249 score cp 0 hashfull 1000 nps 102149 tbhits 0 pv d2d3 f7f6 c2c4 g7g5 b1d2 b8a6 a1b1 a8b8 c4c5
time 9861 nodes 1000012 score cp 0 hashfull 1000 nps 101410 tbhits 0 pv d2d3 b7b6 c2c4 g7g5 b1d2 b8a6 a1b1 a8b8 c4c5
time 10027 nodes 1000234 score cp 0 hashfull 1000 nps 99754 tbhits 0 pv d2d3 f7f6 c2c4 g7g5 b1d2 b8a6 a1b1 a8b8 c4c5


with PR median 9904 ms

time 9760 nodes 1000199 score cp 0 hashfull 1000 nps 102479 tbhits 0 pv b1c3 b8a6 a1b1 a8b8 b1a1 b8a8 a1b1 a8b8 b1a1 b8a8
time 9923 nodes 1000425 score cp 0 hashfull 1000 nps 100818 tbhits 0 pv b1c3 b8a6 a1b1 a8b8 b1a1 b8a8 a1b1 a8b8 b1a1 b8a8
time 9962 nodes 1000284 score cp 0 hashfull 1000 nps 100409 tbhits 0 pv b1c3 b8a6 a1b1 a8b8 b1a1 b8a8 a1b1 a8b8 b1a1 b8a8
time 9851 nodes 1000219 score cp 0 hashfull 1000 nps 101534 tbhits 0 pv g1f3 b8a6 b1a3 a8b8 a1b1 b8a8 b1a1 a8b8 a1b1 b8a8
time 9904 nodes 1000017 score cp 0 hashfull 1000 nps 100971 tbhits 0 pv d2d3 b8a6 b1d2 a8b8 a1b1 b8a8 b1a1 a8b8 a1b1
```

And using the position from the previous comment with the queen sac:
```
./lc0 --backend=random --backend-opts=uniform=true --verbose-move-stats --smart-pruning-factor=0
position fen 3N4/8/2B5/5K2/8/8/Q7/4k3 w - - 0 117
go nodes 1000000


master (b73cb5f) median 3065 ms

time 3065 nodes 1000129 score cp 12800 hashfull 20 nps 326306 tbhits 145 pv a2e2 e1e2
time 3057 nodes 1000264 score cp 12800 hashfull 20 nps 327204 tbhits 145 pv a2e2 e1e2
time 3052 nodes 1000058 score cp 12800 hashfull 20 nps 327673 tbhits 145 pv a2e2 e1e2
time 3168 nodes 1000282 score cp 12800 hashfull 20 nps 315745 tbhits 145 pv a2e2 e1e2
time 3084 nodes 1000283 score cp 12800 hashfull 20 nps 324345 tbhits 145 pv a2e2 e1e2


with PR median 3081 ms

time 3081 nodes 1000072 score mate 501 hashfull 20 nps 324593 tbhits 145 pv a2e2 e1e2 f5e4
time 3137 nodes 1000110 score mate 501 hashfull 20 nps 318810 tbhits 145 pv a2e2 e1e2 f5e4
time 3134 nodes 1000178 score mate 501 hashfull 20 nps 319137 tbhits 145 pv a2e2 e1e2 f5e4
time 3052 nodes 1000398 score mate 501 hashfull 20 nps 327784 tbhits 145 pv a2e2 e1e2 f5e4
time 2977 nodes 1000060 score mate 501 hashfull 20 nps 335928 tbhits 145 pv a2e2 e1e2 f5e4
```

So at least for me with random backend, maybe a 0.5% slowdown?

Mardak:
Also a small thing I added in the second commit: prefer queen promotions if otherwise same terminal score. From an earlier position of that queen sac position just before t40 would have promoted to knight:
  
Screen Shot 2019-07-02 at 1 53 46 AM
https://lichess.org/analysis/standard/Q7/3P4/2B5/5K2/4R3/6k1/8/8_w_-_-_1_114

Here's after 10k visits with multipv 1 preferring a non-tb mate while multipv 2 specially favors queen promotion even though t40 would have a higher prior and more visits for knight promotion.
```
score mate 3 hashfull 12 nps 1010 tbhits 53 multipv 1 pv e4e2 g3h3 f5f4 h3h4 a8h8
score mate 501 hashfull 12 nps 1010 tbhits 53 multipv 2 pv d7d8q g3f3
score mate 501 hashfull 12 nps 1010 tbhits 53 multipv 3 pv d7d8n g3f3
score mate 501 hashfull 12 nps 1010 tbhits 53 multipv 4 pv d7d8b g3f3
score mate 501 hashfull 12 nps 1010 tbhits 53 multipv 5 pv d7d8r g3f3
…
d7d8r N:     160 (P:  1.44%) (Q:  1.00000) (T)  
…
e4e2  N:     162 (P:  1.46%) (Q:  1.00000) (T)  
…
d7d8b N:     171 (P:  1.54%) (Q:  1.00000) (T)  
…
d7d8q N:     178 (P:  1.60%) (Q:  1.00000) (T)  
…
d7d8n N:     403 (P:  3.63%) (Q:  1.00000) (T)  
…
c6d5  N:    1357 (P: 13.78%) (Q:  0.99601)  
```

appveyor.yml
```diff
@@ -38,7 +38,7 @@ install:
 - cmd: IF NOT EXIST c:\cache\protobuf\ git clone -b v3.5.1 --single-branch --depth 1 https://github.com/google/protobuf.git
 - cmd: IF NOT EXIST c:\cache\protobuf\ mkdir protobuf\build_msvc
 - cmd: IF NOT EXIST c:\cache\protobuf\ cd protobuf\build_msvc
-- cmd: IF NOT EXIST c:\cache\protobuf\ cmake -G "Visual Studio 15 2017 Win64" -Dprotobuf_BUILD_SHARED_LIBS=NO -Dprotobuf_MSVC_STATIC_RUNTIME=NO -Dprotobuf_BUILD_TESTS=OFF -DCMAKE_INSTALL_PREFIX=c:/cache/protobuf ../cmake
+- cmd: IF NOT EXIST c:\cache\protobuf\ cmake -G "Visual Studio 15 2017 Win64" -Dprotobuf_BUILD_SHARED_LIBS=NO -Dprotobuf_BUILD_TESTS=OFF -DCMAKE_INSTALL_PREFIX=c:/cache/protobuf ../cmake
```
mooskagh: Changes to to is file (appveyor.yml) and e.g. meson.build seem not to belong to this PR.
It is expected or somehow git branches got mixed up?

Mardak: Nod. They're from #904 but people wanted appveyor builds to test this PR. I'll rebase on latest master when appveyor is working there

borg323: #904 is now merged.

Mardak: Thanks. Rebased and force pushed with just the 3 relevant commits.

Videodr0me:
I see that you added distance to mate calculation by recursively traversing the tree when getting best moves. I believe that this approach is not intrusive to the rest of the codebase, so its a good efficiency/getting-it-commited tradeoff and the benchmarks seems to indicate that the overhead is not much of an issue. 

I am still curious if you had other reasons for doing (uneccessary) tree traversals to obtain distance to mate info. Granted they will only kick when getting best moves and only for terminals, but still it seems somewhat inefficient. Apart from some hypothetical problems like stack size, which with current depths of proven terminals should not really be a concern why did you feel the need for this?

The alternative would be to store distance to mate either by adding it to the node or - if memory is a concern - by overloading q. The info is easily available when doing the backpropagation without any need to do additonal sub-tree traversals. With your approach whenever bestmoves is called you currently redo a calculation/traversal which could be avoided. 

I guess you also wanted to differentiate between TB hits and normal mates, without having to use another flag "propagated tb_hit". But as the terminal flag is using up a full byte (AFAIK) adding 7 additional flags could also be done for free (memory wise).

If you, as i suspect,  did it because of the intrusion factor to the codebase, then i think you found a good balanced solution. Nice work. I did not have time to look if the recursion is bulletproof though, will do later, but it seems we are finally getting mate scores.

Mardak:
The primary concerns were to have no changes to `Node` fields and to have minimal performance impact. Clearly not adding memory concerns in the common case for almost all nodes being non-terminal will have some dynamic runtime trade-offs. I did think about adding some complexity of having search use some memory to cache root children terminal scores as the performance impact is primarily from each visit backing up to the root to determine if the best edge has changed although in the common case, there are no terminal children, so TerminalScore just acts as a wrapper for the Edge.

Videodr0me:
> The primary concerns were to have no changes to `Node` fields and to have minimal performance impact. Clearly not adding memory concerns in the common case for almost all nodes being non-terminal will have some dynamic runtime trade-offs. I did think about adding some complexity of having search use some memory to cache root children terminal scores as the performance impact is primarily from each visit backing up to the root to determine if the best edge has changed although in the common case, there are no terminal children, so `TerminalScore` just acts as a wrapper for the `Edge`.

Fair enough. I think you found a good compromise.

Mardak:
Converted to class as requested and removed the magic score formula to just directly compare plies.

oscardssmith:
Would it make sense as an alternate approach to have 2 different types of nodes: regular and terminal? That way terminal nodes could store these flags and still be smaller as they only need 2 bits for a q field, and don't need a policy or number of visits.

Mardak:
Are you suggesting a conversion constructor or a union of some Node fields? Or maybe a derived class that has *additional* fields, so the extra memory is only associated to terminals?

I suppose to the last point, that is kinda the behavior of the current PR where Terminals have a flag and additional things can be checked (namely edges/children, child_) to determine if it's a true terminal, propagated terminal or TB terminal.

In any case, there would need to be various runtime checks to make sure we're dealing with a terminal or plain node to then use the appropriate state.

Videodr0me:
> 
> 
> Would it make sense as an alternate approach to have 2 different types of nodes: regular and terminal? That way terminal nodes could store these flags and still be smaller as they only need 2 bits for a q field, and don't need a policy or number of visits.

Normal nodes can become terminal during search. If we would have two types of nodes differing in size, these would need to be "converted" when becoming terminal. Then you might have to free memory and reallocate or come up with some scheme to circumvent this. This might not be a large issue performance wise, but might lead to memory fragmentation. Also terminals are kind of rare, so the saved memory would be small. 

But, I think there are other alternative solutions that are memory neutral and avoid traversing the tree to recompute the distance to mate (or TB win). One way, already mentioned here is to overload q, but that is kind of instrusive to the getters and setters of q if you want to shield normal pick node selection from this (if desired). But another one has not yet been discussed: 

- change the terminal bool to an uint8. This takes the same amount of memory.
- non-terminals have terminal_uint8 = 0
- terminals have terminal_uint8 != 0 with the value being the distance to a win, with 255 being the max that is also used for TB wins.

Thats an alternative that would eliminate the tree traversal as the information would be computed when making a node terminal, where it is already available.

That being said, i thing the current solution is also fine, and the recomputation of distance to mate should be negligible. I does everything the other solutions also do: prefer terminal mates over tb wins and prefer shorter over longer mates.

Sopel97:
as far as I can see there is 5 unused bytes in Node that are currently padding

Videodr0me:
> 
> 
> as far as I can see there are currently 5 unused bytes (+6 from EdgeList) in Node that are padding (to a multiple of 8).
> That's plenty space to do it without any 'hacks'

Of course, but the general consensus  has been to keep those in reserve for when really needed. I myself tend to agree though and for example in PR487 added data to the edge structure. But the abovementioned solutions are not really to hacky. Both extending the terminal bool to a variable that represents the minimum (known) plies to a end-of-game terminal and the current approach of recomputing the minimum (or maximum in case of a loss) number of plies when a proven win appears at root are perfectly fine in my book. I hope this gets committed, as i am very curious how it does in training.

jkormu:
Could core devs give some feedback what is blocking this from being merged?
If it is lack of Elo regression tests or some other tests, then I am sure the community would be happy to run some.

src/mcts/node.cc
```diff
@@ -235,7 +235,7 @@ void Node::MakeNotTerminal() {
   n_ = 0;
 
   // If we have edges, we've been extended (1 visit), so include children too.
-  if (edges_) {
+  if (child_) {
```
mooskagh: I don't really understand a change in this line, could you comment?
(also maybe the comment above this line needs change?)

src/mcts/search.h
```diff
@@ -196,6 +197,22 @@ class Search {
   friend class SearchWorker;
 };
 
+// Helper to score and sort terminals for preferred moves.
+class Search::TerminalScore {
+ public:
+  EdgeAndNode edge;
```
mooskagh: A public data member prevents encapsulation and should be private: <https://google.github.io/styleguide/cppguide.html#Access_Control>

src/mcts/search.h
```diff
@@ -196,6 +197,22 @@ class Search {
...
+
+  TerminalScore(EdgeAndNode edge);
+  bool operator<(const TerminalScore& other) const;
+  int GetMate() const { return std::copysign((plies_ + 1) / 2, score_); }
+
+ private:
+  float score_ = 0.0f;
+  int plies_ = 0;
+
+  static int CalcPlies(Node* terminal);
```
mooskagh: There's usually no need for private static function (unless it accesses static private member variables, which in this case it doesn't). A function in anonymous namespace in implementation .cc file is better.

src/mcts/search.h
```diff
@@ -196,6 +197,22 @@ class Search {
...
 };
 
+// Helper to score and sort terminals for preferred moves.
+class Search::TerminalScore {
+ public:
+  EdgeAndNode edge;
+
+  TerminalScore(EdgeAndNode edge);
+  bool operator<(const TerminalScore& other) const;
+  int GetMate() const { return std::copysign((plies_ + 1) / 2, score_); }
```
mooskagh: Add comment what this function does.

src/mcts/search.h
```diff
@@ -196,6 +197,22 @@ class Search {
   friend class SearchWorker;
 };
 
+// Helper to score and sort terminals for preferred moves.
+class Search::TerminalScore {
```
mooskagh: Hmmm there should be better name than TerminalScore..
Long name would be something like `Terminal Edge Ordering Wrapper`.

`TerminalSortingKey`? Meh..

src/mcts/search.h
```diff
@@ -100,12 +100,13 @@ class Search {
...
   void EnsureBestMoveKnown();
 
+  class TerminalScore;
   // Returns a child with most visits, with or without temperature.
   // NoTemperature is safe to use on non-extended nodes, while WithTemperature
   // accepts only nodes with at least 1 visited child.
   EdgeAndNode GetBestChildNoTemperature(Node* parent) const;
-  std::vector<EdgeAndNode> GetBestChildrenNoTemperature(Node* parent,
-                                                        int count) const;
+  std::vector<TerminalScore> GetBestChildrenNoTemperature(Node* parent,
```
mooskagh: Does this mean that `GetBestChildNoTemperature` can only return terminal nodes? That doesn't sound right to me.

src/mcts/search.cc
```diff
@@ -540,42 +544,44 @@ void Search::EnsureBestMoveKnown() REQUIRES(nodes_mutex_)
...
     PopulateRootMoveLimit(&root_limit);
   }
   // Best child is selected using the following criteria:
+  // * Best terminal score (favor wins, avoid losses).
   // * Largest number of playouts.
   // * If two nodes have equal number:
   //   * If that number is 0, the one with larger prior wins.
   //   * If that number is larger than 0, the one with larger eval wins.
-  using El = std::tuple<uint64_t, float, float, EdgeAndNode>;
+  using El = std::tuple<TerminalScore, uint64_t, float, float>;
```
mooskagh: Maybe instead of doing that, add `int GetMovesTillCheckmate()` to `EdgeAndNode`, add that as the first element to the tuple, and leave the rest as it is.
And then when priting info, printing function will have to call that again... Which maybe not that bad as usually there's no multipv or something.

Mardak: A plain `int` ordering of moves-till-checkmate probably isn't sufficient as we want to prefer:
1) shorter winning mates: GMTC() = 1
2) longer winning mates: GMTC() = 12
3) shorter TB win: GMTC() = 1000
4) longer TB win: GMTC() = 1234
5) not terminal win/loss: GMTC() = 0 ?
6) longer TB loss: GMTC() = -1234
7) shorter TB loss: GMTC() = -1000
8) longer losing mates: GMTC() = -12
9) shorter losing mates: GMTC() = -1

We could instead emplace back a transformed GMTC() with some magic number and formula for sorting purposes… ?

Mardak: Oh, I suppose `emplace_back(1.0f / GMTC(), …)` might work…

Sopel97: 0 needs to be handled separately anyway. So maybe to keep it an int something like `if < 0: intmin - gmtc; if > 0: intmax - gmtc`

Mardak: GMTC() could return floats instead including `std::numeric_limits<float>::infinity()` for not terminal win/loss… although that makes other callers, e.g., for reporting mates, slightly more complicated.

Mardak: Ended up adding two EdgeAndNode helpers: `GetMovesTillCheckmate()` and `GetTerminalSortingKey()`

Removed the class

src/mcts/search.h
Mardak: Reverted to just EdgeAndScore.

src/mcts/search.h
Mardak: Removed the class.

src/mcts/search.h
Mardak: Removed the method.

src/mcts/search.h
Mardak: Renamed `CountPliesToTrueTerminal` and moved to anonymous namespace with added comments.

src/mcts/search.h
Mardak: Removed public data member

src/mcts/node.cc
Mardak: Modified the comment to clarify why child_ and the behavior for when there's no child_.

zz4032:
Match vs. current master

Id42850
TB adjudication only
TC: 6.0s/game+0.025s/move
Average node/move: 4063 (lc0) / 4143 (lc0_PR889)
```
   1 lc0          :     0.0   ----    91.8    5000      63.4
   2 lc0_PR889    :    -4.1    5.7     ---    5000      63.4

White ELO advantage = 78.8
Draw rate (equal opponents) = 68.5 %
```

Mardak:
One idea to reduce performance impact is to only compute distance if the root has a winning move and cache the move on the `Search` instance. This is opposed to the current behavior of computing distance for every move at root then sorting. Naively this optimization would only find the one shortest winning move and report its distance and not differentiate losing terminals nor support multipv ranking of shorter to longer mates. E.g.,

multipv 1 = shortest mate; score mate 1
multipv 2 = move visited; score cp 1234
multipv 3 = next most visited ; score mate 4
multipv 4 = next most visited ; score mate 2

So the optimization makes sure the move to play is the shortest mate but the others are sorted normally by visits, etc.

I'm guessing this PR won't get merged unless there's no Elo regression? So the tradeoff is slightly more memory usage on a `Search` tracking shortest winning mate distance and move to avoid the continuous re-computation of those values.

borg323:
@zz4032 do any of the losses stand out? In particular, are there any time losses?

zz4032:
@borg323 I deactivated time forfeits with `timemargin` option in Cutechess assuming there would be no liability to time losses with this PR. If you think that could be an issue, I can run a shorter test with default MoveOverheadMs=200 or lower and check for time forfeits.

Ipmanchess:
Have people check my tests.. i don't get any Time Forfeits anymore!!
Depends on your TC you use..
With 1m+0s i put TimeSteepness=2.5
With 4m+0s TimeSteepness=4
1m+1s i'm using 4 ,some comments are on my spreadsheet..
Running now games with Hace merged Lc0 v21.4+PR889 ..to get Mates and No TF's with my settings! using jhorthos net T40B.4-200
There is maybe a little slowdown but i get better results ,can play SD games for first time without TF's!!
The Mates are not perfect yet..but i see Mardak is working on it..still some trolling games.

@zz4032 if you put here pgn from these 5000games ,i can check in 1sec. if their where TF's.

Mardak:
Looks like first pass of optimization doesn't need to add too much complexity. The most recent commit "uci-only" (on top of the previous "all `GetBest`") only computes the sorting key for UCI output: info and bestmove.

Here's some backend=random uniform **time** stats from startpos of 10 runs of 1 million nodes:

master | all `GetBest` | uci-only
-- | -- | --
8412 | 8475 | 8455
8448 | 8525 | 8468
8490 | 8581 | 8507
8544 | 8591 | 8557
8563 | 8601 | 8569
8580 | 8623 | 8586
8591 | 8646 | 8592
8604 | 8668 | 8617
8699 | 8675 | 8659

So before, there was a slight 0.4% regression from startpos (so no terminals nearby), and the optimized basically has no regression.


And a test position from https://github.com/LeelaChessZero/lc0/pull/889#issuecomment-502174009

master | all `GetBest` | uci-only
-- | -- | --
9664 | 10815 | 10385
10408 | 10948 | 10414
10437 | 10972 | 10532
10448 | 11020 | 10544
10507 | 11039 | 10563
10534 | 11052 | 10566
10554 | 11063 | 10606
10561 | 11101 | 10682
10568 | 11154 | 10695

And for this position with terminal moves, before there was 5.1% speed regression while the uci-only version has 0.5% speed regression.

Mardak:
To be clear, the latest commit "Compute terminal distance sorting only for uci info and best move." still correctly shows mate for all "uci info … multipv" as well as picking shortest mate for "uci bestmove".

The main difference is that there were many intermediate calls (for each visit) to `GetBestChildNoTemperature` -> `GetBestChildrenNoTemperature` that resulted in every root edge computing `GetTerminalSortingKey` without actually changing the uci output.

borg323:
No need for more tests @zz4032. I'm asking because you are getting about 2% more nps with this pr but slightly lower elo. So to a first approximation the backtracking has a similar time penalty and was wondering if it also resulted in time losses, given the tight tc you used.

Mardak:
Could you run a quick test comparing nps with the newest commit? Interesting that there was a 2% increase (?) in nps from the previous commit, so I wonder if the optimization moves the speed significantly for you.

zz4032:
The nodes/move difference is probably statistical noise. Also I tried to deal with huge node values by removing values above a given percentile.
Here an example of 176742 nodes and 4873411 nps after 34 milliseconds calculation time:
https://pastebin.com/raw/JCfC1guh

Maybe median values are better to compare in this case.
From 5000 games match above, median nodes/move without removing any values: 3875 (lc0) / 3887 (lc0_PR889)
With newest commit: 681 (lc0) / 683 (lc0_PR889)
614 games with TC 1.44s/game+0.006s/move. 
```
Score of lc0 vs lc0_PR889: 156 - 146 - 312  [0.508] 614
Elo difference: 5.7 +/- 19.3, LOS: 71.8 %, DrawRatio: 50.8 %
```

jkormu:
There seems to be some comments from Tilps and Crem (mooskagh) regarding this pr in discord and I am not sure how active @Mardak is in discord. It might be easier for the author to address the comments if they are available in this thread. My apologies if these comments were already available on this thread or as a review comments (did a quick check and didn't find).

```
Hace 19.8.2019 9:47 PM (EET):
what is the holdup on #889?

Crem 19.8.2019 10:29 PM (EET):
But actually we are hesitant to merge it because it does recursive tree traversal which may be slow

Hace 20.8.2019 12:07 AM (EET):
@crem I thought only slower (measurable?) in the rare case terminal is found... the benefit of having and using mate info should hopefully outweigh the possible millisecond lost (until some better approach comes along). Even if slightly slower nothing a later netversion can't overcome

Tilps 20.8.2019 12:45 AM (EET):
its the concern that its not going to be a millisecond lost in some rare circumstance
instead it'll be walking almost the entire tree

Hace 20.8.2019 8:03 AM (EET):
@Tilps Well, how can the time concern be validated? Would FENs with some terminals in sight help? Traversal of the tree for calculating Q-avgs happens continuously in mcts, the question is if the (rare) additional terminalscore traversal(s) makes this slower or: faster because of aerlier cutoff decisions. On the other hand traversal is chosen to prevent extending the nodesize (cpu vs memory tradeoff choice). Do you think example FENs will help determine tradeoff considerations for pr889?

Tilps 20.8.2019 8:07 AM (EET):
The difference is traversal for updates is only at most depth steps worth of nodes. Calculating longest loss can involve a much larger number of nodes.
Simplest solution is likely to just track the depths as we go.

Crem 20.8.2019 9:51 AM (EET):
It's also doing so while holding a nodes mutex, so no search is happening when it decides to output uci info line.
```

Mardak:
Oh thanks. Those issues weren't brought up in this PR's review here. If the overall concern is the traversal at time of selecting the bestmove, then the approach here needs to be completely redone.

Is the suggestion then to use additional memory on each Node so that distance to mate is calculated during Backup update when Q and sticky terminals are computed? If so, would this approach even be accepted for the increased memory usage?

Otherwise i suppose an approach more like this current PR but having a cache of computed distances for maybe just the root children?

Would be good to get a reviewer's input before implementing an approach that wouldn't be acceptable anyway.

TesseractA:
Has anyone suggested that lines with more terminal nodes get fewer visits only because they are terminal? Henceforth lines which are drawn out get more visits, therefore are as prioritized first choice? (If so, it also train down P with it when considering faster wins.)

Mardak:
Closing in favor of much simplified #992 for now and with #961 on the horizon, we can revisit to prefer shortest mate then.

PR #890, opened 2019-06-15
Add range validation for q and d values in v4 training data.

Tilps:
One contributor managed to upload corrupt q and d values - unclear why, but lets ensure it doesn't happen in future.

### DIFF
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -1,4 +1,4 @@
-﻿/*
+/*
   This file is part of Leela Chess Zero.
   Copyright (C) 2018 The LCZero Authors
 
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -1,4 +1,4 @@
-/*
+﻿/*
   This file is part of Leela Chess Zero.
   Copyright (C) 2018 The LCZero Authors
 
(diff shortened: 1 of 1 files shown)


src/mcts/node.cc
```diff
@@ -376,11 +376,23 @@ V4TrainingData Node::GetV4TrainingData(GameResult game_result,
 
   // Aggregate evaluation Q.
   result.root_q = -GetQ();
+  if (result.root_q < -1 || result.root_q > 1) {
```
mooskagh: Maybe add a helper function `Check` or `ExceptionUnless` into `utils/exception.h`?

And then call `ExceptionUnless(-1 <= result.root_q  && result.root_q <= 1, "Corrupt root q");` or even `Check(-1 <= result.root_q  &&| result.root_q <= 1) << "Corrupt root q " << result.root_q;` ?

Then it won't look like it's golang. :) And if it's macro like we have for CERR in logging.h, we can also log filename and line.

Naphthalin:
Is this the same cause why we had corrupted training data ~1-2 months ago? Also, is this relevant for the new training data format for MLH and FRC?

Tilps:
I added much more extensive validation than this to rescorer - games corrupted in this way are now safely deleted once they arrive at the training machine.  So possibly no need to have this.

PR #848, opened 2019-05-10
Android build with Cross Compilation for BLAS and OpenCL back-ends

lealgo:
Provides the files and tweaks to **meson** that enable the cross-compilation to Android, and also to other platforms like ARM Linux. This PR depends on the work done by @borg323's on the subprojects: **protobuf** and **opencl**, allowing the build to be performed from source instead of using the build machine's shared libraries. 

You could test this PR as it is right now, but the file `subprojects/protobuf-3.5.1/meson.build` should be replaced with this one: [meson.build](https://github.com/LeelaChessZero/lc0/files/3238627/protobuf.meson.build.zip) . The problem with cross-compiling protobuf is better explained here: https://github.com/LeelaChessZero/lc0/pull/848#issuecomment-497949554

---

Note: This PR started as a method for compiling lc0 on Termux, but then it evolved into cross-compilation. I kept the previous comments for reference. If you want instructions on Termux then follow the [wiki](https://github.com/LeelaChessZero/lc0/wiki/Building-lc0-on-Android,-with-Termux)

PR #891, opened 2019-06-16
Update meson.build

Toniacz93:
Update


borg323:
What is the problem you are trying to fix with this?

lealgo:
@Toniacz93 

You might want to take a look at the following PR:

https://github.com/LeelaChessZero/lc0/pull/848

Naphthalin:
Is this still relevant (if someone understood why it was relevant somewhere in the past)? No explanation by the author for 10 months, so can probably be closed.

PR #892, opened 2019-06-16
fix calculation of max size needed to hold tensors

ankan-ban:
1. kNumFilters can be smaller than kNumInputPlanes. We need to take max of them for computing sizes of intermediate buffers.
2. for wdl, we need 3x the memory for holding value head output.


src/neural/cuda/network_cudnn.cc
```diff
@@ -437,7 +438,8 @@ class CudnnNetwork : public Network {
     // 3. Allocate GPU memory for running the network:
     //    - three buffers of max size are enough (one to hold input, second to
     //      hold output and third to hold skip connection's input).
-    size_t maxSize = resi_last_->GetOutputSize(max_batch_size_);
+    size_t maxSize = max_batch_size_ * std::max(kNumInputPlanes, kNumFilters) *
```
Tilps: Is there also an implicit assumption here that policy channels is less than kNumInputPlanes and/or  that the value fully connected output is smaller 64 times that?
If so might be good to document it.

ankan-ban: I think it's better to avoid assumptions (and run into more such issues in future). Updated the change to take max of all layers.

oscardssmith:
Why does wdl change the policy head?

ankan-ban:
It doesn't. I meant the value head. Updated description.

PR #893, opened 2019-06-20
fix appveyor meson version to 0.50.1

borg323:
Upgrading to 0.51.0 breaks the appveyor build - something changed in the msvc flags used, and the way the protobuf library is built is no longer compatible. This will do for now, until this is analyses better.


borg323:
#904 is the real fix

PR #905, opened 2019-07-12
Add syzygy tablebases for use with selfplay

MelleKoning:
Add syzygy tablebases to use these for selfplay games.

The goal is to save time in training and game matches: for example when contributing to leela-training and matches I've seen games prolong between KRvsKR for 200+ moves, just wasting CPU/GPU cycles.
So the goal here is to use tablebases when those exists for clients; it should just use the tablebases available on your system even if that's only 4 or 5 pieces instead of 6 or 7 piece TB.

Seems to work when starting lco as follows:
lc0 selfplay --visits=800 --syzygy-paths=c:\path_to\tablebases

After this is merged, we probably also need to enable the go-client to pass on the syzygy-path from commandline to lc0 so that local installed client tablebases are indeed being used.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -132,6 +132,7 @@ files += [
   'src/mcts/stoppers/factory.cc',
   'src/mcts/stoppers/legacy.cc',
   'src/mcts/stoppers/smooth.cc',
+  'src/mcts/stoppers/alphazero.cc',
   'src/mcts/stoppers/stoppers.cc',
   'src/mcts/stoppers/timemgr.cc',
   'src/neural/cache.cc',
--- a/src/mcts/stoppers/alphazero.cc
+++ b/src/mcts/stoppers/alphazero.cc
@@ -63,31 +63,25 @@ class AlphazeroTimeManager : public TimeManager {
 std::unique_ptr<SearchStopper> AlphazeroTimeManager::GetStopper(
     const GoParams& params, const Position& position) {
   const bool is_black = position.IsBlackToMove();
-  std::optional<int64_t> time = (is_black ? params.btime : params.wtime);
+  const std::optional<int64_t>& time = (is_black ? params.btime : params.wtime);
   // If no time limit is given, don't stop on this condition.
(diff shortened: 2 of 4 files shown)


AlexisOlson:
What kind of risk to training would there be if some had a corrupted (or intentionally sabotaged) tablebase?

MelleKoning:
That is probably caught in rescoring when games are processed on the server, like with regular tablebase rescoring.

src/selfplay/game.cc
```diff
@@ -89,14 +89,14 @@ void SelfPlayGame::Play(int white_threads, int black_threads, bool training,
...
+
     {
       std::lock_guard<std::mutex> lock(mutex_);
       if (abort_) break;
       search_ = std::make_unique<Search>(
           *tree_[idx], options_[idx].network, options_[idx].best_move_callback,
           options_[idx].info_callback, options_[idx].search_limits,
-          *options_[idx].uci_options, options_[idx].cache, nullptr);
-      // TODO: add Syzygy option for selfplay.
+          *options_[idx].uci_options, options_[idx].cache, syzygy_tb);
```
mooskagh, line 99: Maybe make it a `shared_ptr` in `options_` like other fields, to allow e.g. having tablebase vs no-tablebase training?

MelleKoning: I don't really see why we would want different tablebase setttings for player1 and player2 during training for one client?

In this PR, using TB is already optional. 
That is, some clients who have 4-5 piece TB installed can use it, most clients won't have it installed initially and all those games come together at the server for aggregation into a net.

In other words: what would one client training with and without help? Or is this just for generation of nets on a local system?

MelleKoning: @mooskagh rebase to master was quite difficult due to the number of changes and the time it takes to get these kind of suggestions in. Is there any way I can speed up or make merging easier?

MelleKoning:
Re-open with earlier version to restore branch ⛸️

PR #907, opened 2019-07-22
default stickyendgames to true for training

MelleKoning:
StickyEndgames was set to false for training, let's put it to true for upcoming T60 run.


MelleKoning:
Params can be controlled from the server when trainings are triggered, closing.

PR #917, opened 2019-08-05
minimize max fastmath error considering float32 accuracy

borg323:
The initial fastmath.h functions were optimized analytically to minimize rms error over the full mantissa range. Unfortunately, this was not taking into account the limited float32 accuracy, resulting in a small offset from the desired values. Here I used a numerical optimization approach to tune the constants for minimal error over the full mantissa range, and took the opportunity to minimize max error instead of rms error.

Also improved the comments a bit.

### DIFF
--- a/src/utils/fastmath.h
+++ b/src/utils/fastmath.h
@@ -37,10 +37,8 @@ namespace lczero {
 // Fast approximate log2(x). Does no range checking.
 // The approximation used here is log2(2^N*(1+f)) ~ N+f*(1+k-k*f) where N is the
 // exponent and f the fraction (mantissa), f>=0. The constant k is used to tune
-// the approximation accuracy. Note that the formula used here is transformed to
-// use 1+f for speed reasons: log2(2^N*(1+f)) ~ N+(1+f)*(1+3*k-k*(1+f))-2*k-1.
-// In the final version some constants were slightly modified for better
-// accuracy with 32 bit floating point math.
+// the approximation accuracy. In the final version some constants were slightly
+// modified for better accuracy with 32 bit floating point math.
 inline float FastLog2(const float a) {
   uint32_t tmp;
   std::memcpy(&tmp, &a, sizeof(float));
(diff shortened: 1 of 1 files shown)


oscardssmith:
How much better is the new accuracy?

borg323:
Not a lot, max relative error e.g for log2 was 6.65e-3, with this it is 5.23e-3. However the accuracy was not the motivation here, but to eliminate the small offset: FastLog2(1) would return 7.63e-6 instead of 0 (which was a design objective for the approximation).

borg323:
Due to rounding in an intermediate calculation, `FastLog2()` was not monotonic. The latest commit uses an earlier version that doesn't have this problem. This one was slightly slower with some compilers, but after recent changes I expect the difference to be in the noise.

borg323:
After extensive discussion with @ddobbelaere we concluded that minimizing the absolute error is a better objective for `FastLog2()`.

Absolute error plot for latest (blue) and master (red) versions of `FastLog2()`:
absolute3

Relative error plot for latest (blue) and master (red) versions of `FastPow2()`:
relative3

jjoshua2:
How much different would it be if you optimized it for fp16? Pretty much all tournaments use this instead. It's probably not worth keeping two separate fastLog implementations around though?

ddobbelaere:
Here's a benchmark over 10 runs using the script https://github.com/ddobbelaere/tools/blob/master/lc0/benchmark.py

```
# master 06748a4
min       227501
max       231164
stddev    1115
mean      229461
95% CI    ±691

# PR 917
min       227040
max       230418
stddev    1024
mean      228393
95% CI    ±635
```

Slightly slower, but still not within confidence bounds for the given number of runs.

@jjoshua2 The fastmath routines are only exclusively on CPU, where typically `float` has 32 bits precision.

borg323:
This is only used on the cpu side. Even if a cuda version would improve performance, the only exp calls there are for floats.

borg323:
With the above script and `--smart-pruning-factor=0 --movetime=-1 --nodes=1000000 --backend=random --threads=1` I get:
```
#master 06748a4
min       194346 nps
max       195105 nps
stddev    241 nps
mean      194722 nps
95% CI    ±150 nps

#pr917
min       194121 nps
max       196140 nps
stddev    552 nps
mean      195644 nps
95% CI    ±342 nps
```

So in my case 917 is slightly faster, probably the result is compiler dependent. This is with gcc 8.3.0.

ddobbelaere:
Forgot to mention that I used gcc 9.1.1 on Fedora with lto enabled and pext
enabled (but no fast math enabled). The benchmark command was using random
backend with no other flags set (so all default).

Op do 8 aug. 2019 17:58 schreef borg323 <notifications@github.com>:

Issue #913, opened 2019-08-01
The DeepMind PUCT formula (probably) causes the lack of opening diversity


Naphthalin:
This is supposed to be more or less a summary of a few posts on Discord about the topic.

EatNow (Leela Zero discord) suggested a small change in the PUCT formula which would be in line with the standard UCB, see [here](https://colab.research.google.com/drive/14v45o1xbfrBz0sG3mHbqFtYz_IrQHLTg#scrollTo=1VeRCpCSaHe3 are EatNow's initial considerations), and as it turned out Tilps already realized that there are differences between the DeepMind pseudocode and the formulas, see [https://github.com/LeelaChessZero/lc0/pull/699](https://github.com/LeelaChessZero/lc0/pull/699)

This means:
A) when comparing the DM PUCT formula to their UCB1 origin, one would have to replace a 1/sqrt(parent_visits) by a 1/sqrt(child_visits + 1) to actually match them
B) this means while the formula apparently intends to estimate lower bounds of child nodes, it uses sqrt(parent_visits) to estimate this lower bound
C) one would like to increase the certainty over time which is done by scaling puct by some log(parent_visits) stuff --> this is the same scaling we already do
**D) when using the modified formula, there are actually equilibria for the policies**, so e.g. during the opening the policies don't tend to focus on single moves/lines but on the best moves
especially D) might be very interesting: **while the DeepMind formula promises to have the child_visit ratio as a training target for policies, because of the PUCT formula this training target shifts**, since
`Q_1 + p(s,a_1)*N/(N_1 + 1) * puct/sqrt(N) = P_1 == P_2 = Q_2 + p(s,a_2)*N/(N_2 + 1) * puct/sqrt(N)`
with `p(s,a_1) = N_1/N` and `p(s,a_2) = N_2/N` (assuming we are in a stable equilibrium) can only be true if `Q_1 = Q_2`or the worse moves gets no visits and zero policy.

So whenever Q_1 and Q_2 are different even by the slighest bit, this difference will be amplified by the policies and we get our usual extremely sharp policies.

For Leela Zero Go, the best opening moves 4-4 and 4-3 have evals of 43.76% and 43.67% but policies of 17.6% and 2.6% (which means 70% and 20% including symmetries). the 2 next best moves at 43.54% and 43.28% are at 0.2% and 0.25%

[...]

PR #918, opened 2019-08-05
Implement equilibrium based U formula

kiudee:
This implements the changes mentioned in issue #913 and reuses the flag introduced by #699.

For given puct and parent_visits values we get a unique stable equilibrium solution of the RL process where `child_visits = p(s,a) * parent_visits` for all moves.
Note, that this change can only be evaluated using a full reinforcement learning run.

Let me know if some of the code can be simplified.

Proposals for parameter values
------------------------------
Moved to #924 

TODO
----

- [x] ~~Identify good values for `cpuct`, `cpuct-base` and `cpuct-factor`~~
- [x] ~~What is the effect of fpu, dirichlet noise etc?~~
- [x] ~~Adjust time management~~

### DIFF
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -367,6 +367,11 @@ class EdgeAndNode {
     return numerator * GetP() / (1 + GetNStarted());
   }
 
+  float GetNewU(float numerator) const {
+    const float x = 1 + GetNStarted();
+    return numerator * GetP() * FastInvSqrt(x) / x;
+  }
+
   int GetVisitsToReachU(float target_score, float numerator, float default_q,
                         bool logit_q) const {
     const auto q = GetQ(default_q, logit_q);
@@ -375,6 +380,18 @@ class EdgeAndNode {
     return std::max(
         1.0f,
         std::min(std::floor(GetP() * numerator / (target_score - q) - n1) + 1,
+                 1e9f));
+  }
+
+  int GetVisitsToReachNewU(float target_score, float numerator,
+                           float default_q, bool logit_q) const {
+    const auto q = GetQ(default_q, logit_q);
+    if (q >= target_score) return std::numeric_limits<int>::max();
+    const auto n1 = GetNStarted() + 1;
+    const float inner = std::pow((GetP() * numerator) / (target_score - q), 2. / 3.);
+    return std::max(
+        1.0f,
+        std::min(std::floor(inner - n1) + 1,
                  1e9f));
   }
 
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -212,12 +212,24 @@ std::vector<std::string> Search::GetVerboseStats(Node* node,
   const float fpu = GetFpu(params_, node, node == root_node_);
   const float cpuct = ComputeCpuct(params_, node->GetN());
   const float U_coeff =
-    cpuct * std::sqrt(std::max(node->GetChildrenVisits(), 1u));
+    cpuct * (params_.GetNewUEnabled() ?
+      std::max(node->GetChildrenVisits(), 1u) :
+      std::sqrt(std::max(node->GetChildrenVisits(), 1u)));
   const bool logit_q = params_.GetLogitQ();
 
   std::vector<EdgeAndNode> edges;
   for (const auto& edge : node->Edges()) edges.push_back(edge);
 
+  if (params_.GetNewUEnabled()) {
+  std::sort(
+      edges.begin(), edges.end(),
+      [&fpu, &U_coeff, &logit_q](EdgeAndNode a, EdgeAndNode b) {
+        return std::forward_as_tuple(
+          a.GetN(), a.GetQ(fpu, logit_q) + a.GetNewU(U_coeff)) <
+          std::forward_as_tuple(
+          b.GetN(), b.GetQ(fpu, logit_q) + b.GetNewU(U_coeff));
+      });  
+  } else {
   std::sort(
       edges.begin(), edges.end(),
       [&fpu, &U_coeff, &logit_q](EdgeAndNode a, EdgeAndNode b) {
@@ -226,6 +238,7 @@ std::vector<std::string> Search::GetVerboseStats(Node* node,
           std::forward_as_tuple(
           b.GetN(), b.GetQ(fpu, logit_q) + b.GetU(U_coeff));
       });
+  }
 
   std::vector<std::string> infos;
   for (const auto& edge : edges) {
@@ -249,11 +262,12 @@ std::vector<std::string> Search::GetVerboseStats(Node* node,
     oss << "(D: " << std::setw(6) << std::setprecision(3) << edge.GetD()
         << ") ";
 
-    oss << "(U: " << std::setw(6) << std::setprecision(5) << edge.GetU(U_coeff)
+    oss << "(U: " << std::setw(6) << std::setprecision(5)
+        << (params_.GetNewUEnabled() ? edge.GetNewU(U_coeff) : edge.GetU(U_coeff))
         << ") ";
 
     oss << "(Q+U: " << std::setw(8) << std::setprecision(5)
-        << edge.GetQ(fpu, logit_q) + edge.GetU(U_coeff)
+        << edge.GetQ(fpu, logit_q) + (params_.GetNewUEnabled() ? edge.GetNewU(U_coeff) :edge.GetU(U_coeff))
         << ") ";
 
     oss << "(V: ";
@@ -935,7 +949,10 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend(
     // playout remains incomplete; we must go deeper.
     const float cpuct = ComputeCpuct(params_, node->GetN());
     const float puct_mult =
-        cpuct * std::sqrt(std::max(node->GetChildrenVisits(), 1u));
+        cpuct * (params_.GetNewUEnabled() ?
+            std::max(node->GetChildrenVisits(), 1u) :
+            std::sqrt(std::max(node->GetChildrenVisits(), 1u)));
+
     float best = std::numeric_limits<float>::lowest();
     float second_best = std::numeric_limits<float>::lowest();
     int possible_moves = 0;
@@ -960,7 +977,10 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend(
         ++possible_moves;
       }
       const float Q = child.GetQ(fpu, params_.GetLogitQ());
-      const float score = child.GetU(puct_mult) + Q;
+      const float score = Q +
+        (params_.GetNewUEnabled() ?
+          child.GetNewU(puct_mult):
+          child.GetU(puct_mult));
       if (score > best) {
         second_best = best;
         second_best_edge = best_edge;
@@ -973,7 +993,9 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend(
     }
 
     if (second_best_edge) {
-      int estimated_visits_to_change_best =
+      int estimated_visits_to_change_best = params_.GetNewUEnabled() ?
+          best_edge.GetVisitsToReachNewU(second_best, puct_mult, fpu,
+                                         params_.GetLogitQ()) :
           best_edge.GetVisitsToReachU(second_best, puct_mult, fpu,
                                       params_.GetLogitQ());
       // Only cache for n-2 steps as the estimate created by GetVisitsToReachU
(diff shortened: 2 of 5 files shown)


Naphthalin:
Thanks for taking the time to add the support for the changed formula.

For good values for cpuct etc the equilibrium policies suggest to definitely be in the interval `[0.1, 1.0]` at 800 nodes, and I would recommend to start at 0.5 which would give (at the current evals of the top moves) policies

`e4   35%
d4   23%
Nf3  15%
c4   8%
e3   7%
Nc3  3%`

A lower value like 0.3 or would also be a good choice for a first try since it should increase the convergence speed.

fohristiwhirl:
Someone pointed me at [this autobuild](https://ci.appveyor.com/project/LeelaChessZero/lc0/builds/26590255/job/ej7j2b9rs5otybnt/artifacts) so I tried it out. The effects are quite profound:

image

Normal search at top, new way at bottom.

U is a lot bigger for most moves, which means they get hidden by Nibbler.

fohristiwhirl:
OK, it was noted that I needed to change some other settings. After making my lc0.config look like this:

```
--new-u-enabled=true
--cpuct=0.38
--policy-softmax-temp=1.59
```

Things seem more normal. It's noticably slower in terms of N/s though.

src/mcts/node.h
```diff
@@ -373,6 +377,18 @@ class EdgeAndNode {
                  1e9f));
   }
 
+  int GetVisitsToReachNewU(float target_score, float numerator,
+                           float default_q) const {
+    const auto q = GetQ(default_q);
+    if (q >= target_score) return std::numeric_limits<int>::max();
+    const auto n1 = GetNStarted() + 1;
+    const float inner = std::pow((target_score - q) / (GetP() * numerator), 2. / 3.);
```
Naphthalin: ``target_score - q`` should be in the denominator and vice versa

src/mcts/node.h
```diff
@@ -362,6 +362,10 @@ class EdgeAndNode {
     return numerator * GetP() / (1 + GetNStarted());
   }
 
+  float GetNewU(float numerator) const {
+    return numerator * GetP() / std::pow(1.0 + GetNStarted(), 1.5);
```
Tilps: Should consider using fastpow instead of pow at some point, but maybe not yet.

src/mcts/search.cc
```diff
@@ -212,17 +212,28 @@ std::vector<std::string> Search::GetVerboseStats(Node* node,
...
   for (const auto& edge : node->Edges()) edges.push_back(edge);
 
+  if (params_.GetNewUEnabled()) {
+  std::sort(
+      edges.begin(), edges.end(),
+      [&fpu, &U_coeff](EdgeAndNode a, EdgeAndNode b) {
+        return std::forward_as_tuple(a.GetN(), a.GetQ(fpu) + a.GetNewU(U_coeff)) <
+               std::forward_as_tuple(b.GetN(), b.GetQ(fpu) + b.GetNewU(U_coeff));
+      });  
+  } else {
```
Tilps: Indenting needs adjusting here and below.

kiudee:
> I don't think we should submit this without at least some feeling for how it affects elo. If its hundreds of elo reduction I'm not sure how it could possibly improve in a training run.

Thank you for the review. I will run some Elo tests.

kiudee:
First regression test is finished. I compared the lc0 binary of this branch to v0.22 using the same parameters. This is to test, if some of the new modifications actually worsen performance.

Setup
-----
```LC0-v0.21.3
Time control:   5s + 0.1s
Hardware:  1080TI
Book: Chad 8-ply (sequential)
Adjudication: -draw movenumber=50 movecount=5 score=8 -resign movecount=5 score=1000
Software: Cutechess-cli
Network: 42850
SPRT: Elo0: -10, Elo1: 1, alpha=0.05, beta=0.05
```

Result
------
SPRT:
```
Score of lc0-new-default vs lc0-v0.22: 177 - 157 - 633  [0.510] 967
Elo difference: 7.2 +/- 12.9, LOS: 86.3 %, DrawRatio: 65.5 %
SPRT: llr 2.99 (101.4%%), lbound -2.94, ubound 2.94 - H1 was accepted
Finished match
```
Ordo:
```   # PLAYER             :  RATING  ERROR  POINTS  PLAYED   (%)  CFS(%)
   1 lc0-new-default    :       4      6   493.5     967  51.0      87
   2 lc0-v0.22          :      -4      6   473.5     967  49.0     ---

White advantage = 27.61 +/- 6.41
Draw rate (equal opponents) = 66.10 % +/- 1.50
```
We can confidently say that we did not make lc0 worse using this commit.

Next test
---------
Now I will run a fixed nodes/move test comparing the new-u formula against v0.22 to see if it performs comparably and not much worse. Note, that since the network was not trained using new-u, we expect it to do worse.

kiudee:
First regression test is finished. I compared the lc0 binary of this branch to v0.22 using the same parameters. This is to test, if some of the new modifications actually worsen performance.

Setup
-----
```LC0-v0.21.3
Nodes per move: 800
Hardware:  1080TI
Book: Chad 8-ply (sequential)
Adjudication: -draw movenumber=50 movecount=5 score=8 -resign movecount=5 score=1000
Software: Cutechess-cli
Network: 42850
SPRT: Elo0: -40, Elo1: 0, alpha=0.05, beta=0.05
```

Result
------
SPRT:
```
Score of lc0-new-params vs lc0-v0.22: 20 - 12 - 37  [0.558] 69
Elo difference: 40.5 +/- 56.2, LOS: 92.1 %, DrawRatio: 53.6 %
SPRT: llr 3.03 (102.9%%), lbound -2.94, ubound 2.94 - H1 was accepted
```
Ordo:
```   # PLAYER            :  RATING  ERROR  POINTS  PLAYED   (%)  CFS(%)
   1 lc0-new-params    :      21     29    38.5      69  55.8      92
   2 lc0-v0.22         :     -21     29    30.5      69  44.2     ---

White advantage = 72.51 +/- 29.87
Draw rate (equal opponents) = 57.16 % +/- 6.21
```
Reminder that this is using an SPRT test, thus the Elo estimate is highly biased. The only thing we can conclude is that the formula will not perform much worse than the existing one.

The results are very promising and I think the pull request is ready to be merged and to be tested in a reinforcement learning run.

Tilps:
These results do look great, but to be clear when you say 'the same parameters' - I assume v0.22 are using default puct and policy soft max while the new code is using different values?

kiudee:
The first test uses the default parameters of v0.22 (`NewUEnabled=false`) - just to test whether the code modifications negatively effected performance.

The second test then compares the defaults for v0.22 against `NewUEnabled=true` using the tuned cpuct/policy-temperature values by @jkormu (the test was running a fixed 800 nodes/move, the parameters were tuned for 1064).

src/mcts/node.h
kiudee: Are you referring to [this](https://martin.ankerl.com/2012/01/25/optimized-approximative-pow-in-c-and-cpp/) or is there another `fastpow`?

Tilps:
Default puct and policy softmax is known suboptimal for 800 nodes - so the fact that newU is looking possibly better there may just mean you had better luck with tune - but yes this meets my threshold for acceptable to experiment with.

src/mcts/node.h
Tilps: the old policy_softmax_temperature implementation did fastpow(x,y) by fastexp2(fastlog2(x) * y) IIRC - using the fastlog2 and fastexp2 functions we have in our fastmath utilities file.

oscardssmith:
Have we found optimal tune for 800 nodes? If so, what is it?

kiudee:
@Tilps indeed, the only goal was to sanity check if the search is much worse using the new formula.

@oscardssmith Except for @jkormu, nobody did tuning of the constants yet (his/her values are tuned for 1064 though, which is very close).

src/mcts/node.h
```diff
@@ -362,6 +363,11 @@ class EdgeAndNode {
     return numerator * GetP() / (1 + GetNStarted());
   }
 
+  float GetNewU(float numerator) const {
+    const float x = 1 + GetNStarted();
+    return numerator * GetP() * FastInvSqrt(x) / x;
```
kiudee: @borg323 could you take a look at this function and the referenced `FastInvSqrt` and check if it could be optimized further?

oscardssmith: isn't fastinvsqrt(x)/x just sqrt(x)?

kiudee: It’s equivalent to `1/(x * sqrt(x)) = 1/pow(x, 3./2.)`

Sopel97: some guesses
1. (numerator * GetP()) * (FastInvSqrt(x) / x) may be faster on strict fp because it breaks a dependency
2. GetNStarted has a branch, maybe use a dummy node instead of nullptr?
3. rsqrtss should really be faster, it has latency of 5 and tp of 1 cpi

oscardssmith: The really fast answer is to use the awful sort of floating point bithacks that finvsqrt uses. Not saying that's a good idea though

borg323: `return numerator * GetP() * FastInvSqrt(x * x * x);` should be slightly faster.

src/utils/fastmath.h
```diff
@@ -68,4 +68,18 @@ inline float FastLog(const float a) {
...
+
+  union {
+    float f;
+    uint32_t i;
+  } conv = {number}; // member 'f' set to value of 'number'.
+  conv.i  = 0x5f3759df - ( conv.i >> 1 );
+  conv.f  *= ( threehalfs - ( x2 * conv.f * conv.f ) );
+  return conv.f;
+}
+
```
borg323: The following variation avoids undefined behavior and gives identical code with gcc:
```
inline float FastInvSqrt(const float number)
{ 
  const float x2 = number * 0.5F;
  const float threehalfs = 1.5F;
  float f;
  uint32_t i;
  memcpy(&i, &number, sizeof(float));
  i  = 0x5f3759df - ( i >> 1 );
  memcpy(&f, &i, sizeof(float));
  return f * ( threehalfs - ( x2 * f * f ) );
}
```
Also the `_mm_rsqrt_ss()` alternative should be provided here - this should be faster, maybe the version tested earlier was doing some unnecessary float to double conversions?

kiudee: `return numerator * GetP() * FastInvSqrt(x * x * x);` was tested and 3% slower.

kiudee: @borg323 what would be the proper way to use `_mm_rsqrt_ss()` to avoid conversions?

borg323: The code you suggested seems fine:
```
float FastInvSqrt(const float number)
{ 
    return _mm_cvtss_f32(_mm_rsqrt_ss(_mm_set_ss(number)));
}
```
Similarly the following variation generates identical code:
```
float FastInvSqrt(const float number)
{ 
    return _mm_cvtss_f32(_mm_rsqrt_ss(_mm_load_ss(&number)));
}
```
With gcc, the code is:
```
        pxor    xmm1, xmm1
        movss   xmm1, xmm0
        movaps  xmm0, xmm1
        rsqrtss xmm0, xmm1
        ret
```
Both clang and icc do a better job
```
        rsqrtss xmm0, xmm0
        ret
```
I expect when inlined the gcc code would also be reduced to the minimum.

kiudee: Trying `return _mm_cvtss_f32(_mm_rsqrt_ss(_mm_load_ss(&number)));` using gcc is 2.5% slower than the current implementation.

kiudee: Using clang-8 it is even slower (9% as compared to current implementation in gcc).

src/mcts/node.h
Sopel97: http://quick-bench.com/u2AGvU2ZmWLY2bGqH4ykIpN2_IQ
NPS with new U - random backend (4 runs per build)
I used the newest MSVC compiler, intel i7-920
`lc0 benchmark -w bad-gyal-1.pb.gz --backend=random --threads=1 --smart-pruning-factor=0 --movetime=-1 --nodes=1000000 --new-u-enabled`
The last includes points 1, 2, 3 from my comment above. (in order of builds {nothing; 3; 2,3; 1,2,3}) I'm not sure if I patched correctly all the places that checked whether _node == nullptr (I did all local ones, but I don't know if the node can get 'spilled' outside)
If someone could retest on newer cpu it would be great
profiles before/after
before
after
the `Actualize()` in edge iterator becomes the biggest problem, though it looks hard to improve without architectural changes

changes: https://github.com/kiudee/lc0/compare/new_u...Sopel97:opt_new_u

kiudee: This commit causes a reproducible segmentation fault when running new_u with `LogLiveStats` enabled.

Sopel97: I look at this. Tried a few things. Can't get it to work without breaking other things.

kiudee: Thank you for looking into it @Sopel97.

@Tilps @borg323 The difference in performance (using the random backend) for Sopel97's branch is only ~7%, while being quite error prone due to the changes to the pointers. I think it would be better to have a separate pull request later for this performance optimization, which can be properly reviewed and tested. What is your opinion?

borg323: 7% is more than I expected. This is certainly worth a separate pr for review and discussion.

dtracers:
Will this need to be readjusted to take into account the flags from #925 ?

Naphthalin:
There are conflicts with ``GetVisitsToReachU()`` since both PR's needed an adjustment there.

Naphthalin:
For the version of the PR with resolved conflicts with #925 , see #959 .

kiudee:
To decouple all of the different pull requests, I rebased this one onto master. Thus all of the logit Q changes are incorporated.

Naphthalin:
The main premise of this PR was implemented by using PST > 1.0 in training which had the expected change in behavior discussed in #913 and #918. Actually using this formula in search puts too much weight on lower nodes due to the different powers involved. 

I recommend to close this PR, and point to it when discussing changes on the deepmind PUCT formula (e.g. trying to implement the original PUCT formula).

kiudee:
Closing this since PST was sufficient to gain the effect of this PR.

PR #920, opened 2019-08-09
Add git short revision to version string.

ddobbelaere:
Add unique git short revision to version string. Turns

```
       _
|   _ | |
|_ |_ |_| v0.23.0-dev built Aug  8 2019
```

into:

```
       _
|   _ | |
|_ |_ |_| v0.23.0-dev+gad71ed0 built Aug  9 2019
```

Very handy to avoid any confusion during development/benchmarks/comparison of PRs/logs used for future reference.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -34,16 +34,18 @@ endif
 
 # Use git short revision as build identifier.
 git = find_program('git', required: false)
+short_rev = 'unknown'
 if git.found()
   r = run_command(git, 'rev-parse', '--short', 'HEAD')
   if r.returncode() == 0
     short_rev = r.stdout().strip()
-    message('found git short revision ' + short_rev)
-    add_project_arguments('-DBUILD_IDENTIFIER="g' + short_rev + '"', language : 'cpp')
+  else
+    warning('Failed to parse short revision. Use git clone instead of downloading the archive from GitHub.')
   endif
-else
-  warning('Please install git to add short revision to build identifier in version string.')
 endif
+
+message('Using git short revision "' + short_rev + '" as build identifier.')
+add_project_arguments('-DBUILD_IDENTIFIER="g' + short_rev + '"', language : 'cpp')
 
 # Files to compile.
 deps = []
--- a/meson.build
+++ b/meson.build
@@ -37,7 +37,7 @@ r = run_command('git', 'rev-parse', '--short', 'HEAD')
 if r.returncode() == 0
   short_rev = r.stdout().strip()
   message('found git short revision ' + short_rev)
-  add_project_arguments('-DGIT_SHORT_REVISION="' + short_rev + '"', language : 'cpp')
+  add_project_arguments('-DBUILD_IDENTIFIER="g' + short_rev + '"', language : 'cpp')
 endif
 
 # Files to compile.
--- a/meson.build
+++ b/meson.build
@@ -106,6 +106,7 @@ endif
 # Construct build identifier.
 build_identifier = 'git.' + short_rev
 add_project_arguments('-DBUILD_IDENTIFIER="' + build_identifier + '"', language : 'cpp')
+message('Using build identifier "' + build_identifier + '".')
 
 #############################################################################
 ## Main files
--- a/meson.build
+++ b/meson.build
@@ -66,7 +66,6 @@ endif
 gen = generator(protoc, output: ['@BASENAME@.pb.cc', '@BASENAME@.pb.h'],
   arguments : ['--proto_path=@CURRENT_SOURCE_DIR@/libs/lczero-common', '--cpp_out=@BUILD_DIR@', '@INPUT@'])
 
-git = find_program('git', required: false)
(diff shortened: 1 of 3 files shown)


mooskagh:
It looks good, but to comply with SemVer I think it should be delimeted with `+`, not `-` (it's build identifier). E.g. `v0.23.0-dev+gad71ed0`. That way it will be possible e.g. to parse version on server side.

Also just to confirm, it's still possible to build it even if user happens not to have git installed, right?

Also, maybe it's a good idea to do that for every build, not only `dev`?..

mooskagh:
Continuing to nitpick. :)
Maybe make source code not git aware by renaming `GIT_SHORT_REVISION` to `BUILD_IDENTIFIER` or something like that.

ddobbelaere:
Thanks for the valid points raised.

Right now, if `git` is not installed for some reason, the flag BUILD_IDENTIFIER will not be set by the build system and the old version string with be shown.

meson.build
```diff
@@ -32,6 +32,19 @@ if cc.get_id() == 'clang' or cc.get_id() == 'gcc'
...
+git = find_program('git', required: false)
+if git.found()
+  r = run_command(git, 'rev-parse', '--short', 'HEAD')
+  if r.returncode() == 0
+    short_rev = r.stdout().strip()
+    message('found git short revision ' + short_rev)
+    add_project_arguments('-DBUILD_IDENTIFIER="g' + short_rev + '"', language : 'cpp')
+  endif
+else
+  warning('Please install git to add short revision to build identifier in version string.')
```
borg323: What about making `short_rev` "unknown" in this case?
That being said, I found (the hard way) that without git meson fails when trying to get a subproject, so I have already hardcoded git use in <https://github.com/LeelaChessZero/lc0/blob/master/meson.build#L69>.

ddobbelaere: Done.

ddobbelaere:
Thanks for the helpful suggestions.

Now the behavior is more consistent: the build identifier is always present in the version string now. If for some reason the short revision cannot be extracted (e.g., because `git` is not installed or the root folder is not a git repository because the user most likely downloaded the archives from GitHub) it is set to "unknown" and an appropriate warning is generated.

This leads to:

```
       _
|   _ | |
|_ |_ |_| v0.23.0-dev+gunknown built Aug  9 2019
```

I've also updated the handling of submodules to print a warning if both `git` and the required submodules folder are not found.

borg323:
I'm fine with the changes - @mooskagh is your approval still valid?

ddobbelaere:
Just thinking about something. Maybe it's also an idea to put revision to
'unknown' if working directory is unclean (uncommitted changes)?

Now it will show the HEAD revision but this is not correct for uncommitted
changes during development.

Maybe exclude meson_options.txt from this check if we decide to do this, so
that users can still change include and lib search paths if they like.

borg323:
I like this idea, but excluding files seems overkill and can't think of a simple way to do it. Moreover, do many people edit `meson_options.txt` to override paths? If so we probably need to see what they do and try to support it, so having a dirty/unclean indication may be incentive enough to report the use case.

ddobbelaere:
The latest version of the PR has the following behavior.

If `git` is not installed or if the user downloaded the source archive directly from GitHub (without using `git clone`):

```
v0.23.0-dev+git.unknown
```
Else if the working directory is clean (no uncommitted changes):

```
v0.23.0-dev+git.55fb8a8
```
Else (dirty working, nothing meaningful can be said):
```
v0.23.0-dev+git.dirty
```

This format of the build identifier is SemVer compliant, allows easy parsing of git short revision and is flexible (can be extended with additional identifiers in the future).

meson.build
```diff
@@ -66,20 +66,48 @@ endif
...
+  if r.returncode() == 0
+    # Now let's check if the working directory is clean.
+    if run_command(git, 'diff-index', '--quiet', 'HEAD').returncode() == 0
+      short_rev = r.stdout().strip()
+    else
+      short_rev = 'dirty'
+      warning('Cannot extract valid git short revision from dirty working directory.')
+    endif
+  else
+    warning('Failed to parse short revision. Use git clone instead of downloading the archive from GitHub.')
```
mooskagh: Does it mean that if user has git installed in system, but downloaded source through other means, it will fail to build? I think it still has to be buildable in that case.

borg323: If I parse it correctly, it will still build, with "unknown" tag.

mooskagh: Hm, I was sure it's `error()` but now I see it as `warning()` :)

mooskagh: Ah I saw line 83.. So it's about fetching submodules.
Then still same question about submodules: if `git` is installed but sources are manually downloaded (including submodules), it will fail to build?

Before this change, did it also fail to build?

borg323: Before this, git would be used unconditionally, so it would fail if not installed. This was done because meson requires (required?) git internally for subproject downloading and would fail with really unhelpful errors if git not installed.

If git is installed but repository is downloaded, it will clone the submodule instead of updating and works fine.

PR #921, opened 2019-08-09
also check opencl headers

### LINKED ISSUE #919: [build.sh] Error to run build.sh
Hi, I downloaded the code and followed the steps but I have an error.

Can Anybody help me?

I have Ubuntu 18.04 x64

imagen

This error was produced when I executed CC=clang-6.0 CXX=clang++-6.0 INSTALL_PREFIX=~/.local ./build.sh

borg323:
This should fix #919

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -311,7 +311,7 @@ if get_option('build_backends')
       deps += [ opencl_framework ]
       has_opencl = true
 
-  elif opencl_lib.found() and cc.has_header('CL/opencl.h', args: '-I' + get_option('opencl_include')[0])
+  elif opencl_lib.found() and cc.has_header('CL/opencl.h', args: '-I' + get_option('opencl_include'))
 
       deps += [ opencl_lib ]
       has_opencl = true
--- a/meson.build
+++ b/meson.build
@@ -311,7 +311,7 @@ if get_option('build_backends')
       deps += [ opencl_framework ]
       has_opencl = true
 
-  elif opencl_lib.found()
+  elif opencl_lib.found() and cc.has_header('CL/opencl.h', args: '-I' + get_option('opencl_include')[0])
 
       deps += [ opencl_lib ]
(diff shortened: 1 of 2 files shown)


meson.build
```diff
@@ -311,7 +311,7 @@ if get_option('build_backends')
       deps += [ opencl_framework ]
       has_opencl = true
 
-  elif opencl_lib.found()
+  elif opencl_lib.found() and cc.has_header('CL/opencl.h', args: '-I' + get_option('opencl_include')[0])
```
ddobbelaere: Maybe make it also resilient to future additions of include directories, in the spirit of https://github.com/LeelaChessZero/lc0/blob/06748a4a35b297164e36400a576bc64031fa0415/meson.build#L222
Just an idea, I imagine someone editing `meson_options.txt` later won't recall or even know that this check only considers the first include directory.

borg323: We use the same trick in <https://github.com/LeelaChessZero/lc0/blob/master/meson.build#L160>. 
Maybe we should change the option type from `array` to `string` when only a single argument is supported.

borg323: I did make `opencl_include` a string option with the latest commit. There are a few more candidates but I want to consider them separately

meson_options.txt
```diff
@@ -14,8 +14,8 @@ option('openblas_include',
        description: 'Paths to openblas include directories')
 
 option('opencl_include',
-       type: 'array',
-       value: ['/usr/include/'],
+       type: 'string',
+       value: '/usr/include/',
        description: 'Paths to OpenCL include directories')
```
ddobbelaere: This is really nitpicking, but now it's "Path" :)

borg323: If only every review comment was this easy..

borg323: ...and if only I was more careful when writing patches!

PR #922, opened 2019-08-12
Generate more descriptive zip file names for releases

borg323:
With this the generated zip files will be:
 - lc0-vx.y.z-windows-gpu-nvidia-cuda.zip
 - lc0-vx.y.z-windows-gpu-opencl.zip
 - lc0-vx.y.z-windows-cpu-openblas.zip


MelleKoning:
For me these names are more descriptive and thus an improvement

PR #923, opened 2019-08-13
'-Dblas=false' compiles lc0 for OpenCL without MKL dependency

Naphthalin:
When compiling lc0 with OpenCL for AMD GPUs under Windows, the current ``build-cl.cmd`` creates a lc0.exe binary with an explicit MKL dependency unlike the distributed lc0-opencl binaries which fails to launch without the MKL dll's. I added a ``build-cl-opencl-only.cmd`` file for that purpose.

The flag resolves this, the lc0 binary runs when having the files included in the lc0-opencl.zip 

Additionally, ``meson.py`` needs to be changed to ``meson`` only.


borg323:
There is no need for a separate script for this, just make the changes to `build-cl.cmd` and add a comment above the meson line to "change to -Dblas=true to also build the blas backend with mkl". I expect with these changes anybody who got to this point can figure out the rest - this is the build-cl script after all...

Naphthalin:
Deleted the new script and put the changes into the existing one, so only one file is changed. However it displays 3 commits for these 2 lines now which is a bit confusing :)

borg323:
No need to worry about it, we squash PRs to a single commit when merging.

PR #925, opened 2019-08-17
Transform Q into logit space when determining Q+U best child

AlexisOlson:
**Edit:**
The idea here is that when Q is near +1 or -1, the U term dominates the search since a small change in Q corresponds to a large change in the chance of winning/losing.

This PR transforms Q into logit space ([logit is the function that converts log-odd to probability](https://en.wikipedia.org/wiki/Logistic_regression)) before adding the U term.

In sumary, instead of `Q + U`, I use `logit(Q) + U`.

See also @Naphthalin's explanation on [Discord]( https://discordapp.com/channels/425419482568196106/539960268982059008/614432775164002351)

> The problem it tries to help with: if near Q=1 or Q=-1 (so winning or losing), the search goes super wide because the Q differences are very small. Between 0.97 and 0.98, for example, this wide search is a problem for accurate evaluation of sharper lines. Assume there is a move that in depth 4 increases Q from 0.9 to 0.99. However, the highest policy move there loses momentum, reducing Q from 0.9 to 0.6.
> 
> To get an accurate eval of >0.95 for this line the better line has to get at least 5x the visits of the other line. If we approach 1.0, this problem increases: if a move would increase Q from 0.98 to 0.99 while an alternative drops it to 0.9, it would need 10x the nodes. If the cpuct would decrease once we approach Q=1.0 it would help with this problem because the PUCT search would spend more visits on the top choices, therefore giving a more accurate eval.
> 
> So the originally proposed idea was to effectively reduce cpuct at higher Q. However, especially with including fpu and for symmetry reasons it is better to replace Q+U by something which behaves as an effective Q, which means being always in [-1,1] where -1 and 1 mean definite results. This is done by doing the addition of Q and U in the logit space which is in our case the proper way of thinking about odds of winning/losing. To do this, Q is transformed back into logit (which is what the NN calculates somewhere), adding the exploration term and FPU, and transforming it back by a tanh to get a winrate. The calculation can be simplified a bit so tanh needs to be called only once.
>
> The expected outcome is no different behavior near Q=0 while being more selective near Q=1 and Q=-1 which (hopefully) favors tactical lines where lc0 can already see the progress over staying in a comfortable position because the good line has a higher weight than it would have now. (Also one important detail: one might think about doing the averaging of Q values also in the logit space. This, however, is most likely to be avoided since the Q values are about statistics, and otherwise having a single 1.0 eval somewhere in the tree would destroy the Q.)

---

An example at extreme Q (PR925 on bottom):
```4Rbk1/5p2/p2Q2p1/7p/6N1/2p4P/PP3PPK/8 w - - 0 1```
Mate in 2

---
**Original:**

Scaling factor for `U` term: `2 * Q / ln( (1 + Q) / (1 - Q) )`

LogitScaling
Graph from [Desmos](https://www.desmos.com/calculator/mncok1dmp3)

Cpuct probably needs to be re-tuned along with this change.

It may or may not make more sense to use the root/parent Q value instead.

### DIFF
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -87,72 +87,6 @@ class Edge {
   float GetP() const;
   void SetP(float val);
 
-  void MakeTerminal(GameResult result);
-
-  // Sets flags for certainty, trigger of certainty and result by GameResult.
-  void MakeCertain(CertaintyResult certaintyresult);
-
-  // Sets flags for certainty, trigger of certainty and result (by Q).
-  void MakeCertain(int q, CertaintyTrigger trigger);
-
-  // Sets edge-Q: win = 1; draw = 0; loss = -1.
-  void SetEQ(int eq);
-  // Clears Certainty but keeps bounds.
-  void ClearCertaintyState() {
-    if (certainty_state_ & kCertainMask) {
-      certainty_state_ = 0;
-    } else {
-      certainty_state_ &= kClearKeepBounds;
-      }
-  };
-  // Sets (U)pper and (L)ower bounds.
-  void UBound(int eq) {
-    certainty_state_ |= kUpperBound;
-    SetEQ(eq);
-  };
-  void LBound(int eq) {
-    certainty_state_ |= kLowerBound;
-    SetEQ(eq);
-  };
-
-  // Returns true if only upper bounded.
-  bool IsOnlyUBounded() {
-    return (certainty_state_ & kUpperBound) &&
-           !(certainty_state_ & kLowerBound);
-  };
-
-  // Returns true if only lower bounded.
-  bool IsOnlyLBounded() {
-    return !(certainty_state_ & kUpperBound) &&
-           (certainty_state_ & kLowerBound);
-  };
-  bool IsTerminal() const { return certainty_state_ & kTerminalMask; };
-  bool IsCertain() const { return certainty_state_ & kCertainMask; };
-  bool IsCertainWin() const {
-    return ((certainty_state_ & kCertainMask) &&
-            (certainty_state_ & kGameResultWin));
-  };
-  bool IsCertainLoss() const {
-    return ((certainty_state_ & kCertainMask) &&
-            (certainty_state_ & kGameResultLoss));
-  };
-  bool IsCertainDraw() const {
-    return ((certainty_state_ & kCertainMask) &&
-            !(certainty_state_ & ~kGameResultClear));
-  };
-
-  // Returns true if certainty proof based on a TB hit.
-  bool IsPropagatedTBHit() const { return certainty_state_ & kTBHit; };
-
-  // Query bounds.
-  bool IsUBounded() const { return certainty_state_ & kUpperBound; };
-  bool IsLBounded() const { return certainty_state_ & kLowerBound; };
-
-  // Return all stats flags.
-  uint8_t GetCertaintyState() const { return certainty_state_; };
-  // Returns the edges Q
-  int GetEQ() const;
-
   // Debug information about the edge.
   std::string DebugString() const;
 
@@ -168,28 +102,6 @@ class Edge {
   // network; compressed to a 16 bit format (5 bits exp, 11 bits significand).
   uint16_t p_ = 0;
 
-  // Certainty Propagation attaches a number of flags to each edge.
-  // kTerminalMask    -> edge is terminal.
-  // kCertainMask     -> edge is certain.
-  // kUpperBound      -> edge is upper bounded.
-  // kLowerBound      -> edge is lower bounded.
-  // kTBHit           -> edge is certain because of a TB hit.
-  // kTwoFold         -> edge is certain because of a two-fold.
-  // kGameResultWin   -> edge is a proven win.
-  // kGameResultLoss  -> edge is a proven loss.
-  uint8_t certainty_state_ = 0;
-  enum Masks : uint8_t {
-    kTerminalMask = 0b00000001,
-    kCertainMask = 0b00000010,
-    kUpperBound = 0b00000100,
-    kLowerBound = 0b00001000,
-    kTBHit = 0b00010000,
-    kTwoFold = 0b00100000,
-    kGameResultWin = 0b01000000,
-    kGameResultLoss = 0b10000000,
-    kGameResultClear = 0b00111111,
-    kClearKeepBounds = 0b00001100,
-  };
   friend class EdgeList;
 };
 
@@ -233,11 +145,6 @@ class Node {
   // Returns whether a node has children.
   bool HasChildren() const { return edges_; }
 
-  // Recalculate n_ from real children visits.
-  // This is needed if a node was proved to be certain in a prior
-  // search and later gets to be root of search.
-  void RecomputeNfromChildren();
-
   // Returns sum of policy priors which have had at least one playout.
   float GetVisitedPolicy() const;
   uint32_t GetN() const { return n_; }
@@ -245,74 +152,19 @@ class Node {
   uint32_t GetChildrenVisits() const { return n_ > 0 ? n_ - 1 : 0; }
   // Returns n = n_if_flight.
   int GetNStarted() const { return n_ + n_in_flight_; }
-  // Returns node eval, i.e. average subtree V for non-certain node and -1/0/1
-  // for certain nodes.
-  float GetQ() const;
+  // Returns node eval, i.e. average subtree V for non-terminal node and -1/0/1
+  // for terminal nodes.
+  float GetQ() const { return q_; }
   float GetD() const { return d_; }
 
   // Returns whether the node is known to be draw/lose/win.
-  bool IsTerminal() const {
-    return GetOwnEdge() ? GetOwnEdge()->IsTerminal() : false;
-  }
-  bool IsCertain() const {
-    return GetOwnEdge() ? GetOwnEdge()->IsCertain() : false;
-  }
-
-  // Sets bounds.
-  void UBound(int eq) const {
-    if (GetOwnEdge()) GetOwnEdge()->UBound(eq);
-  }
-  void LBound(int eq) const {
-    if (GetOwnEdge()) GetOwnEdge()->LBound(eq);
-  }
-  // Queries bounds.
-
-  bool IsBounded() const {
-    return GetOwnEdge()
-               ? GetOwnEdge()->IsLBounded() || GetOwnEdge()->IsUBounded()
-               : false;
-  }
-  bool IsOnlyUBounded() const {
-    return GetOwnEdge() ? GetOwnEdge()->IsOnlyUBounded() : false;
-  }
+  bool IsTerminal() const { return is_terminal_; }
   uint16_t GetNumEdges() const { return edges_.size(); }
 
-  // Makes the node terminal or certain and sets its score.
-  void MakeTerminal(GameResult result) {
-    if (GetOwnEdge()) GetOwnEdge()->MakeTerminal(result);
-    if (result == GameResult::DRAW) {
-      q_ = 0.0f;
-      d_ = 1.0f;
-    } else if (result == GameResult::WHITE_WON) {
-      q_ = 1.0f;
-      d_ = 0.0f;
-    } else {
-      q_ = -1.0f;
-      d_ = 0.0f;
-    }
-  }
-  void MakeCertain(CertaintyResult certaintyresult) {
-    if (GetOwnEdge()) GetOwnEdge()->MakeCertain(certaintyresult);
-    if (certaintyresult.gameresult == GameResult::DRAW) {
-      q_ = 0.0f;
-      d_ = 1.0f;
-    } else if (certaintyresult.gameresult == GameResult::WHITE_WON) {
-      q_ = 1.0f;
-      d_ = 0.0f;
-    } else {
-      q_ = -1.0f;
-      d_ = 0.0f;
-    }
-  }
-  void MakeCertain(int q, CertaintyTrigger trigger) {
-    if (GetOwnEdge()) GetOwnEdge()->MakeCertain(q, trigger);
-    q_ = q;
-    if (q == 0) {
-      d_ = 1.0f;
-    } else {
-      d_ = 0.0f;
-    }
-  }
+  // Makes the node terminal and sets it's score.
+  void MakeTerminal(GameResult result);
+  // Makes the node not terminal and updates its visits.
+  void MakeNotTerminal();
 
   // If this node is not in the process of being expanded by another thread
   // (which can happen only if n==0 and n-in-flight==1), mark the node as
@@ -351,6 +203,10 @@ class Node {
   int GetRemainingCacheVisits() {
     return best_child_cache_in_flight_limit_ - n_in_flight_;
   }
+
+  // Calculates the full depth if new depth is larger, updates it, returns
+  // in depth parameter, and returns true if it was indeed updated.
+  bool UpdateFullDepth(uint16_t* depth);
 
   V4TrainingData GetV4TrainingData(GameResult result,
                                    const PositionHistory& history,
@@ -398,8 +254,7 @@ class Node {
   EdgeList edges_;
 
   // 8 byte fields.
-  // Pointer to a parent node. nullptr for the root of tree,
-  // Note: root of tree might not be search->root_node_.
+  // Pointer to a parent node. nullptr for the root.
   Node* parent_ = nullptr;
   // Pointer to a first child. nullptr for a leaf node.
   std::unique_ptr<Node> child_;
@@ -431,8 +286,12 @@ class Node {
   uint32_t best_child_cache_in_flight_limit_ = 0;
 
   // 2 byte fields.
-  // Index of this node in parent's edge list.
+  // Index of this node is parent's edge list.
   uint16_t index_;
+
+  // 1 byte fields.
+  // Whether or not this node end game (with a winning of either sides or draw).
+  bool is_terminal_ = false;
 
   // TODO(mooskagh) Unfriend NodeTree.
   friend class NodeTree;
@@ -484,35 +343,19 @@ class EdgeAndNode {
   float GetD() const {
     return (node_ && node_->GetN() > 0) ? node_->GetD() : 0.0f;
   }
-
-  // Gets the edge's Q, if edge is certain this
-  // is the proven game result (-1, 0, +1).
-  int GetEQ() const { return edge_->GetEQ(); }
-
-  // N-related getters, from node (if exists).
+  // N-related getters, from Node (if exists).
   uint32_t GetN() const { return node_ ? node_->GetN() : 0; }
   int GetNStarted() const { return node_ ? node_->GetNStarted() : 0; }
   uint32_t GetNInFlight() const { return node_ ? node_->GetNInFlight() : 0; }
+
+  // Whether the node is known to be terminal.
+  bool IsTerminal() const { return node_ ? node_->IsTerminal() : false; }
 
   // Edge related getters.
   float GetP() const { return edge_->GetP(); }
   Move GetMove(bool flip = false) const {
     return edge_ ? edge_->GetMove(flip) : Move();
   }
-  bool IsTerminal() const { return edge_->IsTerminal(); }
-  bool IsCertain() const { return edge_->IsCertain(); }
-  bool IsCertainWin() const { return edge_->IsCertainWin(); }
-
-  // Queries bounds.
-  bool IsUBounded() const { return edge_->IsUBounded(); }
-  bool IsLBounded() const { return edge_->IsLBounded(); }
-
-  // Sets bounds.
-  void UBound(int eq) { edge_->UBound(eq); }
-  void LBound(int eq) { edge_->LBound(eq); }
-
-  // Queries if edge's certainty is based on a TB hit.
-  bool IsPropagatedTBHit() const { return edge_->IsPropagatedTBHit(); }
 
   // Returns U = numerator * p / N.
   // Passed numerator is expected to be equal to (cpuct * sqrt(N[parent])).
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -87,6 +87,72 @@ class Edge {
   float GetP() const;
   void SetP(float val);
 
+  void MakeTerminal(GameResult result);
+
+  // Sets flags for certainty, trigger of certainty and result by GameResult.
+  void MakeCertain(CertaintyResult certaintyresult);
+
+  // Sets flags for certainty, trigger of certainty and result (by Q).
+  void MakeCertain(int q, CertaintyTrigger trigger);
+
+  // Sets edge-Q: win = 1; draw = 0; loss = -1.
+  void SetEQ(int eq);
+  // Clears Certainty but keeps bounds.
+  void ClearCertaintyState() {
+    if (certainty_state_ & kCertainMask) {
+      certainty_state_ = 0;
+    } else {
+      certainty_state_ &= kClearKeepBounds;
+      }
+  };
+  // Sets (U)pper and (L)ower bounds.
+  void UBound(int eq) {
+    certainty_state_ |= kUpperBound;
+    SetEQ(eq);
+  };
+  void LBound(int eq) {
+    certainty_state_ |= kLowerBound;
+    SetEQ(eq);
+  };
+
+  // Returns true if only upper bounded.
+  bool IsOnlyUBounded() {
+    return (certainty_state_ & kUpperBound) &&
+           !(certainty_state_ & kLowerBound);
+  };
+
+  // Returns true if only lower bounded.
+  bool IsOnlyLBounded() {
+    return !(certainty_state_ & kUpperBound) &&
+           (certainty_state_ & kLowerBound);
+  };
+  bool IsTerminal() const { return certainty_state_ & kTerminalMask; };
+  bool IsCertain() const { return certainty_state_ & kCertainMask; };
+  bool IsCertainWin() const {
+    return ((certainty_state_ & kCertainMask) &&
+            (certainty_state_ & kGameResultWin));
+  };
+  bool IsCertainLoss() const {
+    return ((certainty_state_ & kCertainMask) &&
+            (certainty_state_ & kGameResultLoss));
+  };
+  bool IsCertainDraw() const {
+    return ((certainty_state_ & kCertainMask) &&
+            !(certainty_state_ & ~kGameResultClear));
+  };
+
+  // Returns true if certainty proof based on a TB hit.
+  bool IsPropagatedTBHit() const { return certainty_state_ & kTBHit; };
+
+  // Query bounds.
+  bool IsUBounded() const { return certainty_state_ & kUpperBound; };
+  bool IsLBounded() const { return certainty_state_ & kLowerBound; };
+
+  // Return all stats flags.
+  uint8_t GetCertaintyState() const { return certainty_state_; };
+  // Returns the edges Q
+  int GetEQ() const;
+
   // Debug information about the edge.
   std::string DebugString() const;
 
@@ -102,6 +168,28 @@ class Edge {
   // network; compressed to a 16 bit format (5 bits exp, 11 bits significand).
   uint16_t p_ = 0;
 
+  // Certainty Propagation attaches a number of flags to each edge.
+  // kTerminalMask    -> edge is terminal.
+  // kCertainMask     -> edge is certain.
+  // kUpperBound      -> edge is upper bounded.
+  // kLowerBound      -> edge is lower bounded.
+  // kTBHit           -> edge is certain because of a TB hit.
+  // kTwoFold         -> edge is certain because of a two-fold.
+  // kGameResultWin   -> edge is a proven win.
+  // kGameResultLoss  -> edge is a proven loss.
+  uint8_t certainty_state_ = 0;
+  enum Masks : uint8_t {
+    kTerminalMask = 0b00000001,
+    kCertainMask = 0b00000010,
+    kUpperBound = 0b00000100,
+    kLowerBound = 0b00001000,
+    kTBHit = 0b00010000,
+    kTwoFold = 0b00100000,
+    kGameResultWin = 0b01000000,
+    kGameResultLoss = 0b10000000,
+    kGameResultClear = 0b00111111,
+    kClearKeepBounds = 0b00001100,
+  };
   friend class EdgeList;
 };
 
@@ -145,6 +233,11 @@ class Node {
   // Returns whether a node has children.
   bool HasChildren() const { return edges_; }
 
+  // Recalculate n_ from real children visits.
+  // This is needed if a node was proved to be certain in a prior
+  // search and later gets to be root of search.
+  void RecomputeNfromChildren();
+
   // Returns sum of policy priors which have had at least one playout.
   float GetVisitedPolicy() const;
   uint32_t GetN() const { return n_; }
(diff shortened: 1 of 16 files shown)

### CODE REVIEW

File: src/mcts/search.cc [hunk 1]
```diff
@@ -956,7 +956,7 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend(
         ++possible_moves;
       }
       const float Q = child.GetQ(fpu);
-      const float score = child.GetU(puct_mult) + Q;
+      const float score = child.GetU(puct_mult) * (2*Q/log((1 + Q)/(1 - Q))) + Q;
```
ddobbelaere: Mathematically, the formula is correct and gives a correct continuous function in [-1,1] if you consider the limits for Q going to -1, 0 and +1.

Translating the formula directly into C++ leads to trouble:
- for Q = 0, we get 0/0 (NaN).
- for Q = -1, we get log(0), which is undefined according to http://www.cplusplus.com/reference/cmath/log/ (may lead to a pole error depending on library implementation).

These numerical aspects aside, you may consider a faster log from the `fastmath.h` which will be considerably faster and will lead to an nps boost on fast backends.

kiudee:
Remember to also change https://github.com/LeelaChessZero/lc0/blob/033cce79a7e39d1c7424524d48e5040be91af444/src/mcts/search.cc#L249 to be able to debug the scaled U values.

src/mcts/search.cc
oscardssmith: For now at least, the speed doesn't matter too much. The numerical issues do though

AlexisOlson: The undefined values did occur to me, but I'm not very fluent in C++, so if anyone wants to propose an alternative formulation to resolve them, it would be much appreciated.

oscardssmith: For now, 2 if statements is probably the easiest

borg323:
How important is the exact shape? Using `1-Q^4` has the same values at -1, 0 ans 1.
curve

AlexisOlson:
@borg323 The behavior near +/- 1 is very different for `1-Q^4`. The [logit function](https://en.wikipedia.org/wiki/Logit) is not an arbitrary choice but rather [the function that converts log-odds to probability](https://en.wikipedia.org/wiki/Logistic_regression), which I think is exactly what we want.

Hopefully, someone with a stronger stats background can chime in.

jhorthos:
I think the logit is probably the right choice. Can't you just use a tiny epsilon to avoid 0?

Naphthalin:
I agree with jhorthos, the logit is definitely the right choice there and should be considered over all alternatives.

However, I find it not as pleasing to scale the U value and would prefer to simply transform the Q into logit(Q), so instead of
``Q + U * 2 * Q / ln( (1 + Q) / (1 - Q) )``
you would have
``0.5*ln( (1 + Q) / (1 - Q) ) + U``
which is the formula divided by that scaling factor.

borg323:
However I think my question is still valid: How important is the exact shape? And to elaborate, is it important over the whole range, or e.g closer to zero?

Naphthalin:
The important behavior implications happen around Q=-1 and Q=1 and not around Q=0. The exact shape is important as in having 0 (or infinity) slope near -1 and 1 to be able to differentiate between let's say 95% win, 99% win an 99.9% win.

src/mcts/search.cc
```diff
@@ -956,7 +956,14 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend(
         ++possible_moves;
       }
       const float Q = child.GetQ(fpu);
-      const float score = child.GetU(puct_mult) + Q;
+      float score = child.GetU(puct_mult);
+      if (Q == 0) {
+        continue;
```
borg323: This doesn't look right.

AlexisOlson: For `Q=0`, `Q+U = U`. Am I missing something?

Naphthalin: Yes, this should just be an empty bracket.

borg323: Or just change the last `else` into `else if (Q != 0)` and remove the first block.

jhorthos:
I agree that the behavior around -1 and +1 is important and near 0 is the least, but I am also interested in the behavior in between. The region where she has occasionally thrown a win for example, possibly because of a rarer milder issue of this sort.

jhorthos:
If we reshaped Q rather than U would that reduce the need to adjust Cpuct?

AlexisOlson:
@jhorthos I don't think so. The CPUCT is all about the balance between Q and U. Whether you change one or the other, you'd still probably need to re-tune since the balance has changed.

jhorthos:
Yeah, you are right. Well if we do short TC tests I can run a CLOP to get a new best Cpuct. The other Cpuct values are nearly impossible to CLOP because they require tons of nodes.

MelleKoning:
Any example FEN positions with search results available that will show Q value improvements to support this change?

AlexisOlson:
@MelleKoning Good question. I'll try to find some relevant ones from position testing logs and see how this change would affect them.

AlexisOlson:
> Any example FEN positions with search results available that will show Q value improvements to support this change?

@MelleKoning,

Here's an interesting FEN:
7R/pb6/1p3P2/5P2/4r1P1/k1p2K2/2P5/8 w - -

**Standard Leela with J8-610 weights after 135k nodes**
Lc0726

**Logit Leela with J8-610 weights after 104k nodes**
Logit726

Standard Leela doesn't show Rb8 as the highest score until after about 1.2 million nodes and isn't the most visited until 2.2 million nodes. Logit Leela settles on Rb8 within the first couple thousand nodes. Rb8 agrees with the chosen move in the ICCF game the position is sourced from.

---

**Standard Leela after 2.2 million nodes**
Lc0726-2 2

**Logit Leela after 1.1 million nodes**

Logit726-1 1

---

Note that Logit Leela likely needs a higher cpuct value since her Q is relatively stronger.

src/mcts/search.cc
```diff
@@ -956,7 +956,12 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend(
         ++possible_moves;
       }
       const float Q = child.GetQ(fpu);
-      const float score = child.GetU(puct_mult) + Q;
+      float score = child.GetU(puct_mult);
+      if (Q == 1 || Q == -1) {
+        score = Q;
+      } else if (Q != 0) {
+        score = tanh ( score + 0.5 * log ( (1 + Q) / (1 - Q) ) );
```
Sopel97: just wanted to point out that function for computing score is equivalent to
https://www.wolframalpha.com/input/?i=(-1+-+(e%5E(2+y)+(1+%2B+x))%2F(-1+%2B+x))%2F(1+-+(e%5E(2+y)+(1+%2B+x))%2F(-1+%2B+x))+for+x+in+-1+to+1
with y:=score; x:=Q

I'll leave this if there is ever need to optimize this.
I have not compared the numerical stability, though the division is probably the biggest issue here and it has to stay

jhorthos:
Nice Alexis - if things look solid please send me the binary and I can run a fixed node CLOP (that way if you squeeze a bit more efficiency it won't matter). - James

jhorthos:
The evaluation of Leela logitU is also more in line with Stockfish, which also picks Rb8 at 53 billion nodes.

AlexisOlson:
Note: I've discovered that
```tanh(0.5 log((Q + 1)/(1 - Q)) + 0.5 log((U + 1)/(1 - U)))```
simplifies to
```(Q + U) / (Q * U + 1)```

I'm testing this approach of applying logit to both terms on [a separate branch](https://github.com/AlexisOlson/lc0/tree/patch-2).

**Comment by Naphthalin** [comment]
Yeah, addition theorems are nice :) However, U is not something which is supposed to be transformed by logit because it doesn't live naturally in [-1,1] or anything. You could however use your simplification to write 
``tanh(U + logit(Q) ) = ( tanh(U) + Q ) / ( tanh(U) * Q + 1)``
so you only have to calculate ``tanh(U)`` once and get the former Q+U values to actually lie in [-1,1].

However, you still have to take care of fpu.

File: src/mcts/search.cc (hunk 3 above)
Comment (AlexisOlson) [reply src/mcts/search.cc]: Using `tanh(x+y) = (tanh x + tanh y)/(tanh x tanh y + 1)`, we can simplify to

`score = ( tanh( score ) + Q) / ( tanh( score ) * Q + 1 )`

which only needs one `tanh` evaluation.

File: src/mcts/search.cc [hunk 4]
```diff
@@ -956,7 +956,11 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend(
         ++possible_moves;
       }
       const float Q = child.GetQ(fpu);
-      const float score = child.GetU(puct_mult) + Q;
+      const float tanhU = tanh(child.GetU(puct_mult));
+      float score = tanhU + Q;
+      if (tanhU * Q != -1) {
+        score = score / (tanhU * Q + 1);
```
ddobbelaere: Division by zero is well-defined, in the sense that it leads to +inf or -inf, depending on the sign of the numerator, so no need for the check. Moreover, abs(tanhU) will never attain 1 for all practical values of U, which would be a necessary condition for the product to be -1.

AlexisOlson: Thanks. I was thinking U*Q could be -1, but tanhU*Q shouldn't have that problem.

src/mcts/search.cc
Sopel97: tanh is painfully slow
http://quick-bench.com/sa29JAi6ODghRHOa8mJcovD3FEY

but actually the more I look at my formula the more I worry about numerical stability
would need to be tested

jkormu:
There might be something wrong with the latest commits. From start position, when Q is still in reasonable range, the visit distribution and tree shape is completely different compared to standard lc0. 

With earlier commit [3e62178](https://github.com/LeelaChessZero/lc0/pull/925/commits/3e62178db5ec98974ad299c2a897d9c3bd0f0070) the visit distribution and tree shape is very similar for start position (as one would expect) and shows clear improvement when Q is close to 1 or -1. In these cases the vanilla lc0 searches very wide and visit distribution is flat. With 3e62178 the search has clear favorite moves that are searched more than others. Looks promising.

AlexisOlson:
@jkormu It may have been the `fpu` term I had a bit off. I've made an adjustment.

src/mcts/search.cc
AlexisOlson: I found a possible faster implementation for tanh here:
<https://www.ipol.im/pub/art/2015/137/article.pdf>

I'm not going to worry about optimizing it right now though

src/mcts/search.cc
```diff
@@ -956,7 +956,7 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend(
         ++possible_moves;
       }
       const float Q = child.GetQ(0);
-      const float tanhU = tanh(child.GetU(puct_mult)) - fpu;
+      const float tanhU = tanh(child.GetU(puct_mult)) - (child.GetQ(fpu) - Q);
```
Naphthalin: IMO the fpu term should be treated as if it was a U value so it should be inside the tanh bracket?

AlexisOlson: I think you're right. I'll move that parenthesis.

fohristiwhirl:
Should one be concerned that the N/s is so much lower?

Naphthalin:
> Should one be concerned that the N/s is so much lower? (Or maybe recent changes already fixed that?)

What do you mean by "so much lower"? I think right now the focus should be to produce evidence that A) the behavior around Q=0 isn't altered and B) the behavior at extreme Q is actually improved.

If we succeed with both we should care about speed, and if we don't get the change without dropping nps we can still think about e.g. using the modification at extreme Q only where the game is practically won but the current MCTS would drag the game.

PowerMove72:
`const float tanhU = tanh(child.GetU(puct_mult)) + (child.GetQ(fpu) - Q);`
I think we have to sum here...

jkormu:
[d62795b](https://github.com/LeelaChessZero/lc0/pull/925/commits/d62795b16fb03270e62562150d3635044c450d28) seems also broken as it results in super shallow search with huge branching factor for start position.

image

jkormu:
@PowerMove72 suggestion seems to fix it. Almost identical search with vanilla lc0 from start position.

AlexisOlson:
@fohristiwhirl @Naphthalin It does appear to be slower even on the latest commit so far (very roughly half the NPS). I may try removing the tanh function to test if that's the bottleneck and then revert back.

AlexisOlson:
@PowerMove72, thanks for that! We'll get it right eventually. (I meant to tag you in the commit message.)

A very broad and shallow search may be the reason for diminished NPS too.

PowerMove72:
@AlexisOlson i am trying some changes starting from your idea.  By the way,  i get the best results with 
`float Q = child.GetQ(fpu);`
`const float tanhU = tanh(child.GetU(puct_mult));`
`const float score = (tanhU + Q) / (tanhU * Q + 1);`
It is simple, elegant and no NPS slow down.

Naphthalin:
@PowerMove72 the problem with this is that in bad situations (Q nearly -1) the GetQ(fpu) can be negative below -1 which is a problem for the formula.

AlexisOlson:
@Naphthalin To be clear, you're saying that we still need the `fpu` term, not that the sign change correction was wrong?

Naphthalin:
@AlexisOlson I am saying that ``child.GetQ(fpu)`` can produce results outside of [-1,1], that's why it shouldn't be used as @PowerMove72 is using it. Your code after the sign correction produces the desired outcome.

borg323:
For faster execution, you can use ~~fhtagn~~FastTanh():
```
inline float FastTanh(const float a) {
  float out = 2.886132f * a;
  int exp = floor(out);
  out -= exp;
  out = 1.0f + out * (0.6602962f + 0.33970374f * out);
  int32_t tmp;
  std::memcpy(&tmp, &out, sizeof(float));
  tmp += exp << 23;
  std::memcpy(&out, &tmp, sizeof(float));
  return (out - 1.0f) / (out + 1.0f);
}
```
The constants have been numerically tuned to minimize max absolute error:
FastTanh2

AlexisOlson:
Here's another example where this PR makes a bit difference:

3r4/1pq2pk1/5Nr1/3pp1P1/p2b4/5Q2/KP4R1/7R w - - 0 0

position_id_33567

Vanilla Leela on top, commit https://github.com/LeelaChessZero/lc0/pull/925/commits/b006698d26966bb6a2f55dd1e0e52a32904e3183 on the bottom. Qf5 agrees with the ICCF move.

Naphthalin:
I'm far from being able to judge the first two examples, but the last one highlights exactly how the squashed evals near -1 and 1 confuse UCT.

Did you also stumble over examples where Vanilleela finds the right move but Logit Leela is trapped in a tactical line?

oscardssmith:
Honestly error of .0015 seems pretty darn negligible to me.

AlexisOlson:
Analyzing CLOP results by @jhorthos (for 1000 npm), it appears that this PR handles higher CPUCT without degrading performance as much, which makes sense from a theoretical perspective.

CLOP

(Size of the bubbles are a function of the count of games in that CPUCT bucket.)

jkormu:
The hope is that this pr reduces trolling and won games would be finished faster.
Tested this by running 3.8k games lc0 vs pr925 with 800 nodes, --minibatch-size=32 (rest as default) and comparing average game length of games pr925 won vs games lc0 won.  

game result | Avg game length (plys) | Game count
--- | --- | ---
pr925 win | 138.45 | 2579
lc0 win | 264.09  | 685
draw | 127.95 | 543
all games | 149.26 | 3807

The average game length of games where lc0 won is almost double of the games where pr925 won.
This indicates that either
1. pr925 is able finish won games faster than lc0
or
2. pr925 is able to delay loss better than lc0
or 
3. both

I would guess it is the option 3. 
To distinguish the effect of 1. and 2. one would need to repeat the test but run lc0 and pr925 vs sf (or some other engine) and compare game lengths of lc0 loss vs pr925 loss and lc0 win vs pr925win.

AlexisOlson:
@jkormu It may also be useful to compare total moves for self-play games too. Ideally, PR925 would train faster.

oscardssmith:
one potential downside with this is it might require higher resign threshold if we want training on imbalanced positions.

zz4032:
I'm getting 20-30% less nps with the latest commit. Is this expected?

MelleKoning:
@jkormu to reduce trolling would prefer focus on certainty propagation like #889

jkormu:
@MelleKoning Why not have both? I think the consensus is that cp reduces the trolling and shortens the games. It also seems that this pr has the same ability. What makes you think that these two together wouldn't be better than cp alone? 
Also, this pr might have better changes to get merged than cp.

MelleKoning:
@jkormu nothing makes me think the two together wouldn't be better. 
It is certain cp reduces trolling and it seems this one also improves choices in certain (pun intended) positions.

jkormu:
The search speed seems to be about equal to vanilla lc0 when the absolute score is between 0.0-2.0. The speed starts to drop once the absolute score goes above 2.0. Following plot show the average move time for 800 nodes search per score interval for pr925 vs lc0 based on jhorthos' pgn from discord "PR925.2_test-match.1.pgn.gz". Used settings:  SmartPruningFactor=1.0 option.Backend="cudnn-fp16", others default. There are 3k games in the pgn.

image

oscardssmith:
Is this caused by narrowing search making batching harder?

jhorthos:
One possible issue there is that the CP score will have to be recalibrated for the PR - it seems to generally be lower CP for the same Q.

oscardssmith:
That seems wrong. Shouldn't more focused search yours higher evals?

jhorthos:
Well maybe I should say they seem often to be somewhat different. This is highly anecdotal from look at a few games in Nibbler.

AlexisOlson:
From what I've seen in nibbler, this PR gives significantly larger evals away from `Q = 0` more in line with Stockfish, but that's purely anecdotal.

jhorthos:
Yeah, I saw that too - bigger away from Q=0 and smaller near Q=0. Also anecdotal.

AlexisOlson:
Confirming @jkormu's observation, based on a PGN of [10,000 games played at 1000 nodes/move](https://discordapp.com/channels/425419482568196106/530486338236055583/614835620543594527), I get this.

|Eval| vs NPS Chart

As noted in the #test-results post, the settings for `cpuct` and`fpu` were significantly different (which can influence speed), so judge the shape of the curve more than the absolute NPS values.

jkormu:
Seems like only part of the slow down can be explained due to more focused search near extreme Q. The majority of slow down seems to be due to messed up GetVisitsToReachU formula in node.h. According to Tilps (from discord) this estimates "how long until the best child will change assuming everything is a collision". He also states that if the score equation is changed and GetVisitsToReachU is not updated then "it'll mess things up". #918 had also initially nps slow down until GetVisitsToReachU was updated.

Unfortunately, I have no idea how to update the formula to address the changes done in this pr. Maybe @kiudee could give some tips how he approached this. 

However, I have caveman "fix" that can be used to show that GetVisitsToReachU is the most likely culprit. As the nps slow down seems to happen only after the score goes above 2.0 (cp-scale), I assume the search of vanilla lc0 and pr925 are quite similar below this limit. Thus, the GetVisitsToReachU-formula is approximately correct in this range. Above score 2.0 the searched are probably too different for GetVisitsToReachU to be accurate anymore. As a "fix" let's just "disable" this estimation when absolute score of the root goes above 2.0. Here by disabling, I mean that instead of trying to estimate nodes required to reach U, lets just use value 1 instead. To my understanding this just disables the speed optimization that GetVisitsToReachU is supposed to provide.

This can be done by replacing following line in search.cc

`int estimated_visits_to_change_best = best_edge.GetVisitsToReachU(second_best, puct_mult, fpu);`

with the line

`int estimated_visits_to_change_best = std::abs(search_->root_node_->GetQ())>0.6752521315 ? 1 : best_edge.GetVisitsToReachU(second_best, puct_mult, fpu);`


Following graph shows the nps as a function of score based on 300 games lc0 vs lc0_p925 vs lc0_pr925_modified with 800 fixed node, --minibatch-size=32, --backend=cudnn-fp16, --smart-pruning-factor=0.0 and rest as defaults. Each engine played total of 200 games.

image

The average nps:
lc0_pr925_modified: 15412 nps
lc0: 15769
lc0_pr925: 14577

Despite the small sample size, the results of the match indicates that this modification did not completely destroy the search:
```
   1 lc0_pr925_PartiallyDisabled_GetVisitsToReachU      12      30     200   51.7%   60.5%
   2 lc0_pr925                       2      31     200   50.2%   59.5%
   3 lc0                           -14      30     200   48.0%   62.0%
```

Based on these observations, it looks like GetVisitsToReachU is here to blame for the slower nps. I don't know enough about how it works to explain why partially "disabling" it works that well. Please feel free to replicate the results (I hope I did not mess up anything while testing this).

AlexisOlson:
@jkormu Nice observation. I'll see if I can rewrite the `GetVisitsToReachU` to be in line with the new formula. @Naphthalin, I may ask for some help with this if I get stuck.

jkormu:
Seems to be fixed. The nps is now on bar with the caveman fix. Also got slight evidence that Elo increased on fixed nodes.

After 2040 games on each engine, the nps graph of latest commit is very similar to my PoC caveman fix (triple checked I have correct engines here). 

image

```
   1 lc0_pr925_new                   7       9    2040   51.0%   65.2%
   2 lc0_pr925_old_modified          6       9    2040   50.9%   63.9%
   3 lc0_pr925_old                  -1       9    2040   49.9%   65.8%
   4 lc0                           -13       9    2040   48.2%   64.0%
```

Average nps:
lc0_pr925_old 14993.940480034002
lc0_pr925_new 15520.498800367164
lc0_pr925_old_modified 15561.549684598134
lc0 15652.325239488735

jkormu:
I think this pr is ready for serious testing to proof or disproof that it gains Elo on equal time control. 
I would suggest a test where we tune vanilla lc0 parameters for low nodes. I would prefer to use --minibatch-size = 32 so that batch size<<number of searched nodes and tune fpu-value, cpuct and maybe also policy temp. Also tune pr925. Then play high number of matches with tuned parameters and fast time controls.

Any thoughts or suggestions? Tuning will take quite some time. Would it be sufficient to tune just cpuct and fpu-value or would we also need policy-softmax-temp?

zz4032:
To give you a thought how that would look like with an RTX2080TI (with LD2 and 10000nodes/move):
```
MinibatchSize      TC         Nodes/MinibatchSize   Time for 5000 games
     128       2.22+0.037             78                    11h
      64       3.00+0.050            156                    14h
      32       4.56+0.076            313                    22h
```

zz4032:
**Gauntlet SF vs. Master / PR925 (both tuned)**
TC (LC0): 1.98s/game+0.033s/move (~3500 nodes/move with LD2)
MinibatchSize=32
Tuning results (8000 games each):
Master: CPuct=1.98, FpuValue=0.52
PR925:  CPuct=2.59, FpuValue=0.63
```
   # PLAYER             :  RATING  ERROR  LOS(%)   GAMES  DRAWS(%)  OppN
   1 stockfish          :     0.0   ----    84.0   10000      40.8     2
   2 lc0-tuned          :    -3.8    7.5    94.7    5000      41.5     1
   3 lc0_PR925-tuned    :   -12.3    7.3     ---    5000      40.1     1

White ELO advantage = 0.3
Draw rate (equal opponents) = 40.8 %
```

AlexisOlson:
Gah. I was trying to merge the other way on the last two commits. Please ignore.

Tilps:
Some of these code changes seem weird.  Passing fpu sometimes but not others - some things guarded by the flag, others not.
At a minimum I would suggest fixing things up so that the flag off case is a no-op - which is certainly not the case.

AlexisOlson:
@Tilps Does the last commit look better?

There's one extra GetQ() call that doesn't get used if the flag is off but I wasn't able to remove that successfully without creating other problems. It should be more readable now anyway.

If anyone with better C++ knowledge wants to refactor it. Please feel free.

Tilps:
The formula for displaying Q+U is not the same as the formula used for calculating best - and also the visits to reach U formula also doesn't match.
I'm unclear why fpu Q should not have logit applied - code would be much simpler if you didn't need to do that...
If it is essential that fpu doesn't have Q applied - it would be more straight forward to do 'if !enabled or child->GetN() == 0) - it would also avoid performing the tanh on 0 for no good reason.

AlexisOlson:
@Tilps Thanks for the suggestion for doing the `N == 0` case. That does look better. There's no `tanh` anymore but your point remains and we don't need to do a useless `FastLogit` eval now.

The issue with the fpu is that if it's relative, then it can fall outside [-1,1] as @Naphthalin [pointed out](https://github.com/LeelaChessZero/lc0/pull/925#issuecomment-523671329).

I've updated the formulas though I haven't handled the `N == 0` case for` GetVisitsToReachU`. It's not clear to me if that actually happens and if it is, what the right way to check call `N` in this context.

src/mcts/params.cc
```diff
@@ -192,6 +195,7 @@ void SearchParams::Populate(OptionsParser* options) {
   // Many of them are overridden with training specific values in tournament.cc.
   options->Add<IntOption>(kMiniBatchSizeId, 1, 1024) = 256;
   options->Add<IntOption>(kMaxPrefetchBatchId, 0, 1024) = 32;
+  options->Add<BoolOption>(kLogitQEnabledId) = true;
```
borg323: This should be overridden to false in selfplay.

src/mcts/node.h
```diff
@@ -363,8 +364,9 @@ class EdgeAndNode {
   }
 
   int GetVisitsToReachU(float target_score, float numerator,
-                        float default_q) const {
-    const auto q = GetQ(default_q);
+                        float default_q, bool logit_q) const {
+    const auto q = (logit_q ? FastLogit(0.99999999 * GetQ(default_q)) :
```
borg323: 0.99999999 is a double constant and doesn't do what you think. You probably want 0.99999994f instead.

src/mcts/search.cc
```diff
@@ -250,7 +250,10 @@ std::vector<std::string> Search::GetVerboseStats(Node* node,
         << ") ";
 
     oss << "(Q+U: " << std::setw(8) << std::setprecision(5)
-        << edge.GetQ(fpu) + edge.GetU(U_coeff) << ") ";
+        << (edge.GetN() == 0 ? fpu :
+            (params_.GetLogitQEnabled() ? FastLogit(0.99999999 * edge.GetQ(0)) :
```
borg323: 0.99999994f here as well.

src/mcts/search.cc
```diff
@@ -955,8 +958,16 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend(
...
       }
-      const float Q = child.GetQ(fpu);
-      const float score = child.GetU(puct_mult) + Q;
+      
+      const float Q = child.GetQ(0);
+      const float U = child.GetU(puct_mult);
+      const float score = (child.GetN() == 0 ? U + fpu :
+                           (params_.GetLogitQEnabled() ?
+                           // Scale by 1-epsilon to avoid infinity
+                           U + FastLogit(0.99999999 * Q) :
```
borg323: Again 0.99999994f.

src/utils/fastmath.h
```diff
@@ -77,4 +77,9 @@ inline float FastLog(const float a) {
 // Fast approximate exp(x). Does only limited range checking.
 inline float FastExp(const float a) { return FastPow2(1.442695040f * a); }
 
+// Fast logit for more readable code.
+inline float FastLogit(const float a) {
+  return 0.5 * FastLog((1.0f + a) / (1.0f - a));
```
borg323: 0.5f here for consistency (although it makes no actual difference).

src/mcts/params.cc
AlexisOlson: Somewhere in `/src/selfplay`?

Veeno: I would guess what line 195 says (`tournament.cc`).

src/mcts/search.cc
```diff
@@ -250,7 +250,10 @@ std::vector<std::string> Search::GetVerboseStats(Node* node,
         << ") ";
 
     oss << "(Q+U: " << std::setw(8) << std::setprecision(5)
-        << edge.GetQ(fpu) + edge.GetU(U_coeff) << ") ";
+        << (edge.GetN() == 0 ? fpu :
+            (params_.GetLogitQEnabled() ? FastLogit(0.99999994f * edge.GetQ(0)) :
+             edge.GetQ(0))) + edge.GetU(U_coeff)
```
Tilps: (!params_.GetLogitQEnabled() || edge.getN() == 0 ? edge.GetQ(fpu)  : FastLogit(0.99999994f * edge.GetQ(fpu))) + edge.GetU(U_coeff)

Produces the identical output and is simpler - no nested ternary.
Separately it would be simpler if this logic just migrated into edge.GetQ(fpu, logit_enabled) - would avoid repetition in the 3 places of the constant.

src/mcts/search.cc
```diff
@@ -955,8 +958,16 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend(
         }
         ++possible_moves;
       }
-      const float Q = child.GetQ(fpu);
-      const float score = child.GetU(puct_mult) + Q;
+      
+      const float Q = child.GetQ(0);
+      const float U = child.GetU(puct_mult);
+      const float score = (child.GetN() == 0 ? U + fpu :
```
Tilps: Same comment as above.

mooskagh:
Could you check nps with random backend
1. before this change
2. after this change with LogitQEnabled=false, and
3. after the change with LogitQEnabled=true

e.g. with `$ ./lc0 benchmark --backend=random`

src/mcts/params.cc
```diff
@@ -49,6 +49,9 @@ const OptionId SearchParams::kMaxPrefetchBatchId{
     "When the engine cannot gather a large enough batch for immediate use, try "
     "to prefetch up to X positions which are likely to be useful soon, and put "
     "them into cache."};
+const OptionId SearchParams::kLogitQEnabledId{
+    "logit-q-enabled", "LogitQEnabled",
```
mooskagh: No need for word "enabled" in boolean flags, "logit-q" = on/off is pretty clear too..

src/mcts/node.h
```diff
@@ -336,8 +337,10 @@ class EdgeAndNode {
   Node* node() const { return node_; }
 
   // Proxy functions for easier access to node/edge.
-  float GetQ(float default_q) const {
-    return (node_ && node_->GetN() > 0) ? node_->GetQ() : default_q;
+  float GetQ(float default_q, bool logit_q = false) const {
+    return (node_ && node_->GetN() > 0) ?
+      (logit_q ? FastLogit(0.99999994f * node_->GetQ()) : node_->GetQ()) :
```
mooskagh: Why such a weird constant? Maybe add comment describing it.

src/mcts/params.cc
```diff
@@ -49,6 +49,9 @@ const OptionId SearchParams::kMaxPrefetchBatchId{
...
+    "Whether to sum Q+U in logit space or not."};
```
mooskagh: "Sum Q+U in logit space." is enough, "whether to" and "or not" are not really informative. :)

src/mcts/node.h
```diff
@@ -336,8 +337,10 @@ class EdgeAndNode {
   Node* node() const { return node_; }
 
   // Proxy functions for easier access to node/edge.
-  float GetQ(float default_q) const {
-    return (node_ && node_->GetN() > 0) ? node_->GetQ() : default_q;
+  float GetQ(float default_q, bool logit_q = false) const {
```
mooskagh: It may make sense to remove `= false` default parameter in order no to forget passing correct value in some important place.
But if it makes lots of places in the code uglier, it's fine to keep it.

src/mcts/params.cc
mooskagh: Or "use-logit-q" is fine too.

src/mcts/params.cc
AlexisOlson: Removed those bits and added a bit of extra information.

src/mcts/node.h
AlexisOlson: There are quite a few places where GetQ is called but only a few places where logit should be applied so I think this way is safer and cleaner.

src/mcts/node.h
AlexisOlson: I used what @borg323 suggested. I'm pretty sure he picked this since it's the largest 32-bit float less than 1. I can take the final 4 off so it doesn't look as weird without affecting the behavior much since it's about 8.4 for logit(0.9999999) instead of about 8.7.

I'll leave a comment as well.

borg323: Yes, this is 0x3f7fffff.

src/mcts/node.h
```diff
@@ -336,8 +337,11 @@ class EdgeAndNode {
...
+    return (node_ && node_->GetN() > 0) ?
+      // Scale Q slightly to avoid logit(1) = infinity
```
mooskagh: Missing period at the end of the sentence. :)

PR #926, opened 2019-08-18
Revert "Restore --noise treating as --noise-epsilon=0.25."

Mardak:
This reverts commit 1753f7c40332121950f818979d9c71bca34ee230.

Followup to #267 but only commit after https://github.com/LeelaChessZero/lc0/commit/8f13ece9698888e5b156c7b2f3d3300832bda004 `has been included in a version which has been made mandatory.` https://github.com/LeelaChessZero/lc0/pull/267#issuecomment-522185245

### DIFF
--- a/src/mcts/params.cc
+++ b/src/mcts/params.cc
@@ -109,12 +109,6 @@ const OptionId SearchParams::kTemperatureVisitOffsetId{
     "Adjusts visits by this value when picking a move with a temperature. If a "
     "negative offset reduces visits for a particular move below zero, that "
     "move is not picked. If no moves can be picked, no temperature is used."};
-const OptionId SearchParams::kNoiseId{
-    "noise", "DirichletNoise",
-    "Add Dirichlet noise to root node prior probabilities. This allows the "
-    "engine to discover new ideas during training by exploring moves which are "
-    "known to be bad. Not normally used during play.",
-    'n'};
 const OptionId SearchParams::kNoiseEpsilonId{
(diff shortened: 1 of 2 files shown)


mooskagh:
Pre-requisite PR is already submitted, right?

mooskagh:
Usually we cherrypick for patch-releases, but it happened several times that we just synced to master for v0.xx.1, so maybe it's better to wait for now.

Mardak:
The server is still requesting `--noise=true` and 0.23 will probably be the first release to have `--noise-epsilon` from #267. So the timing should probably be:

1) release 0.23
2) server requires 0.23 and switch from --noise to --noise-epsilon
3) merge this PR to master

Mardak:
Rebased on latest master resolving adjacent line conflict with #1088 in params.cc.

@Tilps I see match params still have `--noise=true` and training run actually uses both --noise-epsilon and --noise, which actually results in 0.25 overriding the 0.1 value.

If the server already requires 0.23 or will soon require 0.24 for all runs, it should be safe to merge this PR. Otherwise if say training run 1 wants to stay on an older version, we'll need to hold off on this for now.

Mardak:
The server requires 0.24 now, so the server can replace `--noise=true` with `--noise-epsilon=0.25`:
http://training.lczero.org/training_runs

Then with no more usages of `--noise`, this should be fine to merge.

borg323:
Can you merge with latest master to trigger a rebuild?

Tilps:
Training flag changes have been done - ready to merge now.
So doing so.

PR #928, opened 2019-08-21
U-Value exploration factor

AlexisOlson:
https://arxiv.org/pdf/1908.06660.pdf
"This increases the chance of exploring unvisited nodes at least once"


Naphthalin:
While there is a paper linked, the intended behavior can already be reached by changing `fpu`, which we tune anyway since kiudee's tuning, therefore the added complexity isn't justified.

I recommend closing this PR.

PR #930, opened 2019-08-23
Policy Softmax Temp experiment

AlexisOlson:
This equalizes policy values as N increases with the goal of phasing out early bias.

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -1249,7 +1249,7 @@ void SearchWorker::FetchSingleNodeResult(NodeToProcess* node_to_process,
     // Perform softmax and take into account policy softmax temperature T.
     // Note that we want to calculate (exp(p-max_p))^(1/T) = exp((p-max_p)/T).
     p = FastExp((p - max_p) /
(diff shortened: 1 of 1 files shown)


jhorthos:
I don't get what that is doing - it seems to exclude Policy Softmax altogether?

AlexisOlson:
@jhorthos It ignores the parameter in favor of a new temp function.

jhorthos:
Ah okay, I thought it was attempting to modify the softmax. So the hard coded constant is a test value for experiments?

AlexisOlson:
@jhorthos Yes. I should probably use that co-opt that parameter to use as an additive or multiplicative factor instead.

Tilps:
Note that policy values are only populated once per node - so N never increases in this scenario.

AlexisOlson:
@Tilps Thanks for pointing that out. That makes sense.

jjoshua2:
That stopped my last experiment too sadly.

AlexisOlson:
Too much code change required to keep this open.

PR #936, opened 2019-08-25
simple mkldnn backend

borg323:
This is effectively a copy of the blas backend slightly modified to use `mkldnn_sgemm()` (that only supports arrays in row-major order). The idea is to get this working now and introduce high level mkl-dnn functions slowly.

The performance in my tests is very close to MKL and I understand it is legal to distribute windows builds with the required dll, so this is useful even in this stage.

### DIFF
--- a/src/neural/mkldnn/convolution1.cc
+++ b/src/neural/mkldnn/convolution1.cc
@@ -1,0 +1,39 @@
+/*
+ This file is part of Leela Chess Zero.
+ Copyright (C) 2018 The LCZero Authors
+
+ Leela Chess is free software: you can redistribute it and/or modify
+ it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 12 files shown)


src/neural/mkldnn/convolution1.cc
```diff
@@ -0,0 +1,39 @@
+/*
+ This file is part of Leela Chess Zero.
+ Copyright (C) 2018 The LCZero Authors
```
mooskagh: Maybe it makes sense to have 2019 in copyright banners?

borg323:
Marked as WIP since building it on macos proved hard, needs further investigation.

borg323:
Closing, don't plan to work on this anytime soon now that dnnl is available through the blas backend.

PR #949, opened 2019-09-16
add the UCI_Opponent option

borg323:
Testing with cutechess shows that it does use this option. Edited lc0 log excerpt:
```
>> setoption name Ponder value false
>> setoption name UCI_Opponent value none none computer Ethereal
>> position startpos
```


borg323:
This was just a proof of concept.

borg323:
Since there was some interest in this, I added a simple mechanism to override search options depending on the current opponent.

borg323:
This is superseded by #1791

PR #953, opened 2019-09-19
Add a shortsightedness parameter.

Tilps:
I have no idea if this even vaguely does what I suggest in the parameter comment - but it seemed interesting to test out.

### DIFF
--- a/src/mcts/params.cc
+++ b/src/mcts/params.cc
@@ -246,7 +246,7 @@ void SearchParams::Populate(OptionsParser* options) {
   options->Add<ChoiceOption>(kHistoryFillId, history_fill_opt) = "fen_only";
   options->Add<IntOption>(kKLDGainAverageInterval, 1, 10000000) = 100;
   options->Add<FloatOption>(kMinimumKLDGainPerNode, 0.0f, 1.0f) = 0.0f;
-  options->Add<FloatOption>(kShortSightednessId, 0.0f, 0.01f) = 0.0f;
+  options->Add<FloatOption>(kShortSightednessId, 0.0f, 1.0f) = 0.0f;
 
   options->HideOption(kNoiseEpsilonId);
   options->HideOption(kNoiseAlphaId);
--- a/src/mcts/params.cc
+++ b/src/mcts/params.cc
@@ -199,6 +199,9 @@ const OptionId SearchParams::kKLDGainAverageInterval{
     "kldgain-average-interval", "KLDGainAverageInterval",
     "Used to decide how frequently to evaluate the average KLDGainPerNode to "
     "check the MinimumKLDGainPerNode, if specified."};
+const OptionId SearchParams::kShortSightednessId{
+    "short-sightedness", "ShortSightedness",
+    "Used to focus more on short term gains over long term."};
 
 void SearchParams::Populate(OptionsParser* options) {
   // Here the uci optimized defaults" are set.
@@ -243,6 +246,7 @@ void SearchParams::Populate(OptionsParser* options) {
   options->Add<ChoiceOption>(kHistoryFillId, history_fill_opt) = "fen_only";
   options->Add<IntOption>(kKLDGainAverageInterval, 1, 10000000) = 100;
   options->Add<FloatOption>(kMinimumKLDGainPerNode, 0.0f, 1.0f) = 0.0f;
+  options->Add<FloatOption>(kShortSightednessId, 0.0f, 0.01f) = 0.0f;
 
   options->HideOption(kNoiseEpsilonId);
   options->HideOption(kNoiseAlphaId);
@@ -280,6 +284,7 @@ SearchParams::SearchParams(const OptionsDict& options)
       kSyzygyFastPlay(options.Get<bool>(kSyzygyFastPlayId.GetId())),
       kHistoryFill(
           EncodeHistoryFill(options.Get<std::string>(kHistoryFillId.GetId()))),
-      kMiniBatchSize(options.Get<int>(kMiniBatchSizeId.GetId())) {}
+      kMiniBatchSize(options.Get<int>(kMiniBatchSizeId.GetId())),
+      kShortSightedness(options.Get<float>(kShortSightednessId.GetId())) {}
 
 }  // namespace lczero
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -1323,10 +1323,8 @@ void SearchWorker::DoBackupUpdateSingleNode(
       v = n->GetQ();
       d = n->GetD();
     }
-    n->FinalizeScoreUpdate(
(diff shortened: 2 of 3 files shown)


src/mcts/search.cc
```diff
@@ -1313,7 +1314,10 @@ void SearchWorker::DoBackupUpdateSingleNode(
       v = n->GetQ();
       d = n->GetD();
     }
-    n->FinalizeScoreUpdate(v, d, node_to_process.multivisit);
+    n->FinalizeScoreUpdate(
+		v * std::max(0.0f, 1.0f - params_.GetShortSightedness() * depth),
```
AlexisOlson: If `depth = 0`, this returns `v * 1`, but for `depth >= 1` this gives `v * 0` since `1 - depth <= 0`.

Tilps: err shortsightedness is a float parameter with a maximum possible value of 0.03.  there is no calculation of 1 - depth.

AlexisOlson: Ha. For some reason I thought it was a boolean.

Tilps: There are definitely other options for decay. v / (1 + param * depth) for instance would avoid the need to clamp without ever leaving [-1,1] bounds.  But will start with this simple one for some testing to start.  I need to set up a 'vs weaker opponent' test to see if it has a bias towards that scenario.

Tilps: Switched to 1 / ( 1 + param * depth) - seems to work close enough to before, just with a slightly increased parameter value.

Tilps:
I've clopped a tune recommending approx 0.018 (for 10k nodes, guess that maybe it should be lower as nodes increase - but the new version isn't so bad that way, so might be fine!).  Additionally estimate about a 10-20% cpuct reduction at the same time (since the somewhat reduced q value range effectively means cpuct's power is increased).

Note that this does effect the output q scores - which means any tests run with adjudication should be considered questionable.

Tilps:
I'm leaving WIP here mainly because I imagine the effect on output Q could be annoying in practice.

zz4032:
Latest commit (with `1.0f + params_.GetShortSightedness()` and parameter range for  ShortSightedness from 0 to 0.1) tuned vs. SF and tested in a gauntlet with default (and also tuned) LC0:

TC (LC0): 1.98s/game+0.033s/move (~3500 nodes/move)
MinibatchSize=32
lc0-tuned: CPuct=1.98, FpuValue=0.523
lc0_PR953-tuned:  CPuct=2.106, FpuValue=0.568, ShortSightedness=0.02262
```
   # PLAYER             :  RATING  ERROR  CFS(%)   GAMES  DRAWS(%)  SCORE(%)
   1 lc0_PR953-tuned    :     2.9   10.8    70.1    2418      39.5      50.4
   2 stockfish          :     0.0   ----    98.5    4836      40.8      50.6
   3 lc0-tuned          :   -11.5   10.4     ---    2418      42.2      48.4

White advantage = -6.2 +/- 3.9
Draw rate (equal opponents) = 40.9 % +/- 0.7
```
Range for ShortSightedness can be probably decreased to [0;0.1]. Clop diagram shows complete fading upwards from about 0.08.

Tilps:
Slight increase in cpuct goes counter to what my tunes against lc0 were suggesting - but it looks like good evidence that shortsightedness of around 0.02 increases elo. At least for fast time controls.

Tilps:
I'm going to leave the default value at 0 for now, so the Q issue would only affect people who opt in to test it.

zz4032:
One more test vs. weaker SF:
```
   # PLAYER             :  RATING  ERROR  CFS(%)   GAMES  DRAWS(%)  SCORE(%)
   1 lc0-tuned          :   147.1   10.5    67.8    2476      29.5      69.8
   2 lc0_PR953-tuned    :   143.6   10.2   100.0    2476      31.1      69.4
   3 stockfish          :     0.0   ----     ---    4952      30.3      30.4

White advantage = 0.0 +/- 0.0
Draw rate (equal opponents) = 50.0 % +/- 0.0
```
Preferring shorter lines sounds like negative contempt to me (better vs. stronger engines), so I would have expected worse performance vs. weaker SF. But it looks that it's equal to default.
A tuning and match vs. stronger SF might be interesting.

src/mcts/search.cc
```diff
@@ -1322,7 +1323,8 @@ void SearchWorker::DoBackupUpdateSingleNode(
       v = n->GetQ();
       d = n->GetD();
     }
-    n->FinalizeScoreUpdate(v, d, node_to_process.multivisit);
+    n->FinalizeScoreUpdate(v / (1.0f + params_.GetShortSightedness() * depth),
```
mooskagh: Any noticeable `nps` change on random backend?

Tilps: I will get those numbers, but I'm confident the answer will be no.

src/mcts/params.cc
```diff
@@ -243,6 +246,7 @@ void SearchParams::Populate(OptionsParser* options) {
   options->Add<ChoiceOption>(kHistoryFillId, history_fill_opt) = "fen_only";
   options->Add<IntOption>(kKLDGainAverageInterval, 1, 10000000) = 100;
   options->Add<FloatOption>(kMinimumKLDGainPerNode, 0.0f, 1.0f) = 0.0f;
+  options->Add<FloatOption>(kShortSightednessId, 0.0f, 1.0f) = 0.0f;
```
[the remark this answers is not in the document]
Tilps: Hehe.  Testing suggests values over 0.1 are pretty terrible - but I thought someone might like the comic value of the super aggressive play it would invite. (higher than 1 just felt silly :P)

src/mcts/search.cc
Tilps: 10 second random benchmark with 4 threads are 340k +/- 5k on my machine for both master and this branch (with shortsightedness 0).

zz4032:
Tuned vs. stronger SF with fixed CPuct/FpuValue since I expected both parameters don't change much anyway.
Tuning could have run a bit longer, tuned result was ShortSightedness=0.008194.

lc0-tuned: CPuct=1.98, FpuValue=0.523
lc0_PR953-tuned:  CPuct=1.98, FpuValue=0.523, ShortSightedness=0.008194
```
   # PLAYER             :  RATING  ERROR  CFS(%)   GAMES  DRAWS(%)  SCORE(%)
   1 stockfish          :     0.0   ----   100.0    3500      41.3      71.0
   2 lc0-tuned          :  -148.0   12.6    98.2    1750      41.6      30.1
   3 lc0_PR953-tuned    :  -166.8   12.4     ---    1750      41.0      27.9

White advantage = -6.0 +/- 4.6
Draw rate (equal opponents) = 50.3 % +/- 1.0
```

Issue #585, opened 2018-12-14
lto build on Ubuntu 16.04 using GCC broken, works using clang


Aloril:
Works fine:
CC=clang-6.0 CXX=clang++-6.0 ./build.sh

Crashes (uses gcc 5.4.0):
./build.sh
./build/release/lc0 --weights=networks/weights_test35b10-35001.pb.gz
position fen rnbqkb1r/1ppppppp/5n2/p7/8/3PB3/PPP1PPPP/RN1QKBNR w KQkq - 0 1 moves g1f3 d7d6 b1c3 e7e5 g2g3 b8c6 f1g2 d6d5 e1g1 d5d4 a2a3 d4e3 f1e1 a5a4 h2h4 e5e4 b2b4 e4f3 d1d2 e3d2 e2e3 b7b6 g3g4 b6b5 g1h1 g7g6 e1d1 g6g5 h4h5 h7h6 h1g1 f3g2 f2f3 c6d4
go movetime 1000000000 nodes 1

Another test case:
./build/release/lc0 --weights=networks/weights_test35b10-35081.pb.gz
position fen r1bqkbnr/pppppp1p/n5p1/8/6P1/5N2/PPPPPP1P/RNBQKB1R w KQkq - 0 1 moves d2d4 d7d5 h1g1
go movetime 1000000000 nodes 1

Also works fine if build.sh modified as following when using GCC:
if [ -f ${BUILDDIR}/build.ninja ]
then
  meson configure ${BUILDDIR} -Db_lto=false --buildtype ${BUILDTYPE} --prefix ${INSTALL_PREFIX:-/usr/local} "$@"
else
  meson ${BUILDDIR} -Db_lto=false --buildtype ${BUILDTYPE} --prefix ${INSTALL_PREFIX:-/usr/local} "$@"
fi

borg323:
This was the crash @Aloril reported on discord:
Thread 38 "lczero20d1" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fff81fff700 (LWP 127980)]
[...]

PR #954, opened 2019-09-19
enable lto and fast-math - not for merging

borg323:
A few months ago we tried to enable LTO building by default, which resulted in some obscure crashes with older gcc versions (https://github.com/LeelaChessZero/lc0/issues/585), which had been bugging me since. I managed to trace it (`git bisect` rules) to a bad interaction with `-ffast-math` that we had enabled back then. However `-ffast-math` was also breaking standard builds with clang, so it has been disabled for some time. Therefore enabling LTO builds should be safe. However this is not what this PR does: searching through the fast-math sub-options I think I have isolated the bad ones, so this works for me with all compilers tested (including the `/fp:fast` msvc option, although this is less tested) and hope to get some independent testing.
Note to testers: To reproduce the crash reliably I was playing 1000 games (selfplay but with cutechess to stress the full uci interface as well) with 1 node per move, and the crash was happening in the first 200 games, guesstimating a 0.5% crash probability.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -15,7 +15,7 @@
 # along with Leela Chess.  If not, see <http://www.gnu.org/licenses/>.
 
 project('lc0', 'cpp',
-        default_options : ['cpp_std=c++14', 'b_ndebug=if-release'],
+        default_options : ['cpp_std=c++14', 'b_ndebug=if-release', 'b_lto=true'],
         meson_version: '>=0.45')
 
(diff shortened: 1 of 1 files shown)


Naphthalin:
What is the current stance on this PR, especially now that we don't support older gcc versions anymore anyway?

borg323:
This needs extensive testing to see if it is still relevant. Let's keep it around for now.

PR #956, opened 2019-09-20
Patch for PR925

AlexisOlson:
1. Apply logit inside of GetFPU function to handle relative fpu more correctly.
2. Add analogous logit switches inside the *Prefetch into cache* code section for consistency.
3. Make the scaling factor inside FastLogit a hidden parameter.


src/mcts/search.cc
```diff
@@ -192,11 +192,13 @@ int64_t Search::GetTimeToDeadline() const {
 }
 
 namespace {
-inline float GetFpu(const SearchParams& params, Node* node, bool is_root_node) {
+inline float GetFpu(const SearchParams& params, Node* node,
+                    bool is_root_node, bool logit_q = false) {
   const auto value = params.GetFpuValue(is_root_node);
+  const auto Q = logit_q ? node->GetQ() : FastLogit(node->GetQ());
```
dtracers: Why is this opposite of the one in node.h?

AlexisOlson: Wow. I don't think it should be.

dtracers: I think this one is the one that should be flipped

AlexisOlson:
Without the change @dtracers suggested, the PR appear to improve Elo. Afterward, it appears to lose Elo. My hypothesis is that part 2 in the description is the cause for improved Elo since the first effectively wasn't applied and that fpu needs to be re-tuned after making the last commit.

src/utils/fastmath.h
```diff
@@ -78,8 +78,11 @@ inline float FastLog(const float a) {
 inline float FastExp(const float a) { return FastPow2(1.442695040f * a); }
 
 // Fast logit for more readable code.
+// This is actually 0.5 * logit((a+1)/2) = atanh(a).
 inline float FastLogit(const float a) {
-  return 0.5f * FastLog((1.0f + a) / (1.0f - a));
+  // Scale Q slightly to avoid logit(1) = infinity.
+  float b = 0.9999999f * a;
```
borg323: This is not the place for the scaling.

src/mcts/search.cc
```diff
@@ -192,11 +192,13 @@ int64_t Search::GetTimeToDeadline() const {
 }
 
 namespace {
-inline float GetFpu(const SearchParams& params, Node* node, bool is_root_node) {
+inline float GetFpu(const SearchParams& params, Node* node,
+                    bool is_root_node, bool logit_q = false) {
   const auto value = params.GetFpuValue(is_root_node);
+  const auto Q = logit_q ? FastLogit(node->GetQ()) : node->GetQ();
   return params.GetFpuAbsolute(is_root_node)
              ? value
```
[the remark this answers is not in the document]
Naphthalin: @Tilps is right, line 200 should be changed to ``(logit_q ? FastLogit(value) : value)``

src/utils/fastmath.h
Naphthalin: So you suggest it would be better to introduce another function 
```inline float ScaledFastLogit(const float a) {
  return FastLogit(0.9999999f * a);
}```
?

Comment (borg323) [reply src/utils/fastmath.h]: No, just do it like it was before. If you dislike repeating the numeric value, you can define a constant.

File: src/mcts/search.cc (hunk 3 above)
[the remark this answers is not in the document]
Comment (AlexisOlson) [reply src/mcts/search.cc]: OK. How should we handle `value` outside of `[-1,1]` then?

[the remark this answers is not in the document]
Comment (Naphthalin) [reply src/mcts/search.cc]: Probably don't allow values outside of ``[-1,1]`` in the first place. There is no reason I can think of that one would like to initialize unvisited moves as worse than certain loss.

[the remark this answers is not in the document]
Comment (AlexisOlson) [reply src/mcts/search.cc]: The default value is 1.2 though, I think. I'm not sure the best way to handle that.

File: src/utils/fastmath.h (hunk 2 above)
Comment (AlexisOlson) [reply src/utils/fastmath.h]: What's the best way to get a single constant to both `search.cc` and `node.h`? I haven't figured out how to pass it into the latter.

File: src/mcts/search.cc (hunk 3 above)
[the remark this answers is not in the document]
Comment (Tilps) [reply src/mcts/search.cc]: Default value might be 1.2 - but default mode is relative - so they are not related.  Anyone switching to absolute should also be setting a value.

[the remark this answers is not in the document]
Comment (AlexisOlson) [reply src/mcts/search.cc]: OK. I won't worry about it then. If someone is using absolute, presumably they know better.

File: src/utils/fastmath.h (hunk 2 above)
Comment (AlexisOlson) [reply src/utils/fastmath.h]: I decided to try passing it in through the function.

File: src/mcts/search.cc [hunk 4]
```diff
@@ -192,11 +192,14 @@ int64_t Search::GetTimeToDeadline() const {
...
-inline float GetFpu(const SearchParams& params, Node* node, bool is_root_node) {
+inline float GetFpu(const SearchParams& params, Node* node,
+                    bool is_root_node, bool logit_q = false) {
   const auto value = params.GetFpuValue(is_root_node);
+  const auto scale = params.GetOneMinusEps();
+  const auto Q = logit_q ? FastLogit(scale * node->GetQ()) : node->GetQ();
   return params.GetFpuAbsolute(is_root_node)
-             ? value
-             : -node->GetQ() - value * std::sqrt(node->GetVisitedPolicy());
+             ? (logit_q ? FastLogit(scale * value) : value)
```
Naphthalin: I would suggest to replace ``(logit_q ? FastLogit(scale * value) : value)`` by ``(logit_q ? FastLogit(scale * std::max(-1.0f,value)) : value)`` to avoid unwanted behavior since we don't guarantee that ``FpuValue`` is in ``[-1,1]``. Having that value at -1.0 still basically guarantees that that node will never be explored, so one should refrain from using absolute fpu -1.0 with logit.

Tilps: Note that -1 is the value we usually use with absolute.

AlexisOlson: I did `min(1, max(-1, value))` to bound on both sides.

src/mcts/search.cc
```diff
@@ -192,11 +192,17 @@ int64_t Search::GetTimeToDeadline() const {
 }
 
 namespace {
-inline float GetFpu(const SearchParams& params, Node* node, bool is_root_node) {
+inline float GetFpu(const SearchParams& params, Node* node,
+                    bool is_root_node, bool logit_q = false) {
```
Naphthalin: ``logit_q`` doesn't need to be given here as one can already access it by ``params.GetLogitQ()``

AlexisOlson: True, but do I always want to call GetFpu with `logit_q = params.GetLogitQ()`?

Naphthalin: That is indeed a good question. Currently there are 2 calls to ``GetFpu()`` without the ``logit_q`` parameter. In my opinion they should both be called with ``logit_q`` included, as one of them is ``verbosemovestats`` and the other for ``TempValueCutoff`` which we definitely would like to do in logit space anyway -- there still are missing some ``logit_q`` calls anyway.

AlexisOlson: OK. I've switched to using the parameter rather than passing `logit_q` as an argument to the function.

appveyor.yml
```diff
@@ -32,7 +32,7 @@ install:
 - cmd: IF %CUDA%==true IF NOT EXIST C:\cache\cuda appveyor DownloadFile http://developer.download.nvidia.com/compute/redist/cudnn/v7.4.2/cudnn-10.0-windows10-x64-v7.4.2.24.zip
 - cmd: IF %CUDA%==true IF NOT EXIST C:\cache\cuda 7z x cudnn-10.0-windows10-x64-v7.4.2.24.zip -oC:\cache
 - cmd: set PATH=C:\Python36;C:\Python36\scripts;%PATH%
-- cmd: pip3 install --upgrade meson
+- cmd: pip3 install --upgrade meson==0.51.2
```
borg323: This is now in master, please merge again.

src/mcts/params.cc
```diff
@@ -53,6 +53,9 @@ const OptionId SearchParams::kLogitQId{
     "logit-q", "LogitQ",
     "Apply logit to Q when determining Q+U best child. This makes the U term "
     "less dominant when Q is near -1 or +1."};
+const OptionId SearchParams::kOneMinusEpsId{
+    "one-minus-eps", "OneMinusEps",
+    "Scaling factor to apply before computing logit."};
```
borg323: OneMinusEps makes sense as a constant but not as an option. Maybe rename it to something like LogitScale?

src/mcts/params.cc
```diff
@@ -206,6 +209,7 @@ void SearchParams::Populate(OptionsParser* options) {
   options->Add<IntOption>(kMiniBatchSizeId, 1, 1024) = 256;
   options->Add<IntOption>(kMaxPrefetchBatchId, 0, 1024) = 32;
   options->Add<BoolOption>(kLogitQId) = false;
+  options->Add<FloatOption>(kOneMinusEpsId, 0.0f, 1.0f) = 0.99999994f;
```
borg323: Does a lower limit of 0 make sense? 
Also shouldn't the max here be also 0.99999994f? A cleaner approach may be to keep the upper limit as 1, and multiply by 0.99999994f when setting the constant (see later).

src/mcts/params.cc
```diff
@@ -247,11 +251,13 @@ void SearchParams::Populate(OptionsParser* options) {
   options->HideOption(kNoiseEpsilonId);
   options->HideOption(kNoiseAlphaId);
   options->HideOption(kLogLiveStatsId);
+  options->HideOption(kOneMinusEpsId);
 }
 
 SearchParams::SearchParams(const OptionsDict& options)
     : options_(options),
       kLogitQ(options.Get<bool>(kLogitQId.GetId())),
+      kOneMinusEps(options.Get<float>(kOneMinusEpsId.GetId())),
```
borg323: As stated above it may be better to multiply by 0.99999994f here instead of having it as default value.

src/mcts/params.cc
AlexisOlson: I've changed the lower limit to 0.1 where it's close enough to plain Q that using LogitQ doesn't make sense.

Note that I also divide by scale (`logit(scale * Q)/scale`) now so Q behavior isn't changed near 0 when the scale is less than 1. I'll post a graph regarding this momentarily.

AlexisOlson:
Note that the scaling parameter is now applied as

`logit(scale * Q)/scale` rather than just `logit(scale * Q)`

This is to preserve the slope near Q = 0. See gold versus green below:

Scale

When scale = 0.8, we want the green line, not the gold line.

mooskagh:
So by default it's still the old behaviour, right?
Also, would be nice to check again that speed is not affected (with random backend).

borg323:
I see about 1.5% slowdown using the random backend without `--logit-q` and about 2% with it. This is after merging with master (<https://github.com/borg323/lc0/tree/pr956> for the merged tree).

kiudee:
Can we update this pull request to the current state of master? Then it would be easier to test it.

src/mcts/search.cc
```diff
@@ -1031,7 +1038,7 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend(
 
     if (second_best_edge) {
       int estimated_visits_to_change_best = best_edge.GetVisitsToReachU(
-          second_best, puct_mult, fpu, draw_score, params_.GetLogitQ());
+          second_best, puct_mult, fpu, draw_score, params_.GetLogitQ(), scale);
```
borg323: scale is not defined here

src/mcts/search.cc
```diff
@@ -1258,8 +1264,7 @@ int SearchWorker::PrefetchIntoCache(Node* node, int budget, bool is_odd_depth) {
     if (i != scores.size() - 1) {
       // Sign of the score was flipped for sorting, so flip it back.
       const float next_score = -scores[i + 1].first;
-      // TODO: As above - should this use logit_q if set?
-      const float q = edge.GetQ(-fpu, draw_score, /* logit_q= */ false);
+      const float q = edge.GetQ(-fpu, draw_score, /*logit_q= */ false, scale);
```
borg323: scale is not defined here either

src/mcts/node.h
```diff
@@ -345,12 +345,10 @@ class EdgeAndNode {
   Node* node() const { return node_; }
 
   // Proxy functions for easier access to node/edge.
-  float GetQ(float default_q, float draw_score, bool logit_q) const {
+  float GetQ(float default_q, float draw_score, bool logit_q, float scale = 1.0f) const {
```
mooskagh: Instead of carrying two variables everywhere, I think it's better to have one `std::optional<float>`.

Also I believe in this case having default function arguments do more harm than good, it's easy to confuse parameters.

tqa236:
Hello all, I wonder what's the status of this PR? Is it ready to be merged?

Tilps:
Given logit-q was just removed - this PR won't be going anywhere in the immediate future.

Naphthalin:
Closing the PR, as the decision to remove logit_q didn't cause any controversies.

PR #957, opened 2019-09-23
Add a flag to cudnn backend to bypass some batch size non-determinism in result output

Tilps:
I ran into this while trying to set up a system for doing policy head elo calculation and was wondering why I was getting non-zero elo for same network vs itself with mirrored openings whenever I tried to take advantage of more than a little parallelism.

### DIFF
--- a/src/neural/cuda/network_cudnn.cc
+++ b/src/neural/cuda/network_cudnn.cc
@@ -198,6 +198,11 @@ class CudnnNetwork : public Network {
                    pblczero::NetworkFormat::POLICY_CONVOLUTION;
 
     max_batch_size_ = options.GetOrDefault<int>("max_batch", 1024);
+    // Batches smaller than 4 have different results for same input for
+    // cudnn-fp16 mode. Deterministic flag allows this to be disabled by forcing
+    // all batches to pretend to be at least 4 in size.
+    min_batch_size_ =
+        options.GetOrDefault<bool>("deterministic", false) ? 4 : 1;
 
     int total_gpus;
     ReportCUDAErrors(cudaGetDeviceCount(&total_gpus));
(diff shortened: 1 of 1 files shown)


src/neural/cuda/network_cudnn.cc
```diff
@@ -198,6 +198,11 @@ class CudnnNetwork : public Network {
                    pblczero::NetworkFormat::POLICY_CONVOLUTION;
 
     max_batch_size_ = options.GetOrDefault<int>("max_batch", 1024);
+    // Batches smaller than 4 have different results for same input for
+    // cudnn-fp16 mode. Deterministic flag allows this to be disabled by forcing
+    // all batches to pretend to be at least 4 in size.
+    min_batch_size_ =
+        options.GetOrDefault<bool>("deterministic", false) ? 4 : 1;
```
mooskagh: maybe make it `true` by default?

Tilps: I assume there is performance cost to pay there - not sure if its significant though.

mooskagh: I think it's pretty rare that backend ends up evaluating a tiny batch, so I guess it should be fine.. probably.

Tilps:
While testing my value head only variant - I think I have found that there is another scenario where the returned values changes again - but I don't know what threshold it is this time, and assumedly its quite high so maybe this CL isn't worth it.

Naphthalin:
Is this still relevant with the recent developments concerning batch sizes, max-evals-out-of-order-factor etc?

Tilps:
More recent observations have seen that 4 is typically faster than 1 for cudnn-fp16.  I think this will become obsolete once we get the custom winograd on all cards...

PR #959, opened 2019-09-23
PR918+925+956

Naphthalin:
Since #918 is in conflict with the merged #925 and the bug fix #956 I merged the PR's. It might be preferrable though to treat NewU only as a flag for being used in ``GetU()`` and ``GetVisitsToReachU()`` instead of creating new versions of the functions as it is done in #918 .


Naphthalin:
#918 was updated now and resolved all conflicts with the merged #925. This can be closed now.

PR #963, opened 2019-10-06
Beta-MCTS: a soft hybrid of MCTS and Minimax (wip, not for merge)

Naphthalin:
I am implementing a promising idea for a MCTS/Minimax hybrid which will basically do a soft transition from standard MCTS to Minimax. Unlike UCT which theoretically converges to Minimax as well (but without any reasonable bound on the nodes spent...), beta-MCTS actually gives (nearly) Minimax evals for nodes with enough visits. 

Speed and softness of this transition are controlled by two parameters ``--betamcts-trust`` and ``betamcts-percentile``, standard values are ``0.1`` and ``0.05`` (which is rather soft), allowed values are ``[0.0, 1000]`` and ``[0.0, 0.5]``. Too high values of ``trust`` will result in blind spot traps as the NN evals can't be trusted in a minimax way. Lower ``percentile``  will in general put a bit more weight on well explored moves with suboptimal evals, higher percentile on moves with near optimal evals. Setting either parameter to ``0.0`` will result in standard MCTS behavior without transition to Minimax.

There are 4 levels of ``--betamcts-level`` use, affecting how deeply beta-MCTS is used during search, though level 3 isn't implemented yet.

Please note that the size of ``Node`` increased from 80 to 88 Bytes, though I have no idea if that is actually relevant :) Also I expect quite a slowdown on random backend because of some additional CPU operations, any numbers on this would be interesting. Also I don't know if there is any interaction with the ``GetVisitsToReachU()`` logic, especially with the not yet implemented level 3.

By design the biggest differences in displayed eval are to be expected in biased tactical lines where only one side can deviate. This especially includes perpetual checks in losing positions, where one side has to answer the checks while the other can deviate to a losing position.

I hope that no faulty behavior escaped my testing, but please let me know if you find anything and I'll look into it.


jhorthos:
Interesting - would you mind giving a little more high-level description of how search would change? I am guessing that this would reduce the extreme focus on one move that we get with MCTS, or rather it would reduce it as node count gets big.

Naphthalin:
> Interesting - would you mind giving a little more high-level description of how search would change? I am guessing that this would reduce the extreme focus on one move that we get with MCTS when one move emerges with clear best Q in the first 100k nodes or so, or rather it would reduce it as node count gets big. 

Actually, together with a sharp PUCT search due to our lowish cpuct and sharp policies it somehow amplifies one problem of PUCT: getting enough visits for initially seemingly bad moves which only turn out to be good after some while. This most probably stems from the higher sensitivity to changes in lines where evals isn't drifting slowly as we are used to, but fast, similarly to A/B engines. While a found refutation of the top line is propagated way faster since beta-MCTS doesn't need to wait for the refutation line to get enough visits to outweigh the old results, and in a similar fashion a good line for an alternative move is found (i.e. dominates the eval!) with lower number of nodes, it doesn't actually mean that it is expected to solve the blind spots from PUCT coming from secondary nodes not getting enough visits.

> From playing ICCF games, I know that this property is often ridiculously extreme to the point that it seems worrisome.

I agree with that and would extend this to the analysis of human games in general, as a "I would definitely play this move" doesn't mean that other moves have to be inferior, but they aren't analyzed enough. In my opinion beta-MCTS can already help there partially, as it is expected to give reasonable evals even with much higher cpuct values, forced exploration of some lines (people might want to try non-zero stuff like check extensions, though perpetual checks seem to be a prime position which beta-MCTS can already solve on its own) and even other search paradigms as Thompson Sampling. The current paradigm and the reason why ``PUCT>>everything else`` might lie exactly in this entanglement of exploration and precise evals, and I want to break that entanglement by giving a more general "precise eval" module.

Long story short: PUCT is probably not bad, and having a "tactical net" like T20 might actually already work when replacing the standard MCTS eval by beta-MCTS.

jhorthos:
Thanks for the additional comments Naphthalin. If you think you are ready for some testing, let me know - will be happy to help out.

jkormu:
Search seems to hang for certain positions. For example, position `8/pK6/1p6/8/P7/1Pk5/8/8 w - - 7 54` go nodes 600 does not return best move with settings `--betamcts-trust=999.0 --betamcts-percentile=0.05 --minibatch-size=32 --backend=cudnn-fp16 --weights=/path/to/weights/LD2.pb.gz`. Tested 5 times with 600 nodes and hangs each time but with 100 nodes search manages to complete and return the best move.

Naphthalin:
@jkormu Standard value of ``--betamcts-trust`` is for a good reason at  ``0.1``, and I definitely wouldn't recommend any value above 10 at all except for testing purposes, since the NN evals sometimes are a bit erratic. With ``999.0`` you will basically do pure minimax search, and there is a reason minimax and NN currently don't mix well.

jkormu:
Yes, was testing the effect of extreme allowed settings and noticed this. 
I know this PR is WIP but in case this is not a known issue i'll let you know that this is also reproducible with:
--betamcts-trust=9.9 and 50000 nodes
--betamcts-trust=1.0 and 1000000 nodes
--betamcts-trust=0.1 and 5000000 nodes

Naphthalin:
Thx for the responses and for testing in general! However, I think as this position is a 6-piece endgame position, it is likely a fault of the NN.

I would recommend to also test higher cpuct values, e.g. cpuct=4, cpuct-base=1000, cpuct-factor=10 at otherwise standard settings, especially for more tactical positions.

jkormu:
Could you elaborate how you think it is fault of NN? The NN provides value and policy for a given position. Then search uses this information how ever it likes to eventually come up with a "bestmove" (=move suggestion). Shouldn't the search be able to handle robustly also extreme cases where value is very high or low, and cases where policy is very sharp or very flat? In my opinion it should. Latest release handles this position fine (tested up to 20M nodes).

I actually have now a feeling that there is a misunderstanding. I literary mean that this pr will not provide any move suggestion with mentioned settings, i.e. search never completes. Maybe infinite loop or what ever kind of bug. I don't mean that I am not happy with the suggested move but there is no move suggestion at all.

Naphthalin:
> Could you elaborate how you think it is fault of NN? The NN provides value and policy for a given position. Then search uses this information how ever it likes to eventually come up with a "bestmove" (=move suggestion). Shouldn't the search be able to handle robustly also extreme cases where value is very high or low, and cases where policy is very sharp or very flat? In my opinion it should. 

Yes of course, but there is still a horizon effect because of the limited search. beta-MCTS tries to use the searched nodes more efficiently than standard MCTS and giving more accurate evals, but it still needs to explore moves/lines including okish evals from the NN to correctly evaluate positions and pick the right moves. I especially don't change anything about the exploration which is still done by PUCT, and PUCT mostly isn't able to deal with the extreme cases you describe.

> Latest release handles this position fine (tested up to 20M nodes).
> I actually have now a feeling that there is a misunderstanding. I literary mean that this pr will not provide any move suggestion with mentioned settings, i.e. search never completes. Maybe infinite loop or what ever kind of bug. I don't mean that I am not happy with the suggested move but there is no move suggestion at all.

I tested it, and with standard settings search also just stops for me at ~340k nodes. I'll have to find out what the reason is as I can't think of any changes that might result in this behavior. Thanks for bringing that to my attention!

Naphthalin:
> I actually have now a feeling that there is a misunderstanding. I literary mean that this pr will not provide any move suggestion with mentioned settings, i.e. search never completes. Maybe infinite loop or what ever kind of bug. I don't mean that I am not happy with the suggested move but there is no move suggestion at all.

Apparently the implementation of the incomplete beta function I used doesn't like being called with values >1000000, which was easier to fix than to find. Thx for the finding! Should work now.

jkormu:
Thanks for the fix! Seems to be stable now. Before it did not take too many games to hit this issue, now couple hundred games with extreme --betamcts-trust without problems.

src/mcts/node.cc
```diff
@@ -332,9 +332,10 @@ void Node::FinalizeScoreUpdate(float v, float d, int multivisit) {
           const auto n = child.GetNBetamcts();
           const auto r = child.edge()->GetRBetamcts();
```
SunnyWar: move this inside the conditional?

Naphthalin: I think you can't declare variables within conditionals, even when you only use them within the conditional? At least when I tried it the compiler didn't like it...

SunnyWar: I meant inside the body of the condition.
'         if (n > 0) {
            **const auto r = child.GetRBetamcts();** 
            const auto visits_eff = r * n; 
            n_temp += visits_eff; 
            // Flip Q for opponent. 
            q_temp += -child.node()->GetQBetamcts() * visits_eff; 
          }'

gonzalezjo:
Regarding the performance issues with `lgamma` and `xinbta` brought up on Discord: 


The issues with `lgamma` seem to stem from log calculations. A variation of some standard library's lgamma routine, but with log replaced with some fastlog routine, should be sufficient to make `lgamma` workable.


`xinbta` I'm less sure about, but dropping `sae` in `fastxinbta` to ~5 produces a speedup on the order of 20x without any decrease in accuracy that I was able to note.

Naphthalin:
@gonzalezjo thx for that find -- as you can see from the commit history, I played around a bit with the accuracy and got some improvements, but not enough for it to be viable. However I now circumvent it by approximating the beta distribution with a logit normal distribution, which replaces the `betain` calls to `erf` calls, and has on top of this the desired behavior around +1 and -1.

Naphthalin:
Found and extracted a very interesting corner case: Accidentally using the inflate terminals logic where terminal draws got artificially accurate evals, with `--betamcts-prior=42` and `--betamcts-trust=0.1` it is possible that evals don't converge, but jump around periodically in a very small region. When starting with an eval outside of this region, evals converge to the desired point.
```
Repeating score update. Move stats: N_eff 56477.9, q=-0.0416077
Child: q=0.000793022, n_eff=2.73582e+06, r=0.00189466
Child: q=0, n_eff=2.77889e+06, r=0.00158551
Child: q=0.00156321, n_eff=1.81644e+06, r=0.00233607
Child: q=0.0548521, n_eff=34673.5, r=1.22988
```
leads to an oscillation like
```
iteration 100, q_old: -0.0500736, q_new: -0.0521098, diff: -0.00203621
iteration 101, q_old: -0.0521098, q_new: -0.0475661, diff: 0.00454371
iteration 102, q_old: -0.0475661, q_new: -0.0411692, diff: 0.00639687
iteration 103, q_old: -0.0411692, q_new: -0.0417502, diff: -0.000581026
iteration 104, q_old: -0.0417502, q_new: -0.0489617, diff: -0.00721143
iteration 105, q_old: -0.0489617, q_new: -0.0521938, diff: -0.00323215
iteration 106, q_old: -0.0521938, q_new: -0.0488272, diff: 0.00336662
iteration 107, q_old: -0.0488272, q_new: -0.0421346, diff: 0.00669263
```
However, after removing the terminal inflation heuristic, I couldn't produce this behavior anymore.

src/mcts/search.cc
```diff
@@ -1238,10 +1289,13 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend(
...
       }
 
-      const float Q = child.GetQ(fpu, draw_score);
+      const float Q = child.GetQ(fpu, draw_score,
+                                 params_.GetBetamctsLevel() >= 2);
       const float M = m_evaluator.GetM(child, Q);
 
-      const float score = child.GetU(puct_mult) + Q + M;
+      const float score = child.GetU(puct_mult, params_.GetAprilFactor(),
+                              params_.GetAprilFactorParent()) + Q + M;
```
Menkib64, line 1297: Should this use betamcts to select nodes?

My understanding is that weight reducing a weight of branch should mean its visit count should be reduced. The reduced visit count should lead to larger uncertainty here.

src/mcts/search.cc
```diff
@@ -1185,6 +1226,13 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend(
         return NodeToProcess::Visit(node, depth);
       }
     }
+
+    // betamcts::calculate relevances only every X visits
+    if ((node->GetNStarted() + 1 ) % params_.GetBetamctsUpdateInterval() == 0) {
+      node->CalculateRelevanceBetamcts(params_.GetBetamctsTrust(),
```
Menkib64, line 1232: Should relevance update happen only in backup propagation when scores are updated?

src/mcts/node.h
```diff
@@ -151,17 +162,40 @@ class Node {
   // Returns sum of policy priors which have had at least one playout.
   float GetVisitedPolicy() const;
   uint32_t GetN() const { return n_; }
+  float GetNBetamcts() const { return n_betamcts_; } /* betamcts::return effective N */
   uint32_t GetNInFlight() const { return n_in_flight_; }
   uint32_t GetChildrenVisits() const { return n_ > 0 ? n_ - 1 : 0; }
   // Returns n = n_if_flight.
   int GetNStarted() const { return n_ + n_in_flight_; }
-  float GetQ(float draw_score) const { return wl_ + draw_score * d_; }
+  float GetNStartedBetamcts() const { return n_betamcts_ + n_in_flight_; }
```
Menkib64, line 170: Should `n_in_flight_` be scaled based relevance?

src/mcts/search.cc
```diff
@@ -670,6 +696,13 @@ std::vector<EdgeAndNode> Search::GetBestChildrenNoTemperature(Node* parent,
 
         // Neither is terminal, use standard rule.
         if (a_rank == kNonTerminal) {
+          // Prefer largest n_betamcts * relevance when using betamcts.
+          // At level = 0, this just equates to equal visits.
+          if (betamcts_q && a.GetLCBBetamcts(trust, prior, percentile) !=
+              b.GetLCBBetamcts(trust, prior, percentile)) {
```
Menkib64, line 702: Would it make sense to use Q betamcts as the primary selection factor for the best child?

That would mean pruning code and early stopping code would need to be based probability that given child will become the best Q betamcts. The early stopping might also benefit from detecting children which will transposes to the same line late.

src/mcts/node.h
```diff
@@ -307,6 +343,16 @@ class Node {
   Node* best_child_cached_ = nullptr;
 
   // 4 byte fields.
+  // Average value (from value head of neural network) of all visited nodes in
+  // subtree. For terminal nodes, eval is stored. This is from the perspective
+  // of the player who "just" moved to reach this position, rather than from the
+  // perspective of the player-to-move for the position.
+
+  // betamcts needs own Q and N
+  float q_betamcts_ = 0.0f;
```
Menkib64, line 352: Could `q_betamcts_` use `wl_` instead of a new variable?

Menkib64 (commented):
I decided to have a second look into this idea. I still don't fully understand what is going on at all levels. I can also see some problems integrating it with dag.

I'm thinking that this might offer a way to better use extreme multi GPU systems. I'm thinking that early search could be looking wider without fear that bad nodes poison the evaluation.

Too bad there is going to be an issue that score updates are more expensive. That means they would have to run on multiple threads. I'm thinking that it could maybe done.

PR #965, opened 2019-10-12
fix appveyor meson version to 0.51.2 for now

borg323:
This is a temporary fix until meson 0.52.1 is released.


appveyor.yml
```diff
@@ -49,7 +49,7 @@ install:
 - cmd: IF %GTEST%==true IF NOT EXIST KQvKQ.rtbz curl --remote-name-all https://tablebase.lichess.ovh/tables/standard/3-4-5/K{P,N,R,B,Q}vK{P,N,R,B,Q}.rtb{w,z}
 - cmd: cd C:\projects\lc0
 cache:
-  - C:\cache -> appveyor.yml
+  - C:\cache
```
Tilps: Was this intended?

borg323: Yes, to avoid rebuilding the cached libraries for this change.

PR #967, opened 2019-10-17
docs: Contributing guide

nimishbongale:
Made some changes in order for this repository to be more open-source friendly! Will jump into the code soon.


mooskagh:
It doesn't look to me that any of those changes are useful.
While people posting their questions in github issues is sometimes a problem, there are no other problems otherwise. Usually bug reports are clear enough etc.

For example it looks that those templates are not reflecting our needs but just copied from somewhere.  We don't use either slack of stackoverflow.

For now I'm closing this PR but feel free to reopen if you still think there's something to be done.

nimishbongale:
Well, I thought it would help.

.github/docs folder is the recommended way of having contributing guides and other things within it rather than just having a stray CONTRIBUTING.md randomly in the repository. 

The changes I made were mainly to do with this, and the files may need to be modified depending the needs of this repository. These changes are neither necessary, nor do they break the code. But they're just a measure of having the recommended documentation within a repository.

PR #968, opened 2019-10-18
Optional alphazero time

MelleKoning:
Optionally play with alphazero time settings.

After talking with ASilver on Discord, found the explanation that the original alphazero used some different time setting, which is just use 1/20th (thus 5%) of the remaining time available for the upcoming move.

image

Additionally I just added the wtime (the time increment) for use directly for the upcoming move as well here.

Tests done:
This time setting performs worse than the default lczero time setting. 
- In only 10 games (60sec+1sec) the result was 3 wins for lczero time, 1 for aztime, 6 draws.

- Other 10 games (10sec+0.1sec) the result was even worse:

> Score of lc0.net.42850 vs lc0az.net.42850: 5 - 1 - 4 [0.700]
> Elo difference: 147.19 +/- 197.89
> 
> 10 of 10 games finished.

However, this is just an option that does not affect engine strength if you do not enable it, so might be interesting for people who want to play a bit with the setting. You can alter the value used by Alphazero, dividing remaining time always by 20 to other values from 2: take half the time left for upcoming move, to 100: take 1% of time left for upcoming move, to see what difference that makes...

Have fun testing! :)


mooskagh:
It's not clear whether we want this in our code anyway, but in any case it should be done on top of https://github.com/LeelaChessZero/lc0/pull/884 (which is waiting to be approved since long).

fischerandom:
> It's not clear whether we want this in our code anyway, but in any case it should be done on top of #884 (which is waiting to be approved since long).

Please add this so that comparison of Leela with AlphaZero can be done under the same time management scheme.

MelleKoning:
See #987 for implementation, now on top of #884

MelleKoning:
Now implemented as on top of #884

src/mcts/stoppers/factory.cc
```diff
@@ -292,13 +308,52 @@ std::unique_ptr<SearchStopper> LegacyTimeManager::CreateTimeManagementStopper(
   return std::make_unique<LegacyStopper>(deadline, &time_spared_ms_);
 }
 
+std::unique_ptr<SearchStopper>
+LegacyTimeManager::CreateAlphaZeroTimeManagementStopper(
+    const OptionsDict& options, const GoParams& params,
+    const Position& position) {
+  const bool is_black = position.IsBlackToMove();
+  const optional<int64_t>& time = (is_black ? params.btime : params.wtime);
```
borg323: std::optional

src/mcts/stoppers/factory.cc
```diff
@@ -292,13 +308,52 @@ std::unique_ptr<SearchStopper> LegacyTimeManager::CreateTimeManagementStopper(
...
+  // If no time limit is given, don't stop on this condition.
+  if (params.infinite || params.ponder || !time) return nullptr;
+
+  const int64_t move_overhead = options.Get<int>(kMoveOverheadId.GetId());
+  const optional<int64_t>& inc = is_black ? params.binc : params.winc;
```
borg323: std::optional here as well

src/mcts/stoppers/factory.cc
MelleKoning: Thanks I had to install Toolset Visual Studio 2017 (v141), you must install the following components in Visual Studio 2019 Community edition to properly see and solve the build issue:
C++ ATL for v141 build tools (x86 & x64)
C++ MFC for v141 build tools (x86 & x64)
Done and solved

mooskagh:
Now the time manager factory allows adding the alternative time managers.
It may be easier to redo a0 time manager from scratch rather than trying to merge this PR though.

borg323:
Can I ask for an extra feature? I would like two more options to allocate a fixed amount of time for a set number of moves, so that we can say use 75% of the time in the first 40 moves, and then start allocating a fixed percentage of the remaining time to each move.

MelleKoning:
follow-up on #1201

PR #970, opened 2019-10-26
Network compatibility.

mooskagh:
"Fat Fritz" weights file is fully compatible with Lc0, but has poor minimal version signature.
Whitelist it for Lc0 to be able to use that file.

### DIFF
--- a/src/neural/loader.cc
+++ b/src/neural/loader.cc
@@ -30,6 +30,7 @@
 #include <google/protobuf/io/coded_stream.h>
 #include <google/protobuf/io/zero_copy_stream_impl_lite.h>
 #include <zlib.h>
+
 #include <algorithm>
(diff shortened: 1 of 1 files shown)


ASilver:
The reason is because the modified Fat Fritz binary also lists me as the author of the NN weights it uses, and to avoid users then complaining I was not the author if they instead used a set of Leela weights, this was the workaround provided. It also uses a completely different set of defualt UCI settings which would seriously hurt Leela's play if used instead, and since Fritz 17 will also install a fully configured Leela for users who prefer that instead, there was no need to have them somehow come up with a subpar experience. If users want to just pirate the NN weights, they can just easily do that by distributing the Fat Fritz-lc0 binary as well. It has no DRM. I am against DRM.

Issue #786, opened 2019-03-08
Support Chess960


mooskagh:
Let's have a tracking issue for Chess960 support.

The main question is how to encode castling moves as output from NN, and how to encode castling rights as input of NN.

For the output, the agreement seems to be that we should implement it as "king takes own rook" move, which actually we already do for normal chess.
https://discordapp.com/channels/425419482568196106/445928688115122176/502374216180563968
![image](https://user-images.githubusercontent.com/11266455/54036439-e5b45000-41bb-11e9-9233-d654317a45bf.png)

For the input, we need to signalize somehow, which castlings are possible.
Currently it's implemented as separate planes from king-side and queen-side castlings, which are filled with ones if side has right to castle.

For Chess960 this is not enough:
Consider the following position:
  Rooks a1, b1. King e1. Queen-side castling rights are not lost. No pieces at c1, d1.
  Is castling possible in this position?
The answer is "unknown", because it's not possible to distinguish between two situations:
1. Rook a1 was there from the beginning of the game, b1 came later.
1. Rook **b1** was there from the beginning of the game, a1 came later.

So we need different encoding.
Possible options:
1. Create a separate plane per rook location.
   Pros: similar to current approach.
   Cons: Needs NN change. Also plane "castling with rook at d1 is allowed" actually may mean two different allowed moves depending on king location.
2. Use one plane for castling rights, encode allowed castlings with entire column filled with ones.
   Pros: No need for NN change, and two planes (one per player) becomes unused.
   Cons: Encodes two different castling rights the same.
         It's not clear how worse setting column to ones is worse than setting entire plane, as it may make decision making at far corner of the board harder.
3. Use two planes for castling rights, one for king-side castling, and another for queen-side castling. 
(Reuse current castling planes for that)
Encode allowed castlings with entire column filled with ones.
   Pros: No need for NN change, planes even have a bit similar meaning to what we have now (most surely still not compatible play-wise).
         Encodes two different castling types in different planes.
   Cons: It's not clear how worse setting column to ones is worse than setting entire plane, as it may make decision making at far corner of the board harder.

I like option 3 the most.
Please share your ideas/concerns.

RedDenver:
[...]

PR #979, opened 2019-11-02
Add Chess960 support

mooskagh:
* No changes to interaction with NN yet.
* UCI_Chess960 parameter only affects how castling is reported to GUI
* In any mode engine supports both e1g1 and e1h1 style castles
* In any mode engine supports Chess960 FEN positions.


src/utils/bititer.h
```diff
@@ -35,20 +35,40 @@ namespace lczero {
...
+  return result;
+#else
+  return __builtin_ctz(value);
+#endif
+}
+
+inline unsigned long GetHighestBit32(std::uint32_t value) {
+#if defined(_MSC_VER)
+  unsigned long result;
+  return _BitScanReverse(&result, value);
```
borg323: No return here.

src/utils/bititer.h
```diff
@@ -35,20 +35,40 @@ namespace lczero {
...
 #else
-    return __builtin_ctzll(value);
+  return __builtin_ctzll(value);
+#endif
+}
+
+inline unsigned long GetLowestBit32(std::uint32_t value) {
+#if defined(_MSC_VER)
+  unsigned long result;
+  return _BitScanForward(&result, value);
```
borg323: No return here.

src/chess/board.cc
```diff
@@ -979,36 +1034,61 @@ void ChessBoard::SetFromFen(const std::string& fen, int* no_capture_ply,
...
-          break;
-        default:
-          throw Exception("Bad fen string: " + fen);
+        }
+      } else {
+        throw Exception("Bad fen string (unexpected casting symbol): " + fen);
       }
     }
+    castlings_.SetRookPositions(left_rook, right_rook);
   }
```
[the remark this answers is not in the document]
borg323: Maybe there is a better way to handle this, but running the tests I noticed that `rook_positions_` was zero on occasion (undefined behavior sanitizer caught it) and this fixes it.

fischerandom:
I'm happy to see this is being worked on. I have a few comments after briefly looking at the code. Bobby Fischer did not like the name "Chess 960" which was invented in order to not associate the chess variant with his name, after his live radio interviews.

Susan Polgar spent countless hours working on finalizing the Fischer Random rules with Bobby Fischer. A few months ago I asked her on Twitter what terminology she preferred, and she answered "Fischer Random".

I understand that regarding the use of the term Chess 960 there is already some standard defined in software, communication protocols, and documentation that use the term "Chess 960". But where ever it is possible to use the term "Fischer Random", or "F.R. Chess", please use that instead. For the old chess Fischer used the term "Classical Chess".

IMHO constant and variable names containing **ooo** and **oo** is less informative, and looks worse to my eye, than using Bobby Fischer's terminology of **"a"-side** and **"h"-side** castling. Please see https://github.com/LeelaChessZero/lc0/issues/786

I suggest to use: **a_side** and **h_side** in constant and variable names, etc. For example, instead of:
static const Move kChess960oo = Move("e1h1");
static const Move kLegacyOo = Move("e1g1");
static const Move kChess960ooo = Move("e1a1");
static const Move kLegacyOoo = Move("e1c1");
if (move == kChess960oo) return kLegacyOo;
if (move == kChess960ooo) return kLegacyOoo;

You could use something like:
static const Move kFRChess_h_side_castling = Move("e1h1");
static const Move kClassicalChess_h_side_castling = Move("e1g1");
static const Move kFRChess_a_side_castling = Move("e1a1");
static const Move kClassicalChess_a_side_castling = Move("e1c1");
if (move == kFRChess_h_side_castling) {
  return kClassicalChess_h_side_castling;
}
if (move == kFRChess_a_side_castling) {
  return kClassicalChess_a_side_castling;
}

And for:
const OptionId kUciChess960{
"chess960", "UCI_Chess960",
"Castling moves are encoded as \"king takes rook\"."};

You could use something like:
const OptionId kUciFRChess {
"chess960", "UCI_Chess960",
"Castling moves are encoded as \"King-takes-Rook\"."};

Note that in **My 60 Memorable Games**, Fischer uses a capital first character for "King, Queen, Rook, Bishop, Knight, Pawn", and "White, Black". Doesn't that look much nicer? In Bobby Fischer Teaches Chess there are even examples of "King-takes-Rook" (instead of "King takes Rook" or "king takes rook". May be something to consider.

src/utils/bititer.h
mooskagh: Applied the patch that you suggested (with some changes).

fischerandom:
Why use the digit '0' instead of capital 'O'?:
castlings_.reset_we_can_00();
castlings_.reset_we_can_000();

Anyway I would use F.R. Chess terminology:
castlings_.reset_we_can_castle_a_side();
castlings_.reset_we_can_castle_h_side();

Why use magic constants and then a /* comment */ explaining them:
do_castling(6 /* g1 */, to.as_int(), 5 /* f1 */);
do_castling(2 /* c1 */, to.as_int(), 3 /* d1 */);

Better is to use an enum mask defining A..H files, for example:
enum file : uint8_t {
  a_file, b_file, c_file, d_file, e_file, f_file, g_file, h_file
};

Then:
do_castling(g_file, to.as_int(), f_file);
do_castling(c_file, to.as_int(), d_file);

src/chess/board.cc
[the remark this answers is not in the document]
borg323: With the new bitfield encoding this should not be an issue any more.

src/chess/board.cc
```diff
@@ -450,46 +447,42 @@ MoveList ChessBoard::GeneratePseudolegalMoves() const {
...
-          result.back().SetCastling();
+        return true;
+      };
+      // @From may be less or greater than @to. @To is not included in check.
+      auto range_attacked = [this](int from, int to) {
+        const int increment = from < to ? 1 : -1;
+        while (true) {
+          if (IsUnderAttack(from)) return true;
+          if (from == to) break;
+          from += increment;
```
borg323: Despite the comment, `to` is included in the check to cover the case `from == to`. Suggested alternative (saves a call to `IsUnderAttack()`):
```
      // @From may be less or greater than @to. @To is not included in check
      // unless it is the same with @from.
      auto range_attacked = [this](int from, int to) {
        if (from == to) return IsUnderAttack(from);
        const int increment = from < to ? 1 : -1;
        while (from != to) {
          if (IsUnderAttack(from)) return true;
          from += increment;
        }
        return false;
      };
```

src/chess/board.h
```diff
@@ -125,23 +133,46 @@ class ChessBoard {
...
 
+    uint8_t queenside_rook() const { return queenside_rook_; }
+    uint8_t kingside_rook() const { return kingside_rook_; }
+    void SetRookPositions(std::uint8_t left, std::uint8_t right) {
+      queenside_rook_ = left;
+      kingside_rook_ = right;
+    }
+
    private:
+    // Bitmask of initial rook positions.
```
borg323: Still calling it a bitmask is a bit of a stretch.

src/chess/board.cc
```diff
@@ -450,46 +447,42 @@ MoveList ChessBoard::GeneratePseudolegalMoves() const {
...
-        bool can_castle = true;
-        for (int i = 5; i < 7; ++i) {
-          if (our_pieces_.get(i) || their_pieces_.get(i)) {
-            can_castle = false;
-            break;
-          }
+      auto walk_free = [this](int from, int to, int rook, int king) {
+        for (int i = from; i <= to; ++i) {
+          if (i == rook || i == king) continue;
+          if (our_pieces_.get(i) || their_pieces_.get(i)) return false;
```
borg323: The following seems to be marginally faster, but probably not worth it:
```
        uint64_t mask = (2 << to) - (1 << from) - (1 << rook) - (1 << king);
        return !(mask & (our_pieces_.as_int() | their_pieces_.as_int()));
```

mooskagh: Maybe it worth it, but I don't understand how it works. Will look again tomorrow. :)

mooskagh:
Addressed review comments.

For using enums instead of magic constants, good idea, was lazy to do that before!

As for 000 vs OOO vs queenside_castle, I'd prefer shorter string and I just picked what looked neater in that context.
For naming chess pieces from capital letter, I'd stick to English rules, where they are written from lowercase letter.

src/chess/board.cc
```diff
@@ -719,6 +734,37 @@ bool ChessBoard::IsUnderAttack(BoardSquare square) const {
...
+      move1.from().row() != 0 || move1.to().row() != 0 ||
+      move2.to().row() != 0) {
+    return false;
+  }
+  const bool move1_is_castling = our_pieces_.get(move1.to()) ||
+                                 abs(move1.to().col() - move1.from().col()) > 1;
+  const bool move2_is_castling = our_pieces_.get(move2.to()) ||
+                                 abs(move2.to().col() - move2.from().col()) > 1;
+  // If both moves are not castlings, still can be king move.
+  if (!move1_is_castling || !move2_is_castling) return move1 == move2;
```
borg323: We have already checked whether `move1 == move2`, a `return false` is enough.

mooskagh: Done.

src/chess/board.cc
borg323: Before simplification, the left part was `((1 << (to + 1)) - 1) - ((1 << from) - 1)` to make a bitmask with bits from `from` to `to` set, and the right part clears the bits for rook and king.

src/chess/board_test.cc
```diff
@@ -166,6 +178,1947 @@ TEST(ChessBoard, MoveGenPosition6) {
...
+     {26, 848, 23387, 741674, 21591790, 675163653}},  // 476
+    {"rn1bkrb1/1ppppp1p/pn4p1/8/P2q3P/3P4/NPP1PPP1/RN1BKRBQ w FAfa - 1 9",
+     {22, 803, 18322, 632920, 15847763, 536419559}},  // 477
+    {"rn1krbbq/pppp1npp/4pp2/8/4P2P/3P2P1/PPP2P2/RNNKRBBQ w EAea - 1 9",
+     {29, 810, 23968, 670500, 20361517, 575069358}},  // 478
+    {"rnn1rqbb/ppkp1pp1/2p1p2p/2P5/8/3P1P2/PP2P1PP/RNNKRQBB w EA - 0 9",
+     {22, 506, 11973, 292344, 7287368, 189865944}},  // 479
+    {"bbqr1knr/pppppp1p/8/4n1p1/2P1P3/6P1/PPQP1P1P/BB1RNKNR w HDhd - 0 9",
+     {26, 650, 18253, 481200, 14301029, 394943978}},  // 480
+    {"bq1bnknr/pprppp1p/8/2p3p1/4PPP1/8/PP
```
borg323: An additional test:
```
TEST(ChessBoard, CastlingIsSameMove) {
  ChessBoard board;
  board.SetFromFen("r3k2r/ppp1bppp/2np1n2/4p1B1/4P1b1/2NP1N2/PPP1BPPP/R3K2R w KQkq - 0 1");
  EXPECT_TRUE(board.IsSameMove("e1c1","e1c1"));
  EXPECT_TRUE(board.IsSameMove("e1a1","e1a1"));
  EXPECT_TRUE(board.IsSameMove("e1c1","e1a1"));
  EXPECT_FALSE(board.IsSameMove("e1c1","e1g1"));
  EXPECT_FALSE(board.IsSameMove("e1a1","e1h1"));
  EXPECT_FALSE(board.IsSameMove("e1c1","e1h1"));
  EXPECT_FALSE(board.IsSameMove("e1a1","e1g1"));
}
```

src/chess/board.cc
```diff
@@ -719,6 +733,41 @@ bool ChessBoard::IsUnderAttack(BoardSquare square) const {
   return false;
 }
 
+bool ChessBoard::IsSameMove(Move move1, Move move2) const {
```
borg323: The way this is written, "e1a1", "e1b1" and "e1c1" are considered the same.

mooskagh: Only if they all are king's castlings, right? And it's not possible for all of them to be legal king's castlings.

borg323: I meant for normal chess, so only "e1c1" and "e1a1" are legal but "e1b1" is also accepted. What happens is that moving the king two or more places is considered a castling move, even if there is no rook there. This way the "e1c1" notation works OK, but maybe this needs to be stricter, and only return true for `IsSameMove("e1c1", "e1a1")` and `IsSameMove("e1g1", "e1h1")` (and vice versa).

src/chess/board.cc
```diff
@@ -719,6 +734,38 @@ bool ChessBoard::IsUnderAttack(BoardSquare square) const {
...
+    return move2.to().col() > move2.from().col();
+  }
+  return move2.to().col() < move2.from().col();
+}
+
+Move ChessBoard::GetLegacyMove(Move move) const {
+  if (our_king_ != move.from() || !our_pieces_.get(move.to())) {
+    return move;
+  }
+  if (move == Move(E1, H1)) return Move(E1, H1);
```
borg323: `return Move(E1, G1);`

mooskagh: Done.

src/chess/board.cc
```diff
@@ -588,28 +582,67 @@ bool ChessBoard::ApplyMove(Move move) {
...
+      pawns_ &= kPawnMask;
+      our_pieces_.reset(our_king_);
+      our_pieces_.reset(rook_src);
+      rooks_.reset(rook_src);
+      our_pieces_.set(king_dst);
+      our_pieces_.set(rook_dst);
+      rooks_.set(rook_dst);
+      our_king_ = king_dst;
+    };
+    if (from_row == 0 && to_row == 0) {
```
borg323: Use RANK_1 here.

src/chess/board.cc
```diff
@@ -588,28 +582,67 @@ bool ChessBoard::ApplyMove(Move move) {
...
+        // Castling.
+        if (to_col > from_col) {
+          // Kingside.
+          do_castling(G1, to.as_int(), F1);
+        } else {
+          // Queenside.
+          do_castling(C1, to.as_int(), D1);
+        }
+        return false;
+      } else if (to_col - from_col > 1) {
```
borg323: Maybe make this explicit? `to_col == FILE_C && from_col == FILE_E`

src/chess/board.cc
```diff
@@ -588,28 +582,67 @@ bool ChessBoard::ApplyMove(Move move) {
...
+        // Non FRC-style e1g1 castling (as opposed to e1h1).
+        do_castling(G1, H1, F1);
+        return false;
+      } else if (from_col - to_col > 1) {
```
borg323: Same here.

src/chess/board.cc
```diff
@@ -660,11 +677,9 @@ bool ChessBoard::ApplyMove(Move move) {
   }
 
   // Reset castling rights.
-  if (from.as_int() == 0) {
-    castlings_.reset_we_can_000();
-  }
-  if (from.as_int() == 7) {
-    castlings_.reset_we_can_00();
+  if (from_row == 0 && rooks_.get(from)) {
```
borg323: `RANK_1`

src/chess/board.cc
```diff
@@ -885,11 +925,17 @@ bool ChessBoard::IsLegalMove(Move move,
     }
   }
 
-  // Castlings were checked earlier.
-  // Moreover, no pseudolegal king moves to an attacked square are generated.
-  // If it's king's move at this moment, its certainly legal.
+  // King moves.
   if (from == our_king_) {
-    return true;
+    if (from.row() != 0 || to.row() != 0 ||
```
borg323: `RANK_1`

src/chess/board.cc
```diff
@@ -979,43 +1025,68 @@ void ChessBoard::SetFromFen(const std::string& fen, int* no_capture_ply,
   }
 
   if (castlings != "-") {
+    uint8_t left_rook = 0;   // File A by default.
```
borg323: `FILE_A`

src/chess/board.cc
```diff
@@ -979,43 +1025,68 @@ void ChessBoard::SetFromFen(const std::string& fen, int* no_capture_ply,
...
+    uint8_t right_rook = 7;  // File H by default.
```
borg323: `FILE_H`

src/chess/board.cc
```diff
@@ -979,43 +1025,68 @@ void ChessBoard::SetFromFen(const std::string& fen, int* no_capture_ply,
...
-          if (our_king_.as_string() == "e1" && our_pieces_.get(0, 0) &&
-              rooks_.get(0, 0)) {
+      const bool is_black = std::islower(c);
+      const int king_col = (is_black ? their_king_ : our_king_).col();
+      if (!is_black) c = std::tolower(c);
+      const auto rooks = (is_black ? their_pieces_ : our_pieces_) & rooks_;
+      if (c == 'k') {
+        // Finding rightmost rook.
+        for (right_rook = 7; right_rook > king_col; --right_rook) {
+          if (rooks.get(is_black ? 7 : 0, right_rook)) break;
```
borg323: `RANK_?`

src/chess/board.cc
```diff
@@ -979,43 +1025,68 @@ void ChessBoard::SetFromFen(const std::string& fen, int* no_capture_ply,
...
+        }
+        if (is_black) {
+          castlings_.set_they_can_00();
+        } else {
+          castlings_.set_we_can_00();
+        }
+      } else if (c == 'q') {
+        // Finding leftmost rook.
+        for (left_rook = 0; left_rook < king_col; ++left_rook) {
+          if (rooks.get(is_black ? 7 : 0, left_rook)) break;
```
borg323: Here as well.

Issue #976, opened 2019-10-31
Make Lc0 and client Linux-friendly.


mooskagh:
Currently Lc0 tries to write lots of files into binarie's directory, making it not suitable to put Lc0 into linux distros (as natural location for lc0 binary is `/usr/bin/lc0` and an app should not write files there).

Let's make it properly!

Suggestion:
1. For windows keep everything like it is now.

2. For unix and macos do the following:

2.1. Lc0

2.1.1. Make `<autodiscover>` search networks in the following directories (in this order):
`.` (we do that already)
`./networks` (we do that already)
`~/.cache/lc0/networks` (to be added)
`/usr/share/lc0/networks` (to be added)

2.1.2 Change default config file location from `./lc0.config` to `~/.config/lc0/lc0.config`

2.1.3 Store `leelaz_opencl_tuning` file in `~/.cache/lc0/`, make it possible to change filename using `-
-backend-opts`

2.1.4 During training game generation, by default generate games in `~/.cache/lc0/<tmp-dir>` instead of `./<tmp-dir>`, and have a command line parameter to change this. (alternatevely, we can do that in `/tmp`)

2.2 client

2.2.1. Rename `client` to `lc0-client`

2.2.2. Make client first search `lc0` in the current directory, then in `$PATH`

2.2.3. By default make client cache networks in `~/.cache/lc0/client-cache` instead of `./client-cache`

mooskagh:
It's possible to do many of those changes independently, and most of them is quite easy so may be a good way to touch the repo the first time.

borg323:
For 2.2.1 I would suggest an even more descriptive name, say `lc0-training-client`. We regularly see new users trying to use the client instead of lc0. And we should probably do that before the next release.

mooskagh:
For 2.1.1, 2.1.2 and 2.2.2, I believe we currently don't search those files "in current directory", but rather "in directory of a binary".
I guess it makes sense to change it to "current directory" instead, at least for linux.

Not sure about windows.  For example, when you add an engine to GUI, GUI may not change working directory to the engine location, and engine won't be able to find weights file.

mooskagh:
Also I have no idea what file layout on macos look like, so I just assumed it's like on linux.

Ghotrix:
1. It's possible to use %APPDATA% in windows for configs and %TEMP% for networks.
2. macOS X uses ~/Library/Preferences for configs and ~/Library/Caches for not persistent stuff.
[...]

PR #982, opened 2019-11-03
allow changing opencl tuner file

borg323:
This is for the second part of point 2.1.3 in #976.

### DIFF
--- a/src/utils/optionsdict.cc
+++ b/src/utils/optionsdict.cc
@@ -128,7 +128,7 @@ class Lexer {
     }
 
     // Identifier
-    if (std::isalnum(str_[idx_])) {
+    if (std::isalnum(str_[idx_]) || str_[idx_]=='/') {
       ReadIdentifier();
       return;
     }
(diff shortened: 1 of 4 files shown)


src/utils/optionsdict.cc
```diff
@@ -128,7 +128,7 @@ class Lexer {
     }
 
     // Identifier
-    if (std::isalnum(str_[idx_])) {
+    if (std::isalnum(str_[idx_]) || str_[idx_]=='/') {
```
mooskagh: Probably also `\` for those who use windows, and `.` for relative names.
Also I'm not sure whether `ReadIndentifier` accepts `.`, `/` `\`, and `.`, but if not, probably makes sense to do that too.

mooskagh:
This doesn't fully address 2.1.3, right? By default the file is still in the old location.

src/utils/optionsdict.cc
borg323: `ReadIdentifier` accepts `_-./`, see <https://github.com/LeelaChessZero/lc0/blob/master/src/utils/optionsdict.cc#L171>. Unfortunately, `.` will not work as is here since it is already parsed with `ReadNumber`, it would need a more complicated lexer. However there is always `ReadString` for everything not caught here.

borg323:
Yes, the first part of 2.1.3 would need more work, e.g. handling of `~`.

PR #983, opened 2019-11-04
get system specific directories

borg323:
This is for #976, providing access functions to get the users cache, config and data directories using the `$XDG_*` environment variables, or system specific alternatives. An empty string is returned on windows or when no alternative can be provided.

### DIFF
--- a/src/utils/filesystem.posix.cc
+++ b/src/utils/filesystem.posix.cc
@@ -1,6 +1,6 @@
 /*
   This file is part of Leela Chess Zero.
-  Copyright (C) 2018 The LCZero Authors
+  Copyright (C) 2018-2019 The LCZero Authors
 
   Leela Chess is free software: you can redistribute it and/or modify
   it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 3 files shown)


src/utils/filesystem.posix.cc
```diff
@@ -84,4 +84,43 @@ time_t GetFileTime(const std::string& filename) {
 #endif
 }
 
+std::string GetUserCacheDirectory() {
+#ifdef __APPLE__
+  constexpr auto kLocalDir = "/Library/Caches";
```
mooskagh: Nitpicking:
here it looks like those would be absolute directories from root, maybe write it like `"Library/Caches"` here and `std::string(home) + "/" + kLocalDir` below?

Also, maybe it's time to switch to C++17 and use https://en.cppreference.com/w/cpp/filesystem/path/append ?

And also, I think it's better if it contains trailing slash. Then client code can do `GetUserCacheDirectory() + "lc0.config"` and then it will work "fine" if there's no `$HOME` defined (otherwise it will create config at `/`).

mooskagh: Just to clarify, missing trailing slash seems to be a real problem (if calling code will do `GetDir() +"/" + "filename.txt"`, e.g. on windows, it will get resolved into `/filename.txt`, and not `./filename.txt`).

Other comments are totally optional.

PR #985, opened 2019-11-09
Refactor a way UCI response is passed from search.

mooskagh:
It's a partially done refactoring from set of callbacks into callback class.
Selfplay still mostly works the old way.
It has #979 merged in, so only review after #979 is merged into master.

I'm not sure whether it really makes code better, but I tried to add WDL reporting (switchable off with UCI) and adding yet another callback filter felt like too much complication.


mooskagh:
This is now revieweable.

PR #990, opened 2019-11-14
Use actual rooks instead of rook-like pieces when determining castling.

Mardak:
r?@mooskagh or @borg323 Wasn't sure if it's preferable to do `this->rooks()` vs `ChessBoard::rooks()` or just have a different local variable name.


Mardak:
Looks like #943 made the castling rights verification better than no verification but accidentally still allowed invalid queen castling.

PR #1007, opened 2019-11-18
Convert moves to FRC-compatible encoding when building searchmoves list.

### LINKED ISSUE #1001: Crash when "go … searchmoves" for non FRC-style e1c1/e1g1 castling
```
position startpos moves e2e4 e7e6 d2d4 d7d5 c1e3 d5e4 b1d2 g8f6 c2c3 f8e7 d1c2 f6d5 d2e4 f7f5 e4g3 f5f4 e3d2 f4g3 h2g3 g7g6

go nodes 1 searchmoves e1c1
…
bestmove a1a1

go nodes 1 searchmoves e1c1
Segmentation fault: 11
```

To be clear, using e1c1 as part of `position … moves … e1c1` works fine.

mooskagh:
(also do that when unvisited move is entered through uci position command, but it worked fine without that too)

Fixes #1001

### DIFF
--- a/src/chess/board.cc
+++ b/src/chess/board.cc
@@ -759,6 +759,13 @@ Move ChessBoard::GetLegacyMove(Move move) const {
   return move;
 }
 
+Move ChessBoard::GetModernMove(Move move) const {
+  if (our_king_ != E1 || move.from() != E1) return move;
+  if (move == Move(E1, G1) && !our_pieces_.get(G1)) return Move(E1, H1);
+  if (move == Move(E1, C1) && !our_pieces_.get(C1)) return Move(E1, A1);
+  return move;
+}
+
 KingAttackInfo ChessBoard::GenerateKingAttackInfo() const {
   KingAttackInfo king_attack_info;
 
--- a/src/chess/board.h
+++ b/src/chess/board.h
@@ -101,7 +101,7 @@ class ChessBoard {
   bool IsSameMove(Move move1, Move move2) const;
   // Returns the same move but with castling encoded in legacy way.
   Move GetLegacyMove(Move move) const;
-  // Returns the same move but with castling encoded in modern 😎 way.
+  // Returns the same move but with castling encoded in modern way.
   Move GetModernMove(Move move) const;
 
   uint64_t Hash() const {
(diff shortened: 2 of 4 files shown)


src/chess/board.cc
```diff
@@ -759,6 +759,13 @@ Move ChessBoard::GetLegacyMove(Move move) const {
   return move;
 }
 
+Move ChessBoard::GetModernMove(Move move) const {
+  if (our_king_ != E1 || move.from() != E1) return move;
+  if (move == Move(E1, G1) && !our_pieces_.get(G1)) return Move(E1, H1);
```
borg323: Is the `&& !our_pieces_.get(G1)` part needed?

src/chess/board.cc
```diff
@@ -759,6 +759,13 @@ Move ChessBoard::GetLegacyMove(Move move) const {
...
+  if (move == Move(E1, C1) && !our_pieces_.get(C1)) return Move(E1, A1);
```
borg323: Same here

src/chess/board.cc
mooskagh: Yes, if there is a rook at G1, then it's already "modern-style" move and should not be converted.

src/chess/board.h
```diff
@@ -101,6 +101,8 @@ class ChessBoard {
   bool IsSameMove(Move move1, Move move2) const;
   // Returns the same move but with castling encoded in legacy way.
   Move GetLegacyMove(Move move) const;
+  // Returns the same move but with castling encoded in modern 😎 way.
```
borg323: Maybe remove the smiley? Since it is outside the Unicode BMP, I can see it causing issues with some tools.

src/engine.cc
```diff
@@ -61,10 +61,12 @@ const OptionId kShowWDL{"show-wdl", "UCI_ShowWDL",
                         "Show win, draw and lose probability."};
 
 MoveList StringsToMovelist(const std::vector<std::string>& moves,
-                           bool is_black) {
+                           const ChessBoard& board) {
   MoveList result;
   result.reserve(moves.size());
-  for (const auto& move : moves) result.emplace_back(move, is_black);
+  for (const auto& move : moves) {
+    result.push_back(board.GetModernMove({move, board.flipped()}));
```
borg323: Is there a reason for the change to `push_back`?

mooskagh: Probably there's no difference in this case, but my thoughts were that there would be only one construction of Move (inside GetModernMove()) due to RVO, but with emplace_back I was not sure.

UPD. I've just checked on godbolt and indeed they generate identical code in this case, changing to emplace_back.

PR #1016, opened 2019-11-25
build both openblas and dnnl by default

borg323:
Using the dnnl blas functions gives performance similar to mkl on recent processors. However it is considerably slower on old ones (checked on a core 2 against openblas).


mooskagh:
Would it make sense to build both? Or at least support both and switching between them by changing one line near the top of the sciprt?

borg323:
This is trivial to do, I was worried of the possible confusion of having two cpu options, but the more I think about it, the more I like it.

PR #1023, opened 2019-11-26
Allow mixed terminal values when converting parents to terminals.

### LINKED ISSUE #836: Convert more positions to terminals, e.g., all terminal children with different evals
Split from https://github.com/LeelaChessZero/lc0/pull/822#discussion_r276053057 the specific suggestion was that if there's a mix of draw and loss terminal moves, the parent could be made draw.

Potentially there are other situations that could be covered with this issue.

Mardak:
r?@Tilps Replaces #1022 by supporting differing values to also fix #836. The reason why we need to fix even `v == ±1.0f` checks is that during backup, the current terminal node's value (potentially discounted by short-sightedness) replaces the original (potentially non-terminal) value:
https://github.com/LeelaChessZero/lc0/blob/d34bdfb9afd8ab162c256ab36712958c1779600e/src/mcts/search.cc#L1224-L1227

This is necessary because two children could be backed up to a common parent within the same batch, and we don't want the later backup's value to skew the terminal value of the parent (e.g., slightly away from a draw to look slightly losing/winning thus breaking `GetBestChildrenNoTemperature`'s check for a winning terminal).

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -459,7 +459,7 @@ std::vector<EdgeAndNode> Search::GetBestChildrenNoTemperature(Node* parent,
       continue;
     }
     const auto Q = edge.GetQ(0.0f);
-    edges.emplace_back(edge.IsTerminal() && Q == 1.0f, edge.GetN(), Q,
+    edges.emplace_back(edge.IsTerminal() && Q > 0.0f, edge.GetN(), Q,
                        edge.GetP(), edge);
   }
   const auto middle = (static_cast<int>(edges.size()) > count)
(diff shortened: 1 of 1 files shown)


src/mcts/search.cc
```diff
@@ -1236,19 +1236,23 @@ void SearchWorker::DoBackupUpdateSingleNode(
...
 
-    // A non-winning terminal move needs all other moves to have the same value.
-    if (can_convert && v != 1.0f) {
+    // A non-winning terminal move needs all other moves to be similar.
+    auto all_losing = true;
+    if (can_convert && v <= 0.0f) {
       for (const auto& edge : p->Edges()) {
-        can_convert = can_convert && edge.IsTerminal() && edge.GetQ(0.0f) == v;
+        const auto Q = edge.GetQ(0.0f);
+        can_convert = can_convert && edge.IsTerminal() && Q <= 0.0f;
```
Tilps: So technically this permits declaration of draw for mix of draw and and losing terminal edges - which is probably valid - but the old code did not allow that.  Just want to make sure that is intended.

Mardak: Nod. I marked this to fix #836 implementing your suggestion back in https://github.com/LeelaChessZero/lc0/pull/822#discussion_r276053057

E.g.,
```
position fen 8/8/8/8/8/5K2/R7/7k w - - 0 1
go nodes 50000 searchmoves a2h2

Before: a2h2 N:   50135 P:  0.17% Q:  0.01117
After:  a2h2 N:   50373 P:  0.17% Q:  0.00000 (T) 

position fen 8/8/8/8/8/5K2/R7/7k w - - 0 1 moves a2h2
go nodes 1

h1g1 N:     561 P: 45.76% Q: -1.00000 (T) 
h1h2 N:   49684 P: 54.25% Q:  0.00000 (T) 
```

Mardak:
> but I'm not sure if you found every spot

Just making sure, you're referring to places where we check for terminal value matching ±1.0f? This looks to be it. But in general, this does seem a bit fragile making sure we always check for positive/negative terminals instead of the exact values. I suppose we could add various node helpers, e.g., `IsTerminalWin()`, if this becomes problematic…

Tilps:
Yes that was what I was referring to. Good that you double checked. Thanks.

PR #1024, opened 2019-11-27
update subprojects

borg323:
This is for merging after v0.23 is released. Updates various subprojects to the latest wrap. This is mostly interesting for eigen right now. However the gtest update will make including abseil cleaner, if we decide to use it.


Naphthalin:
I remember there recently was an update to subprojects; is this still relevant?

borg323:
Still relevant - the change we did was to bypass wrapdb for downloading, not an update.

PR #1031, opened 2019-12-02
rearrange linking

borg323:
The EncodePositionForNN test fails to build on macos, the linker finds two copies of the protobuf generated files. This rearranges the build process, making a static library with about half of the files that are needed by the tests and linking this to both lc0 and tests avoiding this issue and also reducing build time (with tests enabled) by avoiding compiling some files twice.


Naphthalin:
What is the current status of this PR? It has an approval and seems to be non-invasive and specific for building on mac, so a merge decision should be easy (?)

borg323:
Needs updating and some additional testing.

borg323:
We can revisit this if needed.

PR #1033, opened 2019-12-03
save training data with no visits for single legal move

borg323:


src/mcts/node.cc
```diff
@@ -338,14 +338,16 @@ V4TrainingData Node::GetV4TrainingData(GameResult game_result,
   // Populate probabilities.
   const float total_n = static_cast<float>(GetChildrenVisits());
```
mooskagh: `total_n` is actually an int, and it's only used once as `float` (as denominator in line 350).
Do you think it would be more readable to have it int (or `auto`) here, and do `static_cast` at the point of usage in line 350 instead?

borg323: Indeed, and this also revealed that `GetChildrenVisits()` returns uint32_t, so the test can be simplified.

src/mcts/node.cc
```diff
@@ -336,16 +336,18 @@ V4TrainingData Node::GetV4TrainingData(GameResult game_result,
   result.version = 4;
 
   // Populate probabilities.
-  const float total_n = static_cast<float>(GetChildrenVisits());
+  auto total_n = GetChildrenVisits();
   // Prevent garbage/invalid training data from being uploaded to server.
-  if (total_n <= 0.0f) throw Exception("Search generated invalid data!");
+  if (total_n == 0 && GetNumEdges() != 1) {
```
mooskagh: Maybe also add a short comment?
Like
```cpp
// It's possible to have N=0 when there is only one legal move in position (due to smart pruning).
```

PR #1034, opened 2019-12-03
show build identifier only if not a release

borg323:
In particular, this does not display the build identifier in the uci `id name` reply.


mooskagh:
I'm not sure whether it's a bad idea to show build identifier.. Why not? Especially useful for e.g. custom builds.
Also, there was an idea to check build-id on the server if we want to blacklist particular build of Lc0..

borg323:
If I'm not mistaken, the client gets the version from the logo, so the build identifier is included in the version sent to the server.

borg323:
I forgot to mention that this also fixes the exception text in <https://github.com/LeelaChessZero/lc0/blob/master/src/neural/loader.cc#L109-L110> that is currently wrong, (not an important point, this is trivial to fix).

borg323:
I was wrong, the client gets the version from the uci `id name` response, so with this we lose the ability to check git revisions (or "dirty" builds) on the server.

Naphthalin:
What is the current status of this PR? Was this addressed in any newer PR?

borg323:
I seem to remember we used a git version check once in training, so merging this will reduce our options. Closing for now, will resubmit when(if) we have a plan including the client.

borg323:
@lealgo had a great idea: use the build identifier if you are not on a release (which I approximate with looking if the current commit is an exact git tag).

meson.build
```diff
@@ -99,6 +99,9 @@ if git.found()
     # Now let's check if the working directory is clean.
     if run_command(git, 'diff-index', '--quiet', 'HEAD').returncode() == 0
       short_rev = r.stdout().strip()
+      if run_command(git, 'describe', '--exact-match', '--tags').returncode() == 0
```
mooskagh, line 102: This line doesn't seem to compare that the tag is indeed at the same review id?

borg323: Indeed, it isn't comparing anything - the checks are that if the tree is clean at the current commit (previous check) and that the current commit matches a tag (this check). Should I add a comment about it?

PR #1040, opened 2019-12-11
Some cuda installations use /usr/lib/cuda

borg323:
This has been reported a few times on discord, so adding the path to the defaults.


duelistone:
I was just about to make this pull request myself. It is a little annoying to have to edit meson_options.txt after each new update so leela can compile. The extra paths added in the commit are correct for my system. (I'm on Debian Buster.) Can someone merge this when they have time?

PR #873, opened 2019-06-03
Randomly play opening moves based on priors

Mardak:
r?@Tilps or @mooskagh Fixes #342 by adding selfplay `--random-opening-max-plies` that results in a random number of plies doing a single visit to randomly pick a move proportional to the (noised) priors while not saving training data for these early moves. Can also change `--random-opening-temperature` from 1.0 to larger/flatten or smaller/sharper priors.

Not explicitly mentioned in the issue but implemented here is the skipping of adjudication during this random opening to hopefully allow more opening diversity of positions that otherwise would have been overlooked if assuming opponent will play best move -> resign.

This doesn't change the existing default behavior of 0 prior-based random moves, but it allows the server to change it and even in addition to regular temperature settings.

If the game ends while playing randomly, it's treated like an abort.

This includes/supersedes #773 as I needed to move the `MakeMove` earlier anyway (but still after getting training data).

PR #1046, opened 2019-12-30
Fix training data for odd book lengths.

Tilps:


Mardak:
In https://github.com/LeelaChessZero/lc0/pull/873/files#diff-377454d6dc3a20d8aa99ca111e5cddaaR252 I had an unreviewed PR that just used `chunk.side_to_move` instead of initializing and flipping `black_to_move`

Tilps:
Confirmed via the new rescorer hack that with this change there are no games with win as loss/loss as win in the final ply - vs 20+% of games before change.

Issue #1044, opened 2019-12-30
'badgame split' can create 0 length training data


Tilps:
There is no check to see if discard move would make the game terminal - so discarded games can occasionally start from terminal states, immediately exiting and generating a 0 length training data file.

Tilps:
Note that just checking of the node the move points to is terminal would not be sufficient - would need to also check if currently at 449 ply, as 450 ply isn't node terminal but it is a selfplay game termination condition.

PR #1048, opened 2019-12-30
Don't discard games that are in positions that are already done.

Tilps:
To Fix: #1044

### DIFF
--- a/src/selfplay/game.cc
+++ b/src/selfplay/game.cc
@@ -206,7 +206,7 @@ void SelfPlayGame::Play(int white_threads, int black_threads, bool training,
                                          kMinimumAllowedVistsId.GetId())) {
         break;
       }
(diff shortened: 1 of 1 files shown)


Tilps:
I've run almost 4000 games with current settings, and the if condition has always been true.... (I locally added a log statement to the else path and its never printed.)
So either I have a bug or I maybe I am unlucky. But either way I can't confirm this is working correctly :P

Tilps:
Artificially reducing the draw definition to 10 ply leads this to trigger successfully. But I still don't know if it can trigger correctly for accidentally playing into a draw...

Tilps:
Raising the minimum visits to 200, manages to trigger in a variety of scenarios, not just from ply limit exceeded. No crashes, and its definitely not triggering excessively frequently. So I think its good.

PR #1050, opened 2020-01-09
Extern backend

Tilps:
So it seemed like a cool idea at the time, but I can't get python tensorflow on windows to do inference in under 12ms (regardless of batch size), so in the end I've not actually got a use case for this yet...

Simple spinlock design - assumes that atomic<size_t> is lockfree/addressfree.  Also since there are no yield fallbacks, also assumes the machine isn't under contention for cores...

Memory layout is:
32 byte header, 8 byte source flag, 8 byte dest flag, 8 byte size info. 8 byte padding to ensure remainder is 16 byte aligned. Source and dest flags are assumed access/modified via lock free atomics and are used to signal when data is available/processing is complete.
After header is either serialized net proto of length same as size info or an area with space for 1024 inputs and outputs. Size info in this case indicates how many slots are filled. The area is inputs, policies, wdls.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -128,6 +128,7 @@ files += [
   'src/neural/loader.cc',
   'src/neural/network_check.cc',
   'src/neural/network_demux.cc',
+  'src/neural/network_external.cc',
   'src/neural/network_legacy.cc',
   'src/neural/network_mux.cc',
   'src/neural/network_random.cc',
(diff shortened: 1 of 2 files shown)


Tilps:
Definitely doesn't work on linux - code path for that is not written properly at all.

Issue #869, opened 2019-05-30
Time Management Issue


TesseractA:
https://cd.tcecbeta.club/archive.html?season=15&div=champions&game=29

At TCEC, Most basically, Leela lost to Houdini 3 when she was +1.28 from the initial book position.

This game, Leela went from +3 to +12 in one move. If Leela does that and doesn't spend more time on the move, she will give up too early. It is unreasonable if Leela is such a good engine for her to spend so little time on one move when its clear that she has or is missing something. I'm going to write a TM patch for things like this if I can. There needs to be a gradient that encourages Leela to spend more time on moves that show large deltas between positions. If anyone else can do it before me and/or do it better than I (i'm relatively new to this) then please-- indulge yourself.

I noticed a similar issue with Allie when Allie played KMCTS at the Computer Chess Championship.

borg323:
Lets take this opportunity to do it right. The idea behind the current time management is described in <https://github.com/LeelaChessZero/lc0/pull/271> with a small refinement in <https://github.com/LeelaChessZero/lc0/pull/634>. The basic idea is that we try to predict how many moves are remaining in the game and divide the remaining time by this estimate to allocate time for the next move. (There is also the accumulated time from smart pruning that complicates things, but I'll ignore it for now). There are two problems with this approach:
1. The estimation is only taking into account the current move number and not the score. I expect we have enough data by now to derive a correction term based on the current Q.
2. The more serious issue is the simple division used to calculate the time allocated to the current move. The estimation parameters were optimized to improve Elo, resulting in a formula that is no longer a good predictor of the moves remaining. **I think the first step towards resolving such issues is to separate the remaining move estimator from the move time calculation: A new move time formula is needed** (taking the remaining moves as input) **that is tuned separately to maximize Elo.** This formula can take as additional inputs the current Q and the delta to the previous one to allocate more time when needed. (I expect this will also help a bit when playing without increment.)
3. The actual issue discussed in the first comment is probably best addressed in the smart pruning code (which is orthogonal to the above). It seems to me the best way to handle it is by a temporary reduction of the smart pruning factor to 1 (or even less) where needed.

mooskagh:
Fyi I'm currently refactoring time management code for it to be able to dynamically decide what time is, so it's better to wait with any changes other than simple formula tuning.

mooskagh:
[...]

PR #1065, opened 2020-02-04
Separate remaining moves estimator from time allocation curve

borg323:
This is my effort to separate the time allocation curve from the remaining moves estimator. The existing estimator parameters are renamed to `GameMidpointMove` and `GameMovesSteepness` (still hidden) and the defaults are changed as in <https://github.com/LeelaChessZero/lc0/issues/869#issuecomment-521695618> to serve as an estimator of just the remaining moves, one that can be straightforwardly replaced by the moves left head (https://github.com/LeelaChessZero/lc0/pull/961) if available.
This change results in a flat time allocation that seems to lose strength in most cases, so the resulting time allocation is multiplied by a new time curve to use more time early in the game. This new time curve reuses the old parameter names together with two new ones: `TimeMidpointMove`, `TimeSteepness`, `TimeFactor` and `TimeFactorSlope`. Those define the following shape:
time_curve

@kiudee was kind enough to do a first optimization pass over those parameters to get strong defaults:

I consider the new time curve shape (and parameter names) provisional, so alternatives and any other suggestions are welcome.


zz4032:
Gauntlet vs. SF:
```
   # ENGINE               :  RATING  ERROR  CFS(%)   GAMES  DRAWS(%)
   1 stockfish            :     0.0   ----    63.1    3000      53.6
   2 lc0-master           :    -2.0   11.8    72.4    1500      53.2
   3 lc0-master_PR1065    :    -6.8   11.3     ---    1500      53.9

White advantage = 97.7 +/- 3.9
Draw rate (equal opponents) = 58.7 % +/- 0.9
```
- default time control settings
- 3-move book
- 6-man TBs + adjudication
- network: Id590976
- tc: 30s/game+0.3s/move (Leela), 120s/game+1.2s/move (SF)
- average time/move: 0.8s/move (Leela)
- `concurrency 8` in Cutechess

mooskagh:
I remember that the current time constants were tuned on short TC games too, and then they worked very poor at longer games (basically all time was forced to drain on obvious moves between moves 50 and 60).
We need to make sure somehow that there won't be similar problems now.

src/mcts/stoppers/factory.cc
```diff
@@ -55,17 +55,30 @@ const OptionId kSlowMoverId{
...
-    "The move where the time budgeting algorithm guesses half of all "
-    "games to be completed by. Half of the time allocated for the first move "
-    "is allocated at approximately this move."};
-const OptionId kTimeSteepnessId{
-    "time-steepness", "TimeSteepness",
-    "\"Steepness\" of the function the time budgeting algorithm uses to "
-    "consider when games are completed. Lower values leave more time for "
-    "the endgame, higher values use more time for each move before the "
-    "midpoint."};
+const OptionId kTimeFactorId{"time-factor", "TimeFactor",
```
mooskagh: `initial-slowmover`?
At least the help needs to be changed. s/start allocations/first moves/?

src/mcts/stoppers/factory.cc
```diff
@@ -55,17 +55,30 @@ const OptionId kSlowMoverId{
...
+                             "Like slowmover, but for the start allocations."};
+const OptionId kTimeFactorSlopeId{"time-factor-slope", "TimeFactorSlope",
```
mooskagh: Hm...
Maybe
* initial-timebudget
* time-growth-slope
* time-growth-to-decay-move
* time-decay-slope

Meh.. :\

mooskagh: (feel free to push back on suggestions!)

borg323:
Superseded by #1195

mooskagh:
Actually, the idea is still relevant, it's just that time management code is moved around.

borg323:
#1195 separated the remaining moves estimator from the time management, so it achieves the objective of this. The curve I used here was just an attempt to approximate the legacy time management. The only thing still pending is the default values for the remaining moves estimator, but this is covered in <https://github.com/LeelaChessZero/lc0/issues/869#issuecomment-521695618>.

PR #1066, opened 2020-02-05
Support for non-zero draw scores (aka humbleness).

mooskagh:
aka contempt, aka armageddon (with --draw-score-pov=black --draw-score=100).
Also when trained with (--draw-score-pov=black --draw-score=100), we get Admageddon network.

Didn't test yet. Didn't look into nps impact.


mooskagh:
Also doesn't flip draw score during ponder..

mooskagh:
Discussion link: https://discordapp.com/channels/425419482568196106/425419999096733706/675416674937339944

mooskagh:
Speed test on random backend reliably shows +0.9% speedup (for some reason) on default (zero) or moderate parameter values.
On extreme parameters (+/- 100), there it up to 5% slowdown, most likely caused by tree shape change.

src/mcts/search.cc
```diff
@@ -191,12 +196,23 @@ int64_t Search::GetTimeSinceStart() const {
...
 }
 
+// Root is depth 0, i.e. even depth.
+float Search::GetDrawScore(bool is_odd_depth) const {
+  return (is_odd_depth ? params_.GetOpponentDrawScore()
+                       : params_.GetSidetomoveDrawScore()) +
+         (is_odd_depth == played_history_.IsBlackToMove()
+              ? params_.GetWhiteDrawDelta()
+              : params_.GetBlackDrawDelta());
+}
```
Naphthalin: Should we catch drawscores above 1.0 or below -1.0 in this function and cap them to 1.0 resp. -1.0? The result of the summation could exceed these values.

src/mcts/node.h
```diff
@@ -152,9 +153,13 @@ class Node {
   uint32_t GetChildrenVisits() const { return n_ > 0 ? n_ - 1 : 0; }
   // Returns n = n_if_flight.
   int GetNStarted() const { return n_ + n_in_flight_; }
+  float GetQ(float draw_score) const {
+    if (draw_score) return wl_ + draw_score * d_;
```
borg323: Why do we need the `if` here? Is it to guard against `d_` being NaN or Inf?

mooskagh: I just thought it would be a bit faster in default case that way, but tests shown no difference. Will remove.

src/mcts/params.h
```diff
@@ -99,6 +99,10 @@ class SearchParams {
     return options_.Get<std::string>(kScoreTypeId.GetId());
   }
   FillEmptyHistory GetHistoryFill() const { return kHistoryFill; }
+  int GetSidetomoveDrawScore() const { return kSidetomoveDrawScore; }
+  int GetOpponentDrawScore() const { return kOpponentDrawScore; }
+  int GetWhiteDrawDelta() const { return kWhiteDrawDelta; }
+  int GetBlackDrawDelta() const { return kBlackDrawDelta; }
```
Mardak: These seem to want to be `float`s not `int`s as you divide the provided value by `100.0f`. But why not just make the params from -1.0f to 1.0f instead of integers?

Separately, not sure if it's worthwhile to keep the internal names consistent with the option. E.g., `GetBlackDrawDelta` vs `DrawScoreBlack`

mooskagh: oops, good catch! That's why there was no performance degradation for small values, I'll retest after the fix.
Params are integers because in UCI integers are handled much nicer (we use uci string type for float).

For variable names, probably makes sense to rename them, that's a leftover from the first variant.

src/mcts/search.cc
mooskagh: It seems we work with such Qs just fine? (due to fpu they often happen to be so).
I'm a bit hesitant to put unnecessary checks for something that will be run for every visit, but probably it's negligible, so if there is any moderately valid reason for it, I can add it.

mooskagh:
Addressed review comments. Benchmark at random backend is still 2% faster for me than it used to be.

src/mcts/search.cc
Naphthalin: While it works fine with how we handle Q in general, having a draw score outside of `[-1.0, 1.0]` will place draws either above wins or below losses, therefore not maintaining the order of moves. With `w_1 >= w_2, w_1 + d_1 >= w_2 + d_2` one should always get `Q_1 >= Q_2` in my opinion.

mooskagh: Ok, then I can just add a check the parameters that 
`max(abs(white_draw_score), abs(black_draw_score)) + max(abs(sidetomove_draw_score), abs(opponent_draw_score)) <= 1`.

Then D multiplier will never be more than 1, and neither will Q.
(as `Q = W - L + draw_multiplier*D` and `W+L+D=1`)

Naphthalin: Yes, that check sounds like a sensible solution and completely addresses my concerns.

src/mcts/search.cc
```diff
@@ -502,7 +519,7 @@ EdgeAndNode Search::GetBestChildWithTemperature(Node* parent,
     }
     if (edge.GetN() + offset > max_n) {
       max_n = edge.GetN() + offset;
-      max_eval = edge.GetQ(fpu);
+      max_eval = edge.GetQ(fpu, is_odd_depth);
```
borg323: What is the intention here (and two more places later on)?

mooskagh: Not sure I understand what you ask.
We are limiting low-Q moves from being picked due to temperature. As Q incorporates draw score during visits gathering, it also makes sense to do that during application of the temperature.
`is_odd_depth = false` is because this function is currently only called for the root node.

src/mcts/search.cc
```diff
@@ -492,17 +513,19 @@ EdgeAndNode Search::GetBestChildNoTemperature(Node* parent) const {
 // Returns a child chosen according to weighted-by-temperature visit count.
 EdgeAndNode Search::GetBestChildWithTemperature(Node* parent,
                                                 float temperature) const {
+  // Only works with root node because of draw score handling.
+  assert(parent == root_node_);
```
Tilps: Rather than assert, why not remove the input parameter and rename to GetBestChildOfRootWithTemperature?

src/mcts/search.cc
borg323: I meant that (unless I'm seriously confused) the second argument to GetQ is `draw_score`,so why pass `is_odd_depth`?

mooskagh: Good catch! Fixed.

src/mcts/search.cc

Ttl:
I would like to see some tests that this helps in either regular matches against a weaker opponent or in Armageddon. I tried the adjustable drawscore when I made the WDL head over a year ago (https://github.com/Ttl/lc0/commits/drawscore), but the adjustable drawscore only lost elo against weaker opponents in my tests so I didn't make a PR for it (although it did decrease the draw rate). I never tried Armageddon though, so maybe it's useful for it-

mooskagh:
For Armageddon, I hope someone will try to train the net. The main motivation is that that may help top chess players in their preparations.

As for whether it helps with strength in any other case, I doubt that.

Also there was a suspicion that tuning it to extreme values would make it play more "fun" (but weaker), but this also didn't look like truth.

If these params will not be used, I'll revert that change for future releases.

src/mcts/search.cc
```diff
@@ -218,8 +234,10 @@ inline float ComputeCpuct(const SearchParams& params, uint32_t N,
 }  // namespace
 
 std::vector<std::string> Search::GetVerboseStats(Node* node,
-                                                 bool is_black_to_move) const {
-  const float fpu = GetFpu(params_, node, node == root_node_);
+                                                 bool is_black_to_move,
+                                                 bool is_odd_depth) const {
```
Mardak: nit: if `GetVerboseStats` is only used at root or ponder, seems like the existing and added `bool`s can be derived with some checks that include `node == root_node_`. Otherwise calls with bool arguments can get confusing, e.g., `GetVerboseStats(node, false, true);` does what again (pretend you weren't looking at the parameters here 😛 )

src/mcts/search.cc
```diff
@@ -128,24 +128,29 @@ void Search::SendUciInfo() REQUIRES(nodes_mutex_) {
   common_info.tb_hits = tb_hits_.load(std::memory_order_acquire);
 
   int multipv = 0;
-  const auto default_q = -root_node_->GetQ();
+  const auto default_q = -root_node_->GetQ(0.0);
```
Mardak: nit: pretty sure this is equivalent to just `-root_node_->GetWL();`

src/mcts/search.cc
```diff
@@ -128,24 +128,29 @@ void Search::SendUciInfo() REQUIRES(nodes_mutex_) {
...
   for (const auto& edge : edges) {
     ++multipv;
     uci_infos.emplace_back(common_info);
     auto& uci_info = uci_infos.back();
-    const auto& q = edge.GetQ(default_q);
-    if (score_type == "centipawn") {
+    const auto wl = edge.GetWL();
+    const auto d = edge.GetD();
+    const int w = static_cast<int>(std::round(500.0 * (1.0 + wl - d)));
+    const auto q = edge.GetQ(default_q, GetDrawScore(false));
```
Mardak: nit: `GetDrawScore(false)` is loop invariant

src/mcts/params.cc
```diff
@@ -201,6 +201,18 @@ const OptionId SearchParams::kMaxConcurrentSearchersId{
     "max-concurrent-searchers", "MaxConcurrentSearchers",
     "If not 0, at most this many search workers can be gathering minibatches "
     "at once."};
+const OptionId SearchParams::kDrawScoreSidetomoveId{
+    "draw-score-sidetomove", "DrawScoreSideToMove",
+    "Score of a drawn game, as seen by a player making the move."};
+const OptionId SearchParams::kDrawScoreOpponentId{
+    "draw-score-opponent", "DrawScoreOpponent",
+    "Score of a drawn game, as seen by an opponent the move."};
```
Mardak: nit: probably missing a word somewhere in "an opponent the move"

src/selfplay/game.cc
```diff
@@ -139,11 +139,11 @@ void SelfPlayGame::Play(int white_threads, int black_threads, bool training,
     auto best_eval = search_->GetBestEval();
     if (training) {
       // Append training data. The GameResult is later overwritten.
-      auto best_q = best_eval.first;
+      auto best_wl = best_eval.first;
```
Mardak: Would it be useful for q_ratio to train on Q instead of WL? Probably can be landed without deciding now.

src/mcts/search.cc

src/selfplay/game.cc
mooskagh: I thought as we stored D anyway, it with WL the training site will have more information and options.

Issue #215, opened 2018-07-25
Define a new benchmark position


mooskagh:
Currently we mostly use `startpos` as a position to measure nps.

However, startpos is not very typical position for the chess. Near startpos there are many reasonable moves with similar evals, which makes search wide, node collisions rarer, batches larger, and `nps`es greater.

We should pick another position for benchmark and declare it a standard.
It may be just a middlegame position from an arbitrary game, or we can look at logs (e.g. from WCCC) and pick position with low nps (to be closer to "lower bound" of nps).

Example:
WCCC R5 had ~40000 nps at startpos, and only ~7k at this position:
```
position startpos moves e2e4 c7c5 g1f3 d7d6 d2d4 g8f6 b1c3 c5d4 f3d4 a7a6 f1e2 e7e5 d4b3 f8e7 e1g1 c8e6 f2f4 e8g8 g1h1 f8e8 f4f5 e6d7 e2f3 b7b5 a2a3 h7h6
```

Error323:
Using my 2x 1080Ti and the following argv I'm getting 14185 nps:

`lc0 -t 4 --minibatch-size=512 --backend=multiplexing "--backend-opts=(backend=cudnn,gpu=0),(backend=cudnn,gpu=1)"`

mooskagh:
What do you have on startpos?

It's possible that that position was slow in WCCC for other reasons (long GC after previous move, running our of RAM and starting to swap etc).

Error323:
17836 nps on startpos with everything else equal.

RedDenver:
There's some interesting positions with high and low number of legal moves at: https://www.chessprogramming.org/Perft_Results

Position 2 has 48 legal moves while position 4 has only 6.

jjoshua2:
I would just use the sf ones...

mooskagh:
Hm, I wonder if our `lc0 benchmark` should have some non-startpos fen as default?
@borg323 what do you think?

ankan-ban:
Maybe add these positions:
https://www.chessprogramming.org/Perft_Results

mooskagh:
I think it's better to add some "typical midgame position" rather than tricky position for movegen.
Also it should be just one.

borg323:
I wasn't very happy using startpos by default, but it was what everybody was doing. So I like this idea, but we have to decide which position to use...

I also like the idea of testing multiple positions. This can be done as a script file, or maybe we can add an option to load a file with positions to run

[...]

PR #1069, opened 2020-02-10
Make benchmark more like Stockfish bench 

cn4750:
Add some of the SF bench positions, iterate over them while collecting nodes and times, report time and node count summary at the end.

Final output looks like:
```===========================
Total time (ms) : 430606
Nodes searched  : 27275237
Nodes/second    : 63341
```


jjoshua2:
You say you added some of the positions? How did you pick which ones? And why does comment say // Same positions as Stockfish uses.

cn4750:
@jjoshua2, all the positions but the final three were added as they are terminal or chess 960 positions.

borg323:
We use the current benchmark mode for other stuff, so removing the existing functionality is not likely to be accepted. I would add a new flag for this, say `--auto` (or anything else you like), so that running `lc0 benchmark --auto` will do it the stockfish way (and we can discuss whether it should default to true later). An alternative is to add a new top level mode, so that you run it with `lc0 benchmark-auto`, but this is probably overcomplicating it.

cn4750:
@borg323, I think that having the benchmark default to a more accurate (and longer) bench makes the most sense from a user perspective. I have enabled old bench behavior by using a flag to search only the starting position. I have changed the dx and opencl check scripts to use this flag. Regarding pgo profile generation, we probably want the larger bench to provide a better profile (unless things start taking too long).

mooskagh:
If implemented, fixes #215

cn4750:
@borg323, how about if I remove the `--starting-position` flag and replace it with a `--num-positions` flag which allows you to test a fixed number of positions from the suite? Setting it to one would test only the starting position and if you wanted a quicker than full benchmark run you could choose whatever is appropriate.

borg323:
That would be perfect - in which case we can also update the checks to take about a minute, with 5-6 different positions.

src/benchmark/benchmark.cc
```diff
@@ -25,6 +25,8 @@
   Program grant you additional permission to convey the resulting work.
 */
 
+#include <numeric>
```
borg323: I understand the style used requires this after `#include "benchmark/benchmark.h"` but before `#include "mcts/search.h"`

cn4750: Should be addressed by commit 2257352 now.

PR #1070, opened 2020-02-10
Add option to reduce concurrent search without limiting search worker count.

Tilps:
Completely untested!


Tilps:
It at least 'works' and seems to have marginally better nps with max concurrent searchers set to 1 instead of the default.  I haven't done any elo checks yet though (I can't think of any reason it would be worse though).

Tilps:
Score of lc0 dev local build vs lc0 local dev params2: 39 - 45 - 697 [0.496]
Elo difference: -2.67 +/- 7.98
Partial results with my 2080TI doing 10s+0.2s games from Chad 6 ply book using T59. Basically error bars, but slightly favourable to max concurrent searchers=1 instead of 0 (unlimited).

I don't think my machine is the ideal test case though - I would expect this to be more useful on a machine where lock contention is a big problem, in which case there should be considerably less lock contention at times. Also I think this might be too short a time control to see the improved nps I was measuring. But, it doesn't seem to be a regression :P

src/mcts/search.cc
```diff
@@ -710,6 +715,21 @@ void SearchWorker::GatherMinibatch() {
...
+      if (available > 0 &&
+          search_->pending_searchers_.compare_exchange_weak(
+              available, available - 1, std::memory_order_acq_rel)) {
+        break;
+      }
+      // This is a hard spin lock to reduce latency but at the expense of busy
+      // wait cpu usage. If search worker count is large, this is probably a bad
+      // idea.
+    }
+  }
```
mooskagh: As in multithread situation it can happen that many threads do a busy wait in that place, maybe it makes sense to add `if (!search_->IsSearchActive()) return;` after it?

Otherwise after finishing the search, every thread will do one full evaluation loop with NN request.

src/mcts/search.cc
```diff
@@ -710,6 +715,21 @@ void SearchWorker::GatherMinibatch() {
...
+    while (true) {
+      if (search_->stop_.load(std::memory_order_acquire)) return;
+      int available =
+          search_->pending_searchers_.load(std::memory_order_acquire);
+      if (available > 0 &&
+          search_->pending_searchers_.compare_exchange_weak(
+              available, available - 1, std::memory_order_acq_rel)) {
+        break;
+      }
+      // This is a hard spin lock to reduce latency but at the expense of busy
```
mooskagh: Alternatively, if values more than 1 are not useful, it can be just a mutex lock. :-P

And I kind of doubt that busy wait is faster that cv or mutex.

src/mcts/search.cc
```diff
@@ -1126,7 +1146,12 @@ int SearchWorker::PrefetchIntoCache(Node* node, int budget) {
 
 // 4. Run NN computation.
 // ~~~~~~~~~~~~~~~~~~~~~~
-void SearchWorker::RunNNComputation() { computation_->ComputeBlocking(); }
+void SearchWorker::RunNNComputation() {
+  if (params_.GetMaxConcurrentSearchers() != 0) {
```
mooskagh: Why here and not e.g. in the end of GatherMinibatch?

src/mcts/search.cc
Tilps: Note that IsSearchActive is the same as the first if check in this loop.

src/mcts/search.cc
Tilps: The difference is whether Prefetch can happen in parallel or not.  I kind of think prefetch being single thread at a time also makes sense.

src/mcts/search.cc
Tilps: I could see reason to think values up to the maximum available backend concurrency are plausible, although 1 certainly seems the most promising value.
Busy wait is easily as fast or faster than cv or mutex in best case, its not possible to reliably pass information between cpu cores faster than via atomic checks on memory, and under the assumption the machine has cores to spare, the thread doesn't have to yield with risk of having to wait for OS to reschedule it onto a core with all the associated register backup and restore.
Busy wait may on the other hand not always be the best choice (especially in scenarios other than multiple equivalent waiters, since spinlock isn't a fair lock).  But I think under the cases where this actually potentially adds value, busy wait downsides aren't likely to come into play.

src/mcts/params.cc
```diff
@@ -232,6 +236,7 @@ void SearchParams::Populate(OptionsParser* options) {
   std::vector<std::string> history_fill_opt{"no", "fen_only", "always"};
   options->Add<ChoiceOption>(kHistoryFillId, history_fill_opt) = "fen_only";
   options->Add<FloatOption>(kShortSightednessId, 0.0f, 1.0f) = 0.0f;
+  options->Add<IntOption>(kMaxConcurrentSearchersId, 0, 1000000) = 0;
```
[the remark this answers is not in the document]
Tilps: Done.

PR #1075, opened 2020-02-13
Allow different cpuct at root.

mooskagh:


Naphthalin:
Sry that I come up with this only now, but I thought a bit more about the implementation of cpuct at root, and I think instead of a delta we should implement a factor for cpuct at root.

Reasons:
1.) cpuct-scaling: with a constant difference and notable scaling, a constant difference would lead to an effectively decreasing difference between tree cpuct and root cpuct, as cpuct is always used as a multiplicative parameter
2.) adjusting npm targets in training: if we experiment again with increasing the nodes per move in training, unchanged policy targets need an increased cpuct at root, where a `x2` increase in npm means a `x sqrt(2)` increase in cpuct at root. This would be much easier to follow/implement with a factor than with a difference.

PR #1076, opened 2020-02-14
Rename DirectX 12 backend to dx12

mooskagh:
As ruled by https://discordapp.com/channels/425419482568196106/445928688115122176/676762279634796546

### DIFF
--- a/src/neural/dx/network_dx.cc
+++ b/src/neural/dx/network_dx.cc
@@ -24,6 +24,8 @@
   terms of the respective license agreement, the licensors of this
   Program grant you additional permission to convey the resulting work.
 */
+#include "network_dx.h"
+
 #include <algorithm>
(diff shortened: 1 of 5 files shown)


src/neural/dx/network_dx.cc
```diff
@@ -24,6 +24,8 @@
   terms of the respective license agreement, the licensors of this
   Program grant you additional permission to convey the resulting work.
 */
+#include "network_dx.h"
```
borg323: This one breaks the appveyor build

borg323: Seems this is more complicated... Better move it back for now.

mooskagh: I cannot, my editor formats files on save. :-P

PR #1079, opened 2020-02-15
Get rid of libproto!

mooskagh:
Works except:

* need more cleanup in meson.build, and run autogenerate script. Generated proto/net.pb.h file is temporary, will not be merged.
* Have to think what to do with older networks (I removed those https://github.com/LeelaChessZero/lc0/blob/master/src/neural/loader.cc#L119 lines because it's not possible now to modify a proto.


meson.build
```diff
@@ -38,28 +38,14 @@ has_backends = false
...
-gen = generator(protoc, output: ['@BASENAME@.pb.cc', '@BASENAME@.pb.h'],
-  arguments : ['--proto_path=@CURRENT_SOURCE_DIR@/libs/lczero-common', '--cpp_out=@BUILD_DIR@', '@INPUT@'])
+# Compiling protobufs.
+python = find_program('python3')
+gen = generator(python, output: ['@BASENAME@.pb.h'],
+  arguments : [
+    '@CURRENT_SOURCE_DIR@/scripts/compile_proto.py',
+    '--proto_path=@CURRENT_SOURCE_DIR@/libs/lczero-common',
+    '--cpp_out=@BUILD_DIR@',
+    '@INPUT@'])
```
borg323: I would rewrite this as:
```
compile_proto = find_program('scripts/compile_proto.py')
gen = generator(compile_proto, output: ['@BASENAME@.pb.h'],
  arguments : [
    '--proto_path=@CURRENT_SOURCE_DIR@/libs/lczero-common',
    '--cpp_out=@BUILD_DIR@',
    '@INPUT@'])
```
From the meson docs:
>Meson will also autodetect scripts with a shebang line and run them with the executable/interpreter specified in it both on Windows (because the command invocator will reject the command otherwise) and Unixes (if the script file does not have the executable bit set). Hence, you must not manually add the interpreter while using this script as part of a list of commands.

src/utils/protomessage.h
```diff
@@ -0,0 +1,103 @@
+#pragma once
+
+#include <cstdint>
+#include <cstring>
+#include <functional>
+#include <map>
+#include <string_view>
+#include <vector>
+
```
lealgo: Just a reminder, it needs to `#include <string>` so it can pass cross-compilation.
Also lots of warnings are silenced (not just for cross-compilation) with 
```
#undef minor
#undef major
```

mooskagh: Done!

meson.build
```diff
@@ -79,8 +67,11 @@ if run_command('checkdir.py', 'libs/lczero-common/proto').returncode() != 0
   endif
 endif
 
-pb_files = gen.process('libs/lczero-common/proto/net.proto',
+pb_files = [
+  'src/utils/protomessage.cc',
+  gen.process('libs/lczero-common/proto/net.proto',
   preserve_path_from : meson.current_source_dir() + '/libs/lczero-common/')
```
borg323: The `preserve_path_from` etc. is no longer needed.

mooskagh: I think it is needed. Otherwise meson will expect it in the root of the built directory rather than in proto/ subdir.

borg323: I wasn't sure myself, so I tested it before posting. Turn out we needed it only for the generated c++ file and not the header.

PR #1082, opened 2020-02-17
reduce msvc warnings

borg323:
This mostly silences ~150 warnings while building zlib.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -28,6 +28,10 @@ if cc.get_id() == 'clang' or cc.get_id() == 'gcc'
     add_project_arguments('-march=native', language : 'cpp')
   endif
 endif
+if cc.get_id() == 'msvc'
+  # Silence some zlib warnings.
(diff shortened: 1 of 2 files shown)


meson.build
```diff
@@ -28,6 +28,10 @@ if cc.get_id() == 'clang' or cc.get_id() == 'gcc'
     add_project_arguments('-march=native', language : 'cpp')
   endif
 endif
+if cc.get_id() == 'msvc'
+  # Silence some zlib warnings.
+  add_global_arguments('/wd4131', '/wd4267', '/wd4127', '/wd4244', '/wd4245', language : 'c')
```
mooskagh: Ideally it would be just to disable all warning when building zlib (but keep them enabled for the rest of the project), but as far as I remember it wasn't trivial to do..

borg323: With those, zlib builds without warnings. We could just use `/w` to disable all warnings for the c compiler, I think everything else we build uses the c++ compiler so it should be safe, but I don't like it.

PR #1083, opened 2020-02-17
meson, only add -march=native if compiler supports it.

lealgo:
There are compilers such as the Android NDK's clang cross-compilers that don't support -march=native. So add a check before adding the flag.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -24,7 +24,7 @@ if cc.get_id() == 'clang'
   add_project_arguments('-Wthread-safety', language : 'cpp')
 endif
 if cc.get_id() == 'clang' or cc.get_id() == 'gcc'
-  if get_option('buildtype') == 'release'
+  if get_option('buildtype') == 'release' and cc.has_argument('-march=native')
     add_project_arguments('-march=native', language : 'cpp')
   endif
 endif
--- a/meson.build
+++ b/meson.build
@@ -24,8 +24,8 @@ if cc.get_id() == 'clang'
   add_project_arguments('-Wthread-safety', language : 'cpp')
 endif
 if cc.get_id() == 'clang' or cc.get_id() == 'gcc'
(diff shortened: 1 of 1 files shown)


lealgo:
> FYI, an alternative is to use `cc.get_supported_arguments(['-march=native'])` which works even without a compiler check.

Hmm... I like it, let me give it a try.

borg323:
I don't expect it to be cleaner, it is handy if you have more options that you need to check.

lealgo:
No, but I like it because then we get a nice message from meson like this:
```
Compiler for C++ supports arguments -march=native: NO
```

The syntax I'm using is this:
```
add_project_arguments(cc.get_supported_arguments(['-march=native']), language : 'cpp')
```

Maybe adding a variable?
```
march = cc.get_supported_arguments(['-march=native'])
add_project_arguments(march), language : 'cpp')
```

borg323:
I think the single line version is cleaner.

lealgo:
> I don't expect it to be cleaner, it is handy if you have more options that you need to check.

I see what you mean now. But now we have many separate `add_project_arguments` calls instead of a single one with a variable `flags`.

lealgo:
But wait a minute... I get from the doc that the new syntax is just a helper for the old, just to not need the use of `foreach`

https://mesonbuild.com/Release-notes-for-0-43-0.html

borg323:
See an example here: <https://github.com/borg323/protobuf/blob/3.5.1/meson.build#L5-L7>
This is used to disable warnings, including msvc ones, without compiler checks. The output is a bit ugly (e.g. reporting unsupported windows flags in linux), but is simple and it works.

borg323:
Also the various `add_project_arguments()` we have cannot be merged as they are executed under very specific conditions.

lealgo:
Well, I'm OK with the single-line new syntax. This way we don't add one more condition to the if sentence.

PR #1087, opened 2020-02-21
meson, don't link pthread when host is Android.

lealgo:
Android's `libc, bionic`, provide built-in support for `pthreads`, so no additional linking is necessary.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -124,6 +124,7 @@ files += [
 ]
 includes += include_directories('src')
 
+deps += dependency('threads')
 
 #############################################################################
 ## Platform specific files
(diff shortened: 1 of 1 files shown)


lealgo:
Provided that #1079 is merged, only two PRs are needed to cross-compile to Android: this one and #1083 .
That is for the Eigen back-end. And of course a cross-file would be needed and the NDK toolchain.

borg323:
Can you try if using `deps += dependency('threads')` works unconditionally on android?

lealgo:
> Can you try if using `deps += dependency('threads')` works unconditionally on android?

It works!

```
Dependency threads found: YES
```

borg323:
Do you think this is a better approach?

lealgo:
> Do you think this is a better approach?

Yours, of course it is. This way the dependency is stated explicitly for all targets. I'll push the commit, let's see if appveyor likes it.

borg323:
Seems to work fine under windows as well.

Issue #1085, opened 2020-02-19
Implement cpuct at root as a multiplier instead of as an offset


Naphthalin:
See #1075 , where an `CpuctAtRootOffset` is implemented to allow different cpuct at root and within the tree. While I am very much in support of introducing this feature, I would suggest for the implementation to be changed to a multiplicator instead of an offset. IMO there are two reasons, both of them related to that cpuct is used as a multiplier in the `U` term:

1.) cpuct-scaling: the formula we use for scaling adds a node-dependent scaling term. This however with the current implementation means that the ratio of cpuct at root and cpuct in the tree will shrink towards 1 and therefore be not as efficient in the long run. [note: for this matter it needs to be discussed how the cpuct at root should scale as well, which might require a small code change as well]
2.) increasing nodes per move in training: there is a general notion of `higher npm in training will produce stronger games` which e.g. led to the T49xxx tests. However by now we know that changing npm will drastically change the policy targets, unless we increase cpuct at root by a factor `sqrt(x)` when npm is increased by a factor `x`. Having cpuct at root implemented as a multiplier would make this change trivial, even though calculating an appropriate offset is functionally the same.

Naphthalin:
With #1088 being merged, this becomes obsolete. Thanks for all the work!

PR #1088, opened 2020-02-22
Adjust parameters.

mooskagh:
Parameters change:

- Use the famous kiudee@ v1 params \o/
- Have a separate set of CPuct params at root (enabled by default in uci mode, disabled by default for selfplay)
- Set MaxConcurrentSearchers to 1 by default


mooskagh:
"Fixes" #1085

PR #1095, opened 2020-02-25
Add android to the default builds

borg323:
This is all thanks to @lealgo who made the cross-files and the meson.build changes


lealgo:
This is great!
And thanks to @borg for all the meson & appveyor magic. Also @crem for the libproto work.

PR #1099, opened 2020-02-26
add a default network to the release zip files

borg323:
Specifically 591226.pb.gz is included


Tilps:
I think there is a general consensus that 591226 isn't as good as some of the preceding 27 nets - so maybe consider a different random number to keep the crowd happy :P

mooskagh:
Cannot say anything regarding the network choice though. :)

PR #1105, opened 2020-03-03
simple network embedding mechanism

borg323:
This is based on the great work done by @lealgo in <https://github.com/LeelaChessZero/lc0/pull/868>. The main difference is that this version allows embedding after the fact: It looks for a zip file appended at the end of the executable that stores just an uncompressed network file (e.g. one generated by `zip -0`).

### DIFF
--- a/src/neural/factory.cc
+++ b/src/neural/factory.cc
@@ -117,8 +117,7 @@ std::unique_ptr<Network> NetworkFactory::LoadNetwork(
 
   if (net_path == kAutoDiscover) {
     net_path = DiscoverWeightsFile();
-  }
-  else   if (net_path == kEmbed) {
+  } else if (net_path == kEmbed) {
     net_path = CommandLine::BinaryName();
   } else {
     CERR << "Loading weights file from: " << net_path;
--- a/src/neural/factory.cc
+++ b/src/neural/factory.cc
@@ -48,7 +48,7 @@ const OptionId NetworkFactory::kBackendOptionsId{
     "Exact parameters differ per backend.",
     'o'};
 const char* kAutoDiscover = "<autodiscover>";
-const char* kEmbed = "<embed>";
+const char* kEmbed = "<built in>";
 
 NetworkFactory* NetworkFactory::Get() {
   static NetworkFactory factory;
--- a/src/neural/factory.cc
+++ b/src/neural/factory.cc
@@ -61,7 +61,11 @@ NetworkFactory::Register::Register(const std::string& name, FactoryFunc factory,
 }
 
 void NetworkFactory::PopulateOptions(OptionsParser* options) {
+#if defined(EMBED)
+  options->Add<StringOption>(NetworkFactory::kWeightsId) = kEmbed;
+#else
   options->Add<StringOption>(NetworkFactory::kWeightsId) = kAutoDiscover;
+#endif
   const auto backends = NetworkFactory::Get()->GetBackendsList();
   options->Add<ChoiceOption>(NetworkFactory::kBackendId, backends) =
       backends.empty() ? "<none>" : backends[0];
--- a/src/neural/factory.cc
+++ b/src/neural/factory.cc
@@ -29,6 +29,7 @@
 #include "neural/loader.h"
 
 #include <algorithm>
+#include "utils/commandline.h"
 #include "utils/logging.h"
 
 namespace lczero {
@@ -47,6 +48,7 @@ const OptionId NetworkFactory::kBackendOptionsId{
     "Exact parameters differ per backend.",
     'o'};
 const char* kAutoDiscover = "<autodiscover>";
+const char* kEmbed = "<embed>";
 
 NetworkFactory* NetworkFactory::Get() {
   static NetworkFactory factory;
@@ -111,6 +113,9 @@ std::unique_ptr<Network> NetworkFactory::LoadNetwork(
 
   if (net_path == kAutoDiscover) {
     net_path = DiscoverWeightsFile();
+  }
+  else   if (net_path == kEmbed) {
+    net_path = CommandLine::BinaryName();
   } else {
     CERR << "Loading weights file from: " << net_path;
   }
--- a/src/neural/loader.cc
+++ b/src/neural/loader.cc
@@ -120,7 +120,13 @@ std::string DecompressGzip(const std::string& filename) {
   int bytes_read = 0;
 
   // Read whole file into a buffer.
-  const gzFile file = gzopen(filename.c_str(), "rb");
+  gzFile file = gzopen(filename.c_str(), "rb");
+#ifdef _WIN32
+  if (!file && filename == CommandLine::BinaryName()) {
+    // Try again with a .exe suffix.
+    file = gzopen((filename + ".exe").c_str(), "rb");
+  }
+#endif
   if (!file) throw Exception("Cannot read weights from " + filename);
   while (true) {
     const int sz =
--- a/src/neural/loader.cc
+++ b/src/neural/loader.cc
@@ -44,67 +44,14 @@
 #include "utils/logging.h"
 #include "version.h"
 
+#ifdef _WIN32
+#include <io.h>
+#endif
+
 namespace lczero {
 
 namespace {
 const std::uint32_t kWeightMagic = 0x1c0;
-
-// Read a network file that is appended at the end of the lc0 executable,
-// followed by the network file size and a "Lc0!" magic.
-
-std::string DecompressEmbedded(std::string str) {
-  constexpr int MOD_GZIP_ZLIB_WINDOWSIZE = 15;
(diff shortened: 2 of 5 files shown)


src/neural/factory.cc
```diff
@@ -47,6 +48,7 @@ const OptionId NetworkFactory::kBackendOptionsId{
     "Exact parameters differ per backend.",
     'o'};
 const char* kAutoDiscover = "<autodiscover>";
+const char* kEmbed = "<embed>";
```
lealgo: I liked crem's suggestion of the name <built-in> here.

lealgo:
I'd like to try if this embedding method would trigger some Antivirus alert, on Windows or Android for example.

borg323:
> I'd like to try if this embedding method would trigger some Antivirus alert, on Windows or Android for example.

Good question - I uploaded two windows executables with different embedded nets to both jotti and virustotal. While jotti didn't find anything, virutotal reported a detection from windows defender, but a different one each time, probably matching some byte sequence in the net. The exact same lc0.exe without an embeded net did not trigger any reports.

src/neural/factory.cc
```diff
@@ -111,6 +117,9 @@ std::unique_ptr<Network> NetworkFactory::LoadNetwork(
 
   if (net_path == kAutoDiscover) {
     net_path = DiscoverWeightsFile();
+  }
+  else   if (net_path == kEmbed) {
```
mooskagh: Extra space, and it should be joined to the previous line.

src/neural/loader.cc
```diff
@@ -48,14 +48,97 @@ namespace lczero {
 namespace {
 const std::uint32_t kWeightMagic = 0x1c0;
 
+uint32_t read_le(const uint8_t *addr) {
```
mooskagh: `ReadLittleEndian()` to conform to the style guide.

src/neural/loader.cc
```diff
@@ -48,14 +48,97 @@ namespace lczero {
...
+}
+
+// Read a .zip file containing a network file (just stored, not compressed a
+// second time) that is appended at the end of the lc0 executable. Such a zip
+// file can be generated using, for example, "zip -0 zipfile.zip net.pb.gz",
+// "7z a -mx=0 zipfile.zip net.pb.gz" or the equivalent options of any other
+// compression utility.
+
+std::string DecompressEmbedded(std::string str) {
+  constexpr int MOD_GZIP_ZLIB_WINDOWSIZE = 15;
```
mooskagh: Using zip looks like a huge overcomplication. Why not just append the file + file length + 4-byte magic?

src/neural/loader.cc
```diff
@@ -48,14 +48,97 @@ namespace lczero {
...
   std::string buffer;
   buffer.resize(kStartingSize);
   int bytes_read = 0;
 
   // Read whole file into a buffer.
-  const gzFile file = gzopen(filename.c_str(), "rb");
+  gzFile file = gzopen(filename.c_str(), "rb");
+#ifdef _WIN32
+  if (!file && filename == CommandLine::BinaryName()) {
+    // Try again with a .exe suffix.
```
mooskagh: Why is it needed? `CommandLine()::BinaryName()` doesn't return extension? :-\ It should.

borg323: I found it returns the name as used on the command line. Since using the .exe extension is not mandatory, it wasn't returned I didn't write it - this may be shell dependent.

mooskagh: Uh, that Windows..

Apparently in windows stdlib.h there is https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/get-pgmptr
Which returns proper path to the binary and doesn't require <windows.h> or using windows kernel functions.
There's also GetModuleFileName() which does require windows.h and is a standard windows function..
Anyway, I think it's a bug in `BinaryName()` and must be fixed there somehow rather than working around it at the caller site..

borg323: I started doing something like that, but we don't need the file extension anywhere else. So I decided to avoid platform dependent code - the block here would be safe without the `#ifdef _WIN32`, as it will only trigger on windows.

borg323:
Latest version is a lot simpler, expecting at the end of lc0 the network file + file length (4 bytes in little endian) + "Lc0!" 4-byte magic, as suggested by @mooskagh.

src/neural/loader.cc
```diff
@@ -48,14 +48,78 @@ namespace lczero {
...
+}
+
+// Read a network file that is appended at the end of the lc0 executable,
+// followed by the network file size and a "Lc0!" magic.
+
+std::string DecompressEmbedded(std::string str) {
+  constexpr int MOD_GZIP_ZLIB_WINDOWSIZE = 15;
+  constexpr char magic[4] = {'L', 'c', '0', '!'};
+
+  uint8_t *end_addr = reinterpret_cast<uint8_t *>(str.data()) + str.size() - 8;
```
mooskagh: Maybe `assert(str.size() >= 8);`?

src/neural/loader.cc
```diff
@@ -48,14 +48,78 @@ namespace lczero {
 namespace {
 const std::uint32_t kWeightMagic = 0x1c0;
 
+uint32_t ReadLittleEndian(const uint8_t *addr) {
```
mooskagh: Do we really need this function? There are already places where we rely on system being little endian..
If we really need it, then consider moving it to `utils/bititer.h`.

src/neural/loader.cc
```diff
@@ -75,6 +139,10 @@ std::string DecompressGzip(const std::string& filename) {
   }
   gzclose(file);
 
+  if (filename == CommandLine::BinaryName()) {
+    return DecompressEmbedded(buffer);
```
mooskagh: It just went through gunzip (which didn't do anything as it's known to be not a gz file), and now it's going to be gunzipped again?..
That's not nice.

Ideally we should have just one place where we decompress, whether we do that from the beginning or from some offset.

There is `gzdopen()` function which accepts file descriptor instead of filename.
I think for normal weights file we should open a file ourselves and pass it.
For reading the binary we should open, fseek() and also pass.

Not sure whether it will work on Windows though..

src/neural/loader.cc
mooskagh: What if it's `.com` file? :)
I think it's still better to do that properly, this does look like a hack..

src/neural/loader.cc
borg323: Strangely enough, on windows it works fine, but on linux it decompresses fewer bytes than it should, without returnimg an error.

borg323: This is the diff I'm trying, that works on windows but not on linux (with an embedded net, normal loading is fine):
```diff
diff --git a/src/neural/loader.cc b/src/neural/loader.cc
index acdfb01..adf54ee 100644
--- a/src/neural/loader.cc
+++ b/src/neural/loader.cc
@@ -113,8 +113,26 @@ std::string DecompressGzip(const std::string& filename) {
   int bytes_read = 0;
 
   // Read whole file into a buffer.
-  gzFile file = gzopen(filename.c_str(), "rb");
-  if (!file) throw Exception("Cannot read weights from " + filename);
+  FILE* fp = fopen(filename.c_str(), "rb");
+  if (!fp) {
+    throw Exception("Cannot read weights from " + filename);
+  }
+  if (filename == CommandLine::BinaryName()) {
+    int32_t size, magic;
+    if (fseek(fp, -8, SEEK_END) || fread(&size, 4, 1, fp) != 1 ||
+        fread(&magic, 4, 1, fp) != 1) {
+      throw Exception("Error reading embedded file descriptor.");
+    }
+    if (magic != 0x2130634c) {
+      throw Exception("No embedded file detected.");
+    }
+    fseek(fp, -size-8, SEEK_END);
+  }
+  gzFile file = gzdopen(fileno(fp), "rb");
+  if (!file) {
+    fclose(fp);
+    throw Exception("Cannot process file " + filename);
+  }
   while (true) {
     const int sz =
         gzread(file, &buffer[bytes_read], buffer.size() - bytes_read);
@@ -134,7 +152,7 @@ std::string DecompressGzip(const std::string& filename) {
   gzclose(file);
 
   if (filename == CommandLine::BinaryName()) {
-    return DecompressEmbedded(buffer);
+//    return DecompressEmbedded(buffer);
   }
 
   return buffer;
```
Any ideas?

mooskagh: `fopen()` / `fclose()` are stdlib functions which have their own buffers. Probably it's not a good idea to steal a raw fileno from there.
I think `open()` / `close()` system calls are to be used instead. Not sure if they exist on windows, and whether libgzip does any buffering.

borg323: The solution with `open()`/`close()` etc. works fine on linux but not on windows (while it compiles OK). However a `fflush()` does the trick, and `dup()` makes a new fileno.

mooskagh:
Sorry for being picky in this PR. :)

PR #1107, opened 2020-03-04
Moves left head support for blas

Ttl:
Matches cudnn output with the nets I tested with (Some available in #961). Valgrind output also looks clean.

### DIFF
--- a/src/neural/blas/network_blas.cc
+++ b/src/neural/blas/network_blas.cc
@@ -47,7 +47,8 @@ namespace {
 class BlasComputation : public NetworkComputation {
  public:
   BlasComputation(const LegacyWeights& weights, const size_t max_batch_size,
-                  const bool wdl, const bool conv_policy, const int blas_cores);
+                  const bool wdl, const bool moves_left, const bool conv_policy,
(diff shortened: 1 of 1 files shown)


src/neural/blas/network_blas.cc
```diff
@@ -155,9 +166,11 @@ BlasComputation::BlasComputation(const LegacyWeights& weights,
 void BlasComputation::ComputeBlocking() {
   // Retrieve network key dimensions from the weights structure.
   const auto num_value_channels = weights_.ip1_val_b.size();
+  const auto num_moves_channels = weights_.ip1_mov_b.size();
   const auto num_value_input_planes = weights_.value.biases.size();
   const auto num_policy_input_planes = weights_.policy.biases.size();
-  const auto num_output_policy = kPolicyOutputs;
+  const auto num_moves_input_planes = weights_.moves_left.biases.size();
+  const auto num_output_policy = size_t(kPolicyOutputs);
```
borg323: The code mostly uses `static_cast<size_t>()` instead of `size_t()`

PR #1110, opened 2020-03-04
Refactor moves left head parameters into factor and slope

Naphthalin:
As already discussed on Discord after @kiudee posted a tuning result including MLH, the current `MovesLeftScale` parameter from #961 is very hard to interpret. This PR replaces the `MovesLeftScale` parameter with a `MovesLeftSlope` parameter `slope := factor/scale` which gives the direct Q bonus per shorter move, and the maximum effect is capped to `MovesLeftFactor`.

I think this pair of parameters controlling the effect of MLH on search is the combination which on the one hand is the easiest to interpret, and on the other hand is as orthogonal as possible concerning parameter tunings.

### DIFF
--- a/src/mcts/params.cc
+++ b/src/mcts/params.cc
@@ -206,8 +206,8 @@ const OptionId SearchParams::kHistoryFillId{
     "synthesize them (always, never, or only at non-standard fen position)."};
 const OptionId SearchParams::kMovesLeftFactorId{
     "moves-left-factor", "MovesLeftFactor",
-    "Maximum bonus to add to the score of a node based on how much shorter/longer "
-    "it makes when winning/losing."};
+    "Maximum bonus to add to the score of a node based on how much "
+    "shorter/longer it makes the game when winning/losing."};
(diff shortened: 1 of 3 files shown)


src/mcts/params.cc
```diff
@@ -206,20 +206,18 @@ const OptionId SearchParams::kHistoryFillId{
     "synthesize them (always, never, or only at non-standard fen position)."};
 const OptionId SearchParams::kMovesLeftFactorId{
     "moves-left-factor", "MovesLeftFactor",
-    "Bonus to add to the score of a node based on how much shorter/longer "
+    "Maximum bonus to add to the score of a node based on how much shorter/longer "
```
Ttl: nit: This line exceeds 80 characters.

This should probably be renamed too. For example `MovesLeftLimit`.

Naphthalin: I agree that the name `MovesLeftFactor` doesn't exactly transport what it is supposed to represent, but I didn't want to obfuscate things by changing a name without changing what it is doing on my own.
What do you think about something like `MovesLeftMaxEffect`?

PR #1111, opened 2020-03-05
Prefer shorter winning mates and report mate score.

### LINKED ISSUE #837: Search or move selection should prefer (shorter) mates

Mardak:
r?@mooskagh or @Ttl Fix #837 by making use of `GetM()` on terminals to not need to traverse the tree calculating mate distance. Track the type of Terminal when extending a node as well as converting ancestors to differently rank and score.

https://lichess.org/analysis/4r2k/2q2Q2/5KRb/8/8/8/8/8_w_-_-_0_1
Screen Shot 2020-03-05 at 2 02 48 AM
```
./lc0 -w 59100 -v --multipv=100 --per-pv-counters
position fen 4r2k/2q2Q2/5KRb/8/8/8/8/8 w - - 0 1
go infinite

info nodes 175975 score mate  1 multipv  1 pv g6h6
info nodes 219983 score mate  3 multipv  2 pv f7e8 h6f8 e8f8 h8h7 g6h6
info nodes  53783 score cp 1160 multipv  3 pv f7c7 e8f8 f6e6 f8e8 e6d5 e8e1 c7f7 h6f8 f7g8
…
info nodes     69 score cp -348 multipv 14 pv g6g5 c7d6 f6f5 d6d3 f5f6 d3f3
info nodes     84 score mate -3 multipv 15 pv f7e7 c7e7 f6f5 e7e4 f5f6 e4f4
info nodes    134 score mate -2 multipv 16 pv g6g7 c7f4 f6g6 f4g5
info nodes     93 score mate -2 multipv 17 pv g6g4 c7d6 f6f5 e8e5
info nodes     89 score mate -2 multipv 18 pv f7c4 c7e5 f6f7 e5e7
info nodes     64 score mate -2 multipv 19 pv f7d7 c7e5 f6f7 e8f8
info nodes    157 score mate -1 multipv 20 pv f7a2 c7f4
info nodes     89 score mate -1 multipv 21 pv f7b3 c7f4
info nodes     70 score mate -1 multipv 22 pv f6f5 c7f4
```
Notice the mate-in-1 has fewer visits than the mate-in-3 but is selected as the bestmove. Also pv14+ show not-losses are preferred then longer mates then shorter mates. And within same mate rank, most visits are preferred as before.

(Thanks @Ttl for handling terminals in #961! ~Although unclear what should be done for TB hits.~ Fixed below https://github.com/LeelaChessZero/lc0/pull/1111#issuecomment-595426120)

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -138,7 +138,11 @@ void Search::SendUciInfo() REQUIRES(nodes_mutex_) {
     const auto d = edge.GetD();
     const int w = static_cast<int>(std::round(500.0 * (1.0 + wl - d)));
     const auto q = edge.GetQ(default_q, draw_score);
-    if (score_type == "centipawn_with_drawscore") {
+    if (edge.IsTerminal() && wl != 0.0f) {
+      uci_info.mate = std::copysign(
+          std::round(edge.GetM(0.0f)) / 2 + (edge.IsTbTerminal() ? 101 : 1),
+          wl);
+    } else if (score_type == "centipawn_with_drawscore") {
       uci_info.score = 295 * q / (1 - 0.976953126 * std::pow(q, 14));
     } else if (score_type == "centipawn") {
       uci_info.score = 295 * wl / (1 - 0.976953126 * std::pow(q, 14));
@@ -478,32 +482,81 @@ std::vector<EdgeAndNode> Search::GetBestChildrenNoTemperature(Node* parent,
     PopulateRootMoveLimit(&root_limit);
   }
   // Best child is selected using the following criteria:
-  // * Is terminal win, e.g., checkmate.
+  // * Prefer shorter terminal wins / avoid shorter terminal losses.
   // * Largest number of playouts.
   // * If two nodes have equal number:
   //   * If that number is 0, the one with larger prior wins.
   //   * If that number is larger than 0, the one with larger eval wins.
-  using El = std::tuple<bool, uint64_t, float, float, EdgeAndNode>;
-  std::vector<El> edges;
+  std::vector<EdgeAndNode> edges;
   for (auto edge : parent->Edges()) {
     if (parent == root_node_ && !root_limit.empty() &&
         std::find(root_limit.begin(), root_limit.end(), edge.GetMove()) ==
             root_limit.end()) {
       continue;
     }
-    const auto WL = edge.GetWL();
-    edges.emplace_back(edge.IsTerminal() && WL > 0.0f, edge.GetN(), WL,
-                       edge.GetP(), edge);
+    edges.push_back(edge);
   }
   const auto middle = (static_cast<int>(edges.size()) > count)
                           ? edges.begin() + count
                           : edges.end();
-  std::partial_sort(edges.begin(), middle, edges.end(), std::greater<El>());
+  std::partial_sort(
+      edges.begin(), middle, edges.end(), [](const auto& a, const auto& b) {
+        // The function returns "true" when a is preferred to b.
 
-  std::vector<EdgeAndNode> res;
-  std::transform(edges.begin(), middle, std::back_inserter(res),
-                 [](const El& x) { return std::get<4>(x); });
-  return res;
+        // Lists edge types from less desirable to more desirable.
+        enum EdgeRank {
+          kTerminalLoss,
+          kTablebaseLoss,
+          kNonTerminal,  // Non terminal or terminal draw.
+          kTablebaseWin,
+          kTerminalWin,
+        };
+
+        auto GetEdgeRank = [](const EdgeAndNode& edge) {
+          const auto wl = edge.GetWL();
+          if (!edge.IsTerminal() || wl) return kNonTerminal;
+          if (edge.IsTbTerminal()) {
+            return wl < 0.0 ? kTablebaseLoss : kTablebaseWin;
+          }
+          return wl < 0.0 ? kTerminalLoss : kTerminalWin;
+        };
+
+        // If moves have different outcomes, prefer better outcome.
+        const auto a_rank = GetEdgeRank(a);
+        const auto b_rank = GetEdgeRank(b);
+        if (a_rank != b_rank) return a > b;
+
+        // If both are terminal draws, try to make it shorter.
+        if (a_rank == kNonTerminal && a.IsTerminal() && b.IsTerminal()) {
+          if (a.IsTbTerminal() != b.IsTbTerminal()) {
+            // Prefer non-tablebase draws.
+            return a.IsTbTerminal() < b.IsTbTerminal();
+          }
+          // Prefer shorter draws.
+          return a.GetM(0.0f) < b.GetM(0.0f);
+        }
+
+        // Neither is terminal, use standard rule.
+        if (a_rank == kNonTerminal) {
+          // Prefer largest playouts then eval then prior.
+          if (a.GetN() != b.GetN()) return a.GetN() > b.GetN();
+          if (a.GetWL() != b.GetWL()) return a.GetWL() > b.GetWL();
+          return a.GetP() > b.GetP();
+        }
+
+        // Both variants are winning, prefer shortest win.
+        if (a_rank > kNonTerminal) {
+          return a.GetM(0.0f) < b.GetM(0.0f);
+        }
+
+        // Both variants are losing, prefer longest losses.
+        return a.GetM(0.0f) > b.GetM(0.0f);
+      });
+
+  if (count < edges.size()) {
+    edges.resize(count);
+  }
+  return edges;
 }
 
 // Returns a child with most visits.
@@ -1057,11 +1110,13 @@ void SearchWorker::ExtendNode(Node* node) {
         }
         // If the colors seem backwards, check the checkmate check above.
         if (wdl == WDL_WIN) {
-          node->MakeTerminal(GameResult::BLACK_WON, m);
+          node->MakeTerminal(GameResult::BLACK_WON, m,
+                             Node::Terminal::Tablebase);
         } else if (wdl == WDL_LOSS) {
-          node->MakeTerminal(GameResult::WHITE_WON, m);
+          node->MakeTerminal(GameResult::WHITE_WON, m,
+                             Node::Terminal::Tablebase);
         } else {  // Cursed wins and blessed losses count as draws.
-          node->MakeTerminal(GameResult::DRAW, m);
+          node->MakeTerminal(GameResult::DRAW, m, Node::Terminal::Tablebase);
         }
         search_->tb_hits_.fetch_add(1, std::memory_order_acq_rel);
         return;
@@ -1332,6 +1387,7 @@ void SearchWorker::DoBackupUpdateSingleNode(
 
     // A non-winning terminal move needs all other moves to be similar.
     auto all_losing = true;
+    auto found_tb = n->IsTbTerminal();
     float losing_m = 0.0f;
     if (can_convert && v <= 0.0f) {
       for (const auto& edge : p->Edges()) {
@@ -1339,6 +1395,7 @@ void SearchWorker::DoBackupUpdateSingleNode(
         can_convert = can_convert && edge.IsTerminal() && WL <= 0.0f;
         if (!can_convert) break;
         all_losing = all_losing && WL < 0.0f;
+        found_tb = found_tb || edge.IsTbTerminal();
         losing_m = std::max(losing_m, edge.GetM(0.0f));
       }
     }
@@ -1355,7 +1412,8 @@ void SearchWorker::DoBackupUpdateSingleNode(
       p->MakeTerminal(
           v > 0.0f ? GameResult::BLACK_WON
                    : all_losing ? GameResult::WHITE_WON : GameResult::DRAW,
-          terminal_m);
+          terminal_m,
+          found_tb ? Node::Terminal::Tablebase : Node::Terminal::Terminal);
     }
 
     // Q will be flipped for opponent.
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -514,7 +514,7 @@ std::vector<EdgeAndNode> Search::GetBestChildrenNoTemperature(Node* parent,
 
         auto GetEdgeRank = [](const EdgeAndNode& edge) {
           const auto wl = edge.GetWL();
-          if (!edge.IsTerminal() || wl) return kNonTerminal;
+          if (!edge.IsTerminal() || !wl) return kNonTerminal;
           if (edge.IsTbTerminal()) {
             return wl < 0.0 ? kTablebaseLoss : kTablebaseWin;
           }
@@ -524,7 +524,7 @@ std::vector<EdgeAndNode> Search::GetBestChildrenNoTemperature(Node* parent,
         // If moves have different outcomes, prefer better outcome.
         const auto a_rank = GetEdgeRank(a);
         const auto b_rank = GetEdgeRank(b);
-        if (a_rank != b_rank) return a > b;
+        if (a_rank != b_rank) return a_rank > b_rank;
 
         // If both are terminal draws, try to make it shorter.
         if (a_rank == kNonTerminal && a.IsTerminal() && b.IsTerminal()) {
--- a/src/chess/callbacks.h
+++ b/src/chess/callbacks.h
@@ -67,6 +67,8 @@ struct ThinkingInfo {
   int nps = -1;
   // Hash fullness * 1000
   int hashfull = -1;
+  // Moves to mate.
+  std::optional<int> mate;
   // Win in centipawns.
   std::optional<int> score;
   // Win/Draw/Lose probability * 1000.
--- a/src/chess/uciloop.cc
+++ b/src/chess/uciloop.cc
@@ -248,6 +248,7 @@ void UciLoop::SendInfo(const std::vector<ThinkingInfo>& infos) {
     if (info.seldepth >= 0) res += " seldepth " + std::to_string(info.seldepth);
     if (info.time >= 0) res += " time " + std::to_string(info.time);
     if (info.nodes >= 0) res += " nodes " + std::to_string(info.nodes);
+    if (info.mate) res += " score mate " + std::to_string(*info.mate);
     if (info.score) res += " score cp " + std::to_string(*info.score);
     if (info.wdl) {
       res += " wdl " + std::to_string(info.wdl->w) + " " +
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -190,6 +190,7 @@ class PonderResponseTransformer : public TransformingUciResponder {
     for (const auto& info : *infos) {
       if (info.multipv <= 1) {
         ponder_info = info;
+        if (ponder_info.mate) ponder_info.mate = -*ponder_info.mate;
         if (ponder_info.score) ponder_info.score = -*ponder_info.score;
         if (ponder_info.depth > 1) ponder_info.depth--;
         if (ponder_info.seldepth > 1) ponder_info.seldepth--;
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -210,15 +210,16 @@ Edge* Node::GetOwnEdge() const { return GetParent()->GetEdgeToNode(this); }
 
 std::string Node::DebugString() const {
   std::ostringstream oss;
-  oss << " Term:" << is_terminal_ << " This:" << this << " Parent:" << parent_
-      << " Index:" << index_ << " Child:" << child_.get()
-      << " Sibling:" << sibling_.get() << " WL:" << wl_ << " N:" << n_
-      << " N_:" << n_in_flight_ << " Edges:" << edges_.size();
+  oss << " Term:" << static_cast<int>(terminal_type_) << " This:" << this
+      << " Parent:" << parent_ << " Index:" << index_
+      << " Child:" << child_.get() << " Sibling:" << sibling_.get()
+      << " WL:" << wl_ << " N:" << n_ << " N_:" << n_in_flight_
+      << " Edges:" << edges_.size();
   return oss.str();
 }
 
-void Node::MakeTerminal(GameResult result, float plies_left) {
-  is_terminal_ = true;
+void Node::MakeTerminal(GameResult result, float plies_left, Terminal type) {
+  terminal_type_ = type;
   m_ = plies_left;
   if (result == GameResult::DRAW) {
     wl_ = 0.0f;
@@ -233,7 +234,7 @@ void Node::MakeTerminal(GameResult result, float plies_left) {
 }
 
 void Node::MakeNotTerminal() {
-  is_terminal_ = false;
+  terminal_type_ = Terminal::NonTerminal;
   n_ = 0;
 
   // If we have edges, we've been extended (1 visit), so include children too.
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -130,6 +130,8 @@ class Node {
   using Iterator = Edge_Iterator<false>;
   using ConstIterator = Edge_Iterator<true>;
 
+  enum class Terminal : uint8_t { NonTerminal, Terminal, Tablebase };
+
   // Takes pointer to a parent node and own index in a parent.
   Node(Node* parent, uint16_t index) : parent_(parent), index_(index) {}
 
@@ -161,11 +163,13 @@ class Node {
   float GetM() const { return m_; }
 
   // Returns whether the node is known to be draw/lose/win.
-  bool IsTerminal() const { return is_terminal_; }
+  bool IsTerminal() const { return terminal_type_ != Terminal::NonTerminal; }
+  bool IsTbTerminal() const { return terminal_type_ == Terminal::Tablebase; }
   uint16_t GetNumEdges() const { return edges_.size(); }
 
   // Makes the node terminal and sets it's score.
-  void MakeTerminal(GameResult result, float plies_left = 0.0f);
+  void MakeTerminal(GameResult result, float plies_left = 0.0f,
+                    Terminal type = Terminal::Terminal);
   // Makes the node not terminal and updates its visits.
   void MakeNotTerminal();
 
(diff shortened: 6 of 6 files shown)


Ttl:
I did think about making this exact change, but TB nodes were a problem and they are a problem with this PR too. This gives incorrect mate scores for TB nodes since it only checks if the node is terminal.

One possible solution would be to change `is_terminal_` in `Node` class to enum and assign different values for TB nodes and other terminals. Then mate scores could be given only for non-TB terminal nodes.

Mardak:
Looking back at what I did in a different PR for TB positions, I believe I differentiated TB terminals from actual end of game terminals by adding edges to TB hits:
https://github.com/LeelaChessZero/lc0/pull/889/files#diff-83b7141f4fa67bcfdf95188467895cb5L1067

This then requires some reworking of various checks to account for the additional "type" of terminal (true terminal vs TB terminal vs propagated terminal).

Mardak:
@Ttl Would it be undesirable to have the TB terminals do `MakeTerminal(GameResult:…, m + 100)` pretending terminals have 100 additional ply? Search favoring shorter moves would bias against this terminal over other moves that report a shorter moves left, but with the changes in this PR, the final move selection would still end up picking the winning TB move if it was the only winning Terminal move found.

Mardak:
Here's a quick change to test out https://github.com/LeelaChessZero/lc0/pull/1111#issuecomment-595312166:
```diff
diff --git a/src/mcts/search.cc b/src/mcts/search.cc
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -1060,6 +1060,7 @@ void SearchWorker::ExtendNode(Node* node) {
         if (parent) {
           m = std::max(0.0f, parent->GetM() - 1.0f);
         }
+        m += 200.0f;
         // If the colors seem backwards, check the checkmate check above.
         if (wdl == WDL_WIN) {
           node->MakeTerminal(GameResult::BLACK_WON, m);
```

Same position from the original comment but with TB enabled:
```
info nodes 264315 score mate    1 multipv  1 pv g6h6
info nodes 330416 score mate  102 multipv  2 pv f7e8 h6f8 e8f8
info nodes 320047 score mate  103 multipv  3 pv f7c7 e8f8 c7f7 h6g7 g6g7
info nodes    696 score cp     -7 multipv  4 pv g6g8 e8g8 f7c7
info nodes    255 score cp   -314 multipv  5 pv f7d5 h6g7 f6g5 e8e5 g5f4 e5d5 f4f3 c7f7
info nodes    183 score cp   -277 multipv  6 pv g6g1 c7d6 f6f5 e8f8 g1g8 f8g8 f7g7 g8g7
info nodes    172 score cp   -275 multipv  7 pv g6g3 c7f4 f6g6 f4g3 g6f6 e8f8 f6e7 f8f7
info nodes    128 score cp   -802 multipv  8 pv f7f8 e8f8 f6e6 c7h7 g6g5
info nodes    123 score cp   -272 multipv  9 pv g6g2 c7d6 f6f5 e8f8 g2g8 f8g8 f7g7 g8g7
info nodes    116 score mate -181 multipv 10 pv f7g8 e8g8 g6g8
info nodes    159 score mate -103 multipv 11 pv f7g7 c7g7 f6f5 e8e5 f5g4 g7g6
info nodes    137 score mate -103 multipv 12 pv f7h7 h8h7 g6g5 h6g5
info nodes    118 score mate -102 multipv 13 pv f7e6 e8e6 f6f5 e6g6
info nodes    117 score mate -102 multipv 14 pv f6f5 c7f7 f5g4 f7g6
info nodes    110 score mate -102 multipv 15 pv g6g5 h6g5 f6g6 c7f7
info nodes    141 score mate   -3 multipv 16 pv f7e7 c7e7 f6f5 e7e4 f5f6 e4f4
info nodes    225 score mate   -2 multipv 17 pv g6g7 c7f4 f6g6 f4g5
info nodes    156 score mate   -2 multipv 18 pv g6g4 c7d6 f6f5 e8e5
info nodes    149 score mate   -2 multipv 19 pv f7c4 c7e5 f6f7 e5e7
info nodes    107 score mate   -2 multipv 20 pv f7d7 c7e5 f6f7 e8f8
info nodes    263 score mate   -1 multipv 21 pv f7a2 c7f4
info nodes    149 score mate   -1 multipv 22 pv f7b3 c7f4
```

Ttl:
That would give incorrect M value to nodes before the TB node in the tree when the M values are backpropagated.

Mardak:
Ah oops. The approach from #889 mentioned in https://github.com/LeelaChessZero/lc0/pull/1111#issuecomment-595307114 won't quite work either as the whole point of this PR is to avoid needing to traverse to the tree, so detecting attributes of terminal to determine whether it's a "true" or "tb" terminal isn't useful when converting ancestors to terminals.

Seems like something more along the lines of the suggestion in https://github.com/LeelaChessZero/lc0/pull/1111#issuecomment-595268476 to have is_terminal_ be an enum instead of bool:

1. true terminal
2. tb terminal
3. converted terminal with best descendant leading to true terminal
4. converted terminal with best descendant leading to tb terminal

So in cases (1) and (3), we can take `GetM()` as-is while for (2) and (4), we'll need to treat them as longer mates when ranking moves as well as when reporting mate score.

Mardak:
Ended up just adding 3 enum values as at least right now, we don't treat "true" terminals differently from converted terminals with "true" terminal descendants. And same thing for "tb" terminals. Not quite sure what's a good name for these "combined" behavior. `enum { None, Trueish, Tablebaseish }` 😉 

```
info nodes 486337 score mate    1 multipv  1 pv g6h6
info nodes 607960 score mate  101 multipv  2 pv f7e8 h6f8 e8f8
info nodes 588880 score mate  102 multipv  3 pv f7c7 e8f8 c7f7 h6g7 g6g7
info nodes   1073 score cp     -5 multipv  4 pv g6g8 e8g8 f7c7
info nodes    390 score cp   -311 multipv  5 pv f7d5 h6g7 f6f5 e8e5 f5f4 e5d5 f4f3 c7f7
info nodes    195 score cp   -986 multipv  6 pv f7f8 e8f8 f6e6 h6g7 e6d5 f8f5 d5e6
info nodes    189 score cp   -271 multipv  7 pv g6g2 c7d6 f6f5 e8f8 g2g6 f8f7 g6f6 d6f6
info nodes    267 score mate -102 multipv  8 pv g6g1 c7e5 f6g6 e8e6 f7e6 e5e6
info nodes    244 score mate -102 multipv  9 pv g6g3 c7f4 f6g6 f4g3 g6h6
info nodes    243 score mate -102 multipv 10 pv f7g7 c7g7 f6f5 e8e5 f5g4 g7g6
info nodes    210 score mate -102 multipv 11 pv f7h7 h8h7 g6g5 h6g5
info nodes    177 score mate -102 multipv 12 pv f7g8 e8g8 f6f5 c7f4 f5e6 g8g6
info nodes    180 score mate -101 multipv 13 pv f7e6 e8e6 f6f5 e6g6
info nodes    179 score mate -101 multipv 14 pv f6f5 c7f7 f5g4 f7g6
info nodes    168 score mate -101 multipv 15 pv g6g5 h6g5 f6g6 c7f7
info nodes    215 score mate   -3 multipv 16 pv f7e7 c7e7 f6f5 e7e4 f5f6 e4f4
info nodes    344 score mate   -2 multipv 17 pv g6g7 c7f4 f6g6 f4g5
info nodes    239 score mate   -2 multipv 18 pv g6g4 c7d6 f6f5 e8e5
info nodes    228 score mate   -2 multipv 19 pv f7c4 c7e5 f6f7 e5e7
info nodes    163 score mate   -2 multipv 20 pv f7d7 c7e5 f6f7 e8f8
info nodes    403 score mate   -1 multipv 21 pv f7a2 c7f4
info nodes    228 score mate   -1 multipv 22 pv f7b3 c7f4
```

src/mcts/search.cc
```diff
@@ -138,7 +138,10 @@ void Search::SendUciInfo() REQUIRES(nodes_mutex_) {
     const auto d = edge.GetD();
     const int w = static_cast<int>(std::round(500.0 * (1.0 + wl - d)));
     const auto q = edge.GetQ(default_q, draw_score);
-    if (score_type == "centipawn_with_drawscore") {
+    if (edge.IsTerminal() && wl != 0.0f) {
+      uci_info.mate = std::copysign(
+          edge.GetM(0.0f) / 2 + (edge.IsTbTerminal() ? 100 : 1), wl);
```
Ttl: Rounding mate score to the nearest integer before converting to integer should result in more accurate mate score for TB nodes for nets with moves left head support.

Add 101 instead of 100 for TB nodes to get +100 moves for TB nodes?

Mardak: Okay I'll switch it to `round(GetM / 2.0)`

As for 101 vs 100, I guess it's just someone's preference of whether a TB win should appear as M100 vs M101. Doesn't really matter much either way. Also I guess is 100 or 101 reasonable vs say M1000 or other?

Ttl: Was your intention to only report M100 for TB nodes? With moves left head TB nodes have estimated moves left. Adding 101 would report the mate score as 100 + estimated mate score. On TB positions the estimate should be quite accurate for certain TB nodes, I think it would be easier to parse that for TB nodes M123 means estimated 23 moves left.

Mardak: The behavior I intended and believe I implemented was plain adding 100 moves to TB positions, which with the network I've been testing with has no moves left head, so `GetM` for the converted ancestor is always ply depth to the true/TB terminal, which has a plies_left = 0. So the extra 100 moves is some generic "we need more moves because 0 plies left at the terminal is definitely inaccurate."

But I do see that with moves left head, the TB terminal inherits plies_left based on the parent's value… Which as you suggest would be reasonable to just directly use that `GetM` without the +100 approximation. So then how should we decide when to directly use that value vs approximate?

Ttl: I think the expectation of mate scores is that it's a proven mate. Using the moves left estimate could give too short estimate in some positions. I'm okay with adding 100 in those cases, but I wouldn't report mate score based on the estimate directly.

Certain and true terminal nodes have M that is exactly an integer but TB terminals don't. I was thinking of just rounding the `GetM`. It's the same as before for true terminals which I think worked correctly.

Mardak: Okay, rounding just M and adding 101 instead of 100 for TB.

src/mcts/search.cc
```diff
@@ -478,12 +482,12 @@ std::vector<EdgeAndNode> Search::GetBestChildrenNoTemperature(Node* parent,
...
   }
   // Best child is selected using the following criteria:
-  // * Is terminal win, e.g., checkmate.
+  // * Prefer shorter terminal wins / avoid shorter terminal losses.
   // * Largest number of playouts.
   // * If two nodes have equal number:
   //   * If that number is 0, the one with larger prior wins.
   //   * If that number is larger than 0, the one with larger eval wins.
-  using El = std::tuple<bool, uint64_t, float, float, EdgeAndNode>;
+  using El = std::tuple<float, uint64_t, float, float, EdgeAndNode>;
```
mooskagh: That tuple is becoming so large that it probably makes sense to replace it with a proper struct (with custom `operator<`).
It will probably make the code cleaner because there will be no need to invent values like terminal_rank.

mooskagh: Actually it's likely that just a comparator of `EdgeAndNode` would be enough, without need to store any temporary values.

src/mcts/search.cc
```diff
@@ -1330,13 +1343,15 @@ void SearchWorker::DoBackupUpdateSingleNode(
...
     auto all_losing = true;
+    auto found_tb = n->IsTbTerminal();
     float losing_m = 0.0f;
     if (can_convert && v <= 0.0f) {
       for (const auto& edge : p->Edges()) {
         const auto WL = edge.GetWL();
         can_convert = can_convert && edge.IsTerminal() && WL <= 0.0f;
         if (!can_convert) break;
         all_losing = all_losing && WL < 0.0f;
+        found_tb = found_tb || edge.IsTbTerminal();
```
mooskagh: Does it mean that if there is a short non-tb win, and long tb win, it will mark the node as tb-win, and will show 100 additional moves in mate score?

Mardak: This code path is not for winning `v > 0`, so the behavior here is if the position has all losing moves, assume the tb loss will be played because it's probably longer than an actual loss.

For found winning `v > 0`, the first one found will cause the parent to convert to loss matching the found move's TB-ness-or-not, which indeed can overestimate the moves left because a shorter win might be found that we won't explore from the current search as future PickNodeToExtend will stop at the now-terminal parent loss.
https://github.com/LeelaChessZero/lc0/blob/f6306b350a59be774d2194b0146a3f7657243729/src/mcts/search.cc#L1348-L1351

src/mcts/search.cc
Mardak: Should we leave the `float terminal_rank` temporary value computation outside of `EdgeAndNode::operator<` as there is a little bit of rank computation for terminal nodes that might benefit from a temporary value instead of computing it on each `operator<` call.

mooskagh: This function is only called once per 5 seconds, so it would be fine just with the comparator I guess.. There's not that many math there.

But if you think it makes sense to cache `terminal_rank` then maybe then use `struct`, something like this? :
```cpp
class EdgeRank {
 public:
  EdgeRank(EdgeAndNode* edge) : edge_(edge) {
    checkmate_distance_ = /* compute it */;
  }
  bool operator<(const EdgeRank& other) {
    if (checkmate_distance_ != other.checkmate_distance_) {
      return /* based on some logic */;
    }
    if (edge.GetN() != other.edge.GetN()) return edge.GetN() > other.edge.GetN();  // '>' because we'll want to sort it in descending order.
    // And so on..
  }
 private:
  EdgeAndNode edge_;
  int checkmate_distance_;  // 0 if no checkmate, negative if losing, positive if winning, +100 if tablebase.
};

Comment (Mardak) [reply src/mcts/search.cc]: Or, I think I see what you're suggesting of not having a `terminal_rank` to begin with and have those details in the operator:
```c
bool EdgeAndNode::operator<(other) {
  if (IsTerminal() && other.IsTerminal()) {
    if (WL > 0 && other.WL > 0) return shorter winning mate;
…
```

mooskagh: Yep.
But please not in `operator<` of `EdgeAndNode` but rather an external comparator. :)
Either
```cpp
struct EdgeRankComp{
   bool operator()(const EdgeAndNode& a, const EdgeAndNode& b) const {
    /* compare a and b */
}};
```
or a wrapper struct as I mentioned in the previous comment.

Or the comparator can be actually inline lambda in std::partial_sort:
```cpp
std::partial_sort(edges.begin(), middle, edges.end(), [](const auto& a, const auto& b) {
  // Compare!
});
```

Mardak: I believe https://github.com/LeelaChessZero/lc0/pull/1111/commits/7115097043a443852d6d3ee2561076f7702edd11 implements the explicit comparisons instead of relying on `float terminal_rank` with the appropriate cascading comparisons for N, WL, P.

src/mcts/search.cc
```diff
@@ -478,32 +482,67 @@ std::vector<EdgeAndNode> Search::GetBestChildrenNoTemperature(Node* parent,
...
+                            return a.GetM(0.0f) > b.GetM(0.0f);
+                          // Prefer the win / not-loss terminal.
+                        } else
+                          return a.GetWL() > 0.0f || b.GetWL() < 0.0f;
+                        // Prefer terminal win and avoid loss.
+                      } else if (a.IsTerminal() && a.GetWL() != 0.0f)
+                        return a.GetWL() > 0.0f;
+                      // Prefer other terminal loss and avoid win.
+                      else if (b.IsTerminal() && b.GetWL() != 0.0f)
+                        return b.GetWL() < 0.0f;
```
Mardak: Could these be combined somehow?

Mardak: Could these be combined somehow?

mooskagh: My try from today:
```cpp
std::partial_sort(edges.begin(), middle, edges.end(),
                  [](const auto& a, const auto& b) {
                    // The function returns "true" when a is preferred to b.

                    // Lists edge types from less desirable to more desirable.
                    enum EdgeRank {
                      kTerminalLoss,
                      kTablebaseLoss,
                      kNonTerminal,  // Non terminal or terminal draw.
                      kTablebaseWin,
                      kTerminalWin,
                    };

                    auto GetEdgeRank = [](const EdgeAndNode& edge) {
                      const auto wl = edge.GetWL();
                      if (!edge.IsTerminal() || wl) return kNonTerminal;
                      if (edge.IsTbTerminal()) {
                        return wl < 0.0 ? kTablebaseLoss : kTablebaseWin;
                      }
                      return wl < 0.0 ? kTerminalLoss : kTerminalWin;
                    };

                    // If moves have different outcomes, prefer better outcome.
                    auto a_rank = GetEdgeRank(a);
                    auto b_rank = GetEdgeRank(b);
                    if (a_rank != b_rank) return a > b;

                    // Neither is terminal, use standard rule.
                    if (a_rank == kNonTerminal) {
                      // Prefer largest playouts then eval then prior.
                      if (a.GetN() != b.GetN()) return a.GetN() > b.GetN();
                      if (a.GetWL() != b.GetWL()) return a.GetWL() > b.GetWL();
                      return a.GetP() > b.GetP();
                    }

                    // Both variants are winning, prefer shortest win.
                    if (a_rank > kNonTerminal) {
                      return a.GetM(0.0f) < b.GetM(0.0f);
                    }

                    // Both variants are losing, prefer longest losses.
                    return a.GetM(0.0f) > b.GetM(0.0f);
                  });
```

Initially it also had special rules for draws, but it's hard to make it not to violate a total strict order.

E.g. suppose we have:
a. Terminal draw.
b. Non-terminal loss with N=1000.
c. Non-terminal win with N=100.

`a > b`, `c > a`, `b > c` (because `N`s are compared).
So I think it makes sense to treat as ordinary nodes.
Although maybe only order them specially when both `a` and `b` are terminals..

mooskagh: With draw support:
```cpp
std::partial_sort(edges.begin(), middle, edges.end(),
                  [](const auto& a, const auto& b) {
                    // The function returns "true" when a is preferred to b.

                    // Lists edge types from less desirable to more desirable.
                    enum EdgeRank {
                      kTerminalLoss,
                      kTablebaseLoss,
                      kNonTerminal,  // Non terminal or terminal draw.
                      kTablebaseWin,
                      kTerminalWin,
                    };

                    auto GetEdgeRank = [](const EdgeAndNode& edge) {
                      const auto wl = edge.GetWL();
                      if (!edge.IsTerminal() || wl) return kNonTerminal;
                      if (edge.IsTbTerminal()) {
                        return wl < 0.0 ? kTablebaseLoss : kTablebaseWin;
                      }
                      return wl < 0.0 ? kTerminalLoss : kTerminalWin;
                    };

                    // If moves have different outcomes, prefer better outcome.
                    auto a_rank = GetEdgeRank(a);
                    auto b_rank = GetEdgeRank(b);
                    if (a_rank != b_rank) return a > b;

                    // If both are terminal draws, try to make it shorter.
                    if (a_rank == kNonTerminal && a.IsTerminal() &&
                        b.IsTerminal()) {
                      if (a.IsTbTerminal() != b.IsTbTerminal()) {
                        // Prefer non-tablebase draws.
                        return a.IsTbTerminal() < b.IsTbTerminal();
                      }
                      // Prefer shorter draws.
                      return a.GetM(0.0f) < b.GetM(0.0f);
                    }

                    // Neither is terminal, use standard rule.
                    if (a_rank == kNonTerminal) {
                      // Prefer largest playouts then eval then prior.
                      if (a.GetN() != b.GetN()) return a.GetN() > b.GetN();
                      if (a.GetWL() != b.GetWL()) return a.GetWL() > b.GetWL();
                      return a.GetP() > b.GetP();
                    }

                    // Both variants are winning, prefer shortest win.
                    if (a_rank > kNonTerminal) {
                      return a.GetM(0.0f) < b.GetM(0.0f);
                    }

                    // Both variants are losing, prefer longest losses.
                    return a.GetM(0.0f) > b.GetM(0.0f);
                  });
```

Mardak: Applied your code and fixed up some logic in a separate commit.

Mardak:
Here's the new behavior that is *different* for same mate scores:
```
info nodes 1167030 score mate    1 multipv  1 pv g6h6
info nodes 1458904 score mate  102 multipv  2 pv f7e8 h6f8 e8f8
info nodes 1413100 score mate  103 multipv  3 pv f7c7 e8f8 c7f7 h6g7 g6g7
info nodes    1946 score cp     -4 multipv  4 pv g6g8 e8g8 f7c7
info nodes     379 score mate -103 multipv  5 pv f7h7 h8h7 g6g5 h6g5
info nodes     439 score mate -103 multipv  6 pv f7g7 c7g7 f6f5 e8e5 f5g4 g7g6
info nodes     314 score mate -103 multipv  7 pv g6g2 c7e5 f6g6 e8e6 f7e6 e5e6
info nodes     348 score mate -103 multipv  8 pv f7f8 e8f8 f6e6 c7c6 e6e5 c6g6
info nodes     320 score mate -103 multipv  9 pv f7g8 e8g8 f6f5 c7f4 f5e6 g8g6
info nodes     440 score mate -103 multipv 10 pv g6g3 c7f4 f6g6 f4g3 g6h6
info nodes     467 score mate -103 multipv 11 pv g6g1 c7e5 f6g6 e8e6 f7e6 e5e6
info nodes     303 score mate -102 multipv 12 pv g6g5 h6g5 f6f5 c7f7
info nodes     326 score mate -102 multipv 13 pv f7e6 e8e6 f6f5 e6g6
info nodes     324 score mate -102 multipv 14 pv f6f5 c7f7 f5g4 f7g6
info nodes     389 score mate   -3 multipv 15 pv f7e7 c7e7 f6f5 e7e4 f5f6 e4f4
info nodes     663 score mate   -3 multipv 16 pv f7d5 c7f4 d5f5 f4d6 f5e6 d6e6
info nodes     431 score mate   -2 multipv 17 pv g6g4 c7d6 f6f5 e8e5
info nodes     296 score mate   -2 multipv 18 pv f7d7 c7e5 f6f7 e8f8
info nodes     413 score mate   -2 multipv 19 pv f7c4 c7e5 f6f7 e5e7
info nodes     621 score mate   -2 multipv 20 pv g6g7 c7f4 f6g6 f4g5
info nodes     412 score mate   -1 multipv 21 pv f7b3 c7f4
info nodes     727 score mate   -1 multipv 22 pv f7a2 c7f4
```
Before, most visited same-mate would be preferred. Notice pv22 with 727 visits is below pv21 with 412 visits. Probably doesn't matter much but this makes the "play style" network independent whereas before a network's prior (directing more search visits) to prefer one mate over another mate.

Issue #1109, opened 2020-03-04
OpenCL: Failed to compile opencl source


borg323:
This was reported on discord a few weeks ago (<https://discordapp.com/channels/425419482568196106/427066771627966466/678979495931150346>):
```
Trying KWG=32 KWI=2 MDIMA=16 MDIMC=16 MWG=64 NDIMB=16 NDIMC=16 NWG=64 SA=1 SB=1 STRM=0 STRN=0 VWM=4 VWN=4
/tmp/comgr-337dce/input/CompileSource:653:21: error: passing 'realM *' (aka 'float4 *') to parameter of type 'realM *' (aka 'float4 ') changes address space of pointer
  StoreResults(cgm, cpm, kSizeM);
                    ^~~
/tmp/comgr-337dce/input/CompileSource:480:58: note: passing argument to parameter 'cpm' here
INLINE_FUNC void StoreResults(__global realM cgm, realM cpm[NWI*MWI/VWM], const int kSizeM) {
                                                         ^
1 error generated.
Error: Failed to compile opencl source (from CL or HIP source to LLVM IR).
```
It was obtained using a test lc0 version that includes the following to display the opencl compilation status:
```diff
     std::string args = cl_args;
     args += sgemm_tuners;
     m_program.build(args.c_str());
+    CERR << m_program.getBuildInfo<CL_PROGRAM_BUILD_LOG>(m_device);
   } catch (const cl::Error&) {
     CERR << "Error building kernels: "
          << m_program.getBuildInfo<CL_PROGRAM_BUILD_LOG>(m_device) << ".";
```

Naphthalin:
Seems like once #1112 is merged, this can be closed.

borg323:
This was only seen with one specific driver (and opencl compiler). I'm closing it since I won't merge #1112 without finding a few more instances where it helps.

PR #1112, opened 2020-03-05
Fix a problem with opencl compilers

borg323:
Fixes issue #1109 - this change was suggested by @Ttl.


borg323:
Closing this for now, I'll reconsider it only if I find a few more cases where it helps.

PR #1113, opened 2020-03-06
Add option to only let smart pruning to stop after at least k batches.

Tilps:


src/mcts/stoppers/stoppers.cc
```diff
@@ -211,6 +213,7 @@ bool SmartPruningStopper::ShouldStop(const IterationStats& stats,
   // May overflow if (nps/smart_pruning_factor) > 180 000 000, but that's not
   // very realistic.
   hints->UpdateEstimatedRemainingRemainingPlayouts(remaining_playouts);
+  if (stats.batches_since_movestart < minimum_batches_) return false;
```
mooskagh: Maybe also log batches_since_movestat in the log message below (line 230), so that we notice if e.g. it instamoves as soon as number of required batches comes.

src/mcts/stoppers/stoppers.h
```diff
@@ -124,11 +124,12 @@ class KldGainStopper : public SearchStopper {
 // best move to potentially become the best one, stop the search.
 class SmartPruningStopper : public SearchStopper {
  public:
-  SmartPruningStopper(float smart_pruning_factor);
+  SmartPruningStopper(float smart_pruning_factor, int64_t minimum_batches_);
```
mooskagh: Trailing underscore in parameter name (not needed).

src/mcts/stoppers/stoppers.cc
Tilps: Sounds useful for setting a value that is high enough to do something yes.

src/mcts/stoppers/stoppers.h
Tilps: oops. :)

Tilps:
Actually tested and working as intended :P

PR #1115, opened 2020-03-07
Add max out of order evals parameter

Ttl:
For low batch sizes most of the batches are not full because there are too many NNCache hits. This change allows changing the limit on how many out of order evals can happen before non-full batch is submitted.

Tests with 128x10 network on startpos 1M nodes with maximum batch size 68, max-collisions 68, max-prefetch=68.

Before there are many non-full batches:
bs68_before

After with bigger max out of order evals limit almost all batches are full:
bs68_after

### DIFF
--- a/src/mcts/params.cc
+++ b/src/mcts/params.cc
@@ -176,6 +176,9 @@ const OptionId SearchParams::kOutOfOrderEvalId{
     "in the cache or is terminal, evaluate it right away without sending the "
     "batch to the NN. When off, this may only happen with the very first node "
     "of a batch; when on, this can happen with any node."};
+const OptionId SearchParams::kMaxOutOfOrderEvalsId{
+    "max-out-of-order-evals", "MaxOutOfOrderEvals",
+    "Maximum number of out of order evals during gathering of a batch."};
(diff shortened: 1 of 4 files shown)


src/mcts/params.cc
```diff
@@ -279,6 +282,7 @@ void SearchParams::Populate(OptionsParser* options) {
   options->Add<IntOption>(kMaxCollisionEventsId, 1, 1024) = 32;
   options->Add<IntOption>(kMaxCollisionVisitsId, 1, 1000000) = 9999;
   options->Add<BoolOption>(kOutOfOrderEvalId) = true;
+  options->Add<IntOption>(kMaxOutOfOrderEvalsId, 1, 10000) = 256;
```
mooskagh: Also override this value to 32 for selfplay games, here:
https://github.com/LeelaChessZero/lc0/blob/master/src/selfplay/tournament.cc#L114
So that selfplay doesn't change (but match games actually still will be changed).

Otherwise, looks good!

Ttl: I set the default to 32, but raising it could allow better GPU utilization in self play.

PR #1117, opened 2020-03-08
package android binaries in zip files

borg323:
In case we don't want to merge #1116 for v0.24.0 (which requires cherry-picking #1105 as well), this is a simpler alternative that packages the android binaries to a versioned zip file, which was the most popular option in a discord poll.


borg323:
Since #1116 was merged and is going in the release, this is no longer needed.

PR #1119, opened 2020-03-10
Max out of order evals relative to the maximum batch size

Ttl:
Amount of full batches as function of the maximum out of order evals scales similarly with different maximum batch sizes when it's relative to the maximum batch size. This makes it easier to tune the parameters as batch size and max-out-of-order-evals can be adjusted separately.

### DIFF
--- a/src/mcts/params.cc
+++ b/src/mcts/params.cc
@@ -176,6 +176,9 @@ const OptionId SearchParams::kOutOfOrderEvalId{
     "in the cache or is terminal, evaluate it right away without sending the "
     "batch to the NN. When off, this may only happen with the very first node "
     "of a batch; when on, this can happen with any node."};
+const OptionId SearchParams::kMaxOutOfOrderEvalsId{
+    "max-out-of-order-evals", "MaxOutOfOrderEvals",
+    "Maximum number of out of order evals during gathering of a batch."};
 const OptionId SearchParams::kStickyEndgamesId{
     "sticky-endgames", "StickyEndgames",
     "When an end of game position is found during search, allow the eval of "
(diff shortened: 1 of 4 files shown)


src/selfplay/tournament.cc
```diff
@@ -112,7 +112,7 @@ void SelfPlayTournament::PopulateOptions(OptionsParser* options) {
 
   auto defaults = options->GetMutableDefaultsOptions();
   defaults->Set<int>(SearchParams::kMiniBatchSizeId.GetId(), 32);
-  defaults->Set<int>(SearchParams::kMaxOutOfOrderEvalsId.GetId(), 32);
+  defaults->Set<float>(SearchParams::kMaxOutOfOrderEvalsId.GetId(), 1.0f);
```
Tilps: Could delete this line now its identical to the non-selfplay default.

mooskagh:
I think we should keep it absolute value, to be in line with other similar parameters which logically depend on minibatch size (e.g. `max prefetch`, `number of collision events`, `number of collision visits`, etc).

I think we should not optimize params for ease of people running the tunes, they can compute values themselves in their tuning pipeline if it's believed that the optimal size scales with minibatch size (which I'm not convinced either).

src/mcts/params.cc
```diff
@@ -178,7 +178,8 @@ const OptionId SearchParams::kOutOfOrderEvalId{
     "of a batch; when on, this can happen with any node."};
 const OptionId SearchParams::kMaxOutOfOrderEvalsId{
     "max-out-of-order-evals", "MaxOutOfOrderEvals",
```
Tilps: Might be good for these parameter names to indicate that its a ratio.

Tilps:
As discussed offline - its likely less about tuning and more about whats easiest for end users to configure. I do think the parameter name could be better if its going to be a ratio.
I don't think either crem or I are strongly for/against this PR otherwise. But the fact that historically it was 1.0 ratio is possibly pushing me towards a yes vote if the parameter names are clearer.

Issue #996, opened 2019-11-16
Fix running out of time


mooskagh:
There are quite many reports that Lc0 fails on time in tight time conditions (and sometimes not very tight) on slow backend (and sometimes not very slow).

We need to address that, e.g.:
- If engine sees that it cannot respond immediately when sudden `stop` event comes, only gather batch of size 1. (that will be only the very first batch while thinking, and only if tree was not reused).
- Make sure that bestmove is responded immediately even if batch computation is currently in progress (it should work like that already but worth checking).
- Make sure that when time budget is over, the code that outputs bestmove is run immediately.

It would be helpful to have logs of cases when this happens. (**Note: it should be Lc0 logs, not UCI logs from chess GUI!**).
To record Lc0 logs, either pass `--logfile=filename.txt` or set UCI option `LogFile` to some file name. Then Lc0 will start to write logs there.

borg323:
This is related to #869.

mooskagh:
Indeed, but here it's mostly about minimizing time to response when engine has 0 seconds left, while there it's about distributing time budget amond moves wisely.

Naphthalin:
Is this fixed by any of the new time management variants?

mooskagh:
That's partially fixed.
Not instamoving even when there is reused tree was a real bug, and it was fixed by #1123.

I believe that "stop gathering batch when run out of time" is not fixed... Hmm although the first batch always has size of 1.
Ok, closing this for now.

Issue #1120, opened 2020-03-10
LC0 doesn't start when running a test suite. BUG


Chess321:
Fritz GUI
LC0 v0.24.0-rc2 
2x RTX 2080 Ti
4 of 6 CPU cores
8 of 64 GB RAM
Tested with different nets and net size.

Testsuite is running but LC0 is doing nothing.
[logfile.txt](https://github.com/LeelaChessZero/lc0/files/4310916/logfile.txt)

Looks like LC0 needs to be loaded again and/or run and stop infinite analysis bevor starting a test suite.
This helps sometimes.

mooskagh:
The UCI host seems to be not UCI compiant (no `ucinewgame` and no `isready` after setting the position).
So what happens is the engine starts to load the network on `go infinite` command and while it's loading, it already receives `stop` command.

If still should return the move immediately after that (without starting the search) though, so there seem to be something working incorrectly.

But I'm not able to reproduce it.

mooskagh:
Could you try to produce the process dump when Lc0 is hanging? Here are the instructions how to do it:

How to produce process dump (only useful if you use official release for Windows downloaded from GitHub (<http://lc0.org/releases>)):

1. Download and unpack **ProcDump**: <https://docs.microsoft.com/en-us/sysinternals/downloads/procdump>
2. Reproduce hanging of Lc0.
3. While Lc0 is hanging, run:
`procdump64.exe -mm lc0.exe`  <- this will produce a short dump. Use it if your search was long, millions of nodes.
or
`procdump64.exe -ma lc0.exe` <- this will produce full dump. Useful if your search was short. For long searches it can easily be gigabytes in size.
4. It will create a file named `Lc0.exe_YYMMDD_HHMMSS.dmp` (current date and time instead of YYMMDD_HHMMSS). Attach it to an issue in GitHub or post some other way.

borg323:
I can reproduce it about 50% of the time. Bisecting leads to <https://github.com/LeelaChessZero/lc0/commit/3ff07fea993bca7eca972aeb3a70563a98728f9a>. The obvious candidate is setting the default for `--max-concurrent-searchers` to 1, and indeed setting `--max-concurrent-searchers=0` works fine.

borg323:
@Chess321 can you check if <https://ci.appveyor.com/api/buildjobs/lcrpc8o5wp3mdgtf/artifacts/build%2Flc0.exe> (#1123) fixes the issue?

[...]

PR #1123, opened 2020-03-10
Fix for issue #1120

borg323:


mooskagh:
So in the end all the time there was that bug that caused time outs on short time controls!
I.e. even if there is a tree to reuse, we still do one iteration.

I think the correct fix would be:

1. Add an (atomic) flag "we have at least one visit in a tree". Set it:
1.1. In constructor if there is a tree.
1.2. At the end of every iteration.
2. Also check this flag in GetMaxConcurrentSearchers branch.

Bonus points (although probably better not touch it for the release):
3. Factor it out to a function like "CanStopSearchImmediately()"
4. Replace all other reads of `stop_` with `IsSearchActive()` call.

src/mcts/search.cc
```diff
@@ -756,8 +756,12 @@ void SearchWorker::ExecuteOneIteration() {
   if (params_.GetMaxConcurrentSearchers() != 0) {
     while (true) {
       // If search is stop, we've not gathered or done anything and we don't
-      // want to, so we can safely skip all below.
-      if (search_->stop_.load(std::memory_order_acquire)) return;
+      // want to, so we can safely skip all below. But make sure we have done
+      // at least one iteration.
+      if (search_->stop_.load(std::memory_order_acquire) &&
+          search_->total_playouts_ > 0) {
```
mooskagh: `search_->total_playouts_ + search->initial_visits_`, and it will also fix #996

Tilps: search->GetTotalPlayouts() + search->initial_visits_
to avoid the mutex protection warning issue.

PR #1126, opened 2020-03-14
Drawscore bugfix

Ttl:
GetQ was missing drawscore argument in `GetVisitsToReachU`. Also fix a wrong type for `draw_score` in `GetBestRootChildWithTemperature`.

### DIFF
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -379,8 +379,8 @@ class EdgeAndNode {
   }
 
   int GetVisitsToReachU(float target_score, float numerator, float default_q,
-                        bool logit_q) const {
-    const auto q = GetQ(default_q, logit_q);
+                        float drawscore, bool logit_q) const {
+    const auto q = GetQ(default_q, drawscore, logit_q);
     if (q >= target_score) return std::numeric_limits<int>::max();
     const auto n1 = GetNStarted() + 1;
     return std::max(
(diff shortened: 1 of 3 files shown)


src/mcts/node.h
```diff
@@ -382,8 +382,8 @@ class EdgeAndNode {
   }
 
   int GetVisitsToReachU(float target_score, float numerator, float default_q,
-                        bool logit_q) const {
-    const auto q = GetQ(default_q, logit_q);
+                        float drawscore, bool logit_q) const {
+    const auto q = GetQ(default_q, drawscore, logit_q);
```
mooskagh: Uh, good find!.

..but that bug is bad. :-\
Actually, default arguments were only recently (2 years ago) allowed in the style guide, precisely because of this problem..

I suggest to remove the default value for `logit_q` and type that argument explicitly (as `/* logit_q= */ false`) at every place where it's called. There are many cases where default arguments improve readability, but in this case it does more harm than good.

src/mcts/search.cc
```diff
@@ -1028,7 +1028,7 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend(
 
     if (second_best_edge) {
       int estimated_visits_to_change_best = best_edge.GetVisitsToReachU(
-          second_best, puct_mult, fpu, params_.GetLogitQ());
+          second_best, puct_mult, fpu, even_draw_score, params_.GetLogitQ());
```
Tilps: I don't think even_draw_score is the right value to use here...  Should be just draw_score I think?

Ttl: It should be draw_score. Looks like you already fixed it, so I'll close this PR.

Ttl:
Continued in #1129.

PR #1129, opened 2020-03-14
Remove default for logit_q parameter to GetQ

Tilps:
Also include a fix draw_score being declared bool in GetBestRootChildWithTemperature.

Basically the same as jio's PR, I just wanted to see what removing the parameter default looked like, so I did it myself...

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -289,7 +289,7 @@ std::vector<std::string> Search::GetVerboseStats(Node* node) const {
         << ") ";
 
     oss << "(Q: " << std::setw(8) << std::setprecision(5)
-        << edge.GetQ(fpu, draw_score, false) << ") ";
+        << edge.GetQ(fpu, draw_score, /* logit_q= */ false) << ") ";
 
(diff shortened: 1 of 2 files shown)


src/mcts/search.cc
```diff
@@ -137,7 +137,7 @@ void Search::SendUciInfo() REQUIRES(nodes_mutex_) {
     const auto wl = edge.GetWL();
     const auto d = edge.GetD();
     const int w = static_cast<int>(std::round(500.0 * (1.0 + wl - d)));
-    const auto q = edge.GetQ(default_q, draw_score);
+    const auto q = edge.GetQ(default_q, draw_score, false);
```
kiudee: Can we add a comment here to make clear what is passed here?
```c++
const auto q = edge.GetQ(default_q, draw_score, /* logit_q= */ false);
```

mooskagh: Not really done. :)

Tilps: Heh I got all the other ones but missed this one? Nice.
Fixed.

PR #1130, opened 2020-03-16
Add a flag to allow `go nodes` to be treated as playouts in UCI

farmersrice:
It seems that a lot of users conducting engine/network tests with lc0 are using fixed visits. Currently the `go nodes` UCI command is treated as a visits limit by lc0. This adds a flag `--nodes-as-playouts` which allows `go nodes` to be treated as a playouts limit instead.

Here's why this is important:

Using a node visits limit is problematic in testing because it rewards networks with flatter policy with more computation time. Consider a case where you have two networks of equal strength at time parity, where one is flat, and one is sharp. Suppose you are playing them against one another at fixed nodes. What will happen is that testing with a fixed node limit will unfairly judge the flatter network as being stronger. Whenever a move is actually made in the game, it will be the case that the sharper network has more tree reuse than the flatter one. Suppose that your node limit is 1000 nodes per move. Now, say your sharper network thinks deeper in the selected move and saves 400 nodes from the previous tree, while the flatter one thinks more shallow and only saves 200 nodes from the previous tree. Then when you go to 1000 nodes on the next turn, you've actually given the flatter network 1000 - 200 = 800 playouts of computation, while the sharper one has only received 600! The flatter network is receiving more computational power than the sharper one, so obviously it will win the test. Now you think that the flatter network is superior, when in reality they were the same strength all along. In other cases, you might even select an inferior flatter network over a superior sharper network.

Now, there are a few ways to fix this problem:
1. Discard any notion of playouts, visits, nodes, or whatever from your testing entirely and instead use fixed time. This is definitely the best since in the real world you care about the strongest engine within a given time frame.
2. Count number of neural net evaluations per move (discarding stuff like nncache as well), since neural net eval takes the longest time by far in the game.
3. Limit playouts instead of visits. This isn't perfect, since the time utilized won't always be the same due to nncache, but it has the added advantage that you don't have to worry about stressing your machine and weakening the engines while you're doing testing (although solution 2 has this benefit as well). 

This fix just implements solution 3 as an optional flag because it's quick and easy and retains the vast majority of the benefit.


src/mcts/stoppers/factory.cc
```diff
@@ -114,6 +117,7 @@ void PopulateTimeManagementOptions(RunType for_what, OptionsParser* options) {
     options->Add<FloatOption>(kTimeMidpointMoveId, 1.0f, 100.0f) = 51.5f;
     options->Add<FloatOption>(kTimeSteepnessId, 1.0f, 100.0f) = 7.0f;
     options->Add<FloatOption>(kSpendSavedTimeId, 0.0f, 1.0f) = 1.0f;
+    options->Add<BoolOption>(kNodesAsPlayoutsId) = false;
 
     // Hide time curve options.
     options->HideOption(kTimeMidpointMoveId);
```
mooskagh: Do you think it may make sense to hide the new option too, in order not to overwhelm the UCI settings page?

farmersrice: I think it's probably better not to hide the new option, so that users can be informed. Otherwise I would expect very few people to know and use it.

mooskagh: But why would more than a few people need it? I think that the only scenario people would want this option is to simulate TC game where TC is not possible (e.g. running many games in parallel).

The vast majority of users would not need to run non-TC games.

Maybe the alternative would be to have the opposite (playouts by default, visits through the hidden flag), but I think this needs some discussion in #dev-public as maybe it will break something. E.g. in selfplay it indeed will change the behaviour but there we can have a different default for now..

farmersrice: Here are my thoughts:

- The default behavior should be kept as visits because visits = nodes is a common perception. If we changed that behavior it would probably leave more people confused. Like you said, only a few people need this flag, and having playouts by default could cause some issues in the training pipeline.
- A majority of the test results in #test-discuss are fixed-nodes tests. Fixing nodes or playouts has the advantage that you don't have to worry about weakening lc0 while you're playing a game or watching a stream or something while testing. I'm fairly confident this is the main reason behind most test results not being TC-based tests. 
- Most UCI settings aren't likely to be touched by the user at all unless they're trying to tune or experiment with it. I'd say this flag is more significant and more likely to be used than a lot of other options. The consequences of using playouts instead of visits in testing are decently important and I think this warrants the flag's inclusion.

farmersrice: The polling in #dev-log has stagnated. It should be good to go now.

PR #1131, opened 2020-03-17
Match knights() call more cleanly with other piece calls

tkondrashov:
I looked at this for a long time and haven't been able to find any reason that having two calls, one for their and one for our knights, does anything but make the code a little bit harder to read (specifically because it looks like there must be some reason). This probably isn't central to how anything works but it seems like it wouldn't interfere with what anyone else is doing and makes this easier to follow, unless I'm missing the reason for it.

Similar changes for kings in #1132

Benchmarks:
```
- Master
Benchmark final time 10.0009s calculating 212417 nodes per second.
Benchmark final time 10.0006s calculating 208707 nodes per second.
Benchmark final time 10.0004s calculating 208102 nodes per second.

- Knights
Benchmark final time 10.0005s calculating 208702 nodes per second.
Benchmark final time 10.0004s calculating 204465 nodes per second.
Benchmark final time 10.0006s calculating 206022 nodes per second.

- Kings
Benchmark final time 10.0005s calculating 206489 nodes per second.
Benchmark final time 10.0005s calculating 207881 nodes per second.
Benchmark final time 10.0006s calculating 206204 nodes per second.```

### DIFF
--- a/src/chess/board.cc
+++ b/src/chess/board.cc
@@ -1113,7 +1113,7 @@ bool ChessBoard::HasMatingMaterial() const {
     // K v K, K+B v K, K+N v K.
     return false;
   }
-  if (!our_knights().empty() || !their_knights().empty()) {
+  if (!(knights().empty())) {
     return true;
   }
 
--- a/src/chess/board.h
+++ b/src/chess/board.h
@@ -186,11 +186,9 @@ class ChessBoard {
   BitBoard bishops() const { return bishops_ - rooks_; }
   BitBoard rooks() const { return rooks_ - bishops_; }
   BitBoard queens() const { return rooks_ & bishops_; }
-  BitBoard our_knights() const {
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by tkondrashov** [comment]
There are other, similar changes to make for similar functions. I can add them to this PR if this is useful.

**Comment by oscardssmith** [comment]
Please make separate Pull requests. While it is a little more work, it makes reviewing, testing, and merging much easier. Thanks so much for cleaning stuff up!

**Comment by mooskagh** [comment]
Thanks for the changes, I agree they improve readability.
It's most likely performance neutral, but could you check it just for the case?
Could you run the following command before and after the change and compare numbers:
```
./lc0 benchmark --backend=random
```
The numbers may be quite noisy, so it may make sense to run it 2-3 times.

tkondrashov:
Added benchmarks to OP. Thanks for the command!

PR #1132, opened 2020-03-17
Match kings() call more cleanly with other piece calls

tkondrashov:
same idea as #1131

this also adds an `as_board()` method to BoardSquare to save us from writing `1ULL << *.as_int()` more than once

Benchmarks:
```
- Master
Benchmark final time 10.0009s calculating 212417 nodes per second.
Benchmark final time 10.0006s calculating 208707 nodes per second.
Benchmark final time 10.0004s calculating 208102 nodes per second.

- Knights
Benchmark final time 10.0005s calculating 208702 nodes per second.
Benchmark final time 10.0004s calculating 204465 nodes per second.
Benchmark final time 10.0006s calculating 206022 nodes per second.

- Kings
Benchmark final time 10.0005s calculating 206489 nodes per second.
Benchmark final time 10.0005s calculating 207881 nodes per second.
Benchmark final time 10.0006s calculating 206204 nodes per second.
```

### DIFF
--- a/src/chess/bitboard.h
+++ b/src/chess/bitboard.h
@@ -49,6 +49,7 @@ class BoardSquare {
   BoardSquare(const std::string& str, bool black = false)
       : BoardSquare(black ? '8' - str[1] : str[1] - '1', str[0] - 'a') {}
   constexpr std::uint8_t as_int() const { return square_; }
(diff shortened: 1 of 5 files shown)


mooskagh:
Similarly to another PR, that's probably performance neutral, but as it merges positions of kings in to separate them back, would be nice to run a benchmark to check it.

PR #1133, opened 2020-03-17
make pawn mask code slightly less disruptive

tkondrashov:
kPawnMask sitting in board.cc forces us to write just two of this list of functions in board.cc when the rest of them are in board.h, and it seems like that constant should belong to ChessBoard anyway, since en passant bits in the mask are part of the information requires to understand the contents of the class. Declaring the constant in board.h lets us have a clean block for all of the methods that return these bitboards.

### DIFF
--- a/src/chess/board.cc
+++ b/src/chess/board.cc
@@ -49,6 +49,8 @@ const char* ChessBoard::kStartposFen =
 
 const ChessBoard ChessBoard::kStartposBoard(ChessBoard::kStartposFen);
 
+const BitBoard ChessBoard::kPawnMask = 0x00FFFFFFFFFFFF00ULL;
+
 void ChessBoard::Clear() {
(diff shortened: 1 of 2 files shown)


src/chess/board.cc
```diff
@@ -427,10 +427,6 @@ void InitializeMagicBitboards() {
                     kBishopDirections);
 }
 
-BitBoard ChessBoard::pawns() const { return pawns_ & kPawnMask; }
-
-BitBoard ChessBoard::en_passant() const { return pawns_ - pawns(); }
```
tkondrashov: `pawns_ & ~(pawns_ & kPawnMask)` is the same as `pawns_ & (~pawns_ | ~kPawnMask)`, and ~pawns never does anything. simplified to `pawns_ & ~kPawnMask`, or `pawns_ - kPawnMask`

PR #1135, opened 2020-03-21
Make lc0 output v5 training data.

Tilps:


Tilps:
Note that while writer has been updated to support writing v5 data for FRC nets, selfplay game hasn't yet routed the input fromat from the net to the request to get v5 data.

src/neural/writer.h
```diff
@@ -35,8 +35,9 @@ namespace lczero {
 
 #pragma pack(push, 1)
 
-struct V4TrainingData {
+struct V5TrainingData {
   uint32_t version;
+  uint32_t input_format;
```
Tilps: Inserting this at the beginning is annoying for me in rescorer and training scripts, but it puts the float probabilities at an 8 byte offset that could potentially make for faster memcopies, and it kind of naturally belongs with the version field. I think I can survive the downsides :P

Tilps:
I'm not sure what is up with the formatting changes in search.cc - I upgraded my visual studio fairly recently and I think it must have upgraded clang-format to a newer version, but it is not clear why it seems to be enforcing comments wrap after 79 chars.

src/mcts/node.cc
```diff
@@ -367,14 +368,23 @@ V4TrainingData Node::GetV4TrainingData(GameResult game_result,
...
+  // For non-frc trained nets, just send 1 like we used to.
+  uint8_t queen_side = 1;
+  uint8_t king_side = 1;
+  // If frc trained, send the bit mask representing rook position.
+  if (input_format == pblczero::NetworkFormat::INPUT_112_WITH_CASTLING_PLANE) {
+    queen_side <<= castlings.queenside_rook();
+    king_side <<= castlings.kingside_rook();
+  }
+
+  result.castling_us_ooo = castlings.we_can_000() ? queen_side : 0;
```
mooskagh: As we don't store `input_format`, it's not clear from the training data whether FRC planes are in old or new format.

Maybe keep those values as they were and add additional ones? (either 4 values encoded like this or just `castlings.queenside_rook()` and `castlings.kingside_rook()`, or bitmask `(1 << castlings.kingside_rook()) | (1 << castlings.queenside_rook())` although that probably isn't worth that 1 byte).

Tilps: we do store input_format in training data - its new in v5 format.

src/mcts/search.cc
```diff
@@ -398,14 +398,16 @@ void Search::MaybeTriggerStop(const IterationStats& stats,
 // Return the evaluation of the actual best child, regardless of temperature
 // settings. This differs from GetBestMove, which does obey any temperature
 // settings. So, somethimes, they may return results of different moves.
-std::pair<float, float> Search::GetBestEval() const {
+std::pair<std::pair<float, float>, float> Search::GetBestEval() const {
```
mooskagh: Maybe have `struct BestEval`? Or at least an `std::tuple` instead of nested pairs.
E.g.
```cpp
struct BestEval {
  float wl;
  float d;
  float ml;
};
```

PR #1140, opened 2020-03-22
Moves left head support for DX backend.

Tilps:
Untested, but it compiles, so that is a start...

### DIFF
--- a/src/neural/dx/network_dx.cc
+++ b/src/neural/dx/network_dx.cc
@@ -551,8 +551,8 @@ DxNetwork::DxNetwork(const WeightsFile& file, const OptionsDict& options)
 
     // 1x1 convolution, val_channels output filters
     auto convVal = std::make_unique<ConvLayer>(
-        fp16_, nullptr, nullptr, &dx_context_, resi_last, val_channels, 8,
-        8, 1, kNumFilters, true, true);
+        fp16_, nullptr, nullptr, &dx_context_, resi_last, val_channels, 8, 8, 1,
+        kNumFilters, true, true);
     convVal->LoadWeights(&weights.value.weights[0], &weights.value.biases[0],
                          &dx_context_);
     network_.emplace_back(std::move(convVal));
--- a/src/neural/dx/network_dx.cc
+++ b/src/neural/dx/network_dx.cc
@@ -591,7 +591,7 @@ DxNetwork::DxNetwork(const WeightsFile& file, const OptionsDict& options)
   moves_left_ = file.format().network_format().moves_left() ==
                 pblczero::NetworkFormat::MOVES_LEFT_V1;
   if (moves_left_) {
-    // 1x1 convolution, val_channels output filters
+    // 1x1 convolution, moves_left biases output filters
     auto convMov = std::make_unique<ConvLayer>(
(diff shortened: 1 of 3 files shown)


src/neural/dx/network_dx.cc
```diff
@@ -745,18 +782,45 @@ void DxNetwork::Eval(InputsOutputsDx* io, int batch_size) {
 
   // value FC1.
   network_[l++]->Eval(batch_size, tensor_mem_[1], tensor_mem_[0], DXAlloc(),
-                      tensor_mem_[2], tensor_mem_[3], cl);
+                      DXAlloc(), DXAlloc(), cl);
```
Tilps: I made these changes here because it was 'scary' that tensor_mem_[2] was being passed anywhere when it contains the all important residual final output.  Happily fully connecteds don't use scratch at all.

src/neural/dx/network_dx.cc
```diff
@@ -587,6 +587,43 @@ DxNetwork::DxNetwork(const WeightsFile& file, const OptionsDict& options)
     network_.emplace_back(std::move(FCVal2));
   }
 
+  // Moves left head
+  moves_left_ = file.format().network_format().moves_left() ==
+                pblczero::NetworkFormat::MOVES_LEFT_V1;
+  if (moves_left_) {
+    // 1x1 convolution, moves_left biases output filters
+    auto convMov = std::make_unique<ConvLayer>(
+        fp16_, nullptr, nullptr, &dx_context_, resi_last,
```
Tilps: Note that the value head has getLastLayer() here instead of resi_last - I think that is a bug that luckily has no negative outcomes...

src/neural/dx/network_dx.cc
ankan-ban: Earlier nothing was using values from residual tower after the value head so they were fine to be overwritten, but MLH head would use them, and because FC layers don't use scratch at all the changes are fine.

src/neural/dx/network_dx.cc
ankan-ban: You are right - better change that too (although not affecting anything right now).

ankan-ban:
The changes look reasonable to me.

src/neural/dx/network_dx.cc
```diff
@@ -812,6 +883,14 @@ void DxNetwork::Eval(InputsOutputsDx* io, int batch_size) {
       memcpy(io->op_value_mem_final_ + val_vector_size * i,
              io->op_value_mem_ + kNumOutputValuePadded8 * i,
              val_vector_size * sizeof(float));
+    if (moves_left_) {
+      // Moves left:
+      // Un-pad moves left output.
+      for (int i = 0; i < batch_size; i++)
+        memcpy(io->op_moves_left_mem_final_ + i,
+               io->op_moves_left_mem_ + kNumOutputValuePadded8 * i,
```
Ttl: `kNumOutputMovesLeftPadded8`

src/neural/dx/network_dx.cc
Tilps: Done.

Tilps:
Note I have actually tested it, and it seems able to return values comparable to those returned by cudnn backend.

PR #1141, opened 2020-03-22
Change custom protobuf API

mooskagh:
It turns out that we need to mutate protobufs more than I expected, so I changed the custom protobuf API to be the same as libprotobuf.

* Now parsing/copying happens right away, like in libprotobuf (previous implementation didn't initialize anything, it could access data right from the wire format).
* The code (especially the generated code) is cleaner.
* Sane mutation API (compatible with libprotobuf this time)
* As before, I only implemented features that we use.

Generated [net.pb.h](https://github.com/LeelaChessZero/lc0/files/4364875/net.pb.h.txt)


mooskagh:
Updated [net.pb.h](https://github.com/LeelaChessZero/lc0/files/4364959/net.pb.h.txt)

src/utils/protomessage.cc
```diff
@@ -18,184 +18,67 @@ uint64_t ReadVarInt(const char** iter, const char* const end) {
...
 
-std::string EncodeVarInt(std::uint64_t val) {
-  std::string res;
-  while (true) {
-    char c = (val & 0x7f);
-    val >>= 7;
-    if (val) c |= 0x80;
-    res += c;
-    if (!val) return res;
+// // Kept for serialization part.
```
Tilps: I'm unclear what this means, is it that you are intending to write a serialization part in a future PR?

borg323:
Not looked at the code yet, but run it with `-fsanitize=undefined` and got:
```
lc0@exe/proto/net.pb.h:827:50: runtime error: load of misaligned address 0x7fab54005011 for type 'const std::uint32_t' (aka 'const unsigned int'), which requires 4 byte alignment
0x7fab54005011: note: pointer points here
 00 00 00  0d c0 01 00 00 1a 06 08  00 10 15 18 00 22 0e 08  01 12 0a 08 01 10 02 18  04 20 02 28 02
              ^ 
lc0@exe/proto/net.pb.h:575:45: runtime error: load of misaligned address 0x7fab54005039 for type 'const std::uint32_t' (aka 'const unsigned int'), which requires 4 byte alignment
0x7fab54005039: note: pointer points here
 99 19 3e  1d 22 62 66 3d 25 30 85  00 40 2d 00 18 44 42 52  8b be c6 03 0a d2 e8 0f  0a 8e e0 0f 0d
              ^ 
lc0@exe/proto/net.pb.h:576:48: runtime error: load of misaligned address 0x7fab5400503e for type 'const std::uint32_t' (aka 'const unsigned int'), which requires 4 byte alignment
0x7fab5400503e: note: pointer points here
 62 66 3d 25 30 85  00 40 2d 00 18 44 42 52  8b be c6 03 0a d2 e8 0f  0a 8e e0 0f 0d ac 86 81  be 15
             ^ 
lc0@exe/proto/net.pb.h:577:45: runtime error: load of misaligned address 0x7fab54005043 for type 'const std::uint32_t' (aka 'const unsigned int'), which requires 4 byte alignment
0x7fab54005043: note: pointer points here
 85  00 40 2d 00 18 44 42 52  8b be c6 03 0a d2 e8 0f  0a 8e e0 0f 0d ac 86 81  be 15 ce 27 ff 3f 1a
              ^ 
lc0@exe/proto/net.pb.h:97:46: runtime error: load of misaligned address 0x7fab54005055 for type 'const std::uint32_t' (aka 'const unsigned int'), which requires 4 byte alignment
0x7fab54005055: note: pointer points here
 8e e0 0f 0d ac 86 81  be 15 ce 27 ff 3f 1a 80  e0 0f 8a 18 c0 19 ae 19  85 1a e5 1b 90 1a b1 1b  a1
             ^ 
lc0@exe/proto/net.pb.h:98:46: runtime error: load of misaligned address 0x7fab5400505a for type 'const std::uint32_t' (aka 'const unsigned int'), which requires 4 byte alignment
0x7fab5400505a: note: pointer points here
 86 81  be 15 ce 27 ff 3f 1a 80  e0 0f 8a 18 c0 19 ae 19  85 1a e5 1b 90 1a b1 1b  a1 1c c6 1b 65 1c
              ^ 
```

mooskagh:
Changes since last commit:
* In ParseFromString() has_* variables were not set.
* No unaligned memory access.
* Big endian support (although I'm not sure about float and double)
* Strings are passed as std::string_view rather than const std::string& everywhere (not sure if a good idea)
* Non-message values are returned by value rather than by const reference.

[net.pb.h](https://github.com/LeelaChessZero/lc0/files/4365024/net.pb.h.txt)

mooskagh:
* Non-message arguments are also passed by value rather than by const reference.

[net.pb.h](https://github.com/LeelaChessZero/lc0/files/4365039/net.pb.h.txt)

PR #1142, opened 2020-03-22
Connect up input format when generating training data to allow v5 FRC.

Tilps:
The one line I left out of the last PR, I guess it was easier than I thought it might be...


Tilps:
Also fixed an encoding problem that would have made FRC castling plane values sent to network bad.

PR #1149, opened 2020-03-24
Remember if network has format before making mutable.

### LINKED ISSUE #1148: Test10-30 error Network format 0 is not supported by OpenCL backend.
@mooskagh This regressed with #1141 when using older networks, e.g., 11258, 22202, 32930, but more recent ones, e.g., 42850, are fine.

```
Loading weights file from: 32930
Creating backend [opencl]...
error Network format 0 is not supported by OpenCL backend.
```

There's older t10 pre-protobuf networks that haven't been working for a long time now, so this wouldn't be the first time we dropped support for older networks. Not sure if this is explicitly desired to drop support for pre-test40 networks at this time.

Mardak:
r?@mooskagh or @Tilps  Fix #1148 by saving the value of `has_network_format` before mutable.

### DIFF
--- a/src/neural/loader.cc
+++ b/src/neural/loader.cc
@@ -107,15 +107,15 @@ std::string DecompressGzip(const std::string& filename) {
 void FixOlderWeightsFile(WeightsFile* file) {
   using nf = pblczero::NetworkFormat;
   auto network_format = file->format().network_format().network();
-  if (file->format().has_network_format() &&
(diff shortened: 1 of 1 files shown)


mooskagh:
But where was no bug.
If `not x.has_foo()`, then `x.foo()` returns default value of `foo` but doesn't create it. It's `x.mutable_foo()` that would create it.

mooskagh:
Hmmmmm but what happened then?..

PR #1151, opened 2020-03-25
Add extra params for passing defaults for edge functions that really should consider it.

Tilps:


src/mcts/search.cc
```diff
@@ -130,12 +130,13 @@ void Search::SendUciInfo() REQUIRES(nodes_mutex_) {
 
   int multipv = 0;
   const auto default_q = -root_node_->GetWL();
```
Tilps: This should be default_wl, and there probably should be a separately calculated default_q which takes into account drawscore correctly (since you an't just negate the parent GetQ) - let me know if you want it fixed in this PR.

src/mcts/search.cc
```diff
@@ -545,7 +551,12 @@ std::vector<EdgeAndNode> Search::GetBestChildrenNoTemperature(Node* parent,
         if (a_rank == kNonTerminal) {
           // Prefer largest playouts then eval then prior.
           if (a.GetN() != b.GetN()) return a.GetN() > b.GetN();
-          if (a.GetWL() != b.GetWL()) return a.GetWL() > b.GetWL();
+          // Default doesn't matter here so long as they are the same as either
```
Tilps: So... a separate question here.  Should this be using GetWL instead of GetQ??  Draw score is being ignored in tiebreaking which seems wrong.

src/mcts/search.cc
```diff
@@ -130,12 +130,13 @@ void Search::SendUciInfo() REQUIRES(nodes_mutex_) {
...
   int multipv = 0;
   const auto default_q = -root_node_->GetWL();
+  const auto default_d = root_node_->GetD();
   for (const auto& edge : edges) {
     ++multipv;
     uci_infos.emplace_back(common_info);
     auto& uci_info = uci_infos.back();
-    const auto wl = edge.GetWL();
-    const auto d = edge.GetD();
+    const auto wl = edge.GetWL(default_q);
```
Tilps: This line is a bugfix - since without it the reported score for non-drawscore enabled outputs is always 0 in the 1 node case.

src/mcts/search.cc
Tilps: okay you can just negate parent GetQ (if you negate drawscore at the same time...) - since I made this sound harder than it actually is.

src/mcts/search.cc
mooskagh: I somehow thought that D are equal in that case, but indeed they are not and Q makes more sense then WL.

mooskagh:
Would be nice to check on random backend whether it affects performance (as GetD() / GetWL() are called often and not they have extra parameter and logic).

Tilps:
I'll check, but there isn't actually any extra logic in any of the hot paths - since the function is in a header all the times I pass a constant its trivially inlinable and can have identical code gen to before.  Passing a non-constant is unlikely to make much difference.

src/mcts/search.cc
Tilps: Implemented now.

src/mcts/search.cc
Tilps: Implemented.

Tilps:
Tested go nodes 1000000 with random backend - both at 180k for default settings.

src/mcts/search.cc
```diff
@@ -486,6 +491,10 @@ std::vector<EdgeAndNode> Search::GetBestChildrenNoTemperature(Node* parent,
   if (parent == root_node_) {
     PopulateRootMoveLimit(&root_limit);
   }
+  // Assume this function is only ever called with root or immediate child of
+  // root.
```
mooskagh: `assert(parent==root_node_ || parent->GetParent() == root_node_);` ?

PR #1153, opened 2020-03-26
New windows build script and instructions update

borg323:


cn4750:
Using Meson 0.53.2 and Visual Studio 2017 I am able to successfully build lc0 with the following options set:
```
set CUDNN=true
set DX12=true
set OPENCL=false
set MKL=false
set DNNL=true
set OPENBLAS=false
set EIGEN=false
set TEST=false
```

cn4750:
Building the OpenCL backend errors out with the following error:
```
"C:\Users\Chad\source\repos\lc0\build\lc0.sln" (default target) (1) ->
"C:\Users\Chad\source\repos\lc0\build\lc0@exe.vcxproj" (default target) (30) ->
(Link target) ->
  LINK : fatal error LNK1181: cannot open input file 'C:\Program.obj' [C:\Users\Chad\source\repos\lc0\build\lc0@exe.vcxproj]
```
Changing https://github.com/LeelaChessZero/lc0/blob/master/meson.build#L319 to include `method: 'extraframework'` seems to fix the issue and builds successfully.

borg323:
The opencl breakage is unrelated (and expected). I'm waiting for more info on <https://github.com/mesonbuild/meson/issues/6846> before deciding whether the above workaround is the best option.

borg323:
There hasn't been any progress on the meson issue, so added the workaround to `meson.build`.

build.cmd
```diff
@@ -0,0 +1,62 @@
...
+set DX12=false
+set OPENCL=false
+set MKL=false
+set DNNL=false
+set OPENBLAS=false
+set EIGEN=false
+set TEST=false
+
+rem 2. Edit the paths for the build dependencies.
+set CUDA_PATH=C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v10.0
```
Tilps, line 15: in my experience VS2019 + cuda v10.0 didn't work so well - I switched to 10.1

borg323: Is this the usual cuda header version check? I can add it to the readme files.

Tilps: Yeah I think it was that.

Tilps: Looks good.

PR #1155, opened 2020-03-26
Remove selfplay override for root has own cpuct params

Tilps:
no longer needed as its always sent by training server.

### DIFF
--- a/src/selfplay/tournament.cc
+++ b/src/selfplay/tournament.cc
@@ -127,7 +127,6 @@ void SelfPlayTournament::PopulateOptions(OptionsParser* options) {
                              "multiplexing");
   defaults->Set<bool>(SearchParams::kStickyEndgamesId.GetId(), false);
   defaults->Set<bool>(SearchParams::kLogitQId.GetId(), false);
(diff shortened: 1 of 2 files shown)


src/selfplay/tournament.cc
```diff
@@ -127,7 +127,6 @@ void SelfPlayTournament::PopulateOptions(OptionsParser* options) {
                              "multiplexing");
   defaults->Set<bool>(SearchParams::kStickyEndgamesId.GetId(), false);
   defaults->Set<bool>(SearchParams::kLogitQId.GetId(), false);
-  defaults->Set<bool>(SearchParams::kRootHasOwnCpuctParamsId.GetId(), false);
```
mooskagh: Also remove the parameter from `params.cc`? As the only reason it existed is actually setting it to false for selfplay.

Tilps: We can't remove it - the server still has to send it until the server enforces a binary where the default is true - Which is what I am doing here.

borg323: Add a comment so we don't forget?

Tilps: over in params.cc - sure.

Issue #548, opened 2018-12-02
Lc0 doesn't complain if there are additional symbols after the numeric flag


mooskagh:
E.g. `--threads=5asdf` works and sets threads to 5. Should fail instead.

Naphthalin:
Tested the above example, fails now as intended. Can be closed.

PR #1156, opened 2020-03-27
Validate string values for integer flags

lealgo:
Fixes issue #548 
The Android build needs PR #1166 to be merged.

### DIFF
--- a/src/utils/optionsparser.cc
+++ b/src/utils/optionsparser.cc
@@ -466,7 +466,7 @@ int IntOption::ValidateIntString(const std::string& val) const {
   auto [ptr, err] = std::from_chars(val.data(), end, result);  
   if (err == std::errc::invalid_argument) {
     throw Exception("Flag '--" + GetLongFlag() + "' has an invalid format.");
-  } else if(err == std::errc::result_out_of_range) {
+  } else if (err == std::errc::result_out_of_range) {
     throw Exception("Flag '--" + GetLongFlag() + "' is out of range.");
   } else if (ptr != end) {
     throw Exception("Flag '--" + GetLongFlag() + "' has trailing characters.");
--- a/src/utils/optionsparser.cc
+++ b/src/utils/optionsparser.cc
@@ -409,12 +409,14 @@ IntOption::IntOption(const OptionId& id, int min, int max)
     : Option(id), min_(min), max_(max) {}
 
 void IntOption::SetValue(const std::string& value, OptionsDict* dict) {
+  ValidateIntString(value);
   SetVal(dict, std::stoi(value));
 }
 
 bool IntOption::ProcessLongFlag(const std::string& flag,
                                 const std::string& value, OptionsDict* dict) {
   if (flag == GetLongFlag()) {
+    ValidateIntString(value);
     SetVal(dict, std::stoi(value));
     return true;
   }
@@ -424,6 +426,7 @@ bool IntOption::ProcessLongFlag(const std::string& flag,
 bool IntOption::ProcessShortFlagWithValue(char flag, const std::string& value,
                                           OptionsDict* dict) {
   if (flag == GetShortFlag()) {
+    ValidateIntString(value);
     SetVal(dict, std::stoi(value));
     return true;
   }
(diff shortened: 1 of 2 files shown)


src/utils/optionsparser.cc
```diff
@@ -460,6 +463,19 @@ void IntOption::SetVal(OptionsDict* dict, const ValueType& val) const {
   dict->Set<ValueType>(GetId(), val);
 }
 
+void IntOption::ValidateIntString(const std::string& val) {
+  const int base = 10;
+  char *end;
+  errno = 0;
+  std::strtol(val.c_str(), &end, base);
```
mooskagh: This function already converts the value, why not just return it? :)
Also the more "modern" way of the same function is https://en.cppreference.com/w/cpp/utility/from_chars, consider using it instead.
```cpp
uint64_t result;
const auto end = val.data() + val.size();
auto [ptr, err] = std::from_chars(val.data(), end, &result);
if (err != std::errc() || ptr != end) {
	throw('Blah blah');
}
return result;
```

src/utils/optionsparser.cc
```diff
@@ -460,6 +460,24 @@ void IntOption::SetVal(OptionsDict* dict, const ValueType& val) const {
   dict->Set<ValueType>(GetId(), val);
 }
 
+int IntOption::ValidateIntString(const std::string& val) {
```
mooskagh: This function doesn't access any class data, so no need to make it a member function of a class. I suggest to make it a free-standing function in anonymous namespace instead (like 'EscapeMd()` above).

src/utils/optionsparser.cc
```diff
@@ -405,17 +406,34 @@ void StringOption::SetVal(OptionsDict* dict, const ValueType& val) const {
 // IntOption
 /////////////////////////////////////////////////////////////////
 
+namespace {
+int ValidateIntString(const std::string& flag, const std::string& val) {  
+  int result;
+  const auto end = val.data() + val.size();
+  auto [ptr, err] = std::from_chars(val.data(), end, result);  
+  if (err == std::errc::invalid_argument) {
+    throw Exception("Flag '--" + flag + "' has an invalid format.");
```
mooskagh: Oops I didn't notice it did call the member functions. In that case it would be find to keep it as they were (but make `const`).

Feel free to leave it as it is or undo that change (but then add `const`), as you prefer.

lealgo:
The Android build is failing because Appveyor uses NDK r17 which doesn't contain this header:
```
../src/utils/optionsparser.cc:33:10: fatal error: 'charconv' file not found
#include <charconv>
```
I'm going to prepare another PR for Appveyor, to make it use a more recent NDK.

lealgo:
Another thing I noticed on Ubuntu 18.04 is that the std library for the default compiler `gcc-7` doesn't include the `<charconv>` header either. I had to build with `gcc-8`, maybe we should update the READMEs with this info.

lealgo:
The Android build should be fixed with PR #1166

src/utils/optionsparser.h
```diff
@@ -188,7 +189,7 @@ class FloatOption : public OptionsParser::Option {
                                  OptionsDict* dict) override;
 
   ValueType GetVal(const OptionsDict&) const;
-  void SetVal(OptionsDict* dict, const ValueType& val) const;
+  void SetVal(OptionsDict* dict, const ValueType& val) const;
```
mooskagh: This line has trailing spaces. :)

lealgo: Sorry! Will fix asap. :P

lealgo:
@mooskagh 
I think we're going to have issues with some compiler versions not supporting newer `c++17` features. We're recommending on discord to use gcc-7 and clang-4, but for instance on Ubuntu I had to up to gcc-8 (clang-6 was fine) in order to get the `charconv` header. Another example, I was trying `std::filesystem` and also had to use gcc-8, but had no luck with clang-6. Maybe we should raise a little bit the required compiler versions.

gcp:
Just updating the clang version doesn't work because clang relies on the libstdc++ headers, see the issue I just filed.

PR #1158, opened 2020-03-27
scale mlh bonus with q

Naphthalin:
Basic idea: scale the MLH bonus with the `q` value (of the parent) to allow lower mlh-thresholds, reduce the effect of MLH with lower Q and therefore avoid the discontinuities. 

If child node ~= parent node, this effectively means that the Q value of the child is `Q*(1+M)`; the first commit uses parent_q, while the 2nd commit directly uses the child's Q.

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -1014,8 +1014,7 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend(
         const float m_cap = params_.GetMovesLeftMaxEffect();
         const float parent_m = node->GetM();
         const float child_m = child.GetM(parent_m);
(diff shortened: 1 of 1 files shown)


Naphthalin:
Note that this virtually makes no difference when using high thresholds and in general at high Q values, only when having low Q (and therefore a low mlh-threshold to allow MLH being activated in the first place).

The second commit will also interact with drawscore when threshold is set to 0: with positive drawscore, at WL=0 child nodes resulting in faster draws are preferred, while with negative drawscore they will be prolonged the same way a loss is prolonged.

Naphthalin:
(implemented by #1175, closing now)

PR #1160, opened 2020-03-27
Convert parents that cannot win or cannot lose to help find more draws.

Mardak:
r?@mooskagh Use more of the Terminal enum to store CantWin and CantLose states that are used in a more generalized parent conversion behavior. Notably, a child draw makes the parent CantWin, and this then makes it possible to decide if choosing among moves that Draw or CantWin (or Loss), the best move is Draw -- thus a new parent Draw Terminal that otherwise wouldn't have been converted before. Additionally, if all possible moves CantWin, the parent is CantLose and grandparent is also CantWin. E.g.,

https://lichess.org/analysis/5Qk1/6b1/6KR/6p1/8/8/8/8_b_-_-_0_1
test cant

```
./lc0 -w 591226 -v

position fen 5Qk1/6b1/6KR/6p1/8/8/8/8 b - - 0 1
g7f8 N:     284 (P: 56.96%) (WL: -0.31254) (D:  0.682) (M:  3.7) (L) 
g8f8 N:    4557 (P: 43.04%) (WL: -0.03414) (D:  0.945) (M:  4.6) 

Bxf8 / g7f8 CantWin (L) because…

position fen 5Qk1/6b1/6KR/6p1/8/8/8/8 b - - 0 1 moves g7f8
…
g6f6 N:       2 (P:  2.38%) (WL: -0.48807) (D:  0.502) (M:  0.5) 
g6g5 N:       2 (P:  2.37%) (WL: -0.00017) (D:  0.996) (M:  0.5) (L) 
h6h8 N:       3 (P:  1.68%) (WL:  0.18944) (D:  0.806) (M:  1.0) (W) 
h6h5 N:      15 (P:  9.71%) (WL:  0.18659) (D:  0.810) (M:  1.7) 
…

Rh8 / h6h8 CantLose (W) because…

position fen 5Qk1/6b1/6KR/6p1/8/8/8/8 b - - 0 1 moves g7f8 h6h8
g8h8 N:       2 (P: 100.00%) (WL: -0.00026) (D:  0.998) (M:  0.5) (L) 

Kxh8 / g8h8 is the only move and CantWin (L) because…

position fen 5Qk1/6b1/6KR/6p1/8/8/8/8 b - - 0 1 moves g7f8 h6h8 g8h8
…
g6f5 N:       0 (P: 13.81%) (WL:  0.00000) (D:  0.000) (M:  0.0) 
g6g5 N:       2 (P: 63.82%) (WL:  0.00000) (D:  1.000) (M:  0.0) (T) 

Kxg5 / g6g5 is a terminal Draw with insufficient material
```

### DIFF
--- a/src/chess/position.h
+++ b/src/chess/position.h
@@ -77,7 +77,9 @@ class Position {
   int ply_count_ = 0;
 };
 
-enum class GameResult { UNDECIDED, WHITE_WON, DRAW, BLACK_WON };
+// These are ordered so max() prefers the best result.
+enum class GameResult : uint8_t { UNDECIDED, BLACK_WON, DRAW, WHITE_WON };
+GameResult operator-(const GameResult& res);
 
 class PositionHistory {
  public:
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -214,11 +214,14 @@ std::string Node::DebugString() const {
       << " Parent:" << parent_ << " Index:" << index_
       << " Child:" << child_.get() << " Sibling:" << sibling_.get()
       << " WL:" << wl_ << " N:" << n_ << " N_:" << n_in_flight_
-      << " Edges:" << edges_.size();
+      << " Edges:" << edges_.size()
+      << " Bounds:" << static_cast<int>(lower_bound_) - 2 << ","
+      << static_cast<int>(upper_bound_) - 2;
   return oss.str();
 }
 
 void Node::MakeTerminal(GameResult result, float plies_left, Terminal type) {
+  SetBounds(result, result);
   terminal_type_ = type;
   m_ = plies_left;
   if (result == GameResult::DRAW) {
@@ -255,6 +258,11 @@ void Node::MakeNotTerminal() {
     wl_ /= n_;
     d_ /= n_;
   }
+}
+
+void Node::SetBounds(GameResult lower, GameResult upper) {
+  lower_bound_ = lower;
+  upper_bound_ = upper;
 }
 
 bool Node::TryStartScoreUpdate() {
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -131,10 +131,15 @@ class Node {
   using Iterator = Edge_Iterator<false>;
   using ConstIterator = Edge_Iterator<true>;
 
-  enum class Terminal : uint8_t { NonTerminal, Terminal, Tablebase };
+  enum class Terminal : uint8_t { NonTerminal, EndOfGame, Tablebase };
 
   // Takes pointer to a parent node and own index in a parent.
-  Node(Node* parent, uint16_t index) : parent_(parent), index_(index) {}
+  Node(Node* parent, uint16_t index)
+      : parent_(parent),
+        index_(index),
+        terminal_type_(Terminal::NonTerminal),
+        lower_bound_(GameResult::BLACK_WON),
+        upper_bound_(GameResult::WHITE_WON) {}
 
   // Allocates a new edge and a new node. The node has to be no edges before
   // that.
@@ -166,13 +171,16 @@ class Node {
   // Returns whether the node is known to be draw/lose/win.
   bool IsTerminal() const { return terminal_type_ != Terminal::NonTerminal; }
   bool IsTbTerminal() const { return terminal_type_ == Terminal::Tablebase; }
+  typedef std::pair<GameResult, GameResult> Bounds;
+  Bounds GetBounds() const { return {lower_bound_, upper_bound_}; }
   uint16_t GetNumEdges() const { return edges_.size(); }
 
   // Makes the node terminal and sets it's score.
   void MakeTerminal(GameResult result, float plies_left = 0.0f,
-                    Terminal type = Terminal::Terminal);
+                    Terminal type = Terminal::EndOfGame);
   // Makes the node not terminal and updates its visits.
   void MakeNotTerminal();
+  void SetBounds(GameResult lower, GameResult upper);
 
   // If this node is not in the process of being expanded by another thread
   // (which can happen only if n==0 and n-in-flight==1), mark the node as
@@ -301,9 +309,12 @@ class Node {
   // Index of this node is parent's edge list.
   uint16_t index_;
 
-  // 1 byte fields.
+  // Bit fields using parts of uint8_t fields initialized in the constructor.
   // Whether or not this node end game (with a winning of either sides or draw).
-  Terminal terminal_type_ = Terminal::NonTerminal;
+  Terminal terminal_type_ : 2;
+  // Best and worst result for this node.
+  GameResult lower_bound_ : 2;
+  GameResult upper_bound_ : 2;
 
   // TODO(mooskagh) Unfriend NodeTree.
   friend class NodeTree;
@@ -372,6 +383,10 @@ class EdgeAndNode {
   // Whether the node is known to be terminal.
   bool IsTerminal() const { return node_ ? node_->IsTerminal() : false; }
   bool IsTbTerminal() const { return node_ ? node_->IsTbTerminal() : false; }
+  Node::Bounds GetBounds() const {
+    return node_ ? node_->GetBounds()
+                 : Node::Bounds{GameResult::BLACK_WON, GameResult::WHITE_WON};
+  }
 
   // Edge related getters.
   float GetP() const { return edge_->GetP(); }
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -320,7 +320,16 @@ std::vector<std::string> Search::GetVerboseStats(Node* node) const {
     }
     oss << ") ";
 
-    if (edge.IsTerminal()) oss << "(T) ";
+    const auto [edge_lower, edge_upper] = edge.GetBounds();
+    oss << (edge_lower == edge_upper
+                ? "(T) "
+                : edge_lower == GameResult::DRAW &&
+                          edge_upper == GameResult::WHITE_WON
+                      ? "(W) "
+                      : edge_lower == GameResult::BLACK_WON &&
+                                edge_upper == GameResult::DRAW
+                            ? "(L) "
+                            : "");
     infos.emplace_back(oss.str());
   }
   return infos;
@@ -1448,8 +1457,8 @@ void SearchWorker::DoBackupUpdateSingleNode(
     return;
   }
 
-  // For the first visit to a terminal, maybe convert ancestors to terminal too.
-  auto can_convert =
+  // For the first visit to a terminal, maybe update parent bounds too.
+  auto update_parent_bounds =
       params_.GetStickyEndgames() && node->IsTerminal() && !node->GetN();
 
   // Backup V value up to a root. After 1 visit, V = Q.
@@ -1473,40 +1482,9 @@ void SearchWorker::DoBackupUpdateSingleNode(
     // Nothing left to do without ancestors to update.
     if (!p) break;
 
-    // Convert parents to terminals except the root or those already converted.
-    can_convert = can_convert && p != search_->root_node_ && !p->IsTerminal();
-
-    // A non-winning terminal move needs all other moves to be similar.
-    auto all_losing = true;
-    auto found_tb = n->IsTbTerminal();
-    float losing_m = 0.0f;
-    if (can_convert && v <= 0.0f) {
-      for (const auto& edge : p->Edges()) {
-        // Default_wl doesn't matter as WL is only used if IsTerminal is true.
-        const auto WL = edge.GetWL(0.0f);
-        can_convert = can_convert && edge.IsTerminal() && WL <= 0.0f;
-        if (!can_convert) break;
-        all_losing = all_losing && WL < 0.0f;
-        found_tb = found_tb || edge.IsTbTerminal();
-        losing_m = std::max(losing_m, edge.GetM(0.0f));
-      }
-    }
-
-    // Convert the parent to a terminal loss if at least one move is winning or
-    // to a terminal win if all moves are losing; otherwise there's a mix of
-    // draws and losing, so at best it's a draw.
-    if (can_convert) {
-      // Doesn't give the correct distance to mate because siblings are not
-      // considered but more accurate than doing nothing. This shouldn't
-      // underestimate the distance to mate since at worst we miss shorter
-      // moves.
-      float terminal_m = std::max(losing_m, m) + 1.0f;
-      p->MakeTerminal(
-          v > 0.0f ? GameResult::BLACK_WON
-                   : all_losing ? GameResult::WHITE_WON : GameResult::DRAW,
-          terminal_m,
-          found_tb ? Node::Terminal::Tablebase : Node::Terminal::Terminal);
-    }
+    // Try setting parent bounds except the root or those already terminal.
+    update_parent_bounds = update_parent_bounds && p != search_->root_node_ &&
+                           !p->IsTerminal() && MaybeSetBounds(p, m);
 
     // Q will be flipped for opponent.
     v = -v;
@@ -1524,7 +1502,64 @@ void SearchWorker::DoBackupUpdateSingleNode(
   search_->total_playouts_ += node_to_process.multivisit;
   search_->cum_depth_ += node_to_process.depth * node_to_process.multivisit;
   search_->max_depth_ = std::max(search_->max_depth_, node_to_process.depth);
-}  // namespace lczero
+}
+
+bool SearchWorker::MaybeSetBounds(Node* p, float m) const {
+  auto losing_m = 0.0f;
+  auto prefer_tb = false;
+
+  // Determine the maximum (lower, upper) bounds across all children.
+  // (-1,-1) Loss (initial and lowest bounds)
+  // (-1, 0) Can't Win
+  // (-1, 1) Regular node
+  // ( 0, 0) Draw
+  // ( 0, 1) Can't Lose
+  // ( 1, 1) Win (highest bounds)
+  auto lower = GameResult::BLACK_WON;
+  auto upper = GameResult::BLACK_WON;
+  for (const auto& edge : p->Edges()) {
+    const auto [edge_lower, edge_upper] = edge.GetBounds();
+    lower = std::max(edge_lower, lower);
+    upper = std::max(edge_upper, upper);
+
+    // Checkmate is the best, so short-circuit.
+    const auto is_tb = edge.IsTbTerminal();
+    if (edge_lower == GameResult::WHITE_WON && !is_tb) {
+      prefer_tb = false;
+      break;
+    } else if (edge_upper == GameResult::BLACK_WON) {
+      // Track the longest loss.
+      losing_m = std::max(losing_m, edge.GetM(0.0f));
+    }
+    prefer_tb = prefer_tb || is_tb;
+  }
+
+  // The parent's bounds are flipped from the children (-max(U), -max(L))
+  // aggregated as if it was a single child (forced move) of the same bound.
+  //       Loss (-1,-1) -> ( 1, 1) Win
+  //  Can't Win (-1, 0) -> ( 0, 1) Can't Lose
+  //    Regular (-1, 1) -> (-1, 1) Regular
+  //       Draw ( 0, 0) -> ( 0, 0) Draw
+  // Can't Lose ( 0, 1) -> (-1, 0) Can't Win
+  //        Win ( 1, 1) -> (-1,-1) Loss
+
+  // Nothing left to do for ancestors if the parent would be a regular node.
+  if (lower == GameResult::BLACK_WON && upper == GameResult::WHITE_WON) {
+    return false;
+  } else if (lower == upper) {
+    // Search can stop at the parent if the bounds can't change anymore, so make
+    // it terminal preferring shorter wins and longer losses.
+    p->MakeTerminal(
+        -upper,
+        (upper == GameResult::BLACK_WON ? std::max(losing_m, m) : m) + 1.0f,
+        prefer_tb ? Node::Terminal::Tablebase : Node::Terminal::EndOfGame);
+  } else {
+    p->SetBounds(-upper, -lower);
+  }
+
+  // Bounds were set, so indicate we should check the parent too.
+  return true;
+}
 
 // 7. Update the Search's status and progress information.
 //~~~~~~~~~~~~~~~~~~~~
--- a/src/mcts/search.h
+++ b/src/mcts/search.h
@@ -310,6 +310,8 @@ class SearchWorker {
   void FetchSingleNodeResult(NodeToProcess* node_to_process,
                              int idx_in_computation);
   void DoBackupUpdateSingleNode(const NodeToProcess& node_to_process);
+  // Returns whether a node's bounds were set based on its children.
+  bool MaybeSetBounds(Node* p, float m) const;
 
   Search* const search_;
   // List of nodes to process.
--- a/src/chess/position.cc
+++ b/src/chess/position.cc
@@ -54,6 +54,12 @@ uint64_t Position::Hash() const {
 
 std::string Position::DebugString() const { return us_board_.DebugString(); }
 
+GameResult operator-(const GameResult& res) {
+  return res == GameResult::BLACK_WON
+             ? GameResult::WHITE_WON
+             : res == GameResult::WHITE_WON ? GameResult::BLACK_WON : res;
+}
+
 GameResult PositionHistory::ComputeGameResult() const {
   const auto& board = Last().GetBoard();
   auto legal_moves = board.GenerateLegalMoves();


src/mcts/node.h
```diff
@@ -164,15 +170,24 @@ class Node {
...
 
   // Makes the node terminal and sets it's score.
   void MakeTerminal(GameResult result, float plies_left = 0.0f,
                     Terminal type = Terminal::Terminal);
   // Makes the node not terminal and updates its visits.
   void MakeNotTerminal();
+  // Indicates the node cannot win (e.g., a child can draw).
+  void MakeCantWin() { terminal_type_ = Terminal::CantWin; }
+  // Indicates the node cannot lose (i.e., the best child cannot win).
+  void MakeCantLose() { terminal_type_ = Terminal::CantLose; }
```
Mardak: @borg323 suggested `p->MarkCantWin()` as a better verb/method name. Alternatively, does it make sense to have a single method that takes a `Node::Terminal` parameter, e.g., `p->SetTerminalType(Node::Terminal::CantWin)` although the caller could pass in any valid type… which might be okay?

src/mcts/search.h
```diff
@@ -300,6 +300,7 @@ class SearchWorker {
   void FetchSingleNodeResult(NodeToProcess* node_to_process,
                              int idx_in_computation);
   void DoBackupUpdateSingleNode(const NodeToProcess& node_to_process);
+  void MaybeConvertParent(Node* p, bool& can_convert, float m) const;
```
mooskagh: Non-const reference parameters are not allowed by the style guide (see also https://herbsutter.com/2020/02/23/references-simply/). For out-parameters use pointer instead (and make it the last parameter).
That way it's clear at the calling site that the parameter can be changed.

src/mcts/search.cc
```diff
@@ -1445,7 +1419,87 @@ void SearchWorker::DoBackupUpdateSingleNode(
...
+        max_U = std::max(0, max_U);
+      } else {
+        // Track the longest loss.
+        losing_m = std::max(losing_m, edge.GetM(0.0f));
+      }
+    } else if (edge.CantWin()) {
+      max_U = std::max(0, max_U);
+    } else if (edge.CantLose()) {
+      max_L = std::max(0, max_L);
+      max_U = 1;
```
mooskagh: Why is it so? It doesn't mean this edge can win, does it?

Mardak: Which part are you asking about? Specifically that `max_U = 1` statement? `CantLose` bounds are `(0, 1)` so it'll max the U to 1 like a Win or Regular node.

`CantLose` is introduced when all possible moves are `Loss` or `CantWin`, e.g., your opponent can choose between a capture to tablebase loss or play into a position where you can force a draw. This means before that opponent position, you are in a situation where you `CantLose` because in the worst case, you can play into the draw -- although you might play a different move that does lose similar to how your opponent could suboptimally play into the tablebase loss.

So pretending there was a single edge that's `CantLose`, the parent would then have a lower bound of `-max_U == -1` and upper bound `-max_L == 0` i.e., parent `CantWin`.

src/mcts/search.h
Mardak: I'll just switch this to a `bool` return value

src/mcts/node.h
```diff
@@ -131,7 +131,13 @@ class Node {
   using Iterator = Edge_Iterator<false>;
   using ConstIterator = Edge_Iterator<true>;
 
-  enum class Terminal : uint8_t { NonTerminal, Terminal, Tablebase };
+  enum class Terminal : uint8_t {
```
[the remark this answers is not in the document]
Mardak: Renamed `Terminal -> NodeType`, `Tablebase -> TablebaseTerminal` and `terminal_type_ -> type_`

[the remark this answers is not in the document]
Mardak: And `NonTerminal -> Regular`

src/mcts/search.cc
```diff
@@ -1395,39 +1401,8 @@ void SearchWorker::DoBackupUpdateSingleNode(
...
-      // moves.
-      float terminal_m = std::max(losing_m, m) + 1.0f;
-      p->MakeTerminal(
-          v > 0.0f ? GameResult::BLACK_WON
-                   : all_losing ? GameResult::WHITE_WON : GameResult::DRAW,
-          terminal_m,
-          found_tb ? Node::Terminal::Tablebase : Node::Terminal::Terminal);
-    }
+    // A new terminal among its siblings might require converting the parent.
+    can_convert = MaybeConvertParent(p, can_convert, m);
```
borg323: I think this would be clearer as
```c++
if (can_convert) {
  can_convert = MaybeConvertParent(p, m);
}
```
or similar to the old code
```c++
can_convert = can_convert && MaybeConvertParent(p, m);
```

Mardak: Switched back more to like the old code chaining `&&`s

src/mcts/node.h
Mardak: I switched it to `RemoveOutcome(GameResult)` similar to your suggestion and matching existing `MakeTerminal(GameResult, …)`

Mardak:
Rebased on top of #1151

src/mcts/search.cc
Mardak: Oh, I think I see what you were asking? "If the edge CantLose then why can it win with `max_U = 1`" ? The max_L and max_U values are from the perspective of the children, so in this case, processing a CantLose means we can increase the lower bound up to 0 while increasing the upper bound to the max 1.

The comment block below shows how these bounds are inverted and reversed for the parent.

src/mcts/search.cc
```diff
@@ -1466,7 +1441,88 @@ void SearchWorker::DoBackupUpdateSingleNode(
...
+      max_L = std::max(0, max_L);
+      max_U = 1;
+    } else {
+      // A regular or unvisited node maxes the upper bound.
+      max_U = 1;
+    }
+  }
+
+  auto parent_type =
+      found_tb ? Node::NodeType::TablebaseTerminal : Node::NodeType::Terminal;
```
mooskagh, line 1488: So even if we find tablebase loss and non-tablebase win, the node will still be marked as tablebase?.. I'm not sure that's intended.

src/mcts/search.cc
```diff
@@ -1466,7 +1441,88 @@ void SearchWorker::DoBackupUpdateSingleNode(
...
+  //       Draw ( 0, 0) -> ( 0, 0) Draw
+  // Can't Lose ( 0, 1) -> (-1, 0) Can't Win
+  //        Win ( 1, 1) -> (-1,-1) Loss
+  if (max_L == -1) {
+    if (max_U == -1) {
+      p->MakeTerminal(GameResult::WHITE_WON, std::max(losing_m, m) + 1.0f,
+                      parent_type);
+    } else if (max_U == 0) {
+      // No moves can win, so remove losing from parent.
+      p->RemoveOutcome(GameResult::BLACK_WON);
```
mooskagh, line 1504: It feels like `SetOutcome` or even `SetNodeType` would be more readable.

src/mcts/node.h
```diff
@@ -131,7 +131,13 @@ class Node {
   using Iterator = Edge_Iterator<false>;
   using ConstIterator = Edge_Iterator<true>;
 
-  enum class Terminal : uint8_t { NonTerminal, Terminal, Tablebase };
+  enum class NodeType : uint8_t {
+    Regular,
+    Terminal,
+    TablebaseTerminal,
+    CantWin,
```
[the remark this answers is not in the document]
mooskagh: Sorry for conflicting comments, they evolved as I kept reading the code. :)

I think it really should a bit mask. (or bit fields):

e.g
```
kCanWin  = 0x01
kCanDraw = 0x02
kCanLose = 0x04
kInitial = kCanWin+kCanDraw+kCanLose
kTerminal = 0x08   // Terminal is also certain, bit is set only for true game ending nodes. Maybe even makes sense to rename it. :)
kTablebase = 0x10  // Tablebase is not terminal but also certain
kCertain = 0x20 // Terminals and Tablebases are certain too.
```

Feel free to keep it as it is if you think it's too much hassle. I can do that later then.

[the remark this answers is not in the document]
Mardak: What would be the purpose of kCertain? Just a shortcut for `type_ & kInitial == kInitial`?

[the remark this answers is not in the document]
Mardak: I suppose similarly, the current `IsTerminal()` could be implemented as `type_ == kCanWin || type_ == kCanDraw || type_ == kCanLose` (i.e., one bit set in the bottom 3 bits) although having a `kTerminal` could optimize the check.

Vs having a `kEndsGame` to differently identify game ending terminals (from tablebase or two-fold) would be new functionality that isn't used yet.

src/mcts/search.cc
Mardak: When finding a win, the first one causes the parent to become a loss and no more siblings will be visited. In the cases involving draws and losses, the best outcome is a draw which doesn't care about tablebase-ness. And for all losses, if any are tablebase, the converted parent win would assume the longest/tablebase loss is played.

But yes I see how the bug can arise now. A tablebase loss is found first (and ignored) and later iteration of search finds a low prior win. I guess we need to accumulate `found_win_tb` separately from `found_loss_tb` ?

Mardak:
@mooskagh To be clear, the current implementation with
```c
  enum class NodeType : uint8_t {
    Regular,
    Terminal,
    TablebaseTerminal,
    CantWin,
    CantLose
  };```
takes advantage of reusing WL/Q's {>0, =0, <0}-ness to represent the different types of terminals reduce duplicating state.

There are benefits of decoupling that data into what we actually want: `bounds` (e.g., win, win and draw, initial) and `terminal_type` (non-terminal, end-of-game, tablebase, two-fold), so there does seem to be potential for combining these into a bitfield. Although there are non-nonsensical values that could be introduced when no longer "limited" by enum, e.g., bound = win, loss, !draw or a terminal that is both end-of-game and tablebase.

I'll think a bit more about whether refactoring this now into bits from enum vs cleaning up the names/apis and getting the logic right now to refactor later.

File: src/mcts/search.cc [hunk 9]
```diff
@@ -1466,7 +1441,88 @@ void SearchWorker::DoBackupUpdateSingleNode(
...
+  // Determine the maximum (Lower, Upper) bounds across all children.
+  // (-1,-1) Loss (initial and lowest bounds)
+  // (-1, 0) Can't Win
+  // (-1, 1) Regular node
+  // ( 0, 0) Draw
+  // ( 0, 1) Can't Lose
+  // ( 1, 1) Win (highest bounds)
+  auto max_L = -1;
+  auto max_U = -1;
+  for (const auto& edge : p->Edges()) {
```
Mardak, line 1459: I haven't rewritten this yet, but I would think with each node specifying its `canWin` `canDraw` and `canLose`, there shouldn't need to be a check for terminal/cantwin/etc., and there should be some simple accumulation. Although we probably still need some branching to support tb-terminal-ness.

src/mcts/node.cc
```diff
@@ -257,6 +257,15 @@ void Node::MakeNotTerminal() {
   }
 }
 
+void Node::RemoveOutcome(GameResult result) {
```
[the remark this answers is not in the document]
Mardak: Would `RestrictOutcome` seem less like removing?

src/mcts/node.cc
```diff
@@ -257,6 +257,15 @@ void Node::MakeNotTerminal() {
...
+  if (result == GameResult::WHITE_WON)
+    type_ = NodeType::CantWin;
```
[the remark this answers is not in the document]
Mardak: Added curly braces

src/mcts/search.cc
```diff
@@ -317,7 +317,12 @@ std::vector<std::string> Search::GetVerboseStats(Node* node) const {
     }
     oss << ") ";
 
-    if (edge.IsTerminal()) oss << "(T) ";
+    if (edge.IsTerminal())
```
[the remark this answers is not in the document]
Mardak: Added braces and checking node type

src/mcts/search.cc
```diff
@@ -1416,39 +1422,8 @@ void SearchWorker::DoBackupUpdateSingleNode(
...
-      // underestimate the distance to mate since at worst we miss shorter
-      // moves.
-      float terminal_m = std::max(losing_m, m) + 1.0f;
-      p->MakeTerminal(
-          v > 0.0f ? GameResult::BLACK_WON
-                   : all_losing ? GameResult::WHITE_WON : GameResult::DRAW,
-          terminal_m,
-          found_tb ? Node::Terminal::Tablebase : Node::Terminal::Terminal);
-    }
+    can_convert = can_convert && p != search_->root_node_ && !p->IsTerminal() &&
```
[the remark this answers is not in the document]
Mardak: Renamed to `update_ancestor_outcome` and updated comments that used to refer to "convert"

src/mcts/search.h
```diff
@@ -300,6 +300,7 @@ class SearchWorker {
   void FetchSingleNodeResult(NodeToProcess* node_to_process,
                              int idx_in_computation);
   void DoBackupUpdateSingleNode(const NodeToProcess& node_to_process);
+  bool MaybeConvertParent(Node* p, float m) const;
```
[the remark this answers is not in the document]
Mardak: I've renamed this to `MaybeRestrictOutcome` for now…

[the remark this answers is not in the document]
Mardak: And added comment to describe the return and behavior

src/mcts/search.cc
```diff
@@ -1466,7 +1441,88 @@ void SearchWorker::DoBackupUpdateSingleNode(
...
+  // (-1,-1) Loss (initial and lowest bounds)
+  // (-1, 0) Can't Win
+  // (-1, 1) Regular node
+  // ( 0, 0) Draw
+  // ( 0, 1) Can't Lose
+  // ( 1, 1) Win (highest bounds)
+  auto max_L = -1;
+  auto max_U = -1;
+  for (const auto& edge : p->Edges()) {
+    if (edge.IsTerminal()) {
```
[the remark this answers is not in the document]
Mardak: Converted to switch

src/mcts/node.h
```diff
@@ -372,6 +385,8 @@ class EdgeAndNode {
   // Whether the node is known to be terminal.
   bool IsTerminal() const { return node_ ? node_->IsTerminal() : false; }
   bool IsTbTerminal() const { return node_ ? node_->IsTbTerminal() : false; }
+  bool CantWin() const { return node_ ? node_->CantWin() : false; }
```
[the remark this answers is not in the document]
Mardak: Exposed `GetNodeType` and renamed the enum to have `CanWinOrDraw` and `CanLoseOrDraw`

src/mcts/search.cc
Mardak: I guess part of the reason why it's hard to read is from reusing `GameResult`

src/mcts/search.cc
```diff
@@ -1524,7 +1507,96 @@ void SearchWorker::DoBackupUpdateSingleNode(
...
+      case Node::NodeType::Terminal:
+      case Node::NodeType::TablebaseTerminal: {
+        found_tb = found_tb || edge.IsTbTerminal();
+
+        // Default_wl doesn't matter as WL is only used if IsTerminal is true.
+        const auto WL = edge.GetWL(0.0f);
+        // We can short-circuit after a single win maxes the bounds.
+        if (WL > 0.0f) {
+          max_L = max_U = 1;
+          break;
```
Mardak, line 1536: Oops, this short-circuit is broken with the `switch` but looks like we'll need to iterate over all wins anyway to find the shortest (non-tb) win anyway.

Mardak: Removed comment and break.

src/mcts/search.cc
Mardak: Added `found_mate` for preferring a checkmate (i.e., end of game win terminal) to then make the parent a terminal non-tb loss.

src/mcts/search.cc
Mardak: Rewrote with enum bounds saved in the node to then max directly.

src/mcts/search.cc
Mardak: Changed to `p->SetBounds(…)`

src/mcts/node.h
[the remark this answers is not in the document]
Mardak: I've completely rewritten using `enum GameResult` to store as unsigned 2-bit lower and upper bounds bit fields.

Mardak:
@mooskagh I've rewritten pretty much the whole PR using bit fields with 2 bits each for the terminal type, lower bound, upper bound. This actually simplifies a bunch of the if/switch branching that was necessary for handling the overloaded "NodeType" sometimes identifying terminals and sometimes identifying CantWin/CantLose.

Mardak:
Ehh… Looks like appveyor/cl.exe doesn't like the bit fields:
```
         c:\projects\lc0\src\mcts\node.h(339): error C2338: Unexpected size of Node [C:\projects\lc0\build\lc0@exe.vcxproj]
```

Works fine with macos/clang for me…

Mardak:
Seems okay with ubuntu16.04/gcc8.3 although it does warn but looks like gcc 8.4+ and 9.3+ will stop warning for this https://gcc.gnu.org/bugzilla/show_bug.cgi?id=61414#c29
```
../../src/mcts/node.h:314:29: warning: ‘lczero::Node::terminal_type_’ is too small to hold all values of ‘enum class lczero::Node::Terminal’
   Terminal terminal_type_ : 2;
                             ^
../../src/mcts/node.h:316:29: warning: ‘lczero::Node::lower_bound_’ is too small to hold all values of ‘enum class lczero::GameResult’
   GameResult lower_bound_ : 2;
                             ^
../../src/mcts/node.h:317:29: warning: ‘lczero::Node::upper_bound_’ is too small to hold all values of ‘enum class lczero::GameResult’
   GameResult upper_bound_ : 2;
                             ^
```

Mardak:
`unsigned char` looks good for msvc, gcc, clang. Pushing with `uint8_t` and added `operator-(GameResult)`

Mardak:
Here's a position from https://www.tcec-chess.com/archive.html?season=17&div=sf&game=58 but using 5 piece syzygy to derive a 6 piece draw where there's a mix of tablebase losses and draw and "can't lose." https://lichess.org/analysis/8/8/2k5/8/3qB3/5R2/1p1NK3/8_b_-_-_0_1
derive 6 piece draw
```
position fen 8/8/2k5/8/3qB3/5R2/1p1NK3/8 b - - 0 1
go nodes 10000 searchmoves d4e4

info string d4e4 N:   10030 (P:  2.36%) (WL:  0.00000) (D:  1.000) (M:  3.6) (V:  0.0000) (T) 

position fen 8/8/2k5/8/3qB3/5R2/1p1NK3/8 b - - 0 1 moves d4e4
go nodes 1

info string e2f1 N:      15 (P:  5.20%) (WL: -1.00000) (D:  0.000) (M:  1.5) (V: -1.0000) (T) 
info string e2d1 N:      20 (P:  6.71%) (WL: -1.00000) (D:  0.000) (M:  2.0) (V: -1.0000) (T) 
info string e2f2 N:      21 (P:  6.05%) (WL: -0.83994) (D:  0.153) (M:  2.6) (V: -0.9512) 
info string f3e3 N:      45 (P: 11.97%) (WL: -0.79727) (D:  0.194) (M:  2.4) (V: -0.7587) 
info string d2e4 N:    9059 (P: 70.07%) (WL:  0.00000) (D:  1.000) (M:  0.0) (V:  0.0000) (T) 
```
Black Qxe4 is a draw because the best white move is Nxe4 5-piece draw. Looking at the other white moves, both Kd1 and Kf1 are losing for white because the rook can be captured (5-piece loss because the overloaded knight can't guard the rook and black promotion). The remaining two white moves Kf2 and and Re3 are at best drawn (and are actually 6-piece loss) because black could capture the rook with the queen in both cases to find a 5-piece draw.

The current PR doesn't output (W) or (L) or any bounds which makes things a little harder to debug, but maybe it's not useful for most people anyway.

src/chess/position.h
```diff
@@ -77,7 +77,9 @@ class Position {
   int ply_count_ = 0;
 };
 
-enum class GameResult { UNDECIDED, WHITE_WON, DRAW, BLACK_WON };
+// These are ordered so max() prefers the best result.
+enum class GameResult : uint8_t { UNDECIDED, BLACK_WON, DRAW, WHITE_WON };
+GameResult operator-(const GameResult& res);
```
mooskagh, line 82: That's a bit of an abuse of operator overload. :)

src/mcts/search.cc
```diff
@@ -1524,7 +1493,64 @@ void SearchWorker::DoBackupUpdateSingleNode(
...
+  // (-1,-1) Loss (initial and lowest bounds)
+  // (-1, 0) Can't Win
+  // (-1, 1) Regular node
+  // ( 0, 0) Draw
+  // ( 0, 1) Can't Lose
+  // ( 1, 1) Win (highest bounds)
+  auto lower = GameResult::BLACK_WON;
+  auto upper = GameResult::BLACK_WON;
+  for (const auto& edge : p->Edges()) {
+    const auto bounds = edge.GetBounds();
```
mooskagh, line 1512: In C++17, you can write `auto [edge_lower, edge_upper] = edge.GetBounds()`.

src/mcts/node.h
```diff
@@ -301,9 +309,12 @@ class Node {
   // Index of this node is parent's edge list.
   uint16_t index_;
 
-  // 1 byte fields.
+  // Bit fields using parts of uint8_t fields initialized in the constructor.
   // Whether or not this node end game (with a winning of either sides or draw).
-  Terminal terminal_type_ = Terminal::NonTerminal;
+  Terminal terminal_type_ : 2;
```
mooskagh, line 314: Could you run a benchmark on a random backend to see how it affects performance?

src/chess/position.h
Mardak: As in revert this to the lambda I had before:
https://github.com/LeelaChessZero/lc0/commit/99abcc1587ce7734e75e9fa6941173cfbe7a7590#diff-83b7141f4fa67bcfdf95188467895cb5R1536-R1540
```c
  const auto FlipResult = [](const GameResult& res) {
    return res == GameResult::BLACK_WON ? …
```
Or somehow otherwise defined?

src/mcts/search.cc
Mardak: Neat. Switching to that and using it for verbose output too

src/mcts/node.h
Mardak: Here's 3 runs with master 1e649cb:
```

Comment (Mardak) [reply src/mcts/node.h:314]: Here's 3 runs with master 1e649cb with median 100938 nps:
```
./lc0 --smart-pruning-factor=0 --backend=random --backend-opts=uniform=true
ucinewgame
go nodes 2500000

info depth 6 seldepth 10 time 25001 nodes 2500153 score cp 0 nps 100270 tbhits 0 pv b1a3 g8h6 a1b1 b8a6 b1a1 a8b8 a1b1
info depth 6 seldepth 10 time 24834 nodes 2500155 score cp 0 nps 100938 tbhits 0 pv b1a3 g8h6 a1b1 b8a6 b1a1 a8b8 a1b1
info depth 6 seldepth 10 time 24784 nodes 2500153 score cp 0 nps 101138 tbhits 0 pv b1a3 g8h6 a1b1 b8a6 b1a1 a8b8 a1b1
```
And 3 runs with this PR 529406e with median 100872 nps:
```
info depth 6 seldepth 10 time 25160 nodes 2500260 score cp 0 nps 100082 tbhits 0 pv b1a3 b8a6 a1b1 a8b8 b1a1 b8a8 a1b1
info depth 6 seldepth 10 time 24848 nodes 2500114 score cp 0 nps 100872 tbhits 0 pv b1a3 g8h6 a1b1 b8a6 b1a1 a8b8 a1b1
info depth 6 seldepth 10 time 24735 nodes 2500106 score cp 0 nps 101346 tbhits 0 pv b1a3 g8h6 a1b1 b8a6 b1a1 a8b8 a1b1
```

So 99.9% nps of master

Comment (Mardak) [reply src/mcts/node.h:314]: And using `benchmark --num-positions=6` from #1069:

With master 1e649cb:
```
Total time (ms) : 60013
Nodes searched  : 6023873
Nodes/second    : 100374
```

This PR 529406e:
```
Total time (ms) : 60016
Nodes searched  : 5995393
Nodes/second    : 99895
```
99.5% nps of master

File: src/chess/position.h (hunk 18 above)
Comment (mooskagh) [reply src/chess/position.h:82]: Probably a function with name `FlipGameResult` is better (not as lambda, but here), but if you'd like to keep `operator-`, it's also fine.

File: src/mcts/node.h (hunk 20 above)
Comment (Mardak) [reply src/mcts/node.h:314]: Just to document why the bit fields are fast in our usage. The common case behavior of the bit field is to check `Node::IsTerminal() { return terminal_type != Terminal::NonTerminal; }` which is a comparison against `0`. The difference between when terminal_type_ was 1 byte vs 2 bits x86 assembly:
```diff
-        cmpb    $0, 78(%rdi)
+        testb   $3, 78(%rdi)
         setne   %al
```
Where before it would compare the whole byte against 0. With bit fields, it `and`s the value 3 (binary `0b11`) to get the bottom 2 bits for terminal_type_ which uses the same number of instructions when comparing against 0.

For other comparisons not against `0`, e.g., `upper_bound_ == DRAW`:
```
        movb    78(%rdi), %al
        andb    $48, %al
        cmpb    $32, %al
```
It needs to get the desired 2 bits and compare it to the appropriate value. Here `upper_bound_` are bits 4 and 5 (2^4 + 2^5 = 16 + 32 = 48) and DRAW's value is `2` shifted 4 bits left (2 << 4 = 32).

So something maybe to keep in mind as we add more bit fields and optimization.

PR #1162, opened 2020-03-28
Add input format type 3 support to encoder/training data/search.

Tilps:
I need to test it, and also modify node.cc GetV5TrainingData to make corresponding changes to the parts that aren't just sent in planes.

But most importantly I also need to apply transform on the output probabilities as well... I need to carefully construct 8 1858 element permutation arrays... and pipe the transform value (or the whole identification logic) across to ExpandNode in search.

So a long long way to go for this...  It is interesting that the 'input format' has an effect on the output format, but given they logically have to go together I'm not sure what I can do about that...


Tilps:
Writing tests actually found a bug (wrong mask for detecting king was on queen side), but I could write tests for days trying to get good coverage here.  I think its ready for review.

src/chess/bitboard.cc
```diff
@@ -281,6 +281,19 @@ const int kKingCastleIndex =
     kMoveToIdx[BoardSquare("e1").as_int() * 64 + BoardSquare("h1").as_int()];
 const int kQueenCastleIndex =
     kMoveToIdx[BoardSquare("e1").as_int() * 64 + BoardSquare("a1").as_int()];
+
+BoardSquare Transform(BoardSquare sq, int transform) {
```
mooskagh, line 285: I think it makes sense to introduce `enum` for transform values to have text representation of what every value means.

src/neural/encoder.cc
```diff
@@ -35,17 +35,136 @@ namespace {
 const int kMoveHistory = 8;
 const int kPlanesPerBoard = 13;
 const int kAuxPlaneBase = kPlanesPerBoard * kMoveHistory;
+
+int CompareTransposing(uint64_t value, int initial_transform) {
```
mooskagh, line 39: I guess it's more readable to have `ChessBoard` (or is it `BitBoard`) param rather than `uint64_t`.
Also will reduce amount of `as_int()` in the calling code.

src/neural/encoder.cc
```diff
@@ -35,17 +35,136 @@ namespace {
...
+  if ((initial_transform & 2) != 0) {
+    value = ReverseBytesInBytes(value);
+  }
+  auto alternative = TransposeBitsInBytes(value);
+  if (value < alternative) return -1;
+  if (value > alternative) return 1;
+  return 0;
+}
+
+int ChooseTransform(const ChessBoard& board) {
```
mooskagh, line 52: It's quite a heavy function, would be interesting to see the performance hit, especially near endgames.

src/chess/bitboard.cc
Tilps: Done.

src/neural/encoder.cc
Tilps: Its never called unless input format type is 3.
Its actually not too expensive unless the side to move king is on the diagonal with symmetrical other pieces, but even then, this function is not called within PickNodeToExtend inner loop nor the other significantly expensive loops.  Its only called once per node.  In that context it is not very heavy even if it was called outside of input_type 3.
Actually performing the transforms on all 112 bit planes is going to be considerably more expensive (but still not so expensive as to be a concern).

src/neural/encoder.cc
Tilps: Yeap, I recall thinking much the same myself, but forgot to come back to it.
Done.

src/utils/bititer.h
```diff
@@ -52,6 +52,16 @@ inline unsigned long GetLowestBit(std::uint64_t value) {
 #endif
 }
 
+enum BoardTransform {
```
mooskagh, line 55: It's not really nice to pollute the global namespace with simple names like `Flip`.
Maybe use `enum class` rather than plain `enum`?

src/neural/encoder.cc
mooskagh: Not really done. :)

src/neural/encoder.cc
```diff
@@ -56,74 +56,74 @@ int ChooseTransform(const ChessBoard& board) {
...
   // Our king is now always in bottom right quadrant.
   // Transpose for king in top right triangle, or if on diagonal whichever has
   // the smaller integer value for each test scenario.
   if ((our_king & 0xE0C08000ULL) != 0) {
-    transform |= 4;
+    transform |= Transpose;
   } else if ((our_king & 0x10204080ULL) != 0) {
     auto outcome =
         CompareTransposing((board.ours() | board.theirs()).as_int(), transform);
     if (outcome == -1) {
```
mooskagh, line 81: A reminder that style guide allows writing simple `if `statements which fit one line into one line:
```cpp
if (outcome == -1) return transform;
```

If would save quite a lot of lines, but it's up to you.

src/neural/encoder.cc
Tilps: Looks done to me - did you miss a 'refresh' ?

src/neural/encoder.cc
Tilps: yeah, I'll make the change.

src/utils/bititer.h
Tilps: I'd rather add the word Transform to all of them than deal with 'enum class' for a flags type enum.
So I did that.

PR #1166, opened 2020-03-29
Newer Android NDK for better C++17 support

lealgo:


appveyor.yml
```diff
@@ -44,12 +44,14 @@ install:
 - cmd: set PATH=C:\Python36;C:\Python36\scripts;%PATH%
 - cmd: pip3 install --upgrade meson
 - cmd: call "C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build\vcvarsall.bat" amd64
-- cmd: IF %NAME%==android C:\ProgramData\Microsoft\AndroidNDK64\android-ndk-r17\build\tools\make_standalone_toolchain.py --arch arm64 --api 21 --stl libc++ --install-dir \android-standalone-64
+- cmd: IF %NAME%==android IF NOT EXIST C:\cache\ndk appveyor DownloadFile https://dl.google.com/android/repository/android-ndk-r19c-windows-x86_64.zip
+- cmd: IF %NAME%==android IF NOT EXIST C:\cache\ndk 7z x android-ndk-r19c-windows-x86_64.zip -oC:\cache\ndk
```
borg323: This probably won't work - there is a size limit of 1Gb for caches, which I expect this will exceed. How big are the standalone toolchains? Maybe we can cache them instead.

lealgo: Thanks! I was going to ask you about the cache. But I think maybe we don't need to cache, as the download takes just a couple of seconds. So I guess I just extract to a different folder, right?

The size of the standalone toolchains is roughly the same as the whole ndk, and it's no longer needed to create them for newer ndks, in our case it would duplicate the space as we create two. I'll add another commit to use it in place.

felesays2422:
I will lo e to be  a part

PR #1164, opened 2020-03-28
Fixes #1106

mooskagh:

PR #1170, opened 2020-03-29
Only clear edge list if all children removed.

Tilps:
Fix a crash introduced in #1164.


mooskagh:
That returns the assert failure back though..

mooskagh:
Ok, after staring at that piece of code for 20 more minutes, it seems that it does fix all the ixxues.

PR #1171, opened 2020-03-29
Clearout trailing excess collisions before sending batch to gpu

Tilps:
Seems a good idea...
Currently testing.

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -803,6 +803,9 @@ void SearchWorker::ExecuteOneIteration() {
   // 2. Gather minibatch.
   GatherMinibatch();
 
+  // 2b. Remove any excess collisions.
+  ClearExcessCollisions();
+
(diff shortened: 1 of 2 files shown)


mooskagh:
The idea to do was after the batch was that if you undo it before, there's no new information in the tree, so the next thread just will gather the same collisions again.
Isn't it what happens?

Tilps:
The idea to undo it is so when a second thread lands its batch, there isn't a large number of collisions pointlessly still in the tree.
0 elo after 800 games in testing so far though :P

Tilps:
After 1k games its at -2.8 with +/- 7. (896 draws)
So pretty boring.
Note that benchmarking gives basically same nps on my machine for T70 for both before and after.  NPS is a bit more stable after possibly, but I only ran benchmark 3 times each.

I have a different idea now though...

PR #1177, opened 2020-04-07
Allow user override of input format to random backend.

Tilps:


src/neural/network_random.cc
```diff
@@ -108,7 +106,10 @@ class RandomNetwork : public Network {
   RandomNetwork(const OptionsDict& options)
       : delay_ms_(options.GetOrDefault<int>("delay", 0)),
         seed_(options.GetOrDefault<int>("seed", 0)),
-        uniform_mode_(options.GetOrDefault<bool>("uniform", false)) {}
+        uniform_mode_(options.GetOrDefault<bool>("uniform", false)),
+        capabilities_{static_cast<pblczero::NetworkFormat::InputFormat>(
+                          options.GetOrDefault<int>("input_mode", 1)),
```
borg323, line 111: I think this should be `pblczero::NetworkFormat::INPUT_CLASSICAL_112_PLANE` instead of 1?

Tilps: They are the same value, but I can make the line longer I guess, sure :P

Tilps: Done.

borg323: I wasn't 100% sure the first time I was reading this, so had to check the ,proto file.

PR #1179, opened 2020-04-07
make it possible to call backend without a net

borg323:
Changes all backends to make the WeightsFile passed to them optional, with an error if not there in all except random.
Also included is the hack to fix the c++ version incompatibility with tensorflow.

### DIFF
--- a/src/neural/factory.cc
+++ b/src/neural/factory.cc
@@ -1,6 +1,6 @@
 /*
   This file is part of Leela Chess Zero.
-  Copyright (C) 2018 The LCZero Authors
+  Copyright (C) 2018-2020 The LCZero Authors
 
   Leela Chess is free software: you can redistribute it and/or modify
   it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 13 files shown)


src/neural/factory.cc
```diff
@@ -123,7 +123,10 @@ std::unique_ptr<Network> NetworkFactory::LoadNetwork(
   } else {
     CERR << "Loading weights file from: " << net_path;
   }
-  const WeightsFile weights = LoadWeightsFromFile(net_path);
+  std::optional<WeightsFile> weights;
+  if (!net_path.empty()) {
```
Tilps, line 127: This means random backend needs an explicit --weights= since the default is auto discover.
But I'm fine with that...

Tilps: Err nevermind, I can't read.

src/neural/network_demux.cc
```diff
@@ -235,8 +235,12 @@ void DemuxingComputation::ComputeBlocking() {
   dataready_cv_.wait(lock, [this]() { return dataready_ == 0; });
 }
 
-std::unique_ptr<Network> MakeDemuxingNetwork(const WeightsFile& weights,
-                                             const OptionsDict& options) {
+std::unique_ptr<Network> MakeDemuxingNetwork(
```
Tilps, line 238: Technically these ones could defer failure to the child networks, but I'm fine with it not - given the unlikelyhood of people using a roundrobin/demux/mux of random backends more than once...

borg323: I wanted to have a first version ready quickly.

Tilps: update looks good.

Mardak:
Thanks. Ran into this when testing #1181 😉

Issue #1182, opened 2020-04-08
Compiling broken on Ubuntu 18.04


gcp:
Compilation is broken on Ubuntu 18.04, both with clang-9 and clang-10.

```
../../src/utils/optionsparser.cc:33:10: fatal error: 'charconv' file not found
#include <charconv>
         ^~~~~~~~~~
```

This is caused by the latest commit on master (a89a03e7d371c9a6f363d1e33d4a3e319aa78ec0), which tried to fix #1156 but introduced more problems.

gcp:
Someone pointed out that this guard should probably be improved: https://github.com/LeelaChessZero/lc0/blob/master/meson.build#L23

gcp:
I can confirm installing gcc-8 makes it work, so it is a requirement regardless of the clang in use.

Naphthalin:
Seems like the documentation is still too unprecise and not making clear that gcc >= 7.0 is needed either way.

gcp:
Which part is still ambiguous here?

codesensei2001:
I currently run ubuntu and running the ./build.sh command throws this error.
ERROR: Could not execute command "/usr/bin/python /home/clinton/Downloads/LeelaChessZero-lc0-v0.23.3-0-g29c15a4/LeelaChessZero-lc0-29c15a4/checkdir.py libs/lczero-common/proto".

PR #1183, opened 2020-04-08
Check for presence of charconv headers.

gcp:
Also update the instructions to indicate Ubuntu 18.04 needs GCC 8 or
later regardless of the clang in use, because clang falls back to the
GCC libstdc++ headers.

Fixes issue #1182.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,9 +80,9 @@ in ("Deep Learning").
 
 #### Ubuntu 18.04
 
-For Ubuntu 18.04 you need the latest version of meson and clang-6.0 before performing the steps above:
+For Ubuntu 18.04 you need the latest version of meson, g++-8 and clang-6.0 before performing the steps above:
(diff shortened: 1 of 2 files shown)


gcp:
I'm guessing the ubuntu 16.04 instructions also no longer work, but that's probably even more tricky to update as I'm not sure it'll be able to install gcc-8 without some extra ppas.

lealgo:
> I'm guessing the Ubuntu 16.04 instructions also no longer work, but that's probably even more tricky to update as I'm not sure it'll be able to install gcc-8 without some extra PPA.

#1184

Mardak:
Thanks for your contribution! If you want to participate on Discord (https://lc0.org/chat) you can now ask a moderator for the "Contributor" role to participate in #dev https://lczero.org/about/community/

PR #1190, opened 2020-04-09
Update the rpi section of README

lealgo:
Tested on a rpi zero with raspbian "buster". I looked into adding gcc-8 to an old raspbian but I gave up. It's way easier to upgrade to buster that has the right compilers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,6 +160,8 @@ Now download the lc0 source, if you haven't already done so, following the instr
 
 ### Raspberry Pi
 
+You'll need to be running the latest raspbian "buster".
+
 1. Install OpenBLAS
 
 ```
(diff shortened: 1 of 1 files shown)


rajb245:
Can you check if you really need g++8? See this PR for a similar issue on x86_64 Ubuntu: https://github.com/LeelaChessZero/lc0/pull/1191

The short of it is that you shouldn't need g++8 and clang6 when you're building with clang6. What you really need is libstdc++-8 and clang6. Can you do an apt remove g++8 stuff, autoremove all its dependencies, then just manually install libstdc++-8 and see if lc0 compile still works? And if it does, update your PR to reflect that? Thanks.

lealgo:
> Can you check if you really need g++8? See this PR for a similar issue on x86_64 Ubuntu: #1191
> 
> The short of it is that you shouldn't need g++8 and clang6 when you're building with clang6. What you really need is libstdc++-8 and clang6. Can you do an apt remove g++8 stuff, autoremove all its dependencies, then just manually install libstdc++-8 and see if lc0 compile still works? And if it does, update your PR to reflect that? Thanks.

I will check. Thanks.

lealgo:
> I don't know if its good practice to tell users they HAVE to update to a newer raspbian, but I believe that your instructions will work. I suggest taking the old instructions and putting them under a "Pre-Buster" heading, and putting your suggestions above that under a "Raspbian Buster" heading

Agree. But the thing is the pre-buster instructions won't work anymore since the move to c++17. And I can't find a way to add libstdc++8-dev to raspbian stretch.

PR #1191, opened 2020-04-10
Update README.md

rajb245:
It doesn't make any sense to have to install clang AND g++ on Ubuntu 18.04 when the build instructions explicitly say you're building with clang and the environment variables are enforcing that to the meson call. If you're building with clang, you don't need g++ at all. You DO need one of g++-8.0's dependencies, which is libstdc++-8. So this fix to the documentation makes that more clear and keeps the user from unnecessarily installing a whole compiler when just a library that the compiler depends on will do.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,21 +92,17 @@ Make sure that `~/.local/bin` is in your `PATH` environment variable. You can no
 
 #### Ubuntu 16.04
 
-For Ubuntu 16.04 you need the latest version of meson, ninja and also gcc-8.0 before performing the steps above:
+For Ubuntu 16.04 you need the latest version of meson, ninja, clang-6.0, and libstdc++-8:
 
+    wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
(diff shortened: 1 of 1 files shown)


lealgo:
Can you also update the 16.04 settings? These steps work for me:

```
wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
sudo apt-add-repository 'deb http://apt.llvm.org/xenial/ llvm-toolchain-xenial-6.0 main'
sudo add-apt-repository ppa:ubuntu-toolchain-r/test
sudo apt-get update
sudo apt-get install clang-6.0 libstdc++-8-dev
pip3 install meson --user
pip3 install ninja --user
CC=clang-6.0 CXX=clang++-6.0 INSTALL_PREFIX=~/.local ./build.sh
```

If you do you remember to remove the note about clang-6 below the current text.

rajb245:
Thanks, I think I've incorporated what you suggested. So g++8 no longer works as a compiler?

lealgo:
The g++8 can be installed from the same ppa and it works, just like it is in the README from master. You might want to add a similar alternative for g++ as the previous clang-6 note.

PR #1193, opened 2020-04-10
Update q to cp conversion formula

Ttl:
Based on the current TCEC games, the Q to centipawn conversion formula seems to be too pessimistic at high Q. lc0 cp evaluations are often much lower than the other engines.

I downloaded the TCEC SUFI evaluations and compared to the lc0 cp evaluations to SF evaluations. Below are the results:

lc0_q

The evaluation for Q < 0 comes from only one game and can be ignored. The current formula is the line labeled with centipawn. The first too optimistic conversion function is centipawn_2018 line. I also plotted the conversion function from #841 that used the old equation but fitted to more games. The #841 equation seems to fit very well to the median SF evaluation. I can get slightly better fit with more complex functions, but this one has advantage of being easily invertible which is nice for calculating the Q from centipawns. This PR reverts the conversion function to one from #841.

The average error to median SF eval on this dataset is 0.79 pawns with the current function, 0.28 with this function and 1.85 with the centipawn_2018.

The script for downloading evaluations and plotting the graphs can be found at: https://gist.github.com/Ttl/9926f70800b3fbd7314c150108d4ba61

### DIFF
--- a/src/mcts/params.cc
+++ b/src/mcts/params.cc
@@ -295,6 +295,7 @@ void SearchParams::Populate(OptionsParser* options) {
   options->Add<BoolOption>(kPerPvCountersId) = false;
   std::vector<std::string> score_type = {"centipawn",
                                          "centipawn_with_drawscore",
+                                         "centipawn_2019",
                                          "centipawn_2018",
                                          "win_percentage",
                                          "Q",
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -145,9 +145,11 @@ void Search::SendUciInfo() REQUIRES(nodes_mutex_) {
           std::round(edge.GetM(0.0f)) / 2 + (edge.IsTbTerminal() ? 101 : 1),
           wl);
     } else if (score_type == "centipawn_with_drawscore") {
-      uci_info.score = 295 * q / (1 - 0.976953126 * std::pow(q, 14));
+      uci_info.score = 111.714640912 * tan(1.5620688421 * q);
     } else if (score_type == "centipawn") {
-      uci_info.score = 295 * wl / (1 - 0.976953126 * std::pow(q, 14));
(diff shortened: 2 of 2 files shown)


mooskagh:
It seems that after every TCEC season we start to retune our formula as it doesn't fit the centipawn scale well.

I guess what's really happen is different networks have different Q scale so to say, meaning that different networks will need different tunes.

Maybe it makes sense to have a library of a few hundreds positions with SF eval and run Lc0 evals periodically on them to re-tune the formula..

Ttl:
Don't merge this just yet, I'll tune this further when there are slightly more games played in sufi,

Naphthalin:
I think this formula is much better than our current one for 2 reasons:
1.) it is invertible
2.) if I see that correctly, it reaches +10.0 just below our resign threshold in training (threshold 0.96, +10.0 at 0.934)
and would really appreciate to employ it as our new default formula.

zz4032:
> I guess what's really happen is different networks have different Q scale so to say, meaning that different networks will need different tunes.

I reran the adjusting process on random positions from human games as I did previously for the existing conversion formula.

**Current formula + old sample points (T40):**
previous
Indeed it seems that T60 networks produce much fewer extreme Q values (close to -1 or 1) compared to T40. Current formula (red) and new sample data with T60 don't correlate well starting from CP +/-300.
A small adjustment of the current formula however with exponent parameter changed from 14 to 8 shows a much better fit (green): `uci_info.score = 295 * wl / (1 - 0.976953126 * std::pow(wl, 8));`. Orange is PR1193.
**Current formula + new sample points (SV-T60):**
adjusted
Same diagram with a zoom at CP [-200;200]:
pr-close

Ttl:
All of the SUFI games are now played. The best fit is now slightly closer to the current formula than previously.

lc0_q

About +5 previous evaluation is now +10 with this formula.

PR #1195, opened 2020-04-11
Time management refactoring

mooskagh:


mooskagh:
Description in http://lczero.org/dev/docs/timemgr/

borg323:
I think the moves left estimator should be separated from the time manager. Keep the old one in the legacy time manager for compatibility, and I'll send a separate pr to rename it and add comments that actually reflect what it does, as with current defaults it certainly isn't estimating moves remaining.

PR #1201, opened 2020-04-13
add alphazero time manager

MelleKoning:
as per #968 but now on the new time manager factory.

startup with param:
`--time-manager=alphazero`

Comment copied over from #968 below: 

Optionally play with alphazero time settings.

After talking with ASilver on Discord, found the explanation that the original alphazero used some different time setting, which is just use 1/20th (thus 5%) of the remaining time available for the upcoming move.

image

Additionally I just added the wtime (the time increment) for use directly for the upcoming move as well here.

Tests done:
This time setting performs worse than the default lczero time setting. 
- In only 10 games (60sec+1sec) the result was 3 wins for lczero time, 1 for aztime, 6 draws.

- Other 10 games (10sec+0.1sec) the result was even worse:

> Score of lc0.net.42850 vs lc0az.net.42850: 5 - 1 - 4 [0.700]
> Elo difference: 147.19 +/- 197.89
> 
> 10 of 10 games finished.

However, this is just an option that does not affect engine strength if you do not enable it, so might be interesting for people who want to play a bit with the setting. You can alter the value used by Alphazero, dividing remaining time always by 20 to other values from 2: take half the time left for upcoming move, to 100: take 1% of time left for upcoming move, to see what difference that makes...

Have fun testing! :)

### DIFF
--- a/src/mcts/stoppers/alphazero.cc
+++ b/src/mcts/stoppers/alphazero.cc
@@ -63,31 +63,25 @@ class AlphazeroTimeManager : public TimeManager {
 std::unique_ptr<SearchStopper> AlphazeroTimeManager::GetStopper(
     const GoParams& params, const Position& position) {
   const bool is_black = position.IsBlackToMove();
-  std::optional<int64_t> time = (is_black ? params.btime : params.wtime);
+  const std::optional<int64_t>& time = (is_black ? params.btime : params.wtime);
   // If no time limit is given, don't stop on this condition.
   if (params.infinite || params.ponder || !time) return nullptr;
 
-  std::optional<int64_t> inc = is_black ? params.binc : params.winc;
+  const std::optional<int64_t>& inc = is_black ? params.binc : params.winc;
   const int increment = inc ? std::max(int64_t(0), *inc) : 0;
 
-  // How to scale moves time.
-  float this_move_time = 1.0f;
-  int time_to_squander = 0;
+  auto total_moves_time = *time + increment - move_overhead_;
+  
+  // use the increment in the first upcoming move
+  float this_move_time = increment + (total_moves_time / alphazerotimevalue_);
 
-  auto total_moves_time = *time - move_overhead_;
-  this_move_time =
-      increment +
-      (total_moves_time / alphazerotimevalue_);
-
-  LOGFILE << "Budgeted time for the move: " << this_move_time << "ms(+"
-          << time_to_squander << "ms to squander). Remaining time " << *time
+  LOGFILE << "Budgeted time for the move: " << this_move_time << "ms"
+          << "Remaining time " << *time
           << "ms(-" << move_overhead_ << "ms overhead)";
-  // Use `time_to_squander` time immediately.
-  this_move_time += time_to_squander;
-
+  
   // Make sure we don't exceed current time limit with what we calculated.
   auto deadline =
-      std::min(static_cast<int64_t>(this_move_time), *time - move_overhead_);
+      std::min(static_cast<int64_t>(this_move_time), *time + increment - move_overhead_);
   return std::make_unique<AlphazeroStopper>(deadline, &time_spared_ms_);
 }
 
--- a/src/mcts/stoppers/alphazero.cc
+++ b/src/mcts/stoppers/alphazero.cc
@@ -50,7 +50,7 @@ class AlphazeroTimeManager : public TimeManager {
         alphazerotimevalue_(params.GetOrDefault<float>("alphazero-time-value", 20.0f)),
         spend_saved_time_(params.GetOrDefault<float>("immediate-use", 1.0f)) {}
   std::unique_ptr<SearchStopper> GetStopper(const GoParams& params,
-                                            const Position& position) override;
+                                            const NodeTree& tree) override;
 
  private:
   const int64_t move_overhead_;
@@ -61,7 +61,8 @@ class AlphazeroTimeManager : public TimeManager {
 };
 
 std::unique_ptr<SearchStopper> AlphazeroTimeManager::GetStopper(
-    const GoParams& params, const Position& position) {
+    const GoParams& params, const NodeTree& tree) {
+  const Position& position = tree.HeadPosition();
   const bool is_black = position.IsBlackToMove();
   const std::optional<int64_t>& time = (is_black ? params.btime : params.wtime);
   // If no time limit is given, don't stop on this condition.
--- a/src/mcts/stoppers/alphazero.cc
+++ b/src/mcts/stoppers/alphazero.cc
@@ -35,7 +35,7 @@ class AlphazeroStopper : public TimeLimitStopper {
  public:
   AlphazeroStopper(int64_t deadline_ms, int64_t* time_piggy_bank)
       : TimeLimitStopper(deadline_ms), time_piggy_bank_(time_piggy_bank) {}
-  virtual void OnSearchDone(const IterationStats& stats) override {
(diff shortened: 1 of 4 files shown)


MelleKoning:
output of ```--help``` seems confusing

startup with ```--time-manager=alphazero``` works but ```--help``` gives
```
--alphazero-time-value=2.00..100.00
               Remaining time is divided by this value. Default value of 20 uses alphazero
               time.Lower values will spend more time in beginning of game, higher values will
               save more timefor later in the game
               [UCI: AlphazeroTimeValue  DEFAULT: 20.00  MIN: 2.00  MAX: 100.00]

       --time-manager=STRING
               Name and config of atime manager.
               [UCI: TimeManager  DEFAULT: legacy]

       --alphazero=STRING
               Makes the engine use AlphaZero time. The alphazero engine always budgeted 5
               percent of total time left for the first upcoming move, thereby gradually taking
               less time for each consecutive move in the game..
               [UCI: Alphazero  DEFAULT: alphazero]
```

Any ideas to make this better?

mooskagh:
Like with backend options which are different per different backend, time manager options are not thought to be discoverable for now other than in external documentation (e.g. see http://lczero.org/dev/docs/timemgr/#configuration-parameters for "smooth" time manager).

Also it doesn't really good to have global options that only work in some cases (e.g. in your case alphazero-time-value won't affect anything with the default time manager, but still will be displayed). So for now the recommended way is to access parameters through OptionsDict passed to the TimeManager.

There are some thoughts to change it, but for now that's what we have.

MelleKoning:
@mooskagh updated accordingly, extended the ```--help``` documentation a bit

src/mcts/stoppers/alphazero.cc
```diff
@@ -0,0 +1,100 @@
...
+  std::optional<int64_t> time = (is_black ? params.btime : params.wtime);
+  // If no time limit is given, don't stop on this condition.
+  if (params.infinite || params.ponder || !time) return nullptr;
+
+  std::optional<int64_t> inc = is_black ? params.binc : params.winc;
+  const int increment = inc ? std::max(int64_t(0), *inc) : 0;
+
+  // How to scale moves time.
+  float this_move_time = 1.0f;
+  int time_to_squander = 0;
```
mooskagh, line 75: It seems that this variable is always 0.

src/mcts/stoppers/factory.cc
```diff
@@ -42,9 +43,19 @@ const OptionId kMoveOverheadId{
...
     "communication, etc)."};
-const OptionId kTimeManagerId{"time-manager", "TimeManager",
-                              "Name and config of atime manager."};
-
+const OptionId kTimeManagerId{
+    "time-manager", "TimeManager",
+    "Name and config of atime manager. "
+    "Default is 'legacy'. "
+    "Optional time manager is 'alphazero' "
+    "'alphazero' Makes the engine use AlphaZero time. The alphazero engine always "
```
mooskagh, line 51: It seems that the description gives disproportionally large attention to "alphazero" mode. :)
While leaving out all legacy parameters. I suggest leaving a link to `lczero.org/play/flags/timemgr/` (or `lc0.org/timemgr`) here, and creating a page on a website with the description.

MelleKoning: Moved explanation to the dev documentation, please see
https://github.com/LeelaChessZero/lczero.org/pull/29

MelleKoning:
@mooskagh merged with master; it had a change in TimeManager base class which caused merge conflict, is now resolved.

MelleKoning:
@Naphthalin @mooskagh @Tilps Have merged with latest 0.25, no merge conflicts. 
Can you please approve and merge?

src/mcts/stoppers/factory.cc
```diff
@@ -43,9 +44,12 @@ const OptionId kMoveOverheadId{
     "Amount of time, in milliseconds, that the engine subtracts from it's "
     "total available time (to compensate for slow connection, interprocess "
     "communication, etc)."};
-const OptionId kTimeManagerId{"time-manager", "TimeManager",
-                              "Name and config of atime manager."};
-
+const OptionId kTimeManagerId{
+    "time-manager", "TimeManager",
+    "Name and config of a time manager. "
+    "Default is 'legacy'. Optional time manager is 'alphazero' "
```
mooskagh, line 50: I still think 'alphazero' time manager gets disproportionally more attention here than 'smooth'. :-P

Also http://lczero.org/dev/docs/timemgr is a design doc for my PR, it's not an instuction to users. Better have http://lczero.org/play/configuration/flags/ and describe all parameters there (not really needed to do that before merging though).

Also maybe even better will be to have a short url, https://lc0.org/timemgr.

How about `Name and configuration of a time manager. Possible names are 'legacy' (default), 'smooth-experimental' and 'alphazero'. See https://lc0.org/timemgr for configuration details.` ?

MelleKoning: Updated in github editor as suggested :)

MelleKoning:
Link to (draft) documentation available: 
https://github.com/LeelaChessZero/lczero.org/pull/29/

src/mcts/stoppers/alphazero.cc
```diff
@@ -0,0 +1,95 @@
...
+
+namespace lczero {
+
+namespace {
+
+class AlphazeroStopper : public TimeLimitStopper {
+ public:
+  AlphazeroStopper(int64_t deadline_ms, int64_t* time_piggy_bank)
+      : TimeLimitStopper(deadline_ms), time_piggy_bank_(time_piggy_bank) {}
+  virtual void OnSearchDone(const IterationStats& stats) override {
```
borg323, line 38: Why virtual?

MelleKoning: It's just as copied over from legacy.cc (assuming that's what is wanted)

MelleKoning: have removed the virtual now

src/mcts/stoppers/alphazero.cc
```diff
@@ -0,0 +1,95 @@
...
+
+ private:
+  int64_t* const time_piggy_bank_;
+};
+
+class AlphazeroTimeManager : public TimeManager {
+ public:
+  AlphazeroTimeManager(int64_t move_overhead, const OptionsDict& params)
+      : move_overhead_(move_overhead),
+        alphazerotimevalue_(params.GetOrDefault<float>("alphazero-time-value", 20.0f)),
```
mooskagh, line 50: The variable (both configuration and C++ variable) name is quite too generic. :)
`time-use-fraction` or something like that might be better, but also not good as it's **inverted** fraction..

src/mcts/stoppers/alphazero.cc
```diff
@@ -0,0 +1,95 @@
...
+  float this_move_time = increment + (total_moves_time / alphazerotimevalue_);
+
+  LOGFILE << "Budgeted time for the move: " << this_move_time << "ms"
+          << "Remaining time " << *time
+          << "ms(-" << move_overhead_ << "ms overhead)";
+  
+  // Make sure we don't exceed current time limit with what we calculated.
+  auto deadline =
+      std::min(static_cast<int64_t>(this_move_time), *time + increment - move_overhead_);
+  return std::make_unique<AlphazeroStopper>(deadline, &time_spared_ms_);
```
borg323, line 86: This can be just `return std::make_unique<TimeLimitStopper>(deadline);` since `time_spared_ms_` is updated but not used, and the `AlphazeroStopper` can be removed.

PR #1202, opened 2020-04-14
Fill the history with repeats rather than leaving empty for type 3.

Tilps:
Still early days, but it seems this makes 'all' the difference when it comes to trying to train a network with input type 3.
Will propose this properly if I get over 1k elo :P

I don't like the mess I've made of the code doing this, but I'm not sure how to make it much cleaner.


Tilps:
This did seem quite a lot better, but there was a massive bug which means that both sides of my testing scenario were extremely corrupted.
I'll retest it after I'm done testing the original bug fix.

Tilps:
Positive signs for this PR are starting to fade, still another day of testing to go but, its not really looking super promising.

Tilps:
Testing completed with no notable positive signs.

PR #1203, opened 2020-04-15
Rename training data variables for clarity, and fix bug with training data exposed.

Tilps:
I almost did this rename back when I wrote the original PR - just suffering from too much laziness :P

### DIFF
--- a/src/neural/writer.h
+++ b/src/neural/writer.h
@@ -26,7 +26,9 @@
 */
 
 #include <zlib.h>
+
 #include <fstream>
+
 #include "utils/cppattributes.h"
 
 #pragma once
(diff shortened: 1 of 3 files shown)


src/neural/writer.h
```diff
@@ -44,9 +46,13 @@ struct V5TrainingData {
...
   uint8_t castling_them_oo;
-  uint8_t side_to_move;
+  // For input type 3 contains enpassant column as a mask.
+  uint8_t side_to_move_or_enpassant;
   uint8_t rule50_count;
-  uint8_t deprecated_move_count;  // left in to keep 8 int8 fields.
+  // For input type 3 contains a bit field indicating the transform that was
+  // used and the original side to move info.
+  // Side to move is in the top bit, transform in the lower bits.
+  uint8_t deprecated_move_count_or_invariance_info;
```
borg323, line 55: Why not just rename it to `invariance_info`? Maybe also add a comment that it used to be the `deprecated_move_count` that was kept for padding.

PR #1204, opened 2020-04-15
Fix for starting fens that differ only in 50 move ply.

Tilps:


src/mcts/node.cc
```diff
@@ -468,7 +468,9 @@ bool NodeTree::ResetToPosition(const std::string& starting_fen,
   int no_capture_ply;
   int full_moves;
   starting_board.SetFromFen(starting_fen, &no_capture_ply, &full_moves);
-  if (gamebegin_node_ && history_.Starting().GetBoard() != starting_board) {
+  if (gamebegin_node_ &&
+      (history_.Starting().GetBoard() != starting_board ||
+       history_.Starting().GetNoCaptureNoPawnPly() != no_capture_ply)) {
```
Naphthalin, line 473: replace `GetNoCaptureNoPawnPly()` by `GetRule50Ply()` from #1207

PR #1208, opened 2020-04-15
Apply mlh-max-effect at the end

Naphthalin:
To ensure that `moves-left-head-max-effect` is actually the maximum effect we need to apply the clamping at the end. This PR fixes that.


Tilps:
I don't think we really want this - it means that the number of moves that maximum effect is applied at depends on the Q.

Tilps:
Closing as discussed.

PR #1210, opened 2020-04-17
workaround for partial c++17 compiler support

borg323:


borg323:
@leedavid for Ubuntu 16.04 we have the following instructions in the README:
```
sudo add-apt-repository ppa:ubuntu-toolchain-r/test
sudo apt-get update
sudo apt-get install gcc-8 g++-8
pip3 install meson --user
pip3 install ninja --user
CC=gcc-8 CXX=g++-8 INSTALL_PREFIX=~/.local ./build.sh
```
Assuming this patch works for you, you only need g++-8 on top of your existing setup. The only issue we know is that g++-8 istoo new for cuda 10.0. Is there something else missing?

PR #1212, opened 2020-04-18
Start TempDecay only after a given number of moves

Naphthalin:
This PR allows for temp decay to start only after a given number of moves, and interprets `TempDecayMoves` as the length of the decay, starting at `TempDecayStartMove`. Default value `0` maintains backwards compatibility with old training parameters.

The first commit still allows temperature to be below endgame temperature during the decay, while the 2nd commit fixes this and caps the temp decay at the endgame temperature.


Naphthalin:
> The first commit still allows temperature to be below endgame temperature during the decay, while the 2nd commit fixes this and caps the temp decay at the endgame temperature.

The current formula doesn't check whether temperature has decayed below the specified endgame temperature. As I think this is unintended behavior and we neither did that intentionally in the past nor have any use case for this, the next commit also fixes this behavior.

Tilps:
I think discussion suggested hiding all the temp parameters.

src/mcts/params.cc
```diff
@@ -91,9 +91,14 @@ const OptionId SearchParams::kTemperatureId{
...
     "tempdecay-moves", "TempDecayMoves",
-    "Reduce temperature for every move from the game start to this number of "
-    "moves, decreasing linearly from initial temperature to 0. A value of 0 "
-    "disables tempdecay."};
+    "Reduce temperature for every move from decay start move, decreasing "
+    "linearly over this number of moves from initial temperature to 0. "
+    "A value of 0 disables tempdecay."};
+const OptionId SearchParams::kTempDecayStartMoveId{
+    "tempdecay-start-move", "TempDecayStartMove",
+    "Start the linear decrease of temperature after this number of moves, "
```
mooskagh, line 99: It seems wrong to number moves from 0?
Also to me "starting decay from move 2" sounds like move should already receive reduced temperature at move, and I guess that doesn't happen here.


E.g. currently if --temp-decay-start-move=1, then the first move to recieve reduced temperature is actually move 3.

Tilps: Maybe rename it to temp decay delay? - 0 means the decay line attaches at move 1, 1 means it attaches at move 2 (3rd ply)

Naphthalin: I would be fine with either, renaming the variable or correct the off-by-one error explicitly.

> E.g. currently if --temp-decay-start-move=1, then the first move to recieve reduced temperature is actually move 3.

but this already means that we have an off by one error already in the current code, right?

PR #1213, opened 2020-04-18
add a helpful backendbench assistant

borg323:


Tilps:
I'm fine with the concept - not so sure about the flag name :P

borg323:
I'm more worried about the default settings. Any suggestions (including the name) are appreciated.

Tilps:
I'm not sure what good values would be either.  If anything I would be thinking that the percentage increase without 'run' to trigger new best should have some relationship with the ratio of the current minibatch size with the last peak minibatch size - but again I wouldn't know what to use for the new constants that formula would invoke.

borg323:
I couldn't decide on a better name, so I went all-in:
```
  __
 /  \
 |  |
 +  +    _____________________________________________________
(@)(@) _|                                                     |
 |  |  \  80 looks like the best minibatch-size for this net. |
 || |/  |_____________________________________________________|
 || ||
 |\_/|
 \___/
```

mooskagh:
I tried to run it, and it feels a bit too spammy, especially in the beginning, making useful data to noise ratio on the screen pretty los. Maybe only output it not more frequent than ~once per 10 seconds?

src/benchmark/backendbench.cc
```diff
@@ -128,11 +129,20 @@ void BackendBenchmark::Run() {
...
           best_nps = nps;
           best = i;
-          Clippy(std::to_string(best) +
-                 " looks like the best minibatch-size for this net (so far).");
           run = true;
+          if (!pending) {
+            pending = std::chrono::steady_clock::now();
+          }
+        }
+        if (pending) {
```
mooskagh, line 137: I think it's also needed to check after the loop if anything is pending?
Otherwise the recommendations from the last 10 seconds are lost.

borg323: There is an additional report after the loop (with slightly different text), so no need for that.

PR #1216, opened 2020-04-19
add option to start counting time with 'go'

borg323:


src/engine.cc
```diff
@@ -58,6 +58,9 @@ const OptionId kUciChess960{
     "Castling moves are encoded as \"king takes rook\"."};
 const OptionId kShowWDL{"show-wdl", "UCI_ShowWDL",
                         "Show win, draw and lose probability."};
+const OptionId kStrictUciTiming{"strict-uci-timing", "StrictTiming",
+                                "The UCI host compensates for lag, and waits "
+                                "for the 'readyok' reply before sending 'go'."};
```
Tilps, line 63: 'and only start timing when go arrives'?

borg323: I reworded it a bit.

MelleKoning:
Sorry, too late to the party but thumbs up 👍

PR #1220, opened 2020-04-21
Fix castling in pgn parsing for chess 960 start positions.

Tilps:


src/chess/pgn.h
```diff
@@ -186,10 +186,17 @@ class PgnReader {
     } else if (san[0] == 'O' && san.size() > 2 && san[1] == '-' &&
                san[2] == 'O') {
       Move m;
+      auto king_board = board.kings() & board.ours();
+      BoardSquare king_sq;
+      for (auto sq : king_board) {
+        king_sq = sq;
```
Mardak, line 192: I see `ChessBoard` doesn't have any methods returning `BoardSquare` but this seems unnecessarily roundabout given that `BoardSquare our_king_` exists…
https://github.com/LeelaChessZero/lc0/blob/186c2a2986df482b3187e0ab4ccbf8d008395fac/src/chess/board.h#L194-L195

But then again @mooskagh has suggested to avoid unnecessarily adding methods (at least to Node/EdgeAndNode).

ddobbelaere: Some methods (including `our_king()`) have been removed in the refactoring PR #1132.
Possibly it makes sense to reintroduce a method like our_king_square again (which would just return a number and not loop over a whole bitboard with a bitscan instruction). It indeed feels ugly...

ddobbelaere: But then again, the same refactoring PR changed a call to `board.our_king()` into `board.kings() & board.ours()` in this very file, so it makes no sense to do the opposite now (unless you revert the other PR).

PR #1228, opened 2020-04-22
Cuda winograd

ankan-ban:
An implementation of custom winograd algorithm for cuda backends. Similar to the dx12 backend, it uses custom CUDA C kernels for input and output transforms (fused with optional SE), and cublas gemm for the matrix multiplication.

On RTX 2060, its a bit slower than than cudnn's implicit_gemm algorithm for 256f networks, but it could be faster for bigger networks and on GPUs with higher memory bandwidth like the Tesla V100s.


ankan-ban:
some benchmarks on RTX 2060:

----------------------------------------------------
fp16, new
===========================
Total time (ms) : 341125
Nodes searched  : 5229494
Nodes/second    : 15330

fp16, old
===========================
Total time (ms) : 340658
Nodes searched  : 6964405
Nodes/second    : 20444

----------------------------------------------------

fp32, existing
===========================
Total time (ms) : 342091
Nodes searched  : 2013886
Nodes/second    : 5887

fp32, new
===========================
Total time (ms) : 341627
Nodes searched  : 3105669
Nodes/second    : 9091

-------------------------------------------------------

lp200:
V100
384x30-t40-1573.pb.gz
batchsize=256 
starting position 

PR1228
Avg batch size: 240.2, NN eval time: 2.32663 seconds per 24020 evals. NPS: 10323.9
info depth 20 seldepth 58 time 50891 nodes 589074 score cp 24 nps 20303

24.1
Avg batch size: 245.04, NN eval time: 3.50061 seconds per 24504 evals. NPS: 6999.93
info depth 20 seldepth 58 time 46614 nodes 587051 score cp 24 nps 13797

src/neural/cuda/fp16_kernels.cu
```diff
@@ -35,26 +35,25 @@ namespace cudnn_backend {
...
 // C threads per block.
 // 'HWC' input data processed by thread block.
 // Each thread processes 8x8 elements.
 // K is the no. of outputs of first fully connected layer (same as no. of inputs
-// for second fully connected layer). 
+// for second fully connected layer).
 // The kernel assumes K <= C.
 
-#define readw1(row, col) (w1[(row)*K + (col)])
+#define readw1(row, col) (w1[(row)*se_K + (col)])
```
Tilps, line 48: This is a curious change that doesn't quite feel like its just formatting or anything specific to do with the rest of the PR.

src/neural/cuda/network_cudnn.cc
```diff
@@ -547,22 +585,30 @@ class CudnnNetwork : public Network {
...
+        // as part of SE).
+        if (has_se_) {
+          network_[l++]->Eval(batchSize, tensor_mem_[1], tensor_mem_[0],
+                              nullptr, scratch_mem_, scratch_size_, cudnn_,
+                              cublas_);  // conv2
+        } else {
+          network_[l++]->Eval(batchSize, tensor_mem_[2], tensor_mem_[0],
+                              tensor_mem_[2], scratch_mem_, scratch_size_,
+                              cudnn_,
+                              cublas_);  // conv2
```
Tilps, line 603: This wrapping seems a bit non-standard. But it was also that way before...

src/neural/cuda/fp16_kernels.cu
Tilps: Okay I see why - its because exactly the same defines are in winograd_helper.h and they would be an incorrect redefinition otherwise.

Maybe just remove these and depend on the winograd_helper ones?

src/neural/cuda/winograd_helper.h
```diff
@@ -0,0 +1,387 @@
+/*
```
Tilps, line 1: I'm not sure but this almost feels more like a .inc than a .h - maybe?

ankan-ban: Exactly. I had to put the code in that file and include from two source files because different template instances of the same functions need to be compiled with different SM versions. Renamed to .inc

src/neural/cuda/winograd_helper.inc
```diff
@@ -0,0 +1,387 @@
...
+#define TEMP_INDEX_HWNC(h, w, n, c) \
+  ((h)*6 * GemmN * C + (w)*GemmN * C + (n)*C + c)
+
+// 'C' threads per block
+// 'N' blocks
+// every thread transforms an entire board/plane (8x8 elements)
+// - producing 4 x 6x6 elements
+template <typename T>
+__global__ void InputTransform_kernel(int N, int C, const T* input, T* output) {
+  constexpr int H = 8, W = 8;
```
Ttl, line 143: nvcc warns that H and W are not used.

src/neural/cuda/winograd_helper.inc
```diff
@@ -0,0 +1,387 @@
...
+// output is NCHW
+// 'C' threads per block
+// 'N' blocks
+// every thread generates an entire board/plane (8x8 elements)
+template <typename T, bool use_se, bool relu, bool use_bias, bool use_skip>
+__global__ void OutputTransform_kernel(int N, int C, int se_K, T* output,
+                                       const T* input, const T* skip,
+                                       const T* bias, const T* w1, const T* b1,
+                                       const T* w2, const T* b2) {
+  constexpr int H = 8, W = 8;
```
Ttl, line 249: Same warning here.

src/neural/cuda/winograd_helper.inc
ankan-ban: Thanks for catching. Removed them. There are other assumptions based on W being 8, so no point in defining as constant with the hope that the kernel will still work if someone changes it to something else.

Issue #1200, opened 2020-04-13
When MLH enabled, verbose move stats is not so useful.


Tilps:
We only output Q+U and raw M - no Q+U+(calculated M effect), so its hard to know the actual current active ordering.

mooskagh:
Would be nice to fix that for v0.25.

Naphthalin:
I agree, the `M` term should be included in the verbose `Q+U`.

AlexisOlson:
I've started an attempt at resolving this.

https://github.com/LeelaChessZero/lc0/compare/master...AlexisOlson:issue-1200

mooskagh:
Fixed via #1230

PR #1230, opened 2020-04-22
MLH verbose stats - Issue 1200 

AlexisOlson:
An attempt at fixing this issue:

https://github.com/LeelaChessZero/lc0/issues/1200


src/mcts/search.cc
```diff
@@ -276,6 +281,10 @@ std::vector<std::string> Search::GetVerboseStats(Node* node) const {
 
     oss << std::left << std::setw(5)
         << edge.GetMove(is_black_to_move).as_string();
+    
+    float Q = edge.GetQ(fpu, draw_score, /* logit_q= */ false);
```
Tilps, line 285: The M_effect calculation needs to be gated by whether M effect is applied or not by checking the MovesLeftThreshold like is done in search.
Also not sure why you changed Q here to not pass logit_q, that seems wrong.

PR #1234, opened 2020-04-24
Allow most parts of fen to be optional.

Mardak:
r? @Tilps or @mooskagh Fen handling already allows some leeway, so this makes it even more generous assuming white to move, ~all castling (including 960)~ no castling, no en passant, 0 rule50ply, 1 total moves.

Before:
```
position fen K1k1r b
go nodes 1
error Bad fen string: K1k1r b
```

After:
```
position fen K1k1r b
go nodes 1
info … multipv 14 pv e8h8

position fen K1k1r b moves c8e8
go nodes 1
info … multipv 2 pv a8b7
```
~Where the last position shows the castling of black and still supporting "moves"~
fen castle

### DIFF
--- a/src/chess/board.cc
+++ b/src/chess/board.cc
@@ -969,22 +969,34 @@ MoveList ChessBoard::GenerateLegalMoves() const {
   return result;
 }
 
-void ChessBoard::SetFromFen(const std::string& fen, int* rule50_ply,
-                            int* moves) {
+void ChessBoard::SetFromFen(string fen, int* rule50_ply, int* moves) {
   Clear();
   int row = 7;
   int col = 0;
 
+  // Remove any trailing whitespaces to detect eof after the last field.
+  fen.erase(std::find_if(fen.rbegin(), fen.rend(),
+                         [](char c) { return !std::isspace(c); })
+                .base(),
+            fen.end());
+
   std::istringstream fen_str(fen);
   string board;
-  string who_to_move;
-  string castlings;
-  string en_passant;
-  int rule50_halfmoves;
-  int total_moves;
-  fen_str >> board >> who_to_move >> castlings >> en_passant >>
-      rule50_halfmoves >> total_moves;
-
+  fen_str >> board;
+  string who_to_move = "w";
+  if (!fen_str.eof()) fen_str >> who_to_move;
+  // Assume no castling rights. Other engines, e.g., Stockfish, assume kings and
+  // rooks on their initial rows can each castle with the outer-most rook.  Our
+  // implementation currently supports 960 castling where white and black rooks
+  // have matching columns, so it's unclear which rights to assume.
+  string castlings = "-";
+  if (!fen_str.eof()) fen_str >> castlings;
+  string en_passant = "-";
+  if (!fen_str.eof()) fen_str >> en_passant;
+  int rule50_halfmoves = 0;
+  if (!fen_str.eof()) fen_str >> rule50_halfmoves;
+  int total_moves = 1;
+  if (!fen_str.eof()) fen_str >> total_moves;
   if (!fen_str) throw Exception("Bad fen string: " + fen);
 
   for (char c : board) {
--- a/src/chess/board.cc
+++ b/src/chess/board.cc
@@ -42,8 +42,6 @@
 
 namespace lczero {
 
-using std::string;
-
 const char* ChessBoard::kStartposFen =
     "rnbqkbnr/pppppppp/8/8/8/8/PPPPPPPP/RNBQKBNR w KQkq - 0 1";
 
@@ -981,17 +979,17 @@ void ChessBoard::SetFromFen(std::string fen, int* rule50_ply, int* moves) {
             fen.end());
 
   std::istringstream fen_str(fen);
-  string board;
+  std::string board;
   fen_str >> board;
-  string who_to_move = "w";
+  std::string who_to_move = "w";
   if (!fen_str.eof()) fen_str >> who_to_move;
   // Assume no castling rights. Other engines, e.g., Stockfish, assume kings and
   // rooks on their initial rows can each castle with the outer-most rook.  Our
   // implementation currently supports 960 castling where white and black rooks
   // have matching columns, so it's unclear which rights to assume.
-  string castlings = "-";
+  std::string castlings = "-";
   if (!fen_str.eof()) fen_str >> castlings;
-  string en_passant = "-";
+  std::string en_passant = "-";
   if (!fen_str.eof()) fen_str >> en_passant;
   int rule50_halfmoves = 0;
   if (!fen_str.eof()) fen_str >> rule50_halfmoves;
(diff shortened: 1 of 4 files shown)


src/chess/board.cc
```diff
@@ -982,8 +983,33 @@ void ChessBoard::SetFromFen(const std::string& fen, int* rule50_ply,
...
-      rule50_halfmoves >> total_moves;
+  fen_str >> board;
+  if (fen_str.eof()) {
+    who_to_move = "w";
+  } else {
+    fen_str >> who_to_move;
+  }
+  if (fen_str.eof()) {
+    castlings = "KQkq";
+    ignore_castling_errors = true;
```
Tilps, line 994: I'm not sure about this part - the castling validation logic isn't strong enough to correctly filter castling fen component down to actual plausible scenarios.

Mardak: Should we assume "-" castling instead? The current implementation does have some oddness with differently positioned rooks that lichess/sf happily allows both sides to castle 😉 https://lichess.org/analysis/r2k3r/8/8/8/8/8/8/1R2K1R1

Tilps: assuming '-' seems safer if we're going to do this.

Our implementation only supports chess 960 castling - not castling with independently shuffled back ranks.

Mardak: Defaulting to - instead of KQkq needing ignores.

src/chess/board.cc
```diff
@@ -982,8 +982,32 @@ void ChessBoard::SetFromFen(const std::string& fen, int* rule50_ply,
...
-  fen_str >> board >> who_to_move >> castlings >> en_passant >>
-      rule50_halfmoves >> total_moves;
+  fen_str >> board;
+  if (fen_str.eof()) {
+    who_to_move = "w";
+  } else {
+    fen_str >> who_to_move;
+  }
+  if (fen_str.eof()) {
+    castlings = "-";
```
borg323, line 992: If I understand correctly, this is not what other programs do - if you do another iteration, can you add a comment with any differences you know?

Mardak: Right, the example I linked in https://github.com/LeelaChessZero/lc0/pull/1234#discussion_r414326363 has SF allowing castling with differently positioned "initial" king and rooks whereas currently `ChessBoard::Castlings` assumes the same kingside and queenside rooks for both sides. Keeping with that 960 restriction, I'm not really sure how we would determine if white should get the castling rights or black or neither for https://lichess.org/analysis/r2k3r/8/8/8/8/8/8/1R2K1R1

borg323: I totally agree, just want to have it documented if it isn't too much trouble. I plan to make it a hard error (if it isn't already - need to ceck) to give a FEN with castling rights for both black and while if the rooks are not in the same column.

Mardak: Added comment and details about outer-most rook.

src/chess/board.cc
```diff
@@ -982,8 +982,36 @@ void ChessBoard::SetFromFen(const std::string& fen, int* rule50_ply,
   string en_passant;
   int rule50_halfmoves;
   int total_moves;
-  fen_str >> board >> who_to_move >> castlings >> en_passant >>
-      rule50_halfmoves >> total_moves;
+  fen_str >> board;
+  if (fen_str.eof()) {
```
mooskagh, line 986: Does it really work?.. eof bit is set when there is an attempt to read behind end of string, not when it's just reached.
So if fen = `"k/1R//K ";` (note trailing space, I thought it wouldn't work)

src/chess/board.cc
```diff
@@ -982,8 +982,36 @@ void ChessBoard::SetFromFen(const std::string& fen, int* rule50_ply,
...
+  } else {
+    fen_str >> rule50_halfmoves;
+  }
+  if (fen_str.eof()) {
+    total_moves = 1;
+  } else {
+    fen_str >> total_moves;
+  }
 
   if (!fen_str) throw Exception("Bad fen string: " + fen);
```
mooskagh, line 1016: Also I'm surprised that this works.. But as tests show that it does, it probably does. :)

I thought that when eof bit is set, the stream should return false.

Mardak: The unit test would fail when directly passing in a string with trailing space to `SetFromFen` but `ParseCommand` tokenizes and combines with a single space https://github.com/LeelaChessZero/lc0/blob/9b3c7992aef13a477b084fdaf3304fea5ed9e4b4/src/chess/uciloop.cc#L81-L87

src/chess/board.cc
Mardak: The unit test would fail when directly passing in a string with trailing space to `SetFromFen` but `ParseCommand` tokenizes and combines with a single space https://github.com/LeelaChessZero/lc0/blob/9b3c7992aef13a477b084fdaf3304fea5ed9e4b4/src/chess/uciloop.cc#L81-L87

src/chess/board.cc
Mardak: `operator bool` / `fail` are based on `failbit` and `badbit`: http://www.cplusplus.com/reference/ios/ios/fail/

src/chess/board.cc
Tilps: What happens with a >> std::ws added after each section read?

Tilps: ahh okay that shouldn't work would need to be if (fen_str.eof() || fen_str >> std::ws && fen_str.eof()) or something.

Mardak: Added a commit that removes trailing whitespace so plain `if (fen_str.eof())` will work and spaces ahead / between fields will be handled normally with `skipws`

src/chess/board.cc
```diff
@@ -969,21 +969,54 @@ MoveList ChessBoard::GenerateLegalMoves() const {
...
   string who_to_move;
   string castlings;
   string en_passant;
   int rule50_halfmoves;
   int total_moves;
-  fen_str >> board >> who_to_move >> castlings >> en_passant >>
-      rule50_halfmoves >> total_moves;
+  fen_str >> board;
+  if (fen_str.eof()) {
+    who_to_move = "w";
```
mooskagh, line 992: Maybe initialize those values to their defaults in lines 984-989, and here have:
```
if (!fen_str.eof()) fen_str >> castlings;
// and so on
```

Saves quite a lot of vertical space. :)

Mardak: I interlaced the variable declaration / initialization with the conditional extraction to keep the castling comment closer.

src/chess/board.cc
```diff
@@ -969,22 +969,34 @@ MoveList ChessBoard::GenerateLegalMoves() const {
   return result;
 }
 
-void ChessBoard::SetFromFen(const std::string& fen, int* rule50_ply,
-                            int* moves) {
+void ChessBoard::SetFromFen(string fen, int* rule50_ply, int* moves) {
```
Tilps, line 972: std::string even if you are making it pass by copy.

Mardak: Restored `std::` in the definition

PR #1238, opened 2020-04-25
cuda winograd fixes

ankan-ban:
- don't typecast directly to half datatype in CPU side code as older CUDA runtime doesn't support that.
- don't use gemmEx version on GPUs older than Maxwell generation (not supported).
- modify the check to enable custom_winograd setting. It should be faster in most cases - except on RTX GPUs when using fp16.

### DIFF
--- a/src/neural/cuda/layers.cc
+++ b/src/neural/cuda/layers.cc
@@ -813,10 +813,11 @@ void FusedWinogradConvSELayer<DataType>::LoadSEWeights(float* w1, float* b1,
 }
 
 template <>
-void FusedWinogradConvSELayer<half>::cublasRowMjaorMatrixMul(
+void FusedWinogradConvSELayer<half>::cublasRowMajorMatrixMul(
     const half* A, const half* B, half* Out, int M, int N, int K, int batchSize,
     cublasHandle_t cublas) {
-
+  // Need to initialize 1.0 and 0.0 as hexadecimal for fp16 because typecasting
+  // float to half type doesn't work before CUDA 10.0
   __half_raw one_h{0x3C00};
   __half_raw zero_h{0};
   half halfOne = one_h;
@@ -835,7 +836,7 @@ void FusedWinogradConvSELayer<half>::cublasRowMjaorMatrixMul(
 }
 
 template <>
-void FusedWinogradConvSELayer<float>::cublasRowMjaorMatrixMul(
+void FusedWinogradConvSELayer<float>::cublasRowMajorMatrixMul(
     const float* A, const float* B, float* Out, int M, int N, int K,
     int batchSize, cublasHandle_t cublas) {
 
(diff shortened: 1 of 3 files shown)


src/neural/cuda/network_cudnn.cc
```diff
@@ -272,12 +276,36 @@ class CudnnNetwork : public Network {
...
+    //     cores too, but only for networks with 256 or higher no. of filters.
+    // 
+    //  3. Currently a bug in cublas makes it slower on RTX GPUs with fp16 so
+    //  it's disabled. TODO: Enable it once the bug has been fixed and it's
+    //  tested to be faster.
+
+    if (fp16) {
+      if (!hasTensorCores)
+        use_custom_winograd_ = true;
+      else if (kNumFilters >= 256 && deviceProp.minor != 5) // TODO: remove the second condition.
```
Tilps, line 300: Is it safe to check minor without a major check to go with it?

borg323: Also can we add a check here that cublas version <= 10202 (the currently released one)? I'm assuming that the next release will have the fix and that it won't be ported to cuda 10.1 etc..

ankan-ban: Thanks. modified the check.

src/neural/cuda/layers.cc
```diff
@@ -811,12 +812,15 @@ void FusedWinogradConvSELayer<DataType>::LoadSEWeights(float* w1, float* b1,
   copyTypeConverted((DataType*)b2_, (float*)scratch, num_biases2);
 }
 
+template <>
+void FusedWinogradConvSELayer<half>::cublasRowMjaorMatrixMul(
```
Ttl, line 816: typo `Mjaor`

src/neural/cuda/layers.cc
```diff
@@ -811,12 +812,15 @@ void FusedWinogradConvSELayer<DataType>::LoadSEWeights(float* w1, float* b1,
...
+    const half* A, const half* B, half* Out, int M, int N, int K, int batchSize,
+    cublasHandle_t cublas) {
 
-void cublasRowMjaorMatrixMul(const half* A, const half* B, half* Out, int M,
-                             int N, int K, int batchSize, cublasHandle_t cublas,
-                             int algo = -1) {
-  half halfOne = (half)1.0f;
-  half halfZero = (half)0.0f;
+  __half_raw one_h{0x3C00};
```
Ttl, line 820: Could add a short comment here on why hexadecimals are used.

src/neural/cuda/network_cudnn.cc
```diff
@@ -272,12 +275,37 @@ class CudnnNetwork : public Network {
     // support it).
     ReportCUBLASErrors(cublasSetMathMode(cublas_, CUBLAS_TENSOR_OP_MATH));
 
-    // Check if we want to enable our custom winograd ?
     constexpr bool fp16 = std::is_same<half, DataType>::value;
+    const int kNumInputPlanes = kInputPlanes;
+    const int kNumFilters = weights.input.biases.size();
+    numBlocks_ = weights.residual.size();
+
+    // Use our custom winograd for resudual tower convolutions for most cases:
```
Ttl, line 283: typo `resudual`

src/neural/cuda/layers.cc
ankan-ban: Thanks. Fixed.

src/neural/cuda/network_cudnn.cc
ankan-ban: fixed. Really need to be more careful :-)

PR #1239, opened 2020-04-25
Flat eval at q=0

sqrmax:
Based on discussion at the TCEC chat, it might be slightly better that the eval curve is flat at q=0 to prevent evals such as 0.09 resulting in drawn out draws.  This slightly different approach:

* uses sqrt(x^2 / (x^2 + 1)) to flatten the eval curve at q = 0.
* rescales the multiplier to the maximum intended cp score, 128 (rather than 90).
* sets the tan() argument range by calculating it with atan(), rather than relying on hard coded magic numbers that happen to be close to pi/2 and could also be mistyped --- presumably any optimizing compiler will evaluate atan() at compile time.

Note well: I merely changed the code as a proof of concept, this is provided as a suggestion for consideration.

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -145,7 +145,7 @@ void Search::SendUciInfo() REQUIRES(nodes_mutex_) {
           std::round(edge.GetM(0.0f)) / 2 + (edge.IsTbTerminal() ? 101 : 1),
           wl);
     } else if (score_type == "centipawn_with_drawscore") {
-      uci_info.score = 90 * tan(1.5637541897 * q);
+      uci_info.score = 128 * tan(atan(128 * sqrt(2.0)) * q) * sqrt((q * q) / (q * q + 1));
     } else if (score_type == "centipawn") {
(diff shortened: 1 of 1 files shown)


sqrmax:
Thanks for pointing me to the data leading to the current eval code, let's see how it does first.

Naphthalin:
As the discussion is most likely based on the formula used in v0.24 and we recently accepted a new Q->cp formula in #1193 , this change should be compared to the arguments in #1193.

1.) "taking too long to give <0.08 eval because of TCEC adjudication rule" really isn't anything we should base our decision about the Q->cp conversion formula on [and in fact the new formula already gives lower cp values for low Q values]
2.) if you had read #1193, you would have seen [even though there should be a comment in the code for that] that the "magic number" already is the result of an `atan` calculation
3.) one big argument for that change was the invertibility of the formula. your suggestion doesn't possess this, so you would have to propose much stronger evidence that it is well.

So my summary is: we already "fixed" this, and there are other more important things to do for Lc0.

PR #1241, opened 2020-04-26
Syzygy tablebase for use with selfplay

MelleKoning:
as per #905 now on the c17 version of the code.

Add syzygy tablebases to use these for selfplay games.

The goal is to save time in training and game matches: for example when contributing to leela-training and matches I've seen games prolong between KRvsKR for 200+ moves, just wasting CPU/GPU cycles.
So the goal here is to use tablebases when those exists for clients; it should just use the tablebases available on your system even if that's only 4 or 5 pieces instead of 6 or 7 piece TB.

Seems to work when starting lc0 as follows:
lc0 selfplay --visits=800 --syzygy-paths=c:\path_to\tablebases

After this is merged, this also needs the go-client to pass on the syzygy-path from commandline to lc0 so that local installed client tablebases are indeed being used, or the TB might be configured in lc0.config startup params instead, not sure.

### DIFF
--- a/src/selfplay/game.cc
+++ b/src/selfplay/game.cc
@@ -57,6 +57,11 @@ const OptionId kMinimumAllowedVistsId{
 const OptionId kUciChess960{
     "chess960", "UCI_Chess960",
     "Castling moves are encoded as \"king takes rook\"."};
+const OptionId kSyzygyTablebaseId{
+    "syzygy-paths", "SyzygyPath",
+    "List of Syzygy tablebase directories, list entries separated by system "
+    "separator (\";\" for Windows, \":\" for Linux).",
+    's'};
 }  // namespace
 
 void SelfPlayGame::PopulateUciParams(OptionsParser* options) {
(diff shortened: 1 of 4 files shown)


Tilps:
Despite the description - this PR doesn't lead to a path to saving much processing in training scenarios.
To save processing what is needed is adjudication, not passing syzygy tablebase to search. Much like I have in my policy_tournament/value_tournament branches.
That being said I'm sure there are people who use selfplay for testing purposes (as its more efficient than running lc0 vs lc0 in cutechess), who would value this feature - in addition to the adjudication feature.  They should have separate flags though, it is reasonable to want to test engine without tablebase in search while still adjudicating, or vice versa.

MelleKoning:
Although I see the point of adjudication, this (small) change still makes it possible to train where TBs are in sight, for those clients that have the TBs available, right?
With the given parameter `--syzygy-paths` it can either be used or not. 

I guess the adjudication can be added in a separate PR. Are you referring to the code around these lines <https://github.com/Tilps/lc0/blob/value_tournament/src/selfplay/game.cc#L110> ?

Tilps:
Yes that is the code for adjudication.

cn4750:
I've been using this code for my own training run for some time now without issue. I think this is safe to merge once the conflicts are addressed. Any adjudication features can be added in a future PR.

MelleKoning:
Have rebased the changes on master to solve the merge conflict.

PR #1250, opened 2020-04-29
memory optimization for cudnn custom_winograd

ankan-ban:
- don't save untransformed weights
- print warning message when low memory is detected.

### DIFF
--- a/src/neural/cuda/layers.cc
+++ b/src/neural/cuda/layers.cc
@@ -720,7 +720,6 @@ FusedWinogradConvSELayer<DataType>::FusedWinogradConvSELayer(
       use_gemm_ex_(use_gemm_ex) {
   // Allocate memory for weights (filter tensor) and biases.
   const size_t weight_size = sizeof(DataType) * c_input_ * C * 3 * 3;
-  ReportCUDAErrors(cudaMalloc(&weights_, weight_size));
 
   if (use_bias_) {
     const size_t blas_size = sizeof(DataType) * C;
@@ -755,12 +754,15 @@ void FusedWinogradConvSELayer<DataType>::LoadWeights(float* pfilter,
   const size_t weight_size = sizeof(float) * c_input_ * C * 3 * 3;
   const size_t blas_size = sizeof(float) * C;
 
+  // Store untransformed weights in scratch.
+  const DataType* weights = (DataType*)scratch + weight_size + blas_size;
+
   // first copy from CPU memory to scratch space in GPU memory
   // and then do the type conversion using a kernel
   assert(scratch);
   ReportCUDAErrors(
       cudaMemcpy(scratch, pfilter, weight_size, cudaMemcpyHostToDevice));
-  copyTypeConverted((DataType*)weights_, (float*)scratch, C * c_input_ * 3 * 3);
+  copyTypeConverted((DataType*)weights, (float*)scratch, C * c_input_ * 3 * 3);
 
   if (pBias) {
     ReportCUDAErrors(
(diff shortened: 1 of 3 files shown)


src/neural/cuda/layers.cc
```diff
@@ -755,12 +754,15 @@ void FusedWinogradConvSELayer<DataType>::LoadWeights(float* pfilter,
   const size_t weight_size = sizeof(float) * c_input_ * C * 3 * 3;
   const size_t blas_size = sizeof(float) * C;
 
+  // Store untransformed weights in scratch.
+  const DataType* weights = (DataType*)scratch + weight_size + blas_size;
```
Tilps, line 758: I'm guessing its unlikely, but shouldn't there be a logic somewhere that ensures that scratch is at least 2x weights size or something in order to do this safely?

src/neural/cuda/network_cudnn.cc
```diff
@@ -310,6 +310,26 @@ class CudnnNetwork : public Network {
     if (!options.IsDefault<bool>("custom_winograd"))
       use_custom_winograd_ = options.Get<bool>("custom_winograd");
 
+    // Warn if the memory required for storing transformed weights is
+    // going to exceed 60% of total video memory, force custom_winograd off
+    // if it's going to exceed 80% of memory.
+    size_t residual_weight_size =
+        3 * 3 * kNumFilters * kNumFilters * numBlocks_ * sizeof(DataType);
+    size_t transformed_residual_weight_size = residual_weight_size * 4;
```
Tilps, line 318: Shouldn't this logic be conditional on use_custom_wingrad not already being false?

borg323: I would split this in two, moving the logic for `use_custom_winograd_ = false;` before line 310 (`if (!options.IsDefault<bool>("custom_winograd"))`) and keep only the warning here (conditional as Tilps says).

src/neural/cuda/layers.cc
ankan-ban: definitely - was just me being lazy. Updated.

src/neural/cuda/network_cudnn.cc
ankan-ban: Yes. I wanted to have some warning for people with very less memory even with custom_winograd truend off but we probably don't want people to get confused by useless messages. Updated the condition so that we print only useful messages.

src/neural/cuda/network_cudnn.cc
```diff
@@ -310,6 +310,32 @@ class CudnnNetwork : public Network {
...
+    size_t residual_weight_size =
+        residual_single_layer_weight_size * numBlocks_ * 2;
+    size_t transformed_residual_weight_size = residual_weight_size * 4;
+    if (residual_weight_size > 0.6 * deviceProp.totalGlobalMem) {
+      CERR << "Low video memory detected. You may run into OOM errors. Please "
+              "consider using a smaller network.";
+      // No hope of using custom winograd - even the fallback path might not run.
+      use_custom_winograd_ = false;
+    } else if (use_custom_winograd_) {
+      if (transformed_residual_weight_size > 0.8 * deviceProp.totalGlobalMem) {
```
borg323, line 327: I was thinking about an additional check here to only disable the custom code if it wasn't explicitly requested by the user, allowing to overriding this (with the warning in the else branch) but this is probably overkill.

Tilps: I'm going assume that we can do such things in a follow up if they end up looking useful and just merge this for now.

borg323: As I said, overkill. Will only consider it if we get complaints about it.

PR #1260, opened 2020-05-01
fix weight memory check to be more conservative

ankan-ban:
- 60% memory usage for transformed weights is too optimistic. Try 40% for warning, and 50% for auto-disable of custom_winograd.
- Address @borg323 's comment for my previous pull request (allow user to override with warning).
- also fix (size_t)->(int) conversion warnings in cudnn backend

### DIFF
--- a/src/neural/cuda/network_cudnn.cc
+++ b/src/neural/cuda/network_cudnn.cc
@@ -276,8 +276,8 @@ class CudnnNetwork : public Network {
 
     constexpr bool fp16 = std::is_same<half, DataType>::value;
     const int kNumInputPlanes = kInputPlanes;
-    const int kNumFilters = weights.input.biases.size();
-    numBlocks_ = weights.residual.size();
+    const int kNumFilters = (int)weights.input.biases.size();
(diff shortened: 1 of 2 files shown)


src/neural/cuda/network_cudnn.cc
```diff
@@ -306,34 +306,36 @@ class CudnnNetwork : public Network {
...
         residual_single_layer_weight_size * numBlocks_ * 2;
     size_t transformed_residual_weight_size = residual_weight_size * 4;
+
     if (residual_weight_size > 0.6 * deviceProp.totalGlobalMem) {
       CERR << "Low video memory detected. You may run into OOM errors. Please "
               "consider using a smaller network.";
-      // No hope of using custom winograd - even the fallback path might not run.
+    }
+    if (use_custom_winograd_ &&
+        transformed_residual_weight_size > 0.5 * deviceProp.totalGlobalMem) {
```
borg323, line 323: Need to add `&& options.IsDefault<bool>("custom_winograd")` here to avoid this specific warning if an override is given.

PR #1266, opened 2020-05-03
Cuda Winograd kernel optimization

Ttl:
One Winograd convolution optimization ported from Leela Zero Go OpenCL backend. About 6% faster with 256x20 network and 4% faster with 128x10 network in benchmark using GTX 1080 Ti GPU.

### DIFF
--- a/src/neural/cuda/common_kernels.cu
+++ b/src/neural/cuda/common_kernels.cu
@@ -30,7 +30,6 @@
 #include "cuda_common.h"
 #include "winograd_helper.inc"
 
-
 namespace lczero {
 namespace cudnn_backend {
 namespace {
--- a/src/neural/cuda/common_kernels.cu
+++ b/src/neural/cuda/common_kernels.cu
@@ -30,6 +30,7 @@
 #include "cuda_common.h"
 #include "winograd_helper.inc"
 
+
 namespace lczero {
 namespace cudnn_backend {
 namespace {
--- a/src/neural/cuda/network_cudnn.cc
+++ b/src/neural/cuda/network_cudnn.cc
@@ -40,7 +40,7 @@
 #include "utils/bititer.h"
 #include "utils/exception.h"
 
-#define DEBUG_RAW_NPS
+//#define DEBUG_RAW_NPS
 
 namespace lczero {
 using namespace cudnn_backend;
@@ -387,7 +387,7 @@ class CudnnNetwork : public Network {
         cudnn_, xDesc, wDesc, convDesc, xDesc, conv_algo, &scratch_size_));
 
     // Have some minumum as we also use this for transforming weights.
-    size_t max_weight_size = 128 * 1024 * 1024;
+    int max_weight_size = 128 * 1024 * 1024;
 
     // parts from scratch allocation are suballocated to hold various weights
     // and biases when transforming winograd weights (one layer at a time), 128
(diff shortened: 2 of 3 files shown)


Ttl:
This compiles for me locally but doesn't work on CI. @ankan-ban how should half precision literals be handled? Do we even want to do non-tensor core computation in half precision? I see some existing kernels cast to single precision for computation. I can't test with GPU that supports half precision at the moment.

ankan-ban:
@Ttl I think what you have done in the next commit should be fine (provided that it compiles).

>>Do we even want to do non-tensor core computation in half precision?
It can provide upto 2x boost in performance (in theory) on GPUs that support half precision math (like P100, V100, GTX16xx, RTX series and later GPUs). Note that even on GPUs with tensor cores, unless you use WMMA intrinsics, you won't get tensor core acceleration and max theoretical speedup vs FP32 is only 2x. Practically the speedup could be less - or even zero (if the kernel is totally memory bound). 

On GPUs that don't have half precision support (like consumer Pascal GPUs or older), it makes more sense to typecast everything to float before doing computation in the kernel so that you get benefit of bandwidth saving loading data in fp16, and also don't run into issues with hardware/compiler trying to emulate fp16 math using fp32. 

For the winograd transform kernels, my suggestion would be to try doing all the computations in float and measure if there is any perf loss compared to doing them in half even on GPUs that support FP16 math. If there isn't you can probably have just a single path.

Ttl:
According to the table in https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#arithmetic-instructions, fp16 does have 2x throughput on RTX GPUs. On my GTX 1080 Ti however half precision performance is significantly worse than single precision performance. It's probably okay as it is now, since GTX 1080 Ti doesn't use the half precision kernels but good to keep in mind for potential half precision storage, single precision compute kernels.

The code seems to compile now on CI environments after the latest commit.

ankan-ban:
Unfortunately it looks like this optimization makes the fp16 path slower! 

On RTX 2060 (baseline is r25.1 build, new is appveyor build for this PR), with  lc0 backendbench --backend-opts=custom_winograd=true -w 256x20-t40-1541.pb.gz --batches=10

```
FP32:
Baseline: Benchmark batch size 256 with inference average time 56.3886ms - throughput 4539.93 nps.
New:      Benchmark batch size 256 with inference average time 55.7229ms - throughput 4594.17 nps.

FP16:
Baseline: Benchmark batch size 256 with inference average time 32.6387ms - throughput 7843.46 nps.
New:      Benchmark batch size 256 with inference average time 34.8069ms - throughput 7354.87 nps.

```
So a tiny bit faster for fp32, but quite a bit slower for fp16.

One possible explanation could be that it was easier/possible for the compiler to convert the dumb matrix multiply into HFMA2 instructions (vector half precision fused multiply-add that needs two parallel multiply-add operations to get the 2X throughput with fp16), and with the new path although the number of math operations are less, it's no longer possible for compiler to find paris of multiply-add that can be vectorized :-(

I will try to test it more and confirm this indeed if the problema and whether we can do anything to fix it.

ankan-ban:
it's also possible that appveyor build is slower than v25.1 release build for some reason (given that I don't see the speedup on FP32 either), so I will try to build it locally and test more.

Ttl:
This is how the backendbench results look like for me on GTX 1080 Ti with 256x20 network:

bt256

Issue #1071, opened 2020-02-11
It's not possible to specify absolute path to config


mooskagh:
Because of
https://github.com/LeelaChessZero/lc0/blob/3ac7e7e2e08f29a4b2864701d3e9e9e949d5b671/src/utils/configfile.cc#L90

Naphthalin:
Was there any attempted fix to this so far, given it is marked as a release blocker?

mooskagh:
It's not a release blocker as everyone forgot. It would be nice to have it fixed eventually.

PR #1276, opened 2020-05-07
Add check for absolute path of config file

KipHamiltons:
#1071 
Added absolute path capability for specifying config file.
This is my first time addressing an issue with a GitHub commit - if I have screwed up anything, please let me know! :)

### DIFF
--- a/src/utils/configfile.cc
+++ b/src/utils/configfile.cc
@@ -94,7 +94,7 @@ bool ConfigFile::Init(OptionsParser* options) {
   return true;
 }
 
-bool ConfigFile::ParseFile(const std::string& filename,
+bool ConfigFile::ParseFile(std::string& filename,
                            OptionsParser* options) {
 
   // Check to see if we are using the default config file or not.
--- a/src/utils/configfile.cc
+++ b/src/utils/configfile.cc
@@ -42,7 +42,7 @@ const OptionId kConfigFileId{
     "Path to a configuration file. The format of the file is one command line "
     "parameter per line, e.g.:\n--weights=/path/to/weights",
     'c'};
-const char* kDefaultConfigFile = "lc0.config";
+const char* kDefaultConfigFile = "<default>";
 }  // namespace
 
 std::vector<std::string> ConfigFile::arguments_;
@@ -88,6 +88,15 @@ bool ConfigFile::Init(OptionsParser* options) {
   // loading the default configuration file.
   if (filename == "") return true;
 
+  // Parses the file into the arguments_ vector.
+  if (!ParseFile(filename, options)) return false;
+
+  return true;
+}
+
+bool ConfigFile::ParseFile(const std::string& filename,
+                           OptionsParser* options) {
+
   // Check to see if we are using the default config file or not.
   const bool using_default_config = 
       filename == std::string(kDefaultConfigFile);
@@ -98,20 +107,12 @@ bool ConfigFile::Init(OptionsParser* options) {
     filename = CommandLine::BinaryDirectory() + "/" + filename;
   }
 
-  // Parses the file into the arguments_ vector.
-  if (!ParseFile(filename, options)) return false;
-
-  return true;
-}
-
-bool ConfigFile::ParseFile(const std::string& filename,
-                           OptionsParser* options) {
   std::ifstream input(filename);
 
   if (!input.is_open()) {
     // It is okay if we cannot open the default file since it is normal
-    // for it to not exist. Check it is default by comparing to <default>.
-    if (filename == std::string(kDefaultConfigFile)) return true;
+    // for it to not exist.
+    if (using_default_config) return true;
 
     CERR << "Could not open configuration file: " << filename;
     return false;
--- a/src/utils/configfile.cc
+++ b/src/utils/configfile.cc
@@ -42,7 +42,8 @@ const OptionId kConfigFileId{
     "Path to a configuration file. The format of the file is one command line "
     "parameter per line, e.g.:\n--weights=/path/to/weights",
     'c'};
-const char* kDefaultConfigFile = "<default>";
+const char* kDefaultConfigFile = "lc0.config";
+const char* kDefaultConfigFileParam = "<default>";
 }  // namespace
 
 std::vector<std::string> ConfigFile::arguments_;
@@ -55,7 +56,7 @@ void ConfigFile::PopulateOptions(OptionsParser* options) {
 // ProcessAllFlags() that should be called only once, and needs the config file.
 std::string ConfigFile::ProcessConfigFlag(
     const std::vector<std::string>& args) {
-  std::string filename = kDefaultConfigFile;
+  std::string filename = kDefaultConfigFileParam;
   for (auto iter = args.begin(), end = args.end(); iter != end; ++iter) {
     std::string param = *iter;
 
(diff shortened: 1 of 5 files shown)


src/utils/configfile.cc
```diff
@@ -88,7 +88,9 @@ bool ConfigFile::Init(OptionsParser* options) {
   // loading the default configuration file.
   if (filename == "") return true;
 
-  filename = CommandLine::BinaryDirectory() + "/" + filename;
+  if (filename.at(0) != '/') {
```
ddobbelaere, line 91: This fix will work on Unix systems, but seems to leave Windows (who likes to prefix paths with `C:\`, `D:`, etc) users in the cold.

KipHamiltons: I have added a preprocessor directive to check for windows, then check for a colon in that path. I don't have a way to whether the windows section is robust though.

ddobbelaere:
Now that we support C++17, I think https://stackoverflow.com/a/43278300 is even more robust (probably there exist corner cases where the second character on Windows is not a colon, are you sure the current code does not throw on exception for single character filenames now?)

ddobbelaere:
CI on Windows systems passes. Only Android on Appveyor and Linux on Circle CI fail, probably because of this (see https://en.cppreference.com/w/cpp/filesystem):

> Notes
> 
> Using this library may require additional compiler/linker options. GNU implementation prior to 9.1 requires linking with -lstdc++fs and LLVM implementation prior to LLVM 9.0 requires linking with -lc++fs 

Probably @borg323 and/or @lealgo have ideas to fix this?

borg323:
Just a quick hint for now, this should probably be handled in `src/utils/filesystem.posix.cc` and `src/utils/filesystem.win32.cc`.

KipHamiltons:
I initially pushed that commit because I had no idea why my compiler couldn't find `<filesystem>`, then I realised that I was using the wrong compiler just after I pushed.
Now that I can actually test my code, I am getting a segfault whenever I call the `std::filesystem::path` constructor.
for example, just having the line `  std::filesystem::path test = "";` in there, without using the variable creates a segfault when the `ParseFile` is returning true (I didn't add a --config option).
>Just a quick hint for now, this should probably be handled in src/utils/filesystem.posix.cc and src/utils/filesystem.win32.cc.
I'll port it over now.

ddobbelaere:
Seems `std::filesystem` causes some headaches and/or is poorly supported. I wasn't aware of the OS specific files @borg323 pointed to. This seems indeed the cleanest and most consistent approach to do thing there (e.g. make an "IsRelativePath" (or whatever) function with OS specific implementation, for Windows you can make use of its specific API calls).

mooskagh:
Uh I was going to suggest https://en.cppreference.com/w/cpp/filesystem/path/is_absrel but see it's already being discussed. :)

mooskagh:
Note that for Windows, both `/` and `\` is allowed, and it's also allowed to omit `D:` part.
So technically what's absolute path in Linux is a subset of what's absolute path in Windows.

Maybe regex-based approach would work? :) Something like `^(?:\w:)?[\\/]`. Although at least as of 3 years ago, support of regexes in C++ were roughly as problematic as filesystem part of STL seems to be now.

And technically speaking `C:\windows/system32` is actually a valid relative path in Linux.. :)

mooskagh:
How about making it similar to weights file autodiscover? Basically if the value is `<default>`, search for Lc0.config in the directory of binary. Otherwise just use the path (relative to cwd if it's relative).

KipHamiltons:
I have used the check for first character `/` in the posix filesystem, and let the windows API decide about the windows path, as per https://stackoverflow.com/a/5377662/11108223 - although I am not sure if I need to `#include <shlwapi.h>` on the windows file.

>How about making it similar to weights file autodiscover? Basically if the value is <default>, search for Lc0.config in the directory of binary. Otherwise just use the path (relative to cwd if it's relative).

I have retained this functionality I think, by not messing with how it works currently for the most part.

I really appreciate all of your help everyone (not that this is necessarily done). Please keep the feedback coming - I am really enjoying my first open source experience :)

lealgo:
IMO the main issue here is that the current functionality (in the release) is confusing: the config file can only reside in a path that is relative to the directory where the binary is located. This is a very weird convention. Normally paths are either absolute or relative to the current working dir. So I like @mooskagh's idea.

KipHamiltons:
Ahh I think I understand how that is different to what I have done/what happens at the moment. Do you want me to go full steam ahead on that @lealgo ?

mooskagh:
Yeah (unless someone objects) I think having it only search in the directory of the binary if it's "`<default>`" is good enough to be implemented. :)

That way we'll also be able to search multiple locations by default in future, which is useful for linux (i.e. `lc0.config` in the directory of the binary, then '~/.config/lc0/lc0.config`, then `/etc/lc0.config`, then `./lc0.config`).

ddobbelaere:
The current solution comes forward to what @mooskagh suggested and also allows users to specify an absolute path (with relative path w.r.t. the working directory, which is more logical), LGTM.

*Nit:* there are still some stale unneeded `#include` statements lingering.

KipHamiltons:
> _Nit:_ there are still some stale unneeded `#include` statements lingering.

Well spotted! cheers!

src/utils/configfile.cc
```diff
@@ -81,14 +81,22 @@ std::string ConfigFile::ProcessConfigFlag(
...
 
-  // If filename is an empty string then return true.  This is to override
+  // If filename is an empty string then return true. This is to override
   // loading the default configuration file.
   if (filename == "") return true;
 
-  filename = CommandLine::BinaryDirectory() + "/" + filename;
+  // Check to see if we are using the default config file or not.
+  const bool using_default_config = 
+      filename == std::string(kDefaultConfigFile);
```
mooskagh, line 93: I think it's better to have a separate value for default behaviour, like "`<default>`" or "`<auto>`" (which still will read `$LC0_DIRECTORY/lc0.config`), similarly to what we have for weights file (there we have "`<autodiscover>`").

Otherwise when run like `./lc0 -c lc0.cfg` it will search in the current directory, while for `./lc0 -c lc0.config` it will search in the directory of the binary, which is a confusing inconsistency.

Also in future we'll try several locations by default (possibly with different file names, not only directories, e.g. `/etc/lc0.cfg` vs `~/.lc0rc`), and it's more logical if the parameter doesn't have file name in it.

KipHamiltons: I have made changes along these lines. Let me know if this is what you were thinking :slightly_smiling_face:

src/utils/configfile.cc
```diff
@@ -42,7 +42,7 @@ const OptionId kConfigFileId{
     "Path to a configuration file. The format of the file is one command line "
     "parameter per line, e.g.:\n--weights=/path/to/weights",
     'c'};
-const char* kDefaultConfigFile = "lc0.config";
+const char* kDefaultConfigFile = "<default>";
```
mooskagh, line 45: Sorry for nitpicking, but I suggest moving both constants here:
```cpp
const char* kDefaultConfigFile = "lc0.config";
const char* kDefaultConfigFileParam = "<default>";
```
(or if it's one, better it be the filename and not the option value)

KipHamiltons: Not a problem. Do these changes match what you were thinking?

mooskagh: Looks good!

PR #1277, opened 2020-05-08
Set bounds for lonely kings.

Mardak:
r? @Tilps Here's setting bounds as suggested in staying strict. (Vs assuming king + bishop/knight also can't mate, which could happen in special cases.)

lonely king
```
position fen 2KQ/k
  
go nodes 1
node N:       1 (P:  0.00%) (WL: -0.00000) (D: 0.000) (M:  0.0) (W)  

go nodes 2 searchmoves d8b6
d8b6 N:       1 (P:  5.00%) (WL:  0.00000) (D: 0.000) (M:  0.0) (W)  
node N:       2 (P:  5.00%) (WL: -0.00000) (D: 0.000) (M:  0.5) (W)  
--- Opponent moves after: d8b6
a7a8 N:       0 (P: 50.00%) (WL:  -.-----) (D: -.---) (M:  -.-)  
a7b6 N:       0 (P: 50.00%) (WL:  -.-----) (D: -.---) (M:  -.-)  
node N:       1 (P:  0.00%) (WL: -0.00000) (D: 0.000) (M:  0.0) (L)  

go nodes 3 searchmoves d8b6
d8b6 N:       2 (P:  5.00%) (WL:  0.00000) (D: 0.000) (M:  0.5) (W)  
node N:       3 (P:  5.00%) (WL: -0.00000) (D: 0.000) (M:  1.0) (W)  
--- Opponent moves after: d8b6
a7b6 N:       0 (P: 50.00%) (WL:  -.-----) (D: -.---) (M:  -.-)  
a7a8 N:       1 (P: 50.00%) (WL:  0.00000) (D: 0.000) (M:  0.0) (L)  
node N:       2 (P: 50.00%) (WL: -0.00000) (D: 0.000) (M:  0.5) (L)  

go nodes 4 searchmoves d8b6
d8b6 N:       3 (P:  5.00%) (WL:  0.00000) (D: 1.000) (M:  1.0) (T)  
node N:       4 (P:  5.00%) (WL: -0.00000) (D: 0.250) (M:  1.2) (W)  
--- Opponent moves after: d8b6
a7a8 N:       1 (P: 50.00%) (WL:  0.00000) (D: 0.000) (M:  0.0) (L)  
a7b6 N:       1 (P: 50.00%) (WL:  0.00000) (D: 1.000) (M:  0.0) (T)  
node N:       3 (P:100.00%) (WL: -0.00000) (D: 1.000) (M:  1.0) (T)   
```

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -1234,6 +1234,13 @@ void SearchWorker::ExtendNode(Node* node) {
     }
   }
 
+  // A lonely king cannot mate, so indicate that with restricted bounds.
+  if (board.ours().count() < 2) {
+    node->SetBounds(GameResult::DRAW, GameResult::WHITE_WON);
+  } else if (board.theirs().count() < 2) {
+    node->SetBounds(GameResult::BLACK_WON, GameResult::DRAW);
+  }
+
   // Add legal moves as edges of this node.
   node->CreateEdges(legal_moves);
 }
@@ -1501,9 +1508,11 @@ void SearchWorker::DoBackupUpdateSingleNode(
     return;
   }
 
-  // For the first visit to a terminal, maybe update parent bounds too.
+  // For the first visit to a bounded node, maybe update parent bounds too.
+  const auto [lower, upper] = node->GetBounds();
   auto update_parent_bounds =
-      params_.GetStickyEndgames() && node->IsTerminal() && !node->GetN();
+      params_.GetStickyEndgames() && !node->GetN() &&
+      (lower > GameResult::BLACK_WON || upper < GameResult::WHITE_WON);
 
   // Backup V value up to a root. After 1 visit, V = Q.
   float v = node_to_process.v;


src/mcts/search.cc
```diff
@@ -1234,6 +1234,13 @@ void SearchWorker::ExtendNode(Node* node) {
     }
   }
 
+  // A lonely king cannot mate, so indicate that with restricted bounds.
+  if (board.ours().count() < 2) {
+    node->SetBounds(GameResult::DRAW, GameResult::WHITE_WON);
+  } else if (board.theirs().count() < 2) {
```
Mardak, line 1240: For those curious, clang11 optimizes this to…
```asm
blsrq   -104(%r14), %rax
```
because the 64-bit board will have a single bit set for the lonely king, and the instruction resets the lowest bit resulting in 0 for branching.

src/mcts/search.cc
```diff
@@ -1501,9 +1508,11 @@ void SearchWorker::DoBackupUpdateSingleNode(
     return;
   }
 
-  // For the first visit to a terminal, maybe update parent bounds too.
+  // For the first visit to a bounded node, maybe update parent bounds too.
+  const auto [lower, upper] = node->GetBounds();
   auto update_parent_bounds =
-      params_.GetStickyEndgames() && node->IsTerminal() && !node->GetN();
+      params_.GetStickyEndgames() && !node->GetN() &&
+      (lower > GameResult::BLACK_WON || upper < GameResult::WHITE_WON);
```
Mardak, line 1515: and this optimizes to…
```asm
        movb    78(%rax), %cl
        andb    $56, %cl # 0b11 10 00 to get both "upper" bits and high bit of "lower"
        cmpb    $48, %cl # 0b11 00 00 checks "upper" is WHITE_WON and "lower" is *not* DRAW/WHITE_WON
```

Tilps:
I presume this is a pretty small nps impact - for probably an almost negligable gain in elo, but if you can put some numbers on that I think I'm willing to approve.

Mardak:
10 benchmark runs with master 9158a67a917d630caf7fbd22baeaebd31686db24:
```
Nodes/second    : 132125
Nodes/second    : 132194
Nodes/second    : 132252
Nodes/second    : 133328
Nodes/second    : 133421
Nodes/second    : 133459
Nodes/second    : 133552
Nodes/second    : 133592
Nodes/second    : 133821
Nodes/second    : 134150
```
and with this PR:
```
Nodes/second    : 131987
Nodes/second    : 132332
Nodes/second    : 132400
Nodes/second    : 132936
Nodes/second    : 133071
Nodes/second    : 133136
Nodes/second    : 133197
Nodes/second    : 133341
Nodes/second    : 133919
Nodes/second    : 134162
```

So looking at the median numbers, maybe 0.25% regression although the difference between high and low within a run ranges ~1.5%? But I suppose even if that's an acceptable amount, maybe we shouldn't add the code complexity as it probably only helps really early networks that incorrectly evaluate "no more pieces" to help find more draws.

src/mcts/search.cc
```diff
@@ -1234,6 +1234,13 @@ void SearchWorker::ExtendNode(Node* node) {
     }
   }
 
+  // A lonely king cannot mate, so indicate that with restricted bounds.
+  if (board.ours().count() < 2) {
+    node->SetBounds(GameResult::DRAW, GameResult::WHITE_WON);
```
MelleKoning, line 1239: How does `SetBounds` work? The best we have if only a king is left is between a loss and a draw, right?
Shouldn't this be something like (pseude code) `node->SetBounds(lower=Loss, upper=Draw)`

For some reason it looks as if SetBounds takes the values backward, but there must be a reason for this and most probably I did not yet understand how SetBounds works, how does it work?

Tilps: I don't know in this particular case for certain, but we do some things where values are stored from the parents perspective, especially around how we deal with GameResult it can be quite unintuitive at times.

Mardak: Yeah, if you look at other outcomes in `ExtendNode`:
https://github.com/LeelaChessZero/lc0/blob/9158a67a917d630caf7fbd22baeaebd31686db24/src/mcts/search.cc#L1174-L1179
https://github.com/LeelaChessZero/lc0/blob/9158a67a917d630caf7fbd22baeaebd31686db24/src/mcts/search.cc#L1221-L1227

MelleKoning: Yeah well... seen that before but it is pretty hard to read, I'm not entirely sure why code is like we are always talking to the parent of a leaf node (or so it seems). 

is this part responsible for that?

```
Node* cur = node;
  while (cur != search_->root_node_) {
    Node* prev = cur->GetParent();
    to_add.push_back(prev->GetEdgeToNode(cur)->GetMove());
    cur = prev;
  }
```

And to be honest I'm not sure how the code could be more readable either...

liamlim:
What about this position? 
8/8/8/7k/8/8/8/1BBK4 w - - 0 1

It's a position with two bishops against lonely knight. This is verbose move stats output after 1000 nodes (irrelevant columns filtered out)
info string c1g5  (67  ) N:       2 (+ 0) (WL:  0.35749) (D:  0.639) (M: 19.8) (L)
info string c1h6  (69  ) N:       2 (+ 0) (WL:  0.26993) (D:  0.727) (M: 18.5) (L)
info string b1g6  (43  ) N:       2 (+ 0) (WL:  0.39916) (D:  0.599) (M: 19.4) (L)
info string b1f5  (41  ) N:      63 (+11) (WL:  0.99863) (D:  0.001) (M: 28.7)
info string c1f4  (65  ) N:      64 (+19) (WL:  0.99714) (D:  0.003) (M: 31.6)
info string b1h7  (45  ) N:      70 (+14) (WL:  0.99836) (D:  0.002) (M: 30.7)
info string b1a2  (30  ) N:      71 (+20) (WL:  0.99957) (D:  0.000) (M: 31.7)
info string d1c2  (80  ) N:      73 (+15) (WL:  0.99955) (D:  0.000) (M: 32.1)
info string b1e4  (39  ) N:      73 (+14) (WL:  0.99868) (D:  0.001) (M: 29.8)
info string c1a3  (59  ) N:      74 (+22) (WL:  0.99953) (D:  0.000) (M: 31.0)
info string c1d2  (57  ) N:      75 (+19) (WL:  0.99857) (D:  0.001) (M: 29.9)
info string c1b2  (55  ) N:      76 (+21) (WL:  0.99959) (D:  0.000) (M: 30.3)
info string b1d3  (37  ) N:      77 (+18) (WL:  0.99968) (D:  0.000) (M: 29.4)
info string d1e1  (75  ) N:      77 (+17) (WL:  0.99857) (D:  0.001) (M: 29.5)
info string b1c2  (32  ) N:      78 (+15) (WL:  0.99862) (D:  0.001) (M: 29.9)
info string c1e3  (63  ) N:      79 (+20) (WL:  0.99946) (D:  0.001) (M: 30.0)
info string d1d2  (81  ) N:      80 (+15) (WL:  0.99965) (D:  0.000) (M: 29.9)
info string d1e2  (82  ) N:      85 (+16) (WL:  0.99887) (D:  0.001) (M: 27.6)

So we already know that bishop sascrifice is (L). We don't know that this position is (W) though and can't propagate this information up in the tree. Do I understand correctly that after this change, we would have bishop sacrifice moves as terminal draws? 

I think it would help network to understand BB+K and NN+K games much better than it does now, because the parent would be set to W and grandparent would be set to L almost without any search.

Mardak:
Sounds like this will only help in the unbalanced corner cases so not worth the complexity.

PR #1280, opened 2020-05-09
Make `wl_` double.

### LINKED ISSUE #875: Bug in Q calculation
I ran a long-ish search and got this output which clearly can't be right: https://pastebin.com/VY8H0NGh

If you calculate the Q that this d3a3 move should have then unless I made an error calculating, it should be around 0.063, rather than the 0.075 reported. No idea why it happens but that is quite a difference.

mooskagh:
And get rid of `EdgeList` (took 6 bytes extra space than it should have).

NPS drop: -1.6% at random backend.
Memory usage: unchanged! \o/ (was offset by EdgeList refactoring)

Fixes #875.

### DIFF
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -194,8 +194,8 @@ void Node::CreateEdges(const MoveList& moves) {
   num_edges_ = moves.size();
 }
 
-Node::ConstIterator Node::Edges() const { return {this, &child_}; }
-Node::Iterator Node::Edges() { return {this, &child_}; }
+Node::ConstIterator Node::Edges() const { return {*this, &child_}; }
+Node::Iterator Node::Edges() { return {*this, &child_}; }
 
 float Node::GetVisitedPolicy() const { return visited_policy_; }
 
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -167,14 +167,11 @@ std::string Edge::DebugString() const {
   return oss.str();
 }
 
-/////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


mooskagh:
Ah and I also changed num_edges from uint16 to uint8, that should be fine, right?

ddobbelaere:
Hmm, why not convert `d_` to double also, while we are at it? The added memory overhead will still be negligible.

I know search issues are now at least solved by this PR but people doing long analysis will also suppose they can at least trust the WDL values. This would also allow extension to search (e.g. different draw value) to be performed reliably...

ddobbelaere:
We also forgot `m_` btw :)

mooskagh:
We remember about both `m_` and `d_`. :)

`d_` is currently not used in search (unless drawscore is non-zero).
And `m_` can wait, it still has to demonstrate its usefullness first. :-P

I discussed that with Tilps, we decided to be conservative as usually we don't like growing `Node` structure.
Maybe worth discussing in Discord.

Naphthalin:
I see a 2-3% nps increase in benchmark with random backend:

```
Lc0_7886bfc with --backend=random:

Total time (ms) : 340239
Nodes searched  : 43740344
Nodes/second    : 128557

Lc0_v0.25-rc1 with --backend=random:

Total time (ms) : 340135
Nodes searched  : 42765331
Nodes/second    : 125730

Lc0_c4c92f6 with --backend=random:

Total time (ms) : 340111
Nodes searched  : 42331240
Nodes/second    : 124463
```

PR #1284, opened 2020-05-10
Allow smart pruning to terminate search if win is known.

Tilps:
Not going to gain any elo, but it seems consistent with the meaning of smart pruning being enabled.


Tilps:
Amphoria did a 1k games test with T60 at fixed nodes. 3 elo, well within error bounds. I presume its a 0.

PR #1285, opened 2020-05-10
Clear policy for terminal loss.

Tilps:
This probably is worthy of a fixed node elo test to rule out any strange interactions, but I would estimate positive elo in the 0-1 range ;)
This seems a cheaper alternative to adding more logic to the pick node loop with regard to terminals.

### DIFF
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -232,6 +232,9 @@ void Node::MakeTerminal(GameResult result, float plies_left, Terminal type) {
   } else if (result == GameResult::BLACK_WON) {
     wl_ = -1.0f;
     d_ = 0.0f;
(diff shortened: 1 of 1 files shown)


Tilps:
1k nodes, 2000 games of master vs this PR, chad 6 ply book. Final scoreline is within error bars, slightly positive.
Score of lc0 newbaseline vs lc0 dev local build: 368 - 386 - 1246 [0.495]
Elo difference: -3.13 +/- 9.34

2000 of 2000 games finished.

Tilps:
Given the positive test (but within errorbars) and the extremely low cost and that this should have some additional positive effects to do with quality of MLH target values near mate outside of TB rescoring potential.
Merging.

Issue #1283, opened 2020-05-10
Caching of best_edge serves double duty that are not consistent.


Tilps:
We cache best_edge for two reasons, one to know when the PV has changed to force a new info, and secondly to know the best child n without needing to loop the root node edges in order to correctly apply smart pruning at the root.
We update the cached best edge whenever a visit occurs which matches or exceeds the current N value for the currently cached best edge.
However we use GetRootChildNoTemperature to decide the best child - and GetRootChildNoTemperature no longer always uses N as the primary tie breaker.  This is good for the PV in mate scenarios, but may potentially mess with the logic of smart pruning, and also the logic of when to update best edge as the 'shortest mate' isn't necessarily guaranteed to have the highest N.

This has a possible relationship with 1272 as if the edge getting most of the visits, isn't the proven mate that has been cached into best_child, an even greater percentage of visits will run the GetRootChildNoTemperature path to try and update it which would have have triggered the root probe even more frequently than it could before. (Although now that 1272 has been fixed, its less of a performance issue.)

Tilps:
1272 GetRootChildNoTemperature being called so much was for a different much more obvious reason. But I fixed that as well in 1286 while fixing the fact that best_child could get out of sync as part of terminal propagation.
This leaves using 'best child n' for pruning even though the best child may be a terminal win and not the maximum n.  Its debatable whether that is a real issue, but as of 1284 smart pruning now terminates immediately if best move is proven win, so that really is a non-issue. (There is also the case where all options are terminal loss, but that scenario hardly matters...)

PR #1286, opened 2020-05-10
Fix bug where pv might not update for best move change.

Tilps:
Specifically, if a visit proves terminal win status, it should become the best edge even though the n might still be far behind the best.
Also don't recalculate if already the best edge other than in the change to terminal case. Doing this introduces another corner case, but the new logic should handle it. Specifically if a visit proves terminal loss status, it should stop being the best edge even if it has the maximum n.

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -1524,6 +1524,8 @@ void SearchWorker::DoBackupUpdateSingleNode(
     // Nothing left to do without ancestors to update.
     if (!p) break;
 
+    bool old_update_parent_bounds = update_parent_bounds;
+
     // Try setting parent bounds except the root or those already terminal.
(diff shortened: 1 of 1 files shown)


Tilps:
This in combination with #1284 should fix #1283 - I think... since terminal wins will immediately terminate search so if the best_edge isn't the maximum n, that is fine in the search loop, since that should only happen for terminal win, where we are going to immediately exit search anyway.

Tilps:
1k fixed node test with a pre mlh T70 - borderline positive well inside error bars.  I expect this is 0 elo unless I made a mistake, so looks good.
Score of lc0 dev local build vs lc0 newbaseline: 395 - 388 - 1217 [0.502]
Elo difference: 1.22 +/- 9.52

2000 of 2000 games finished.

PR #1287, opened 2020-05-10
alphazero input value to pct 

MelleKoning:
refactored the input value to pct.
Also updated documentation at
https://github.com/LeelaChessZero/lczero.org/pull/44/

### DIFF
--- a/src/mcts/stoppers/alphazero.cc
+++ b/src/mcts/stoppers/alphazero.cc
@@ -60,8 +60,7 @@ std::unique_ptr<SearchStopper> AlphazeroTimeManager::GetStopper(
 
   auto total_moves_time = *time + increment - move_overhead_;
   
-  // use the increment in the first upcoming move
-  float this_move_time = increment + (total_moves_time * (alphazerotimepct_ / 100.0f));
+  float this_move_time = total_moves_time * (alphazerotimepct_ / 100.0f);
 
   LOGFILE << "Budgeted time for the move: " << this_move_time << "ms"
           << "Remaining time " << *time
--- a/src/mcts/stoppers/alphazero.cc
+++ b/src/mcts/stoppers/alphazero.cc
@@ -61,7 +61,7 @@ std::unique_ptr<SearchStopper> AlphazeroTimeManager::GetStopper(
   auto total_moves_time = *time + increment - move_overhead_;
   
   // use the increment in the first upcoming move
-  float this_move_time = increment + (total_moves_time / (100.0f / alphazerotimepct_));
+  float this_move_time = increment + (total_moves_time * (alphazerotimepct_ / 100.0f));
 
   LOGFILE << "Budgeted time for the move: " << this_move_time << "ms"
           << "Remaining time " << *time
--- a/src/mcts/stoppers/alphazero.cc
+++ b/src/mcts/stoppers/alphazero.cc
@@ -55,10 +55,7 @@ std::unique_ptr<SearchStopper> AlphazeroTimeManager::GetStopper(
   // If no time limit is given, don't stop on this condition.
   if (params.infinite || params.ponder || !time) return nullptr;
 
-  const std::optional<int64_t>& inc = is_black ? params.binc : params.winc;
-  const int increment = inc ? std::max(int64_t(0), *inc) : 0;
-
-  auto total_moves_time = *time + increment - move_overhead_;
+  auto total_moves_time = *time - move_overhead_;
   
   float this_move_time = total_moves_time * (alphazerotimepct_ / 100.0f);
 
@@ -66,10 +63,7 @@ std::unique_ptr<SearchStopper> AlphazeroTimeManager::GetStopper(
           << "Remaining time " << *time
           << "ms(-" << move_overhead_ << "ms overhead)";
   
-  // Make sure we don't exceed current time limit with what we calculated.
-  auto deadline =
-      std::min(static_cast<int64_t>(this_move_time), *time + increment - move_overhead_);
-  return std::make_unique<TimeLimitStopper>(deadline);
+  return std::make_unique<TimeLimitStopper>();
 }
 
 }  // namespace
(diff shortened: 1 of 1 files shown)


src/mcts/stoppers/alphazero.cc
```diff
@@ -83,7 +80,7 @@ std::unique_ptr<SearchStopper> AlphazeroTimeManager::GetStopper(
   // Make sure we don't exceed current time limit with what we calculated.
   auto deadline =
       std::min(static_cast<int64_t>(this_move_time), *time + increment - move_overhead_);
```
borg323, line 82: With the change suggested above, this is not needed (and increment shouldn't be added).

src/mcts/stoppers/alphazero.cc
```diff
@@ -74,7 +71,7 @@ std::unique_ptr<SearchStopper> AlphazeroTimeManager::GetStopper(
   auto total_moves_time = *time + increment - move_overhead_;
   
   // use the increment in the first upcoming move
-  float this_move_time = increment + (total_moves_time / alphazerotimevalue_);
+  float this_move_time = increment + (total_moves_time / (100.0f / alphazerotimepct_));
```
borg323, line 74: You shouldn't use increment here. I think this should just be:
```
float this_move_time = (*time - move_overhead_) * alphazerotimepct_ / 100.0f
```

MelleKoning: Let me get to the bottom of this.

I think we do need the increase see screenshot below.. 

Of course, alphazero just used a budget and did not have an inrease per move, but if we have increase per move we should add it to the budgeted time for the upcoming move immediately. Below is the sheet I made to test out the difference. The blue line stalls at 240 seconds, because 5% of 240 seconds is the 12 seconds increase, which are budgeted and used and then next move again 12 seconds is added, therefore the line does not get below the 240 seconds (asymptote).

The red line is what we want to achieve: even if there is an increase we should use it.

Think it's correct but if you can point out a possible mistake please do :)

image

src/mcts/stoppers/alphazero.cc
```diff
@@ -33,8 +33,6 @@ namespace {
 
 class AlphazeroStopper : public TimeLimitStopper {
  public:
-  AlphazeroStopper(int64_t deadline_ms, int64_t* time_piggy_bank)
-      : TimeLimitStopper(deadline_ms), time_piggy_bank_(time_piggy_bank) {}
   void OnSearchDone(const IterationStats& stats) override {
     *time_piggy_bank_ += GetTimeLimitMs() - stats.time_since_movestart;
   }
```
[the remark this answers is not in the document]
MelleKoning: have removed the class

src/mcts/stoppers/alphazero.cc
```diff
@@ -47,18 +45,17 @@ class AlphazeroTimeManager : public TimeManager {
  public:
   AlphazeroTimeManager(int64_t move_overhead, const OptionsDict& params)
       : move_overhead_(move_overhead),
-        alphazerotimevalue_(params.GetOrDefault<float>("alphazero-time-value", 20.0f)),
-        spend_saved_time_(params.GetOrDefault<float>("immediate-use", 1.0f)) {}
+        alphazerotimepct_(params.GetOrDefault<float>("alphazero-time-pct", 5.0f)) {
+    if (alphazerotimepct_ < 1.0f || alphazerotimepct_ > 99.0f) 
+      throw Exception("alphazero-time-pct value to be in range [1.0, 99.0]");
```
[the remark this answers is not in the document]
MelleKoning: adjusted

src/mcts/stoppers/alphazero.cc
MelleKoning: Have created a sheet with some test values, see other comment

src/mcts/stoppers/alphazero.cc
borg323: You add the increment before the move, which is not correct. So at the first move the total time in your example is 300 seconds, not 312. (This is a quick comment, I'm a bit busy right now, will take a deeper look later).

MelleKoning: It looks as if the Inc is added twice and that looks strange  but it works out when UCI comes in with

Wtime 300000 ms winc 12000 ms.

Total time = wtime + winc

Move time = winc + (totaltime * perc)

Totaltime also includes the winc, but otherwise we don't use all possible given milliseconds.

borg323: Think of an edge case, e.g. 100% perc, no move overhead. Then for the first move you will have
```
Total time = 312
Move time = 324
```
I agree that the behavior with percentages that give a move time under the increment is strange, but I'm not sure this needs fixing.

MelleKoning: Hmm, when somebody would budget close to all time to the first move then the graph begins to look strange anyways indeed.

Here is an example for 75%.

image

Basically I think when an increment is set, that increment should be used, as is the case with the current implementation. While if during testing the clock would not go below the threshhold of increment+chosen percentage (like in this example go from 16 to 4 seconds and thus total time stays at16 seconds) that gives a wrong impression. Therefore I would prefer to keep the calculation as it is now.

src/mcts/stoppers/alphazero.cc
```diff
@@ -73,8 +60,7 @@ std::unique_ptr<SearchStopper> AlphazeroTimeManager::GetStopper(
 
   auto total_moves_time = *time + increment - move_overhead_;
```
borg323, line 61: You shouldn't be adding the increment here either.

MelleKoning: Ehm. Can you explain where the increment would be used if it would also be removed here?

borg323: Nowhere really, since we don't try to predict number of remaining moves to estimate total game duration. Note that uci time control is not like in fide rules, so the increment is not added to the current move available time.

MelleKoning: ah, that clarifies

PR #1173, opened 2020-04-01
Policy scaling -- now with logit!

Naphthalin:
As policies are the output of a softmax layer, it makes sense to think of the "true" policies as logits. This PR adds a logarithmic term `c * log(1 + a * N_b)` to the logit of the policy dependent on the visits already spent on a move and recomputes the softmax. The result of this is that policies of moves which get explored will grow (up to a maximum of 1.0), of course not adding up to 1 anymore -- but this is just the role of cpuct scaling: effectively avoid the "policies add up to 1" restriction, but in a way that only moves worth exploring profit, so it basically is a different approach to policy flattening + cpuct scaling.

Expected effects include:

- when a low policy move is discovered to be the best, its policy will rise quickly, accelerating its exploration -- both at root and in the tree
- when a move was explored but is found to be bad, it will have a high policy, but not explored much anymore because of the worse value
- in general, reasonable moves profit from this much more than from cpuct scaling, while bad moves don't profit --> **this and** 
- when multiple moves are identical best moves (e.g. when a transposition happens), after some time all of them will have the same number of visits, ignoring the initial policies --> **this might heavily increase the usefulness of using Lc0 as an analysis tool for human play** 
- and finally a rather theoretical argument: this treats the policies as priors of how reluctantly the UCT should explore these moves; when searching for infinite times, all policies will be 1 (while cpuct is constant), and the theoretical results for convergence actually apply, unlike for cpuct scaling

Example 1: Start position with net 591215 at default values. First is with standard settings, second is with standard settings but cpuct scaling turned off at root as well, third is still with cpuct scaling off, but policy logit scaling with `c=1` and `a=0.0016`, all searched for 1M nodes:

Default:
```
info string g2g4  (378 ) N:     203 (+ 0) (P:  1.17%) (WL: -0.37563)
info string f2f3  (346 ) N:     287 (+ 0) (P:  1.26%) (WL: -0.25841)
info string g1h3  (161 ) N:     389 (+ 0) (P:  1.46%) (WL: -0.20402)
info string b1a3  (34  ) N:     507 (+ 0) (P:  1.67%) (WL: -0.16423)
info string h2h4  (403 ) N:     559 (+ 0) (P:  1.68%) (WL: -0.13910)
info string f2f4  (351 ) N:     692 (+ 0) (P:  1.80%) (WL: -0.10390)
info string b2b4  (234 ) N:     869 (+ 0) (P:  2.02%) (WL: -0.07957)
[...]

PR #1288, opened 2020-05-10
Policy Scaling with tunable exponent

AlexisOlson:
This is a copy of Naphthalin's https://github.com/LeelaChessZero/lc0/pull/1173 with an extra parameter and "April" naming changed.


Naphthalin:
Nice :) if you want, take the `fastpow` function in `utils/fastmath.h` from https://github.com/Naphthalin/lc0/blob/april2/src/utils/fastmath.h#L87

Naphthalin:
Just a few benchmark results on my machine. i7-4770, RX Vega 56, Win10, dx12, random backend

`lc0_1288_aa9fce6` (= pre fastpow): 
```
Total time (ms) : 340168
Nodes searched  : 26117601
Nodes/second    : 76778
```

`lc0_1288_17bc85f` (= with fastpow): 
```
Total time (ms) : 340182
Nodes searched  : 27454097
Nodes/second    : 80704
```

And as a comparison:
`lc0_1173` (= sqrt fixed): 
```
Total time (ms) : 340101
Nodes searched  : 39490513
Nodes/second    : 116114
```

`lc0_05ee599`:
```
Total time (ms) : 340252
Nodes searched  : 43511546
Nodes/second    : 127880
```

These changes heavily affect the inner loop, so have a performance implication. However this general concept definitely doesn't need the precision given by the various function, and just cares about having a result in the right order of magnitude, and in the case of "big numbers" the result will be close to 1 anyway. So I would hope for some possible simplifications here.

AlexisOlson:
After FastPow update from @borg323 with `benchmark --backend=random`:

```
lc0_1288_948026e  :   65,594 nps
lc0_25.1_69105b4e :  168,376 nps 
```

With `benchmark --weights=11258-32x4-se.pb.gz` (cudnn backend)
```
lc0_1288_04be91e  :  38,249 nps
lc0_25.1_69105b4e :  38,270 nps 
```

GPU Cuda Appveyor builds for [v25.1-69105b4e](https://ci.appveyor.com/project/LeelaChessZero/lc0/builds/32564767) and [PR1288-04be91e](https://ci.appveyor.com/project/LeelaChessZero/lc0/builds/32809098/job/e59chm7cx5xlwf4j/artifacts)
CPU = i7-8700T CPU @ 2.40GHz
GPU = NVIDIA Quadro P620

---

Using the power function is a serious slowdown on the random backend, even with a fast approximation, but less so using a weights file given that it's not as CPU limited and the tree shape might benefit from the policy scaling instead of cpuct scaling.

src/mcts/node.h
```diff
@@ -378,25 +378,42 @@ class EdgeAndNode {
...
+  float GetPEffect(float policy_factor, float policy_factor_parent,
+                  float policy_exponent) const {
+    auto visits = GetN();
+    auto visits_parent = node_ ? node_->GetParent()->GetN() : 0;
+    auto psa = GetP();
+    return ( ( policy_factor > 0.0 ) || ( policy_factor_parent > 0.0 ) )
+      ? ( psa > 0.0f
+        ? 1.0f / ( 1.0f + (1.0f / psa - 1.0f) *
+            FastPow( 1.0f + policy_factor * visits +
+                          policy_factor_parent * visits_parent, policy_exponent )
```
borg323, line 390: Unless I'm missing something (not unlikely), this should be `/ FastPow()` instead of `* FastPow()` to be equivalent to #1173. And no, `FastPow()` **will not** work with negative exponents.

Naphthalin: You are completely correct there :)

AlexisOlson: Yep. I just realized this. Thanks guys.

AlexisOlson: I've updated the benchmark above.

borg323:
What about something like the following?
```diff
diff --git a/src/mcts/node.h b/src/mcts/node.h
index 0e0145c..59e02cd 100644
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -379,18 +379,32 @@ class EdgeAndNode {
   // Edge related getters.
   float GetP() const { return edge_->GetP(); }
   float GetPEffect(float policy_factor, float policy_factor_parent,
-                  float policy_exponent) const {
+                   float policy_exponent) const {
     auto visits = GetN();
     auto visits_parent = node_ ? node_->GetParent()->GetN() : 0;
     auto psa = GetP();
-    return ( ( policy_factor > 0.0 ) || ( policy_factor_parent > 0.0 ) )
-      ? ( psa > 0.0f
-        ? 1.0f / ( 1.0f + (1.0f / psa - 1.0f) /
-            FastPow( 1.0f + policy_factor * visits +
-                          policy_factor_parent * visits_parent, policy_exponent )
-          )
-        : 0.0f )
-      : psa ; }
+    float m;
+    if (policy_exponent == 0.0f) {
+      m = 1.0f;
+    } else if (policy_exponent == 0.5f) {
+      m = FastInvSqrt(1.0f + policy_factor * visits +
+                      policy_factor_parent * visits_parent);
+    } else if (policy_exponent == 1.0f) {
+      m = 1.0f / (1.0f + policy_factor * visits +
+                  policy_factor_parent * visits_parent);
+    } else if (policy_exponent == 2.0f) {
+      m = 1.0f / (1.0f + policy_factor * visits +
+                  policy_factor_parent * visits_parent);
+      m *= m;
+    } else {
+      m = FastPow(
+          1.0f + policy_factor * visits + policy_factor_parent * visits_parent,
+          -policy_exponent);
+    }
+    return ((policy_factor > 0.0) || (policy_factor_parent > 0.0))
+               ? (psa > 0.0f ? 1.0f / (1.0f + (1.0f / psa - 1.0f) * m) : 0.0f)
+               : psa;
+  }
   Move GetMove(bool flip = false) const {
     return edge_ ? edge_->GetMove(flip) : Move();
   }
diff --git a/src/utils/fastmath.h b/src/utils/fastmath.h
index 37a582c..996579f 100644
--- a/src/utils/fastmath.h
+++ b/src/utils/fastmath.h
@@ -84,13 +84,20 @@ inline float FastLogit(const float a) {
 
 // Fast approximate a^x.
 inline float FastPow(const float a, const float b) {
-  int32_t tmp;
-  std::memcpy(&tmp, &a, sizeof(float));
-  tmp = (tmp - 0x3f800000) * b;
-  tmp += 0x3f800000;
-  float out;
-  std::memcpy(&out, &tmp, sizeof(float));
-  return out;
+  return FastPow2(FastLog2(a) * b);
+}
+
+// Fast inverse sqrt.
+inline float FastInvSqrt(const float number)
+{
+  const float x2 = number * 0.5F;
+  const float threehalfs = 1.5F;
+  float f;
+  uint32_t i;
+  memcpy(&i, &number, sizeof(float));
+  i  = 0x5f3759df - ( i >> 1 );
+  memcpy(&f, &i, sizeof(float));
+  return f * ( threehalfs - ( x2 * f * f ) );
 }
 
 }  // namespace lczero
```

AlexisOlson:
@borg323 I don't think I'll put in special cases at this point but do you think I should revert to the other FastPow?

borg323:
Yes, the `FastPow()` hack I suggested was just to see how much it improved nps, but I wouldn't use it in practice. Note that this version uses `FastPow2(FastLog2(a) * b);` which is a bit faster than `FastExp(FastLog(a) * b);` and is called with negative `b` to avoid one division.

borg323:
This is the faster I have so far (with lower accuracy). Also saturates instead of underflowing.
```cpp
diff --git a/src/utils/fastmath.h b/src/utils/fastmath.h
index 553401f..bca095d 100644
--- a/src/utils/fastmath.h
+++ b/src/utils/fastmath.h
@@ -82,9 +82,21 @@ inline float FastLogit(const float a) {
   return 0.5f * FastLog((1.0f + a) / (1.0f - a));
 }
 
-// Fast approximate a^x.
+// Fast approximate x^y. Works for x>0 and saturates on underflow.
+// The approximation used here is a combination of log2(2^N*(1+f)) ~ N+f+k
+// and 2^(N+f) ~ 2^N*(1+f-k) with N integer and f the fractional part, f>=0.
+// The constant k is used to tune the approximation accuracy.
 inline float FastPow(const float a, const float b) {
-  return FastPow2(FastLog2(a) * b);
+  int32_t tmp;
+  std::memcpy(&tmp, &a, sizeof(float));
+  tmp -= 0x3f78aa40;
+  float out;
+  tmp = static_cast<int>(tmp * b + 1064872512.0f);
+  if (tmp < 0x800000) {
+    return 1.17549435E-38;
+  }
+  std::memcpy(&out, &tmp, sizeof(float));
+  return out;
 }
 
 }  // namespace lczero
```

src/utils/fastmath.h
```diff
@@ -82,9 +82,21 @@ inline float FastLogit(const float a) {
   return 0.5f * FastLog((1.0f + a) / (1.0f - a));
 }
 
-// Fast approximate a^x.
+// Fast approximate x^y. Works for x>0 and saturates on underflow.
+// The approximation used here is a combination of log2(2^N*(1+f)) ~ N+f+k
+// and 2^(N+f) ~ 2^N*(1+f-k) with N integer and f the fractional part, f>=0.
+// The constant k is used to tune the approximation accuracy.
```
oscardssmith, line 88: what constant k?

AlexisOlson: I'm guessing a magical constant that you can tune to optimize accuracy for a particular x,y distribution. The maximal error will depend on what range you expect to see.

oscardssmith: right, but the code should probably have a constant k, not just the comment.

borg323: This is the high level description of the algorithm, similar to the ones earlier in the file. Are you asking for the exact value of the constant or to see it in the code (where it is indeed folded)? BTW in this case there is nothing magical about it, the error of the approximation is very one sided, so the constant is just an offset added to make it symmetrical. I expect there is still some improvement possible by tuning the two k independently (they are identical now) but this has no effect on the speed so is not done yet.

AlexisOlson:
Fresh test with fresh reboot and other programs closed:

```
lc0_1288_5bb36bc  :  129,238 nps
lc0_1173_f1ce36d  :  129,850 nps
lc0_25.1_69105b4e :  179,140 nps
```

AlexisOlson:
I forked [@borg323 's brach](https://github.com/borg323/lc0/tree/dag-master-pr1821) here to try to bring this up to date:
https://github.com/AlexisOlson/lc0/tree/policy-scaling

Naphthalin:
#2093 is a reimplementation of this in the current code base without the child visit boost part.

PR #1292, opened 2020-05-11
Use explicit int for printing verbose and debug num edges.

### LINKED ISSUE #1291: Bytes being inserted into the verbose "node" lines
When introduced, Mardak's node lines in the verbose stats looked like this:

```
info string node  (   7) etc
```

Where 7 is the number of moves available. However, at the current commit, it now inserts a byte into the string instead, e.g. byte 0x07.

If you want to test this, you might want to use a position with enough moves to generate a non-whitespace character, e.g.

```
r1bqkbnr/ppp2ppp/2n5/3pp3/3PP3/2N5/PPP2PPP/R1BQKBNR w KQkq - 0 4
```

Mardak:
r? @mooskagh or @Tilps Fix #1291. This regressed with #1280's:
```diff
-  uint16_t GetNumEdges() const { return edges_.size(); }
+  uint8_t GetNumEdges() const { return num_edges_; }
```

```
After #1291:
info string node  (   ) N:       1 (+ 0) (P:  0.00%) (WL:  0.10861) (D: 0.325) (M:  0.0) (Q:  0.10861) (V:  0.1086) 
Term:0 This:0x7fc1bb004230 Parent:0x0 Index:0 Child:0x0 Sibling:0x0 WL:-0.108615 N:1 N_:0 Edges: Bounds:-1,1


After this PR:
info string node  (  20) N:       1 (+ 0) (P:  0.00%) (WL:  0.10861) (D: 0.325) (M:  0.0) (Q:  0.10861) (V:  0.1086) 
Term:0 This:0x7ffc81f27710 Parent:0x0 Index:0 Child:0x0 Sibling:0x0 WL:-0.108615 N:1 N_:0 Edges:20 Bounds:-1,1
```

### DIFF
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -213,7 +213,7 @@ std::string Node::DebugString() const {
       << " Parent:" << parent_ << " Index:" << index_
       << " Child:" << child_.get() << " Sibling:" << sibling_.get()
       << " WL:" << wl_ << " N:" << n_ << " N_:" << n_in_flight_
-      << " Edges:" << num_edges_
+      << " Edges:" << static_cast<int>(num_edges_)
       << " Bounds:" << static_cast<int>(lower_bound_) - 2 << ","
       << static_cast<int>(upper_bound_) - 2;
   return oss.str();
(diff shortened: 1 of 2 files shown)


Mardak:
Actually, node's DebugString is broken too. Fixing…

https://github.com/LeelaChessZero/lc0/blob/2400dabfd0cc7f3c057254ff12f29d25dc7c1668/src/mcts/node.cc#L216

PR #1294, opened 2020-05-12
adjust minibatch-size for appveyor cpu builds

borg323:
4 seems a good value for openblas and dnnl, while for android 7 looked better in some tests.


mooskagh:
Maybe pass it as a macro?.. seding is way too hacky..

borg323:
The latest version adds a mechanism to override some parameters, and changes minibatch-size to 1 and max-prefetch to 0 for cpu builds, as this seems to be a good combination.

borg323:
I find that in practice a minibatch-size of 1 works quite well on intel cpus when all cores are working, which they should if we care about high nps. The arm optimum may be different, which is the reason I kept the android setting on a separate line.

PR #1302, opened 2020-05-14
Use info string to display some important information

nguyenpham:
Use UCI command "info string" to display information about numbers of Sygyzy endgames and the network file.

**Discussion:**

Even though consoles/terminals can display all information Lc0 wants to show, almost all users use chess GUIs to play with Lc0. There are a few problems for some GUIs to display some information:

- They (information) are treated as log information, displayed in the log window only. However, that window may be not activated or hidden in some GUIs (not easy to access)

-  They may be easily scrolled out /deleted when playing since there is too much log info
- They may use std::err which some GUIs totally ignore thus they have lost forever

This PR converts from simple using std::err to use UCI "info string" to display some vital (users highly referent) information about Syzygy endgames and the network name.

That information could be displayed in the string window of a GUI as bellow image.

Screen Shot 2020-05-14 at 8 55 38 pm


mooskagh:
Wrong destination branch (should be `master`, not `release/0.25`).

nguyenpham:
Corrected!

mooskagh:
I don't think it's right to create `UciLoop()` in arbitrary places of the program.
Search/syzygy loading/etc is not really dependent from UCI.

E.g. I have unfinished python PR with search wrapper, and it also loads tablebases. But it's not connected with UCI in any way, it's weird if it will suddenly output info string. Or actually since recently we have a python backend wrapper, which also needs to load weights, and it's not expected for it to print anything to stdout.
Search doesn't mean UCI, sysygy doesn't mean UCI. Also weights loaded in selfplay would output uci `info` now, even though it's not an `uci` session.

I think the correct way to do that is to pass info callback around. (or pass a simplified wrapper of this callback which is `std::function<(const std::string&)>` rather than `std::function<(const ThinkingInfo&)>`.
And then the host code can decide what to do with this. E.g. if we a in Uci context, this can be output as info string, if it's python wrapper, it can be written into some variable, if it's selfplay context, it can be printed to the screen without info string prefix, or if we even have a json extension, as info json {"message":"blah"}.
If there are multiple games going on in parallel (for selfplay), we can attribute that message to a particular game, if we ever need it.

nguyenpham:
UciLoop is actually a class with totally empty constructor and deconstructor. It means the call UciLoop() just allocates a bit tiny memory for the object without any extra work. It is quite similar to command `int number`.

By doing as my code:
- the code runs quickly since it is directly, not extra nor redundant work
- we can avoid complicated ways (as the one you suggest ;)  ). The code is much easy to maintain and to be consistent (the code for sending `into` is in one place only)

mooskagh:
It is wrong to output something to `stdout` when clients don't expect that.

For example, I have a python script which loads networks one by one, runs evals on them, and outputs results as `.csv` to stdout.
After merging this PR, it will have `info string weights file is /path/to/file` after every line.

There are also ideas of supporting other protocols (namely gRPC). gRPC is actually not stdin/stdout based so it wouldn't be so much of a problem, but i.e. for winboard it would.

nguyenpham:
OK, got it. If you want to support more protocols and more output the code is a problem. I closed it. I will come back later to implement it in another way.

Issue #1306, opened 2020-05-21
Ramping up the node limit vastly changes the distribution of nodes


fohristiwhirl:
Suppose I gently ramp up the number of nodes, e.g. `go nodes 200` (let it finish) `go nodes 400` (let it finish) `go nodes 600` etc and eventually reach 4000. Is it expected that the resulting distribution of nodes in the tree is drastically different than if I'd just asked for 4000 nodes in the first place? 

Discussion [here on Discord](https://discordapp.com/channels/425419482568196106/425419999096733706/712965973661188116).

To reproduce exactly what I did, use net 63409. Note that SmartPruningFactor is 0.

```
uci
ucinewgame
setoption name WeightsFile value 63409
setoption name SmartPruningFactor value 0
setoption name Threads value 1
position fen 8/5pk1/p1b1pqp1/P1P4r/5PN1/4Q1P1/1R1R1K2/7r b - - 4 42

go nodes 200
(let it finish)

go nodes 400
(let it finish)

go nodes 600
(let it finish)

go nodes 800
(let it finish)

... etc etc until reaching 4000.

setoption name VerboseMoveStats value true
go nodes 4000

info string h5h4  (888 ) N:      86 (+ 0) (P:  0.71%) (WL:  0.78393)
info string f6e7  (571 ) N:     102 (+ 0) (P:  5.18%) (WL:  0.23908)
info string h1c1  (1787) N:    3491 (+ 0) (P:  1.35%) (WL:  0.35197)
```

(The process of ramping up the node limit is also available in recent versions of Nibbler.) Now compare the distribution with what you get from the following:

```
ucinewgame
position fen 8/5pk1/p1b1pqp1/P1P4r/5PN1/4Q1P1/1R1R1K2/7r b - - 4 42
go nodes 4000

info string f6e7  (571 ) N:     118 (+ 0) (P:  5.18%) (WL:  0.21227)
info string h1c1  (1787) N:     520 (+ 0) (P:  1.35%) (WL:  0.35927)
info string h5h4  (888 ) N:    2988 (+ 0) (P:  0.71%) (WL:  0.72386)
```

jkormu:
The inability to disable smart pruning can lead to even more extreme results in certain positions where N nodes will give 0 nodes to move X but when searching with N+1 node, the X will be the most visited.

[...]

PR #1307, opened 2020-05-21
Don't do any smart pruning if smart pruning factor is 0 or lower.

Tilps:
Also disable this smart pruning in benchmark and selfplay that don't support self pruning setting.

### DIFF
--- a/src/benchmark/benchmark.cc
+++ b/src/benchmark/benchmark.cc
@@ -91,7 +91,7 @@ void Benchmark::Run() {
         stopper->AddStopper(std::make_unique<TimeLimitStopper>(movetime));
       }
       if (visits > -1) {
-        stopper->AddStopper(std::make_unique<VisitsStopper>(visits));
+        stopper->AddStopper(std::make_unique<VisitsStopper>(visits, 0.0f));
       }
 
       NNCache cache;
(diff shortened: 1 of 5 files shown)


Tilps:
This logic change reverts behavior to before PR884, where playouts_remaining_ was never set if smart pruning factor was <=0

Tilps:
Hopefully fixes #1306 when smart pruning is disabled.

jkormu:
Seems to fix the issue in #1306 regarding position 8/5pp1/3k2p1/1p6/pPB5/P4Kb1/5PP1/8 w - - 0 1, where a move jumps from 0 to hero when go node limit is increased just by 1 node.

mooskagh:
Hmm it feels like instead of passing that to stoppers, it's cleaner to somehow either ignore it in search or maybe unset in the last stopper in chain..
Because logically it's just telling how many nodes are left, and it's up to search to decide what to do with that information.
I'll look more tomorrow morning.

Tilps:
my first thought was to add it to the if statement in search - but since smart pruning factor is a 'stopper parameter' now and no longer part of params.cc - this seemed much simpler :P

In some sense I actually like this way better - because playouts remaining estimate is already not updated based on time control and nps estimate if smart pruning factor is 0 - so this just makes it consistently not updated rather than inconsistently not updated. At least when its consistently not updated its clearer that it isn't usable outside of a smart pruning scenario.

mooskagh:
Ok, then probably it makes sense at least to pass `bool populate_remaining_moves` instead of `float` (as that float is only used to be compared with 0).

Tilps:
Done.

Tilps:
Just realized I used your recommended name without thinking - should be populate_remaining_playouts - will change.

PR #1309, opened 2020-05-23
basic line editing

borg323:
I seems on windows the console provides some basic line editing functionality to the uci interface. This adds it to other systems as well.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -14,7 +14,7 @@
 # You should have received a copy of the GNU General Public License
 # along with Leela Chess.  If not, see <http://www.gnu.org/licenses/>.
 
-project('lc0', 'cpp',
+project('lc0', ['c', 'cpp'],
         default_options : ['cpp_std=c++17', 'b_ndebug=if-release', 'warning_level=3'],
         meson_version: '>=0.52')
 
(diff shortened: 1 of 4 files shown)


mooskagh:
I wonder if:
1. We should add some more copyrights to the license file that we bundle,
2. Whether it still works fine in all non-interactive scenarios (i.e. chess GUIs)
3. Whether it doesn't slow down input..

borg323:
> 1. We should add some more copyrights to the license file that we bundle,

We certainly should add copyrights somewhere but maybe not in the license file. However, I see we are only bundling a README with cuda, which seems like a bug. 
Moreover, I am also worried about your other two questions (and 4. MacOS portability, 5. Is it needed for android) so I didn't bother doing it before we decide that we really want this.

mooskagh:
I thought that it may make sense to just have a wrapper tool:
Which redirects stdout and stderr to stdout and stderr.
Which provides line editing interface for stdin and then redirects it.

I've just found https://github.com/thejoshwolfe/consoline, seems to be it..

borg323:
Yes, this is good enough.

PR #1311, opened 2020-05-23
Editing installation instructions for Mac for Different OSes

TesseractA:
Went through some rather frustrating troubleshooting... turns out that Xcode Developer tools now contains these SDK headers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,16 +160,16 @@ First you need to install some required packages through Terminal:
 2. Install python3: `brew install python3`
 3. Install meson: `brew install meson`
 4. Install ninja: `brew install ninja`
-5. (For Mac OS 10.14 Mojave, or if the other step 5 fails)
-    Install developer tools: ``xcode-select --install``
-    When using Mojave install SDK headers: `installer -pkg /Library/Developer/CommandLineTools/Packages/macOS_SDK_headers_for_macOS_10.14.pkg -target /` (if this doesn't work, use `sudo installer` instead of just `installer`.)
+5. (For Mac OS 10.14 Mojave, or if the other step 5 fails):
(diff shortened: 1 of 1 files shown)


TesseractA:
Before someone merges this change, can someone with 10.15 check whether the 5b step for MacOS Catalina is necessary? 

5a. Install developer tools: ``xcode-select --install``
5b. Install Xcode Developer Tools through the app store. You may need to create an apple store account to do this. (The correct application should be the first one that appears with the blueprint and hammer in the Apple Store Search.)
5c. Use in terminal: `export CPATH=`xcrun --show-sdk-path`/usr/include`

TesseractA:
@mooskagh could this thread get a "help needed" label?

mooskagh:
I'm not sure if that's ready, so didn't hit the `merge` button.

TesseractA:
I'm putting a 10-day time limit on the help, because it's better to have these instructions than not to have these instructions included at all. Less is not better here for those that are trying to figure out how to install on Catalina. After the 10-day time limit, I'll request that this be merged because these instructions worked for at least one individual.

TesseractA:
Given that it has been 12 days, I think that if this ever gets tested, it will be a while. So I did a bit of research just to make sure I was right. According to these sources, it looks like there was a migration of SDK headers from the default location to XCode11 app. So these steps are necessary.
https://apple.stackexchange.com/questions/372032/usr-include-missing-on-macos-catalina-with-xcode-11

https://stackoverflow.com/questions/53169970/where-are-the-c-headers-in-macos-mojave

https://developer.apple.com/documentation/macos-release-notes/macos-catalina-10_15-release-notes

Could someone please merge?

TesseractA:
Thanks!

PR #1312, opened 2020-05-24
Terminal correction

Tilps:
Adjust parent node Q/D/M values to act as if newly converted terminals always had the Q/D/M values that they now have.
This will 'momentarily' be confusing for search - so may need to be combined with something like crem's recent time management adjustment that waits for best N to catch up with best Q to avoid missing out on taking advantage of the information gain applied.

I've only done minimal validation that the values move the direction I expect them to when terminal status gets determined.
Given this is in the return loop it probably will need nps or check at time control in addition to any proof of improvement at fixed nodes.
But first I'll try to demonstrate at fixed nodes if I can find a spare minute or two on my machine :P

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -1544,7 +1544,7 @@ void SearchWorker::DoBackupUpdateSingleNode(
     }
     n->FinalizeScoreUpdate(v / (1.0f + params_.GetShortSightedness() * depth),
                            d, m, node_to_process.multivisit);
-    if (n_to_fix > 0) {
+    if (n_to_fix > 0 && !n->IsTerminal()) {
       n->AdjustForTerminal(
           v_delta / (1.0f + params_.GetShortSightedness() * depth), d_delta,
           m_delta, n_to_fix);
@@ -1557,8 +1557,9 @@ void SearchWorker::DoBackupUpdateSingleNode(
     // If parent already is terminal further adjustment is not required.
     if (p->IsTerminal()) n_to_fix = 0;
     // Try setting parent bounds except the root or those already terminal.
-    update_parent_bounds = update_parent_bounds && p != search_->root_node_ &&
-                           !p->IsTerminal() && MaybeSetBounds(p, m, &n_to_fix, &v_delta, &d_delta, &m_delta);
+    update_parent_bounds =
+        update_parent_bounds && p != search_->root_node_ && !p->IsTerminal() &&
+        MaybeSetBounds(p, m, &n_to_fix, &v_delta, &d_delta, &m_delta);
 
     // Q will be flipped for opponent.
     v = -v;
@@ -1587,7 +1588,9 @@ void SearchWorker::DoBackupUpdateSingleNode(
   search_->max_depth_ = std::max(search_->max_depth_, node_to_process.depth);
 }
 
-bool SearchWorker::MaybeSetBounds(Node* p, float m, int* n_to_fix, float* v_delta, float* d_delta, float* m_delta) const {
+bool SearchWorker::MaybeSetBounds(Node* p, float m, int* n_to_fix,
+                                  float* v_delta, float* d_delta,
+                                  float* m_delta) const {
   auto losing_m = 0.0f;
   auto prefer_tb = false;
 
@@ -1632,9 +1635,7 @@ bool SearchWorker::MaybeSetBounds(Node* p, float m, int* n_to_fix, float* v_delt
   } else if (lower == upper) {
     // Search can stop at the parent if the bounds can't change anymore, so make
     // it terminal preferring shorter wins and longer losses.
-    int old_n_to_fix = *n_to_fix;
-    // This assumes the active visit is a multivist of size 1.
-    *n_to_fix = p->GetN() - 1;
+    *n_to_fix = p->GetN();
     assert(*n_to_fix > 0);
     float cur_v = p->GetWL();
     float cur_d = p->GetD();
@@ -1643,9 +1644,11 @@ bool SearchWorker::MaybeSetBounds(Node* p, float m, int* n_to_fix, float* v_delt
         -upper,
         (upper == GameResult::BLACK_WON ? std::max(losing_m, m) : m) + 1.0f,
         prefer_tb ? Node::Terminal::Tablebase : Node::Terminal::EndOfGame);
-    *v_delta = p->GetWL() - cur_v + old_n_to_fix * *v_delta / *n_to_fix;
-    *d_delta = p->GetD() - cur_d + old_n_to_fix * *d_delta / *n_to_fix;
-    *m_delta = p->GetM() - cur_m + old_n_to_fix * *m_delta / *n_to_fix;
+    // Negate v_delta because we're calculating for the parent, but immediately
+    // afterwards we'll negate v_delta in case it has come from the child.
+    *v_delta = -(p->GetWL() - cur_v);
+    *d_delta = p->GetD() - cur_d;
+    *m_delta = p->GetM() - cur_m;
   } else {
     p->SetBounds(-upper, -lower);
   }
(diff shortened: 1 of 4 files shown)


Tilps:
Trre did a quick partial test.
   # PLAYER        :  RATING  ERROR  POINTS  PLAYED   (%)  CFS(%)    W    D    L  D(%)
   1 lc0-pr1312    :     3.4   15.1   422.0     836    50      67  177  490  169    59
   2 lc0-master    :     0.0   ----   414.0     836    50     ---  169  490  177    59
I'll try and get a more comprehensive one in later.

Tilps:
Score of lc0 dev local build vs lc0 newbaseline: 410 - 405 - 1185 [0.501]
Elo difference: 0.87 +/- 9.71

2000 of 2000 games finished.
T70 at 1k nodes without MLH parameters.  
(It was over +12 elo when I walked away at ~600 games, came back to a much less interesting score line :P)
With MLH parameters enabled... almost identical - up 12 elo around 400 then decayed back while I wasn't watching :P 
Score of lc0 dev local build vs lc0 newbaseline: 400 - 396 - 1204 [0.501]
Elo difference: 0.69 +/- 9.59

2000 of 2000 games finished.

Naphthalin:
Can we get some more insight into what exactly the problem is this attempts to fix?

Tilps:
When a node becomes a terminal we suddenly change its Q value - which for the immediate parent means it suddenly becomes as if it was that Q value for all of its previous visits.
This PR makes it so that all parents also see that change not just the immediate parent.
I would just call it a 'bug fix' but the sudden shift in Q value has some risk which was more localized before, and performing the calculation isn't completely free (although its no worse than sticky endgames itself).
More specifically I think this will result in better Q and M values at root for training (for q_ratio mixing and M value at resign), in cases where conversion to terminal doesn't quite propagate all the way to the root before search terminates in training games. (And even when it does it should be better sooner.)

AlexisOlson:
**LC0 version:** Appveyor builds with defaults
**Hardware:**  i7-8700K @ 2.4 GHz + NVIDIA Quadro P620
**Time control:** 6" + 0.1"
**Book:** Balsa_v2724.pgn - random with reverse
**Tablebases:** 5-man syzygy
**Adjudication:** Syzygy TB; Draw after 50 moves and 10 under 10cp; Resign at 10 over 1000cp
**Software:** Cutechess (GUI)

**Ordo:**
```
# PLAYER                       :  RATING  ERROR   (%)  PLAYED  CFS(%)     W     L     D  D(%)
1 lc0_master_32x4 (39,436 nps) :     4.7   11.1    53    3600      80  1213   987  1400  39
2 lc0_PR1312_32x4 (40,764 nps) :     0.0   ----    52    3600     100  1221  1059  1320  37
3 lc0_master_GG5  (30,983 nps) :   -22.3   11.2    48    3600      91  1079  1220  1301  36
4 lc0_PR1312_GG5  (32,005 nps) :   -30.1   11.2    47    3600     ---  1019  1266  1315  37

White advantage = 24.69 +/- 3.27
Draw rate (equal opponents) = 37.34 % +/- 0.57
```

**Head to head statistics:**
```
1) lc0_master_32x4   4.7 :   3600 (+1213,=1400,-987),  53.1 %

   vs.                    :  games (    +,    =,   -),   (%) :    Diff,    SD, CFS (%)
   lc0_PR1312_32x4        :   1200 (  311,  565, 324),  49.5 :    +4.7,   5.7,   79.7
   lc0_master_GG5         :   1200 (  466,  401, 333),  55.5 :   +27.0,   5.7,  100.0
   lc0_PR1312_GG5         :   1200 (  436,  434, 330),  54.4 :   +34.8,   5.7,  100.0

2) lc0_PR1312_32x4   0.0 :   3600 (+1221,=1320,-1059),  52.3 %

   vs.                    :  games (    +,    =,    -),   (%) :    Diff,    SD, CFS (%)
   lc0_master_32x4        :   1200 (  324,  565,  311),  50.5 :    -4.7,   5.7,   20.3
   lc0_master_GG5         :   1200 (  437,  387,  376),  52.5 :   +22.3,   5.7,  100.0
   lc0_PR1312_GG5         :   1200 (  460,  368,  372),  53.7 :   +30.1,   5.7,  100.0

3) lc0_master_GG5  -22.3 :   3600 (+1079,=1301,-1220),  48.0 %

   vs.                    :  games (    +,    =,    -),   (%) :    Diff,    SD, CFS (%)
   lc0_master_32x4        :   1200 (  333,  401,  466),  44.5 :   -27.0,   5.7,    0.0
   lc0_PR1312_32x4        :   1200 (  376,  387,  437),  47.5 :   -22.3,   5.7,    0.0
   lc0_PR1312_GG5         :   1200 (  370,  513,  317),  52.2 :    +7.8,   5.7,   91.5

4) lc0_PR1312_GG5  -30.1 :   3600 (+1019,=1315,-1266),  46.6 %

   vs.                    :  games (    +,    =,    -),   (%) :    Diff,    SD, CFS (%)
   lc0_master_32x4        :   1200 (  330,  434,  436),  45.6 :   -34.8,   5.7,    0.0
   lc0_PR1312_32x4        :   1200 (  372,  368,  460),  46.3 :   -30.1,   5.7,    0.0
   lc0_master_GG5         :   1200 (  317,  513,  370),  47.8 :    -7.8,   5.7,    8.5
```

src/mcts/search.cc
```diff
@@ -1540,18 +1544,26 @@ void SearchWorker::DoBackupUpdateSingleNode(
     }
     n->FinalizeScoreUpdate(v / (1.0f + params_.GetShortSightedness() * depth),
                            d, m, node_to_process.multivisit);
+    if (n_to_fix > 0 && !n->IsTerminal()) {
+      n->AdjustForTerminal(
+          v_delta / (1.0f + params_.GetShortSightedness() * depth), d_delta,
```
mooskagh, line 1549: Unless we have plans to do something else with ShortSighteness, it would make sense to revert it.

Naphthalin: I strongly vote for reverting ShortSightedness, as the problem it originally attempted to solve (depth dependent adjustment of drawish scores) didn't work out the way it was hoped to, and the moves left head so far seems to work good enough.

Tilps: I'm on board with reverting short sightedness - but that will be a separate PR.
I think I need to run a bigger test on this before I'm willing to be happy that it has 0 regression risk to offset the training data quality gains.

mooskagh: Yeah I didn't suggest to revert it now, it probably sounded like I did. :)

Tilps:
Switched to chad's 8 ply book to get smaller error bars.
Score of lc0 dev local build vs lc0 newbaseline: 4045 - 3923 - 12032 [0.503]
Elo difference: 2.12 +/- 3.03

20000 of 20000 games finished.
This time it was up outside of error bars when I went to sleep (at over 3k games) - came back to near zero again.
Since it is again positive, I'm going to try a 10k nodes run (only 2k games though).

Tilps:
Rerun of the second 2k games fixed nodes test, but with 10k nodes per move instead.
Score of lc0 dev local build vs lc0 newbaseline: 244 - 209 - 1547 [0.509]
Elo difference: 6.08 +/- 7.23

2000 of 2000 games finished.
I actually kept the pgn this time, ordo says 91% CFS for this result.
I'll note that the pattern continued - up much larger elo margins at the ~1/5th way mark. If I had have selected to only run 400 games sequential the outcome would have been 20 elo with >95% CFS. But I didn't so I don't have that ;)

Based on these results in combination with the fact that its a logical 'bugfix' I'm happy that its non-regression with likely improvement of training data quality.  Obviously the time control tests performed came out with opposite results to fixed nodes, but they are also within error bars of zero - and I have a slight bias that these tiny nets are not a great representative of what we usually use lc0 for, which gives me some grounds to give their results lower weight.
I haven't measured nps myself as my machine is heavily loaded with other tasks which make that difficult to do reliably, but I will note that the small net time control tests include benchmark details that suggest that this change is an NPS increase which suggests the additional code path execution cost is pretty negligible.
On the balance of the evidence and details I'm moving forwards.

PR #1317, opened 2020-05-26
Save the last sent nps info to use when missing nps.

### LINKED ISSUE #1316: Cutechess calculates nps if one isn't provided resulting in a speed spike
nps
spike

https://tcec-chess.com/#div=p&game=1&season=18&x=archive Also shown in the screenshot and log from #1176.

```
position fen 6k1/3n2P1/4p3/1pp1p3/r7/P2q1N2/1P4Q1/K6R b - - 2 39
go …
bestmove a4a3 ponder b2a3

position fen 6k1/3n2P1/4p3/1pp1p3/r7/P2q1N2/1P4Q1/K6R b - - 2 39 moves a4a3
go nodes 1
info depth 1 seldepth 0 time 1162 nodes 7357 score cp 301 tbhits 0 pv b2a3 d3a3 g2a2 a3c3 a1b1 c3b4 a2b2 b4e4 b2c2 e4b4 b1a2 e5e4 f3d2 d7e5 d2b3 c5c4 b3c1 b4a5 a2b1
```
Notice no nps is reported, so cutechess calculates it:
https://github.com/cutechess/cutechess/blob/25e6ed6204698932b6db577316d2c088d2a0a0a6/projects/lib/src/moveevaluation.cpp#L156-L160

https://github.com/LeelaChessZero/lc0/blob/9883568795e772a540a8f399971e6d44280e7d1f/src/mcts/search.cc#L151-L152

Mardak:
r?@borg323 or @Tilps  Fix #1316 with a nps cache on `UciLoop`

```
Before:
position fen 6k1/3n2P1/4p3/1pp1p3/r7/P2q1N2/1P4Q1/K6R b - - 2 39
go nodes 2 searchmoves a4a3
info depth 1 seldepth 2 time 1068 nodes 2 score cp -17 nps 25 tbhits 0 pv a4a3 b2a3
bestmove a4a3 ponder b2a3

position fen 6k1/3n2P1/4p3/1pp1p3/r7/P2q1N2/1P4Q1/K6R b - - 2 39 moves a4a3
go nodes 1
info depth 1 seldepth 0 time 0 nodes 1 score cp 17 tbhits 0 pv b2a3
bestmove b2a3


After:
position fen 6k1/3n2P1/4p3/1pp1p3/r7/P2q1N2/1P4Q1/K6R b - - 2 39
go nodes 2 searchmoves a4a3
info depth 1 seldepth 2 time 1071 nodes 2 score cp -17 nps 22 tbhits 0 pv a4a3 b2a3
bestmove a4a3 ponder b2a3

position fen 6k1/3n2P1/4p3/1pp1p3/r7/P2q1N2/1P4Q1/K6R b - - 2 39 moves a4a3
go nodes 1
info depth 1 seldepth 0 time 0 nodes 1 score cp 17 nps 22 tbhits 0 pv b2a3
bestmove b2a3
```

### DIFF
--- a/src/chess/uciloop.cc
+++ b/src/chess/uciloop.cc
@@ -259,7 +259,11 @@ void UciLoop::SendInfo(const std::vector<ThinkingInfo>& infos) {
              std::to_string(info.wdl->d) + " " + std::to_string(info.wdl->l);
     }
     if (info.hashfull >= 0) res += " hashfull " + std::to_string(info.hashfull);
-    if (info.nps >= 0) res += " nps " + std::to_string(info.nps);
+    if (info.nodes >= 0) {
+      // Always report nps if we report nodes defaulting to the last reported.
+      if (info.nps >= 0) last_nps_ = info.nps;
+      res += " nps " + std::to_string(last_nps_);
+    }
     if (info.tb_hits >= 0) res += " tbhits " + std::to_string(info.tb_hits);
     if (info.multipv >= 0) res += " multipv " + std::to_string(info.multipv);
 
--- a/src/chess/uciloop.h
+++ b/src/chess/uciloop.h
@@ -89,6 +89,7 @@ class UciLoop {
   bool DispatchCommand(
       const std::string& command,
       const std::unordered_map<std::string, std::string>& params);
+  int last_nps_ = 0;
 };
 
 }  // namespace lczero


borg323:
I don't like keeping nps around in UciLoop just for this. 
What about the alternative of making `if (info.time >= 0) res += " time " + std::to_string(info.time);` use just `>` instead of `>=`? I can't find the code in cutechess that fails, so can't tell if this will work, but seems conceptually better (to me).

Mardak:
The `info.nps` value is default -1, so it's not an issue of reporting `0 nps` as cutechess will display that only if time is also 0:
https://github.com/cutechess/cutechess/blob/25e6ed6204698932b6db577316d2c088d2a0a0a6/projects/lib/src/moveevaluation.cpp#L156-L160

nguyenpham:
UCI is stateless, keeping to report any info from the last calculation is not really right. Furthermore, speed reporting is not a vital factor. I suggest to just ignore or do a simple such as nps = 1 (the logic is a kind of rounding up the elapsed time to 1 s). It is good in the long term.  Cutechess is a live project, just report any problem and they will fix.

Tilps:
UCI might be stateless - but the engine is not. Keeping the nps carrying over is an approximation of the nps that was used to build the tree that is cached and reported in the nodes count.
Not that I think this is the way we want to do it, but I do think it outputs the 'correct' value for the scenario.

mooskagh:
I don't like very much the idea of fixing something that seems like a GUI bug in Lc0, except:
- When fix is trivial (and need for additional variable doesn't make it trivial)
or
- When our engine can crash/lose because of that (like our workaround with starting time from `position` command).

PR #1319, opened 2020-05-27
alternative way to show hidden options

borg323:


src/utils/optionsparser.cc
```diff
@@ -120,6 +120,9 @@ bool OptionsParser::ProcessAllFlags() {
 
 bool OptionsParser::ProcessFlags(const std::vector<std::string>& args) {
   auto show_help = false;
+  if (CommandLine::BinaryName().find("expert") != std::string::npos) {
```
mooskagh, line 123: Feel free also to pick some funnier/mysterious substring for that. :)

fohristiwhirl: "galaxybrain"

borg323: I was leaning towards "pro" for a more ironic approach.

mooskagh: Have no objections to either of options. Or `hackerman` to get "pro" irony even further. :)

Naphthalin: I propose `sudo`, as changing hidden options should require knowledge about the inner workings and be done on your own risk.

PR #1320, opened 2020-05-29
skip building dnnl and opencl if not a release

borg323:


Naphthalin:
Good choice; all fast and PR testers I know of use either cudnn or dx12 backend, especially given that the newer AMD GPUs under windows don't even work with OpenCL.

borg323:
I also kept openblas to runs the tests and android that checks cross compilation and 32 bit builds.

PR #1322, opened 2020-05-30
Remove shortsightedness

Tilps:
It never quite proved itself and it is almost certainly obsolete now that MLH is a real thing.

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -1542,12 +1542,9 @@ void SearchWorker::DoBackupUpdateSingleNode(
       d = n->GetD();
       m = n->GetM();
     }
-    n->FinalizeScoreUpdate(v / (1.0f + params_.GetShortSightedness() * depth),
-                           d, m, node_to_process.multivisit);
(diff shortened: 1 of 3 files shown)


src/mcts/search.cc
```diff
@@ -1542,12 +1542,9 @@ void SearchWorker::DoBackupUpdateSingleNode(
       d = n->GetD();
       m = n->GetM();
     }
-    n->FinalizeScoreUpdate(v / (1.0f + params_.GetShortSightedness() * depth),
```
Mardak, line 1545: Looks like this mostly reverts https://github.com/LeelaChessZero/lc0/commit/c0705b77faa20b042ed8d24a57f63c36f46f80c4 although that commit did add `depth` which is now unused but the optimizer knows to remove it anyway.

I believe the only other code interaction with short sightedness was to account for adjusted terminal win/loss values, e.g., wl > 0 vs wl == 1.

Tilps: Depth removed.

PR #1323, opened 2020-05-30
Calculate verbose M effect with delta-M like PickNode.

Mardak:
r?@AlexisOlson or @Tilps  Following up from #1230. I do see `child_m - parent_m` is redundant if the child is not visited…

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -366,9 +366,9 @@ std::vector<std::string> Search::GetVerboseStats(Node* node) const {
   };
 
   std::vector<std::string> infos;
+  const auto parent_m = node->GetM();
   for (const auto& edge : edges) {
     float Q = edge.GetQ(fpu, draw_score, logit_q);
-    const auto parent_m = node->GetM();
     const auto child_m = edge.GetM(parent_m);
     float M_effect =
         do_moves_left_adjustment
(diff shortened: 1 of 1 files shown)


src/mcts/search.cc
```diff
@@ -368,9 +368,11 @@ std::vector<std::string> Search::GetVerboseStats(Node* node) const {
   std::vector<std::string> infos;
   for (const auto& edge : edges) {
     float Q = edge.GetQ(fpu, draw_score, logit_q);
+    const auto parent_m = node->GetM();
```
Tilps, line 371: move outside the loop?

src/mcts/search.cc
```diff
@@ -368,9 +368,11 @@ std::vector<std::string> Search::GetVerboseStats(Node* node) const {
...
+    const auto child_m = edge.GetM(parent_m);
```
Tilps, line 372: Just in case anyone else is wondering - yes this is wrong, should be parent_m+1, but the same 'error' exists in the pick node code.
Same applies on the line below - child_m - parent_m should really be child_m - parent_m - 1 - but same problem exists in pick node.

src/mcts/search.cc
Mardak: Hoisted.

src/mcts/search.cc
Mardak: I believe you meant the child move default m has an off-by-1  error, but it should be 1 move fewer (not `parent_m+1`). Filed #1324

PR #1324, opened 2020-05-31
Calculate difference of expected M (1 fewer move than parent).

Mardak:
r?@Ttl or @Tilps As noted in https://github.com/LeelaChessZero/lc0/pull/1323#discussion_r432898085 the default M and the matching M delta calculation is off by 1.


Tilps:
This is what I was meaning to ask for yes.  But I'm not sure if we're confident that its better with the bug fixed or not.

Mardak:
Sounds like we'll just keep the off-by-1 calculation.

PR #1326, opened 2020-06-03
minor cleanup

borg323:


borg323:
Yes, I only noticed because of the warning...

PR #1328, opened 2020-06-04
Adjust default MLH settings and unhide options for next release.

AlexisOlson:
Having hidden options that are set to not-recommended defaults is a major usability issue.

The defaults here still have MLH disabled but only require changing two parameters to match the settings recommended by the current !mlh bot command on Discord:

```
MovesLeftMaxEffect 0.15
MovesLeftThreshold 0.3
MovesLeftSlope 0.015
MovesLeftQuadraticFactor 1.0
MovesLeftConstantFactor 0.0
MovesLeftScaledFactor 0.0
```

### DIFF
--- a/src/mcts/params.cc
+++ b/src/mcts/params.cc
@@ -295,6 +295,7 @@ void SearchParams::Populate(OptionsParser* options) {
   options->Add<BoolOption>(kPerPvCountersId) = false;
   std::vector<std::string> score_type = {"centipawn",
                                          "centipawn_with_drawscore",
+                                         "centipawn_2019",
                                          "centipawn_2018",
                                          "win_percentage",
                                          "Q",
--- a/src/mcts/params.cc
+++ b/src/mcts/params.cc
@@ -235,9 +235,6 @@ const OptionId SearchParams::kMovesLeftQuadraticFactorId{
     "moves-left-quadratic-factor", "MovesLeftQuadraticFactor",
     "A factor which is multiplied by the square of Q of parent node and the "
     "base moves left effect."};
-const OptionId SearchParams::kShortSightednessId{
-    "short-sightedness", "ShortSightedness",
-    "Used to focus more on short term gains over long term."};
 const OptionId SearchParams::kDisplayCacheUsageId{
     "display-cache-usage", "DisplayCacheUsage",
     "Display cache fullness through UCI info `hash` section."};
@@ -257,6 +254,11 @@ const OptionId SearchParams::kDrawScoreWhiteId{
 const OptionId SearchParams::kDrawScoreBlackId{
     "draw-score-black", "DrawScoreBlack",
     "Adjustment, added to a draw score of a black player."};
+const OptionId SearchParams::kNpsLimitId{
+    "nps-limit", "NodesPerSecondLimit",
+    "An option to specify an upper limit to the nodes per second searched. The "
+    "accuracy depends on the minibatch size used, increasing for lower sizes, "
+    "and on the length of the search. Zero to disable."};
 
 void SearchParams::Populate(OptionsParser* options) {
   // Here the uci optimized defaults" are set.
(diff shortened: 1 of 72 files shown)


Naphthalin:
The above values were tuned for 0.07 slope for T70, which now uses a sharper slope so whenever this is relevant, policies should already be sharper. I therefore would prefer `slope=0.01` as the default value and not higher. Default threshold should be 0.99 or so so it would be off by default (until terminal nodes), and to activate it we only need to reduce the threshold. Also, I am strongly against unhiding the 3 factors, once we have reasonable default values. These were only intended to adjust the Q curve without having to change the binary.

AlexisOlson:
A different slope is fine. We can adjust to whatever is considered best when it comes time to merge.

I left the threshold at 1.0 because since MLH is disabled by default, the `do_moves_left_adjustment` boolean should be false to prevent unnecessary calculation of `M_effect`. It might make sense to add an `m_cap > 0.0` condition in the [definition](https://github.com/LeelaChessZero/lc0/blob/b1597b51c1b21f93c2ff4febfc1ea6661fab3fe6/src/mcts/search.cc#L295) of `do_moves_left_adjustment`. Then we could set a default threshold of less than 1.0 without triggering `do_moves_left_adjustment` when `m_cap = 0` (and you'd only need to adjust one parameter to use MLH with recommended settings).

Long term, hiding the 3 factors is fine but they should be unhidden as long as there isn't a clear consensus on their settings. It's not clear to me whether or not we have that consensus yet. If there is consensus, I'm fine with keeping them hidden.

mooskagh:
I'm not sure about unhiding part, especially given that there will be an easier way to unhide all hidden params ("pro version").
Actually, I'm more for hiding more parameters that most of people don't touch. There were multiple complaints about that including from top chess players.

AlexisOlson:
Hidden is fine as long as they don't need tweaking for best results.

However, it seems like best results for the factor settings aren't completely settled yet:
https://discord.com/channels/425419482568196106/530486338236055583/719100480885751808

src/mcts/params.cc
```diff
@@ -313,10 +313,10 @@ void SearchParams::Populate(OptionsParser* options) {
   options->Add<ChoiceOption>(kHistoryFillId, history_fill_opt) = "fen_only";
   options->Add<FloatOption>(kMovesLeftMaxEffectId, 0.0f, 1.0f) = 0.0f;
   options->Add<FloatOption>(kMovesLeftThresholdId, 0.0f, 1.0f) = 1.0f;
-  options->Add<FloatOption>(kMovesLeftSlopeId, 0.0f, 1.0f) = 0.001f;
-  options->Add<FloatOption>(kMovesLeftConstantFactorId, -1.0f, 1.0f) = 1.0f;
+  options->Add<FloatOption>(kMovesLeftSlopeId, 0.0f, 1.0f) = 0.015f;
```
Naphthalin, line 316: I would prefer a more conservative value here like `0.005f`; the current TCEC shows that 0.007 with linear factor to be a bit too much for drawish positions around Q=0.4, and this value would reduce it to 1/3 there without influencing the effect on decided games too much.

src/mcts/params.cc
```diff
@@ -313,10 +313,10 @@ void SearchParams::Populate(OptionsParser* options) {
   options->Add<ChoiceOption>(kHistoryFillId, history_fill_opt) = "fen_only";
   options->Add<FloatOption>(kMovesLeftMaxEffectId, 0.0f, 1.0f) = 0.0f;
```
Naphthalin, line 314: I also think we should set `max-effect=0.1` per default, so to activate MLH one only needs to set the threshold and nothing else.

AlexisOlson: Agreed

src/mcts/params.cc
AlexisOlson: Sounds reasonable

Naphthalin:
It seems like you pushed an old branch again, as the MLH options are unhidden again?

AlexisOlson:
@Naphthalin No, I never re-hid them as it hasn't been clear which factor values should be the default. Is there a consensus that quadratic = 1 should be default with other options not directly accessible?

Tilps:
I think we can argue about hide vs unhide while the RC process is going on - most important is the default values are improved.

PR #1330, opened 2020-06-06
Per color settings.

mooskagh:
It's a bit too unwieldy and ugly, but it's the best I could come up with.

So, before:
```
Root_params
  |
  + player1
  |
  + player2
```

After:
```
Root_params
  |
  + player1
  |       |
  |       + white
  |       |
  |       + black
  |       
  + player2
  |       |
  |       + white
  |       |
  |       + black
  |
  + white
  |
  + black
```

`root_params.white` and `root_params.black` are "aliases" (maybe a poor term) for `root_params.player*.white` (and `root_params.player*.black`).

When you e.g. access the parameter `player1.black.backend`, it's looked up in the following order:
1. `(default.root.)player1.black.backend` (self)
2. `(default.root.)black.backend` (alias) Note that parents of the alias are not looked up.
3. `(default.root.)player1.backend` (parent of the self)
4. `(default.root.)backend` (parent of parent of self)
4. `(default.)backend` (parent of parent of parent of self)

### DIFF
--- a/src/selfplay/tournament.cc
+++ b/src/selfplay/tournament.cc
@@ -140,10 +140,10 @@ SelfPlayTournament::SelfPlayTournament(
     CallbackUciResponder::BestMoveCallback best_move_info,
     CallbackUciResponder::ThinkingCallback thinking_info,
     GameInfo::Callback game_info, TournamentInfo::Callback tournament_info)
-    : player_options_{{options.GetSubdict("player1.white"),
-                       options.GetSubdict("player1.black")},
-                      {options.GetSubdict("player2.white"),
-                       options.GetSubdict("player2.black")}},
+    : player_options_{{options.GetSubdict("player1").GetSubdict("white"),
+                       options.GetSubdict("player1").GetSubdict("black")},
+                      {options.GetSubdict("player2").GetSubdict("white"),
+                       options.GetSubdict("player2").GetSubdict("black")}},
       best_move_callback_(best_move_info),
       info_callback_(thinking_info),
       game_callback_(game_info),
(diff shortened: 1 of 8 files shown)


src/utils/optionsdict.h
```diff
@@ -181,27 +202,49 @@ template <typename T>
...
+  return {};
+}
+template <typename T>
+std::optional<T> OptionsDict::OwnGet(const OptionId& option_id) const {
+  return OwnGet<T>(GetOptionId(option_id));
+}
+
+template <typename T>
+bool OptionsDict::Exists(const std::string& key) const {
+  if (OwnExists<T>(key)) return true;
```
Tilps, line 221: aliases contains this - so this is redundant right?

src/utils/optionsdict.h
```diff
@@ -144,6 +160,9 @@ class OptionsDict : TypeDict<bool>,
   // Returns list of subdictionaries.
   std::vector<std::string> ListSubdicts() const;
 
+  // Adds alias dictionary.
+  void AddAliasDict(const OptionsDict* dict);
```
Tilps, line 164: This (trivial?) methods implementation oddly seems missing?

src/selfplay/tournament.cc
```diff
@@ -209,10 +218,10 @@ void SelfPlayTournament::PlayOneGame(int game_number) {
     Mutex::Lock lock(mutex_);
     player1_black = ((game_number % 2) == 1) != first_game_black_;
     if (!openings_.empty()) {
-      if (player_options_[0].Get<bool>(kOpeningsMirroredId)) {
+      if (player_options_[0][!player1_black].Get<bool>(kOpeningsMirroredId)) {
```
Tilps, line 221: I don't think these openings flags make sense to try and lookup a specific sub dictionary.  Much like below where you did [0][0] for an openings flag, I think the same could be done for this one, and for the 'random' check immediately below.

src/utils/optionsdict.h
mooskagh: It's there in `optionsdict.cc`.

src/utils/optionsdict.h
mooskagh: Indeed, removed.

src/selfplay/tournament.cc
mooskagh: Done.

Tilps:
all good, I guess we should have someone test it works with current training parameters like it used to :)

PR #1332, opened 2020-06-07
Use info string to display some important information

nguyenpham:
This PR is a re-implementation of canceled PR #1302.

Whenever there is a UCI loop, it uses command "info string" to display information about numbers of Sygyzy endgames and the network file. Otherwise, it uses cerr to display that information as before.

The technique uses std::function as the suggestion by mooskagh.

### DIFF
--- a/src/chess/uciloop.cc
+++ b/src/chess/uciloop.cc
@@ -129,6 +129,12 @@ bool ContainsKey(const std::unordered_map<std::string, std::string>& params,
 }  // namespace
 
 void UciLoop::RunLoop() {
+  LogInfo::logFunc = [=](const std::string& msg) {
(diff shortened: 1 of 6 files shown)


fohristiwhirl:
Don't change the grammar of "Loading weights file" please, it's correct already.

nguyenpham:
"Loading" shows action in progress. It is suitable if it is displayed only once and then be scrolled up. However, when sending the info to a chess GUI in the command "info string" it may be displayed in a special place, e.g. a panel for string info. Users can re-read it many times. Clearly it is not the progressing action later, thus the verb-ing is not suitable.

nguyenpham:
Too long to exist

PR #1333, opened 2020-06-07
Add UCI `movesleft` info.

mooskagh:


src/mcts/search.cc
```diff
@@ -189,6 +189,10 @@ void Search::SendUciInfo() REQUIRES(nodes_mutex_) {
     // Using 1000-w-l instead of 1000*d for D score so that W+D+L add up to
     // 1000.0.
     uci_info.wdl = ThinkingInfo::WDL{w, 1000 - w - l, l};
+    if (network_->GetCapabilities().has_mlh()) {
+      uci_info.moves_left = static_cast<int>(
```
Tilps, line 193: Should this be rounded before casting?

Tilps: Nevermind I think the +1 is doing that job...

PR #1337, opened 2020-06-10
Unhide MLH parameters

AlexisOlson:


mooskagh:
I'm not sure it's right to expose all those parameters most of people will never know what they mean. I'd actually prefer to hide more parameters than we currently have.

We currently have "pro" mode in master (rename `lc0.exe` to `lc0-pro.exe` to unhide all parameters), I think that's good enough.

PR #1338, opened 2020-06-11
change default directories

borg323:
Does the lc0 part of #976.

### DIFF
--- a/src/neural/opencl/network_opencl.cc
+++ b/src/neural/opencl/network_opencl.cc
@@ -250,6 +250,7 @@ class OpenCLNetwork : public Network {
       std::string user_cache_path = GetUserCacheDirectory();
       if (!user_cache_path.empty()) {
         user_cache_path += "lc0/";
+        CreateDirectory(user_cache_path);
       }
       params_.tuner_file = user_cache_path + "leelaz_opencl_tuning";
     } else {
--- a/src/neural/opencl/network_opencl.cc
+++ b/src/neural/opencl/network_opencl.cc
@@ -32,6 +32,7 @@
 #include "neural/shared/winograd_filter.h"
 #include "utils/bititer.h"
 #include "utils/exception.h"
+#include "utils/filesystem.h"
 #include "utils/logging.h"
 #include "utils/weights_adapter.h"
 
@@ -245,8 +246,11 @@ class OpenCLNetwork : public Network {
     params_.tune_only = options.GetOrDefault<bool>("tune_only", false);
     params_.tune_exhaustive =
         options.GetOrDefault<bool>("tune_exhaustive", false);
-    params_.tuner_file =
-        options.GetOrDefault<std::string>("tuner_file", "leelaz_opencl_tuning");
+    if (options.IsDefault<std::string>("tuner_file")) {
+      params_.tuner_file = GetUserCacheDirectory() + "lc0/leelaz_opencl_tuning";
+    } else {
+      params_.tuner_file = options.Get<std::string>("tuner_file");
+    }
 
     wdl_ = file.format().network_format().output() ==
            pblczero::NetworkFormat::OUTPUT_WDL;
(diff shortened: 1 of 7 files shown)


borg323:
What this does:
1. For network file search with `<autodiscover>`, the following directories are searched in order:
- `.` and `networks`
- `lc0` and `lc0/networks` under `$XDG_DATA_HOME` (or `~/.share/local` if not set)
- `lc0` and `lc0/networks` under every path in `$XDG_DATA_DIRS` (or `/usr/local/share/:/usr/share/` if not set)
2. For the config file by default the following directories are searched in order for `lc0.config`:
- `.`
- `lc0` under `$XDG_CONFIG_HOME` (or `~/.config` if not set)
- `lc0` under every path in `$XDG_CONFIG_DIRS` (or ` /etc/xdg` if not set)
Note that currently the help text shows `lc0.config` but `<default>` is used internally. This is not ideal, maybe we should change the help text to `<search for lc0.config>`.
3. Training game generation is done in `lc0/data-<random>` under `$XDG_CACHE_HOME` (or `~/.cache` if not set).

4. The `leelaz_opencl_tuning` file is searched and saved in `lc0` under `$XDG_CACHE_HOME` (or `~/.cache` if not set), with an already existing backend option to override.

mooskagh:
That will probably mean that older/newer client won't be compatible with newer/older lc0?
Because lc0 writes files not where the client expects them.
Probably not a problem, but would be good to know whether that's the case.

borg323:
>  Because lc0 writes files not where the client expects them.

The client gets the path by parsing lc0 output, this works fine (tested). The opposite may not be true, with new client running old lc0 from the path, but this is a new use case, so requiring new lc0 seems reasonable.

src/neural/opencl/network_opencl.cc
```diff
@@ -245,8 +246,11 @@ class OpenCLNetwork : public Network {
     params_.tune_only = options.GetOrDefault<bool>("tune_only", false);
     params_.tune_exhaustive =
         options.GetOrDefault<bool>("tune_exhaustive", false);
-    params_.tuner_file =
-        options.GetOrDefault<std::string>("tuner_file", "leelaz_opencl_tuning");
+    if (options.IsDefault<std::string>("tuner_file")) {
+      params_.tuner_file = GetUserCacheDirectory() + "lc0/leelaz_opencl_tuning";
```
Tilps, line 250: This means multiple installs of lc0 get the same tuner file unless you use backend-opts to override?
I wonder if that will ever cause us pain...

Tilps: hmm only on linux it seems - windows this is empty string, so maybe I am fine with that :P

src/neural/writer.cc
```diff
@@ -36,9 +36,20 @@
...
+}
+
+}  // namespace
+
 TrainingDataWriter::TrainingDataWriter(int game_id) {
   static std::string directory =
-      GetUserCacheDirectory() + "lc0/data-" + Random::Get().GetString(12);
+      GetLc0CacheDirectory() + "data-" + Random::Get().GetString(12);
   // It's fine if it already exists.
   CreateDirectory(directory.c_str());
```
Tilps, line 54: This needs to be recursive create.

src/neural/opencl/network_opencl.cc
```diff
@@ -247,7 +247,11 @@ class OpenCLNetwork : public Network {
     params_.tune_exhaustive =
         options.GetOrDefault<bool>("tune_exhaustive", false);
     if (options.IsDefault<std::string>("tuner_file")) {
-      params_.tuner_file = GetUserCacheDirectory() + "lc0/leelaz_opencl_tuning";
+      std::string user_cache_path = GetUserCacheDirectory();
+      if (!user_cache_path.empty()) {
+        user_cache_path += "lc0/";
+      }
+      params_.tuner_file = user_cache_path + "leelaz_opencl_tuning";
```
Tilps, line 254: If user_cache_path is not empty - it needs a CreateDirectory before this.

PR #1340, opened 2020-06-15
Add support for canonical_hectoplies and canonical_hectoplies_armageddon

Tilps:
Just to get the code written to show I'm ready for it to be in 0.26...

I'll do a cleanup to introduce some helper functions - just need to decide where to put them...


Tilps:
Added helper functions.  Do need to test them sometime ;)

Issue #1313, opened 2020-05-25
Number of visits goes above 4000000000


mooskagh:
Link to discussion: https://discord.com/channels/425419482568196106/425419999096733706/714373730587115531

Naphthalin:
#1342 was merged -- can this be closed then?

PR #1341, opened 2020-06-16
Fix the issue Number of visits goes above 4000000000

nguyenpham:
Fix the issue Number of visits goes above 4000000000 #1313

This PR is a replacement for PR #1339, closed by accidentally deleting the commit.


nguyenpham:
Closed because pushing from master

Issue #706, opened 2019-01-26
Uses of exit instead of Exception


cn4750:
The following code uses `exit` instead of the more flexible [Exception](https://github.com/LeelaChessZero/lc0/blob/407be4d0bd480cb1a56e8bd86dcc3d5d8d58c6e7/src/utils/exception.h#L36):

1. [syzygy.cc#L1049](https://github.com/LeelaChessZero/lc0/blob/master/src/syzygy/syzygy.cc#L1049)
2. [syzygy.cc#L1063](https://github.com/LeelaChessZero/lc0/blob/master/src/syzygy/syzygy.cc#L1063)
3. [syzygy.cc#L1070](https://github.com/LeelaChessZero/lc0/blob/master/src/syzygy/syzygy.cc#L1070)

Naphthalin:
This still applies, right? If yes, a PR should be created, if not, we can close this issue.

mooskagh:
Fixed by #1343

PR #1343, opened 2020-06-16
Uses of exit instead of Exception

nguyenpham:
Fix issues #706 (Uses of exit instead of Exception)

### DIFF
--- a/src/syzygy/syzygy.cc
+++ b/src/syzygy/syzygy.cc
@@ -1070,7 +1070,7 @@ class SyzygyTablebaseImpl {
     *mapping = mmap;
     base_address = MapViewOfFile(mmap, FILE_MAP_READ, 0, 0, 0);
     if (!base_address) {
-      throw Exception("MapViewOfFile() failed, name = " + fname + ", error = " + GetLastError());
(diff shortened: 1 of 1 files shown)


src/syzygy/syzygy.cc
```diff
@@ -1066,15 +1065,12 @@ class SyzygyTablebaseImpl {
...
-      exit(1);
+      throw Exception("CreateFileMapping() failed");
     }
     *mapping = mmap;
     base_address = MapViewOfFile(mmap, FILE_MAP_READ, 0, 0, 0);
     if (!base_address) {
-      CERR << "MapViewOfFile() failed, name = " << fname
-           << ", error = " << GetLastError();
-      exit(1);
+      throw Exception("MapViewOfFile() failed, name = " + fname + ", error = " + GetLastError());
```
mooskagh, line 1073: This doesn't compile for windows, as it's not possible to add number to the string.
It should be `std::to_string(GetLastError())`.

nickodell: @mooskagh Does 213cf4d fix it?

mooskagh (approved):

Issue #705, opened 2019-01-26
Uses of runtime_error instead of Exception


cn4750:
The following code uses `std::runtime_error` instead of the more flexible [Exception](https://github.com/LeelaChessZero/lc0/blob/407be4d0bd480cb1a56e8bd86dcc3d5d8d58c6e7/src/utils/exception.h#L36):

1. [OpenCL.cc#L302](https://github.com/LeelaChessZero/lc0/blob/98a80caac1a29640dfc640f7c0e42f10ca1a5f36/src/neural/opencl/OpenCL.cc#L302)
2. [OpenCL.cc#L315](https://github.com/LeelaChessZero/lc0/blob/98a80caac1a29640dfc640f7c0e42f10ca1a5f36/src/neural/opencl/OpenCL.cc#L315)
3. [OpenCL.cc#L331](https://github.com/LeelaChessZero/lc0/blob/98a80caac1a29640dfc640f7c0e42f10ca1a5f36/src/neural/opencl/OpenCL.cc#L331)
4. [OpenCL.cc#L348](https://github.com/LeelaChessZero/lc0/blob/98a80caac1a29640dfc640f7c0e42f10ca1a5f36/src/neural/opencl/OpenCL.cc#L348)
5. [OpenCLTuner.cc#L361](https://github.com/LeelaChessZero/lc0/blob/e4cfc8729403a5f38526fb8ec2e5316f8901cdf7/src/neural/opencl/OpenCLTuner.cc#L361)

Naphthalin:
This still applies, right? Is it still an issue? If yes, we would need PR; if not, this issue could be closed.

Issue #657, opened 2019-01-04
Add additional permission to GPL in opencl and blas directories


borg323:
Following the recent change to Leela zero licensing (https://github.com/gcp/leela-zero/commit/808bb43df34654d357be2dd278eba19c03f07094) we can similarly update the license in the opencl, blas and shared directories that reuse Leela zero code. This will allow the distribution of all backends compiled in the same executable. The additional permission added by Leela zero is:
```
    Additional permission under GNU GPL version 3 section 7

    If you modify this Program, or any covered work, by linking or
    combining it with NVIDIA Corporation's libraries from the
    NVIDIA CUDA Toolkit and/or the NVIDIA CUDA Deep Neural
    Network library and/or the NVIDIA TensorRT inference library
    (or a modified version of those libraries), containing parts covered
    by the terms of the respective license agreement, the licensors of
    this Program grant you additional permission to convey the resulting
    work.
```
This is compatible with the current license but also adds permission to link against TensorRT. To add this we need permission from all contributors in the code in these directories (in alphabetical order):
@borg323 (OK for all contributions)
@dubslow 
@frpays 
@mooskagh 
@ra1u 
@ttl

Please add a comment saying whether you **agree or not with this change**, and whether you **agree to extend this to all your contributions** in case we want to add a TensorRT backend in the future. Also please indicate if I missed any contributors. I will keep this list updated with the latest information.

ra1u:
Ok for all contributions.

ra1u:
Just to be clear, can someone explain in simple terms how  line "the licensors of this Program grant you additional permission to convey the resulting work."  is to be interpreted?

borg323:
I think the explanation in the GPL FAQ is the best we can hope for: <https://www.gnu.org/licenses/gpl-faq.en.html#GPLIncompatibleLibs>

frpays:
Ok for all contributions.

mooskagh:
Ok for all contributions!

Naphthalin:
Any update on this? @dubslow doesn't seem to have responded.

borg323:
This needs a careful code review, what we have so far may be sufficient.

nguyenpham:
I am always a bit scared whenever we do modify default licenses since that doesn’t look as safe and simple as original ones. However, this project is an open source, everything is clear and only for communities. Thus I support things which may help more for communities:
OK for all contributions

borg323:
[...]

PR #1344, opened 2020-06-16
Uses of runtime_error instead of Exception

nguyenpham:
Fix for issue #705

Even OpenCL is being abandoned its issues should be fixed in cases we need later and for closing issues.


borg323:
@nguyenpham can you state you opinion in #657?

nguyenpham:
Too long to exist

PR #1347, opened 2020-06-17
Display board

nguyenpham:
Add new command "d" to display the current chess board. It is a convenient function for developers.

Discussion: This implementation uses an existent function DebugString() of ChessBoard to display the board in the text form. We may improve later for showing more pretty as well as having more info such as endgame probe result.

Screen Shot 2020-06-19 at 7 21 00 pm


nguyenpham:
Too long to exist

PR #1351, opened 2020-06-19
remove blas_cores backend option

borg323:
I haven't found a configuration where a value more than 1 is generally useful, while the confusion created has been considerable.


cn4750:
I'm fine with this.

PR #1352, opened 2020-06-19
update to dnnl 1.5.0

borg323:


Tilps:
Number is bigger, must be better?

borg323:
It is measurably faster than 1.1.1, the version we were using so far.

Issue #831, opened 2019-04-12
Ignore restricted files while scanning for network file


mooskagh:
When people run lc0.exe from `C:\`, it fails to start because it tires to open hiberfill.sys, see https://groups.google.com/d/topic/lczero/I_T-2dVlYbo/discussion.

![image](https://user-images.githubusercontent.com/11266455/56019889-1ed56800-5d06-11e9-8f95-b5b251faa0d1.png)

We should ignore files that we cannot open (can be also e.g. missing read permissions).

Naphthalin:
Is that still an issue now, or was it fixed accidentally by something else? For me it tries access `pagefile.sys` instead.

mooskagh:
> Is that still an issue now, or was it fixed accidentally by something else? For me it tries access `pagefile.sys` instead.

tries to access and returns an error? Then it's still an issue.
In Linux, I'm not able to reproduce similar bug.


So, still a bug, easy to fix, not very high priority, good first issue.

AlexisOlson:
This appears to be resolved by #1353.

PR #1353, opened 2020-06-20
Fixed issue 'Cannot stat file'

nguyenpham:
Fixed the issue #831
Fixed by returning 0 instead of throwing the exception. Two involved functions are GetFileSize and GetFileTime. All code which called those functions can work well with returns of zero.


borg323:
You also need to update the comments in `filesystem.h`.

PR #1354, opened 2020-06-21
Unique Names For Training Games

### LINKED ISSUE #1226: Feature Request: Selfplay training games generated with unique names
Since all games generated in training of selfplay mode start with `game_000000.gz` you can't easily mix the games from multiple runs due to name collisions. Add the same randomness from the random folder name to the file names, e.g. `game_gibberish_000000.gz`.

nguyenpham:
Fix #1226

### DIFF
--- a/src/neural/writer.cc
+++ b/src/neural/writer.cc
@@ -37,13 +37,14 @@
 namespace lczero {
 
 TrainingDataWriter::TrainingDataWriter(int game_id) {
-  static std::string directory =
-      CommandLine::BinaryDirectory() + "/data-" + Random::Get().GetString(12);
+  static const std::string randomString = Random::Get().GetString(12);
+  static const std::string directory =
+      CommandLine::BinaryDirectory() + "/data-" + randomString;
   // It's fine if it already exists.
   CreateDirectory(directory.c_str());
 
   std::ostringstream oss;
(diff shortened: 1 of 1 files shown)


Tilps:
Does this break the client?
(I know it'll break some of my scripts :P)

nguyenpham:
I have taken a look at the client code. The code to take the game filename:
```
file := line[idx1+13 : idx2-1]
```
It takes whatever between two keywords `trainingfile` and `gameid`, regardless of the format nor length of the name. Thus I believe (hopefully ;) ) it may work.

Tilps:
I was doing a bit more thinking about this - and I think it would be good to have an option to specify the training file output prefix - if not specified then we can add this randomness possibly.
The reason I suggest this was that the issue suggests this is useful because it was 'games from different runs get confusing' - but a given run is going to contain games from multiple executions of lc0 selfplay, its very easy to get confused as to which run the multiple executions belong to.

borg323:
Also note that the location we write has changed (at least for linux and mac) see #1338 and #976.

nguyenpham:
Can you suggest how to setup prefix for game training? Command param? E.g. —outputprefix=mygame_ ?

Some code should be changed, we should consider benefits vs complexity.

The PR code modified only file name without touching to folder name thus I think it won’t affect the PRs you mentioned

nguyenpham:
Too long to exist

PR #1355, opened 2020-06-21
selfplay logging

### LINKED ISSUE #1225: Feature Request: Add logging support for selfplay
There doesn't seem to be any support for a log file when in selfplay mode. In Windows it is not possible to see all errors without logging. Equivalent support for a logfile option to that of uci mode is desired.

nguyenpham:
Fix #1225

### DIFF
--- a/src/selfplay/tournament.cc
+++ b/src/selfplay/tournament.cc
@@ -32,6 +32,7 @@
 #include "mcts/stoppers/factory.h"
 #include "neural/factory.h"
 #include "selfplay/game.h"
+#include "utils/optionsparser.h"
 #include "utils/random.h"
 
 namespace lczero {
--- a/src/selfplay/tournament.cc
+++ b/src/selfplay/tournament.cc
@@ -32,7 +32,6 @@
 #include "mcts/stoppers/factory.h"
 #include "neural/factory.h"
 #include "selfplay/game.h"
-#include "utils/optionsparser.h"
 #include "utils/random.h"
 
 namespace lczero {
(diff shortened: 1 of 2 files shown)


Tilps:
I think this change more properly belongs in SelfPlayLoop (loop.cc) rather than in Tournament.

nguyenpham:
The logging function is in SelfPlayTournament which belongs only to SelfPlayLoop. Almost all other options such as Threads, Playout, Visits... are coded in that class. Thus, the logging option should be stored there for having same logic and be easier to maintain.

Tilps:
Logging functions can happen anywhere at any time - so the most important thing is for it to be set ASAP - which means immediately after command line parsing before anything can potentially go wrong and mean that the logging doesn't get configured.

nguyenpham:
Good idea. Fixing!

src/selfplay/tournament.cc
```diff
@@ -32,7 +32,6 @@
 #include "mcts/stoppers/factory.h"
 #include "neural/factory.h"
 #include "selfplay/game.h"
-#include "utils/optionsparser.h"
```
Tilps, line 35: Not sure why this was removed.

nguyenpham: I removed it by accident (though I have added it before) and it was still compiled well. I have checked, that including file is in the file tournament.h already - thus the removing is actually a kind of redundancy

Tilps: It can be good style to include headers you use rather than relying on transitive includes, although in this case since its already in the 'primary' header its less of an issue.
However my main recommendation for reverting is that the PR is cleaner if it doesn't contain extraneous file changes.

nguyenpham: Added it back.

PR #1356, opened 2020-06-21
Hacky backend to record/replay another backend for performance testing of search

Tilps:
needs more work - at least to do a better job of recording the output faster. At the moment I haven't gone over 700nps during recording. Maybe pushing them to a writing thread queue to stream more effectively. 
The idea is for this to surplant the need for random backend for performance test.
Probably need to do a better job in replay performance for that to be realistic, right now too - I've not gotten over 40k nps...


Tilps:
recording and replaying with prefetch disabled I get over 160knps - so it doesn't seem the replay backend is a bottleneck.

Either prefetch itself is - or (maybe more likely) the fact that the request pattern is not consistent when prefetch is enabled is causing too much tree structure corruption which in turn makes prefetching worse.  I did say it was a hack - and the hack can't handle the difference in request patterns between a prefetched node and a normal node. ;)

Tilps:
Seems to be prefetch itself.  I created a version that captures all 1861 values during ComputeBlocking rather than only capturing as requested (35 times larger disk storage required) - and playback performance difference was identical pattern of 4 times faster with prefetch disabled.

Tilps:
Ahh possibly I'm putting too much on prefetch there - I mistaken had run all my tests with max batch size of 1.
which was also why I was only getting 700nps recording.  I can get over 13k recording with a 30b 2080TI now.  Seems recording is not significant overhead for a 30b on 2080TI :P

Tilps:
Just to confirm - I've not seen any clear nps issues related to prefetch once I disabled my minibatch size override.
And I've seen over 200k nps peak with replay backend even with several million positions in the dictionary.

src/neural/network_rr.cc
```diff
@@ -91,5 +93,197 @@ std::unique_ptr<Network> MakeRoundRobinNetwork(
 
 REGISTER_NETWORK("roundrobin", MakeRoundRobinNetwork, -999)
 
+class RecordComputation : public NetworkComputation {
```
borg323, line 96: This should go to a separate file, say network_record.cc

src/neural/network_rr.cc
```diff
@@ -91,5 +93,197 @@ std::unique_ptr<Network> MakeRoundRobinNetwork(
...
+  float GetPVal(int sample, int move_id) const override {
+    return Capture(inner_->GetPVal(sample, move_id), sample);
+  }
+  float GetMVal(int sample) const override {
+    return Capture(inner_->GetMVal(sample), sample);
+  }
+  virtual ~RecordComputation() {
+    Mutex::Lock lock(mutex_);
+    std::fstream output(record_file_, std::ios::app | std::ios_base::binary);
+    for (int i = 0; i < hashes_.size(); i++) {
```
borg323, line 143: Change to `size_t i` to avoid warning.

src/neural/network_rr.cc
```diff
@@ -91,5 +93,197 @@ std::unique_ptr<Network> MakeRoundRobinNetwork(
...
+  void ComputeBlocking() override {}
+  // Returns how many times AddInput() was called.
+  int GetBatchSize() const override { return hashes_.size(); }
+  float Replay(int index) const {
+    const auto& entry_ptr = lookup_->find(hashes_[index]);
+    if (entry_ptr == lookup_->end()) {
+      return 0.0f;
+    }
+    const auto& entry = entry_ptr->second;
+    int counter = replay_counter_[index];
```
borg323, line 183: Make `unsigned` to fix warning.

src/neural/network_rr.cc
```diff
@@ -91,5 +93,197 @@ std::unique_ptr<Network> MakeRoundRobinNetwork(
...
+      return 0.0f;
+    }
+    replay_counter_[index]++;
+    return entry[counter];
+  }
+  // Returns Q value of @sample.
+  float GetQVal(int sample) const override { return Replay(sample); }
+  float GetDVal(int sample) const override { return Replay(sample); }
+  // Returns P value @move_id of @sample.
+  float GetPVal(int sample, int move_id) const override {
```
borg323, line 194: Here `move_id` is unused.

Tilps: Done.

src/neural/network_rr.cc
Tilps: Made replay_counter_ size_t and changed.

src/neural/network_rr.cc
Tilps: Done.

src/neural/network_rr.cc
Tilps: Done.

PR #1357, opened 2020-06-21
Show Training In Progress

### LINKED ISSUE #783: Unclear if games are being produced until they are complete
As the title indicates, upon running client.exe, there is no indication whatsoever as to whether the games have actually started being produced or not. The only time you would realize this is if a game has completed. 

Of course, if the games complete relatively quickly, this is not an issue but on my computer, it took about 15 minutes for one game to complete so I must have closed and opened the client file at least 5 times before I finally ended up letting it sit long enough for the game to finish.

Thus, my suggestion is essentially this:
For the first game run on a given client, when the game starts being produced, print something to the console that indicates this, i.e., "Running..." or "Starting game creation... Please be patient" etc

nguyenpham:
Fix #783

Interval displaying "Game playing" is about 20s.

Screen Shot 2020-06-21 at 8 38 24 pm

### DIFF
--- a/src/selfplay/tournament.cc
+++ b/src/selfplay/tournament.cc
@@ -203,6 +203,8 @@ SelfPlayTournament::SelfPlayTournament(
 }
 
 void SelfPlayTournament::PlayOneGame(int game_number) {
+  ShowWorkInProcess(WorkInProgressType::newgame, game_number);
+
   bool player1_black;  // Whether player1 will player as black in this game.
   Opening opening;
   {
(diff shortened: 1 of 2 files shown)


Tilps:
All output of lc0 is filtered by the client code itself. So I am not sure if this will address 783's original use case or not.
In any case - directly handling output rather than piping it through the standard mechanisms we currently use for other outputs means you risk interleaving your output with the standard outputs - which will break the client at random.

Also I would like to see this be optional. Even just printing the game ids is too much extra spam for many use cases for myself.

mooskagh:
Probably showing some stats (number of games, moves, nps, etc) every 5 seconds would be fine?
Without spamming for every game start/stop.

And yeah, need to ensure that it gets through the client.

nguyenpham:
Too long to exist

PR #1358, opened 2020-06-21
Replay backend robust

Tilps:
A variant of record replay backend that trades massive increase in disk storage requirement with not being sensitive to whether prefetch is enabled.


Tilps:
I have a different idea for robustness that isn't so crazy expensive - so closing this for now even if I'm not going to implement it immediately.

PR #1360, opened 2020-06-22
Solid tree

Tilps:
There are multiple known bugs as of right now - but surprisingly I haven't managed to crash it in over 4 moves of trying!

So I'm sharing it early.


Tilps:
Note this is branched on top of network_recordreplay since I needed that to performance test it properly.

Tilps:
Apparently this gives a 25% increase in nps after 2 minutes of searching from start position on a noob 4xv100 machine that was the passive mirror for the actual tcec machine. And the rate the nps was decaying was still slower.
So seems very good for high gpu to cpu ratio scenarios.

On the other hand I've not managed to complete a full game without crash yet - so more debugging is required before this is ready for review.

Tilps:
This is still not stable with the last fix - but its better :P

Tilps:
seems stable finally.

src/mcts/node.cc
```diff
@@ -215,10 +232,67 @@ std::string Node::DebugString() const {
...
+      << " Solid:" << static_cast<bool>(solid_children_);
   return oss.str();
 }
 
+bool Node::MakeSolid() {
+  if (solid_children_ || num_edges_ == 0 || IsTerminal()) return false;
+  // Can only make solid if no immediate leaf childredn are in flight since we
+  // allow the search code to hold references to leaf nodes across locks.
+  Node* old_child_to_check = child_.get();
+  int total_in_flight = 0;
```
borg323, line 245: `unsigned int total_in_flight = 0;`

src/mcts/node.cc
```diff
@@ -74,13 +75,24 @@ class NodeGarbageCollector {
...
         Mutex::Lock lock(gc_mutex_);
         if (subtrees_to_gc_.empty()) return;
         node_to_gc = std::move(subtrees_to_gc_.back());
         subtrees_to_gc_.pop_back();
+        solid_size = subtrees_is_solid_.back();
+        subtrees_is_solid_.pop_back();
+      }
+      // Solid is a hack...
+      if (solid_size != 0) {
+        for (int i = 0; i < solid_size; i++) {
```
borg323, line 91: `for (size_t i = 0; i < solid_size; i++) {`

src/mcts/search.cc
```diff
@@ -1533,6 +1542,7 @@ void SearchWorker::DoBackupUpdateSingleNode(
   float v_delta = 0.0f;
   float d_delta = 0.0f;
   float m_delta = 0.0f;
+  int solid_threshold = params_.GetSolidTreeThreshold();
```
borg323, line 1545: `unsigned int solid_threshold = params_.GetSolidTreeThreshold();`

src/mcts/node.cc
Tilps: Done.

src/mcts/node.cc
Tilps: Done.

src/mcts/search.cc

src/mcts/node.cc
```diff
@@ -215,10 +232,67 @@ std::string Node::DebugString() const {
...
+    gNodeGc.AddToGcQueue(std::move(old_child));
+    Node* to_update_parent = new_children[index].child_.get();
+    if (!new_children[index].solid_children_) {
+      while (to_update_parent != nullptr) {
+        to_update_parent->parent_ = &new_children[index];
+        to_update_parent = to_update_parent->sibling_.get();
+      }
+    } else {
+      for (int i = 0; i < new_children[index].num_edges_; i++) {
+        to_update_parent[i].parent_ = &new_children[index];
```
Mardak, line 284: The name `to_update_parent` was a bit confusing initially, so a comment about fixing references to the solidified child could help or maybe name it `grandchild` ?

(If only there was some node iterator that could handle solid children or sibling chained 😜😜😜 )

src/mcts/node.h
```diff
@@ -220,15 +226,12 @@ class Node {
   ConstIterator Edges() const;
   Iterator Edges();
 
-  class NodeRange;
-  // Returns range for iterating over nodes. Note that there may be edges
-  // without nodes, which will be skipped by this iteration.
-  NodeRange ChildNodes() const;
```
Mardak, line 226: Just documenting some history that this was added in #145 and the last use was removed in #251.

And apparently imaginary friends are okay 😉 
https://github.com/LeelaChessZero/lc0/blob/b4f1dbc60a2befe4e56384441ac4a91e2972354b/src/mcts/node.h#L313-L314
(`Node_Iterator` is gone and `Edge` doesn't actually need to be a friend.)

src/mcts/node.cc
```diff
@@ -215,10 +232,67 @@ std::string Node::DebugString() const {
       << " WL:" << wl_ << " N:" << n_ << " N_:" << n_in_flight_
       << " Edges:" << static_cast<int>(num_edges_)
       << " Bounds:" << static_cast<int>(lower_bound_) - 2 << ","
-      << static_cast<int>(upper_bound_) - 2;
+      << static_cast<int>(upper_bound_) - 2
+      << " Solid:" << static_cast<bool>(solid_children_);
```
Mardak, line 236: nit: no need to cast as it's already a `bool`. (The others need cast from enum or avoiding being treated as `char`)

src/mcts/node.h
```diff
@@ -305,6 +326,8 @@ class Node {
   // Best and worst result for this node.
   GameResult lower_bound_ : 2;
   GameResult upper_bound_ : 2;
+  // Whether the child_ is actually an array of equal length to edges.
+  bool solid_children_ : 1;
```
Mardak, line 330: Definitely for later, but any thoughts on what else can be done when all children are visited? E.g., `GetFpu` is unnecessary saving a `sqrt` for reduction.

src/mcts/node.h
```diff
@@ -240,6 +243,22 @@ class Node {
   // Debug information about the node.
   std::string DebugString() const;
 
+  // Reallocates this nodes children to be in a solid block, if possible and not already done.
+  // Returns true if the transformation was performed.
+  bool MakeSolid();
```
Mardak, line 248: The comment at the top should be updated too:
https://github.com/LeelaChessZero/lc0/blob/b4f1dbc60a2befe4e56384441ac4a91e2972354b/src/mcts/node.h#L51-L52

Bonus points for the diagram too! 😛

src/mcts/search.cc
```diff
@@ -1547,6 +1557,14 @@ void SearchWorker::DoBackupUpdateSingleNode(
     if (n_to_fix > 0 && !n->IsTerminal()) {
       n->AdjustForTerminal(v_delta, d_delta, m_delta, n_to_fix);
     }
+    if (n->GetN() >= solid_threshold) {
+      if (n->MakeSolid() && n == search_->root_node_) {
+        // If we make the root solid, the current_best_edge_ becomes invalid and
+        // we should repopulate it.
+        search_->current_best_edge_ =
+            search_->GetBestChildNoTemperature(search_->root_node_, 0);
```
Mardak, line 1565: What if we solidified parent `p` and cleared `current_best_edge_` instead of the next iteration of the loop after `n->FinalizeScoreUpdate` (i.e., GetN() is 1 smaller)? Then the existing best move logic would be reused:
https://github.com/LeelaChessZero/lc0/blob/b4f1dbc60a2befe4e56384441ac4a91e2972354b/src/mcts/search.cc#L1575-L1580

src/mcts/node.h
Tilps: FYI - solid_children doesn't mean all children are visited - it means all children are allocated.  They can still have 0 visits.

This change is not memory neutral for small values of threshold - memory usage goes up a lot because there are many unused nodes in each node block.  In practice the ideal value doesn't cause a large memory usage increase though, since performing those memory allocations is itself expensive if you start doing it for every node.
So there is unlikely to be much logic that can depend on this flag.

src/mcts/search.cc
Tilps: Solidifying p makes n an invalid pointer and it would have to be reconstructed.  I think its probably clearer the way it is.

src/mcts/node.h
Tilps: I think maybe I'll give the diagram a miss, but updated the text :)

src/mcts/node.cc
Tilps: done.

src/mcts/search.cc
Tilps: Also note - typically root node solidification happens extremely rarely - since tree reuse almost always means its already solid before the search even begins.

src/mcts/node.h
Tilps: Good point about Node_Iterator. Fixed.

src/mcts/node.cc
Tilps: This code section does actually turn up twice, so I am going to refactor it and make it a bit nicer. It should be much clearer now.

src/mcts/search.cc
```diff
@@ -1174,11 +1177,14 @@ void SearchWorker::ExtendNode(Node* node) {
   // Could instead reserve one more than the difference between history_.size()
   // and history_.capacity().
   to_add.reserve(60);
-  Node* cur = node;
-  while (cur != search_->root_node_) {
-    Node* prev = cur->GetParent();
-    to_add.push_back(prev->GetEdgeToNode(cur)->GetMove());
-    cur = prev;
+  {
```
borg323, line 1180: Add a short comment here

src/mcts/search.cc
```diff
@@ -1232,9 +1238,12 @@ void SearchWorker::ExtendNode(Node* node) {
       if (state != FAIL) {
         // TB nodes don't have NN evaluation, assign M from parent node.
         float m = 0.0f;
-        auto parent = node->GetParent();
-        if (parent) {
-          m = std::max(0.0f, parent->GetM() - 1.0f);
+        {
```
borg323, line 1241: and here

src/mcts/node.cc
```diff
@@ -94,6 +106,7 @@ class NodeGarbageCollector {
 
   mutable Mutex gc_mutex_;
   std::vector<std::unique_ptr<Node>> subtrees_to_gc_ GUARDED_BY(gc_mutex_);
+  std::vector<size_t> subtrees_is_solid_ GUARDED_BY(gc_mutex_);
```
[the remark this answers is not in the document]
Tilps: yes it would :)

src/mcts/search.cc
Tilps: done.

src/mcts/search.cc
Tilps: Done.

src/mcts/params.h
```diff
@@ -112,6 +112,9 @@ class SearchParams {
 
   int GetMaxOutOfOrderEvals() const { return kMaxOutOfOrderEvals; }
   float GetNpsLimit() const { return kNpsLimit; }
+
```
mooskagh, line 115: It's somehow inconsistent regarding where empty lines are inserted in the list of functions.
Consider removing all the empty lines (or grouping them using some logic other than random).

src/mcts/node.h
```diff
@@ -266,8 +292,10 @@ class Node {
   // Pointer to a parent node. nullptr for the root.
   Node* parent_ = nullptr;
   // Pointer to a first child. nullptr for a leaf node.
+  // As a 'hack' actually a unique_ptr to Node[] if solid_children.
   std::unique_ptr<Node> child_;
```
mooskagh, line 296: Maybe make it `std::unique_ptr<Node[]>` (with just one element by default)?
That way there need to be no hacks, everything will will automatically work..

It probably may be possible even not to touch iteration, we'd just need to make sibling_ a raw pointer rather than unique_ptr and rename flag to `sigling_is_owned_`.

Tilps: I considered both of these. But I was cautious due to a couple of reasons.
1) IIUC Node[] allocations unlike allocator.allocate(n) can store the length of the allocation so that delete[] knows how many items to destroy. I was concerned that the 'default case' this additional length storage might reduce likelihood that memory for a single node will be cacheline aligned and also waste more ram. I haven't tested it.
2) I prefer the iteration code being a separate path - advancing a pointer is considerably cheaper than dereferencing a sibling chain (let alone including the index checks to ensure they line up with the edge index desired) even when the sibling chain points directly to the next element.

I'd prefer to investigate 1) as a followup.  It doesn't change the overall behavior with respect to lifetime of node pointers being valid, so more people can be testing whether I have missed any race conditions sooner.

src/mcts/params.h
Tilps: Done.

PR #1362, opened 2020-06-23
Negative wdl

### LINKED ISSUE #1350: WDL displays negative values
As it says in the title, W in WDL gives negative values for certain positions. My current machine uses a 1080 Max-Q. Using latest binary with sv4175 net using cudaNN backend. 

8/8/8/2p5/1pp5/brpp4/qpprpK1P/1nkbn3 w - - 0 1
bug

negW

nguyenpham:
Fix #1350

The problem is about the precision of double/float variants when their values are not exactly but actually approximated, thus a value 0.123 could be stored as 0.123 or 0.1231. We won't notice the difference when floating variants w, d, l are far from 0.0 and 1.0 (so they all look good). However, when one is closed to 0, it could be presented as a positive zero (+0.0) or a negative zero (-0.0) which both are accepted and considered as 0.0 but when we multi it with 1000 and convert into an integer it may become a small negative integer. When that (a negative float zero) occurs, it may be propagated and created some a bit larger negative values.

There are two solutions:
- check and limit variants in the range. For example, after updating the value d_ (in Node class) we can do:
`d_ = std::max<float>(d_, 1.0 + wl);`

IMO, this way doesn't improve the precisions of variants but requiring more computing (than the second one)

- make the displaying values are not negative as below (d is an integer in WDL):
`d = std::max<int>(d, 0);`

This PR uses the second way.


Naphthalin:
Thanks for attempting to fix the issue. However, I think I remember that the problem wasn't negative `d` values, but actually a negative `w` or `l` value in very one-sided positions. This doesn't seem to be fixed here to me.

nguyenpham:
Check the code:
‘  const auto w = std::max(0, static_cast<int>(std::round(500.0 * (1.0 + wl - floatD))));’

LouisZulli:
Even with the two proposed commits, I see this:

```
info depth 26 seldepth 54 time 88185 nodes 1958147 score cp 103 wdl 644 259 97 nps 23093 tbhits 0 pv f3e5 g4d1 f1b5 c7c6 d5c6 a7a6 c6c7 a6b5 c7d8n a8d8 e5f7 e8f7 e1d1 g8f6 h1e1 e7e6 a2a3 d6d5 e4e5 f6d7 c1d2 d5d4 d2b4 g7g5 b1d2 f7g6 d2e4 b7b6 e4d6 f8d6 e5d6 e6e5 c2c3 h8f8 f2f3 f8f4 d1c2 g5g4 c3d4 g4f3 g2f3 f4d4 a1d1 d4d1 e1d1 d8f8 d1g1 g6f7
info depth 26 seldepth 54 time 93204 nodes 2071360 score cp -2147483648 wdl 0 1000 0 nps 23063 tbhits 0 pv f3e5 g4d1 f1b5 c7c6 d5c6 a7a6 c6c7 a6b5 c7d8n a8d8 e5f7 e8f7 e1d1 g8f6 h1e1 e7e6 a2a3 d6d5 e4e5 f6d7 c1d2 d5d4 d2b4 g7g5 b1d2 f7g6 d2e4 b7b6 e4d6 f8d6 b4d6 d7c5 d1d2 d8d7 a1d1 d7f7 f2f3 h8c8 d2c1 f7f4 c1b1 h7h5 d1d2 h5h4 h2h3 g6f5 e1d1

```

borg323 (approved):

borg323:
@nguyenpham  I merged with master and updated a comment, OK to merge?

src/mcts/search.cc
```diff
@@ -250,10 +249,12 @@ void Search::SendUciInfo() REQUIRES(nodes_mutex_) {
...
     }
-    const int l = static_cast<int>(std::round(500.0 * (1.0 - wl - d)));
-    // Using 1000-w-l instead of 1000*d for D score so that W+D+L add up to
-    // 1000.0.
-    uci_info.wdl = ThinkingInfo::WDL{w, 1000 - w - l, l};
+
+    const auto w = std::max(0, static_cast<int>(std::round(500.0 * (1.0 + wl - floatD))));
+    const auto l = std::max(0, static_cast<int>(std::round(500.0 * (1.0 - wl - floatD))));
+    // Using 1000-w-l so that W+D+L add up to 1000.0.
+    const auto d = std::max(0, 1000 - w - l);
```
mooskagh, line 256: Maybe this? (ensures they always add up to 1000)
```cpp
auto d = 1000 - w - l;
if (d < 0) {
  w += d / -2;
  l = 1000 - w;
  d = 0;
}
```

ddobbelaere: Isn't the requirement it should add up to 1000 artificial? E.g. for 1.4 + 1.4 + 997.2 this is hard, but 1 + 1 + 997, although only 999, is still "best" rounding IMHO...

mooskagh: WDL is for chess UI, and all UIs that I saw (including visualizations/graphs that I made) assume w+d+l=1000.
It's actually pretty hard to avoid, usually it's some kind of stacked graph, there e.g. L and D is taken from the WDL output, and W takes all the remaining space.

src/mcts/search.cc
```diff
@@ -258,10 +257,19 @@ void Search::SendUciInfo() REQUIRES(nodes_mutex_) REQUIRES(counters_mutex_) {
...
-    uci_info.wdl = ThinkingInfo::WDL{w, 1000 - w - l, l};
+
+    auto w =
+        std::max(0, static_cast<int>(std::round(500.0 * (1.0 + wl - floatD))));
+    auto l =
+        std::max(0, static_cast<int>(std::round(500.0 * (1.0 - wl - floatD))));
+    // Using 1000-w-l so that W+D+L add up to 1000.0.
+    auto d = 1000 - w - l;
+    if (d < 0) {
+      w = std::min(1000, std::max(0, w + d / 2));
```
Naphthalin, line 268: This is supposed to be `w + d / -2`, right?

Naphthalin: Nevermind, confused myself.

Naphthalin (approved):

PR #878, opened 2019-06-07
change q from float to double for averaging reasons

MelleKoning:
#875 
This PR is for changing Q from float to double, to try out if this makes for different (better?) gameplay.

Have a beautiful testset of two games reversed 👍  whereby Qdouble wins as white while it holds the draw as black against Q-float.
Game settings 5 sec for the game +5sec for each move.

```
[Event "My Tournament"]
[Site "HaceTourney"]
[Date "2019.06.07"]
[Round "1"]
[White "lc0.Net.42475"]
[Black "lc0.Net.42475-Qdouble"]
[Result "1/2-1/2"]
[ECO "A00"]
[Opening "Mieses Opening"]
[PlyCount "132"]
[Termination "adjudication"]
[TimeControl "5+5"]

1. d3 {book} g6 {book} 2. c3 {book} d6 {book} 3. e4 {+0.07/6 4.0s}
Nf6 {+0.04/6 3.4s} 4. g3 {+0.04/7 4.5s} Bg7 {+0.03/7 4.2s} 5. Bg2 {+0.02/7 4.1s}
O-O {+0.03/7 4.9s} 6. Ne2 {+0.05/7 4.5s} e5 {0.00/7 4.9s} 7. c4 {+0.08/8 7.5s}
c5 {+0.01/7 5.3s} 8. Nbc3 {+0.03/7 4.8s} Nc6 {+0.02/7 5.9s}
9. O-O {+0.04/7 3.4s} a6 {0.00/7 5.7s} 10. Nd5 {+0.10/7 6.5s} Nxd5 {0.00/8 3.7s}
11. cxd5 {+0.07/8 2.4s} Nd4 {-0.03/8 3.8s} 12. Nxd4 {+0.12/8 3.6s}
exd4 {-0.05/7 3.4s} 13. Bf4 {+0.09/7 9.2s} f5 {-0.03/7 8.0s}
14. Qd2 {+0.15/7 2.8s} fxe4 {-0.06/8 4.0s} 15. Bxe4 {+0.14/7 3.0s}
Bh3 {-0.04/8 6.6s} 16. Rfe1 {+0.14/7 4.3s} b5 {-0.03/8 2.3s}
17. a4 {+0.06/8 7.6s} c4 {+0.04/9 5.7s} 18. axb5 {+0.01/10 3.9s}
c3 {-0.05/11 2.9s} 19. bxc3 {+0.34/11 4.3s} dxc3 {-0.19/10 3.3s}
[...]

PR #349, opened 2018-09-11
WIP make Q and U doubles for issue #328 test

borg323:
This is just for testing #328.

PR #1363, opened 2020-06-23
Change variant wl_ in Nodes from double to float

nguyenpham:
**Edited:** 
The PR:
- add a warning comment for the variant wl_ in Node class
- changes the return type of the function GetWL from float to double and do clear castings if needed

Discuss: the original function GetWL is a data bottleneck for the variant wl_ since sometimes the engine has to keep converting from double to float then back to double, losing both computing and precision. We should keep the data unchanged type as long as possible to maximize the benefit of changing from float to double for wl_

**Origin:**
Using double for the variant wl_ (in the class Nodes) is redundant both on size and computing (for converting) since all in/out and operations with other variants using floats only.

By changing from double to float the precisions of all functions are almost the same but we can reduce the size of Nodes from 80 (for 64-bit system) to 72 bytes, a 10% decreasing.

### DIFF
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -430,7 +430,7 @@ V5TrainingData Node::GetV5TrainingData(
   }
 
   // Aggregate evaluation WL.
-  result.root_q = -GetWL();
+  result.root_q = static_cast<float>(-GetWL());
   result.best_q = best_q;
 
   // Draw probability of WDL head.
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -252,14 +252,6 @@ class Node {
   // padding when new fields are added, we arrange the fields by size, largest
   // to smallest.
 
-  // 8 byte fields.
-  // Average value (from value head of neural network) of all visited nodes in
-  // subtree. For terminal nodes, eval is stored. This is from the perspective
-  // of the player who "just" moved to reach this position, rather than from the
-  // perspective of the player-to-move for the position.
-  // WL stands for "W minus L". Is equal to Q if draw score is 0.
-  double wl_ = 0.0f;
-
   // 8 byte fields on 64-bit platforms, 4 byte on 32-bit.
   // Array of edges.
   std::unique_ptr<Edge[]> edges_;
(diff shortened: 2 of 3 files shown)


ddobbelaere:
If you are wondering why we use double, please check out #1280 and #875.

Maybe it's a good idea to discuss "holy grail solutions without any apparent drawback" in Discord chat #dev or #dev-public first, because more likely than not there is a drawback/good reason for it :P.

nguyenpham:
Oh I see. Look like I am reinventing the wheel or even worse, trying to destroy it ;)
Thanks for the info. Reading...

ddobbelaere:
No problem (and no offense meant), I think it is very much appreciated to have multiple critical eyes looking at the codebase and taking nothing for granted...

borg323:
Actually, can you re-purpose this pr into one adding a comment with a pointer to the relevant discussion?

nguyenpham:
@borg323: can you help to do that, I don’t know how to do. Thanks a lot

borg323:
Just saying "This has to be a double" with a link to #875 should be enough

nguyenpham:
Sorry, I totally lost. Do you mean just edit the introduction/explanation of this PR (so it’s still a PR) or add a new issue about this one?

borg323:
I meant change this pr to one just adding the above comment to node.h, so I can review it quickly.

liamlim:
Just revert all changes in this PR and add explaining comment to definition of _wl which says that this variable must be double and also mentions #875. Then everyone who wants to touch this part of code in future will know the reasons why it has to be double

nguyenpham:
Oops, got it now. How silly of me!

nguyenpham:
I have committed:
- revert all changes
- add a comment for variant wl_ (as the suggestion by @borg323)
- change the return type of the function GetWL from float to double, do clear castings if needed

IMO, the original function GetWL is a data bottleneck for the variant wl_ since sometimes the engine has to keep converting from double to float then back to double, losing both computing and precision. We should keep the data unchanged type as long as possible to maximize the benefit of changing from float to double for wl_

borg323:
There were at least two earlier PRs doing something similar: #878 and #349

nguyenpham:
IMO, it is quite different since the PR didn't touch to any variants, only return type of a function. That deals with operations in temporary memory only

PR #1367, opened 2020-06-25
change included net to 703350

borg323:


lealgo:
Maybe add another appveyor variable for the network hash?

PR #1368, opened 2020-06-26
Minor optimizations for non-LTO builds

Tilps:
These methods are fairly trivial and used in or near the hottest paths of our code - so having them not be inline in the header forces LTO to be required to inline them if the compiler feels it could be a good idea - when there isn't a real compelling reason to keep them in the implementation file.


borg323:
I think #1369 is a better alternative.

Tilps:
I'll see how that goes with my debugging :)

borg323:
I would prefer to keep the Edge::GetP() implementation together with the Edge::SetP() one (and the comment above).

Tilps:
I actually plan on abandoning this PR - LTO being default on may not quite work with VS out of the box, but now that I know how to fix that I no longer worry about these details personally - and our release binaries use PGO anyway.

PR #1370, opened 2020-06-26
Reduce record backend stored file size by a factor of 3.

Tilps:
This takes advantage of the specific pattern of requests made to any given computation to avoid having to store the entire list of requests.
It is a bit 'more' fragile, but given how fragile it already is, I think the disk space and load time savings are worth it.
10 million nodes search takes about 1GB file now meaning 40 million node searches are a plausible replay possibility.

Also make it more explicit that only the first instance of a hash value in the replay file is used.  This was needed to make the behavior correct, but its good to be clear anyway. (It also slightly reduces the memory usage at runtime during replay, even above the 3 fold reduction from reduced recording data.)

### DIFF
--- a/src/neural/network_record.cc
+++ b/src/neural/network_record.cc
@@ -53,6 +53,7 @@ class RecordComputation : public NetworkComputation {
   void AddInput(InputPlanes&& input) override {
     hashes_.push_back(make_hash(input));
     requests_.emplace_back();
+    q_count_.push_back(0);
     inner_->AddInput(std::move(input));
   }
   // Do the computation.
@@ -60,11 +61,15 @@ class RecordComputation : public NetworkComputation {
   // Returns how many times AddInput() was called.
   int GetBatchSize() const override { return inner_->GetBatchSize(); }
   float Capture(float value, int index) const {
+    // Only capture until we see Q again - the rest can be infered from that
+    // set.
+    if (q_count_[index] > 1) return value;
     requests_[index].push_back(value);
     return value;
   }
   // Returns Q value of @sample.
   float GetQVal(int sample) const override {
+    q_count_[sample]++;
     return Capture(inner_->GetQVal(sample), sample);
   }
   float GetDVal(int sample) const override {
@@ -95,6 +100,7 @@ class RecordComputation : public NetworkComputation {
   std::unique_ptr<NetworkComputation> inner_;
   std::string record_file_;
   std::vector<uint64_t> hashes_;
+  mutable std::vector<int> q_count_;
   mutable std::vector<std::vector<float>> requests_;
   static Mutex mutex_;
 };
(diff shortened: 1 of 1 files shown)


borg323:
Wouldn't this be better done in the search/cache, just reading once from the backend?

Tilps:
Much easier said than done... The first read happens in CachingBackend during ComputeBlocking.
The second and third are much further a field - I don't think I could plausibly get it down to one loop without changing CachingBackend API to be much more ugly. And I think getting that last factor of 2 (after the 2/3rds) is well worth the complexity in this corner case backend that only a handful of people are ever going to use and notice if it ever stops working...

borg323:
The following gets rid of the last round (the more fragile one) and is ~5% faster with the random backend for me:
```
index 6a771e8..733aa62 100644
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -1472,26 +1472,24 @@ void SearchWorker::FetchSingleNodeResult(NodeToProcess* node_to_process,
   // ...and secondly, the policy data.
   // Calculate maximum first.
   float max_p = -std::numeric_limits<float>::infinity();
-  for (auto edge : node->Edges()) {
-    max_p = std::max(max_p, computation_->GetPVal(
-                                idx_in_computation,
-                                edge.GetMove().as_nn_index(
-                                    node_to_process->probability_transform)));
-  }
   // Intermediate array to store values when processing policy.
   // There are never more than 256 valid legal moves in any legal position.
   std::array<float, 256> intermediate;
-  float total = 0.0;
   int counter = 0;
   for (auto edge : node->Edges()) {
     float p = computation_->GetPVal(
         idx_in_computation,
         edge.GetMove().as_nn_index(node_to_process->probability_transform));
+    intermediate[counter++] = p;
+    max_p = std::max(max_p, p);
+  }
+  float total = 0.0;
+  for (int i = 0; i < counter; i++) {
     // Perform softmax and take into account policy softmax temperature T.
     // Note that we want to calculate (exp(p-max_p))^(1/T) = exp((p-max_p)/T).
-    p = FastExp((p - max_p) / params_.GetPolicySoftmaxTemp());
-
-    intermediate[counter++] = p;
+    float p =
+        FastExp((intermediate[i] - max_p) / params_.GetPolicySoftmaxTemp());
+    intermediate[i] = p;
     total += p;
   }
   counter = 0;
```

Tilps:
Yes, I was thinking much the same thing. And I would be happy to see that as a separate PR - in which case I can remove the 3rd if statement as a follow up.
However random backend is a likely poor choice for benchmarking here, since it calls HashCat in its computation getters, where for real backends its just an array indexing operation.
So I wouldn't get too excited about that 5%...

borg323:
I tested with cache, so the last GetPVal should have been just a cache lookup. Anyway, we'll find out soon enough.

borg323:
BTW, the easy way is to skip the first round of calls if NNCacheSize=0, but the added complexity will probably not be worth it in normal use.

Tilps:
> I tested with cache, so the last GetPVal should have been just a cache lookup. Anyway, we'll find out soon enough.

A given CachingComputation either uses the cache or does not use the cache for a given input for its lifetime - it doesn't switch to using the cache part way through.

borg323:
It looks to me that CachingComputation::ComputeBlocking() calls the Get?Val functions even if the cache size is 0, which is how you get the 1st round.

Tilps:
I was meaning that regardless of whether caching is enabled or not - the number of calls received by the parent computation for a given sample index is either 0 or all of them. There is no case that some of the calls to a CachingComputation for a specific sample index go through to parent, and some don't.

PR #1372, opened 2020-06-26
Remove logit q.

Tilps:
Profiling on windows with LTO enabled, I noticed that edge.GetQ was not being inlined in our hottest loop. If I removed the nested logit_q path it was inlined and there was a dramatic performance increase in CPU limited scenarios. (Combined with one other small change it was ~15% increase in nps over a 5 million node search).
Given that logit-q has failed to prove itself in practice, having the potential for this much CPU bottleneck when the feature is disabled doesn't seem a worthwhile situation.
Until such time as there seems a new reason for it - it seems prudent to remove it entirely.

### DIFF
--- a/src/utils/fastmath.h
+++ b/src/utils/fastmath.h
@@ -77,9 +77,4 @@ inline float FastLog(const float a) {
 // Fast approximate exp(x). Does only limited range checking.
 inline float FastExp(const float a) { return FastPow2(1.442695040f * a); }
 
-// Fast logit for more readable code.
-inline float FastLogit(const float a) {
-  return 0.5f * FastLog((1.0f + a) / (1.0f - a));
-}
-
 }  // namespace lczero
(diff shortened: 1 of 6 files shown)


AlexisOlson:
Should this also remove `FastLogit` from `utils/fastmath.h`?

src/utils/fastmath.h
```diff
@@ -77,9 +77,4 @@ inline float FastLog(const float a) {
 // Fast approximate exp(x). Does only limited range checking.
 inline float FastExp(const float a) { return FastPow2(1.442695040f * a); }
 
-// Fast logit for more readable code.
-inline float FastLogit(const float a) {
```
Mardak, line 81: The include is unnecessary now https://github.com/LeelaChessZero/lc0/blob/770ca152e168b0f8d28a4e8e7eb7ef8bfcb546b9/src/mcts/node.h#L42

PR #1373, opened 2020-06-26
skip repeated GetPVal() calls

borg323:


Tilps:
This is probably worth it for not having to iterate over edges twice alone...

PR #1197, opened 2020-04-11
Focus search with --two-fold-draw repetition except for root children.

Mardak:
r?@mooskagh or @Tilps Fix #710 and improves on #1161 by correctly not playing moves that aren't actually draw. E.g., a dummy situation where white could checkmate but decides to shuffle the rook and the previous PR would have black repeat thinking it's a draw:
could mate

```
# with 1161 treating any two-fold repetition as draw
position fen k1r5/8/K1R5/8/8/8/8/8 w - - 0 1 moves c6b6 c8b8 b6c6

info string b8b6  (35  ) N:    1365 (+ 0) (P:  7.69%) (WL: -1.00000) (D:  0.000) (M:  5.0) (Q: -1.00000) (U: 0.87158) (Q+U: -0.12842) (V: -1.0000) (T) 
…
info string b8c8  (24  ) N:  108972 (+ 0) (P:  7.69%) (WL:  0.00000) (D:  1.000) (M:  0.0) (Q:  0.00000) (U: 0.01093) (Q+U:  0.01093) (V:  0.0000) (T) 
bestmove b8c8

# with this PR
info string b8c8  (24  ) N:     422 (+ 0) (P:  7.69%) (WL: -1.00000) (D:  0.000) (M:  1.0) (Q: -1.00000) (U: 1.01206) (Q+U:  0.01206) (V: -1.0000) (T) 
info string b8b6  (35  ) N:    1365 (+ 0) (P:  7.69%) (WL: -1.00000) (D:  0.000) (M:  5.0) (Q: -1.00000) (U: 0.31340) (Q+U: -0.68660) (V: -1.0000) (T) 
…
```

Otherwise, the behavior is the same as the previous PR #1196 original comment copied below.

---
At CCC13 Heptagonal Game 155, 384x30-t60-3180 blundered the win with 68… Bf6 that allows white to keep checking and shuffling but not mate with the knights:
2fold
https://www.chess.com/computer-chess-championship#event=ccc13-heptagonal&game=155

Here's master (b5bedf4) with 384x30-t60-3180 at various nodes:
```
./lc0 -w 384x30-t60-3180 --multipv=2 --per-pv-counters --smart-pruning-factor=0
[...]

Issue #710, opened 2019-01-30
Blundered win into perpetual check forcing stalemate (cause: resign and temperature?)


Mardak:
Here's a position from CCC 4 game 450 where a few moves earlier, 32742 says +28 and SF says +153.
![screen shot 2019-01-29 at 10 08 29 pm](https://user-images.githubusercontent.com/438537/51961771-b66b3000-2412-11e9-9695-873b94e2499a.png)

This position 32742 captures a rook still unaware that black will perpetually check with the lone rook as white capturing will result in stalemate.

If self-play got into this position, here's what it would look close to (but without noise):
```
[...]

PR #1375, opened 2020-06-27
Twofold repetition draw scoring

Naphthalin:
EDIT: It is fully functional now, tests are welcome! Also, the tree reuse problem with level 1 being complicated has been solved successfully, so the heuristic of level 2 probably isn't needed anymore.

Original post:
This is a draft for a twofold draw scoring patch. For development + testing purposes, this PR introduces an `int TwoFoldDrawLevel = 0...3` where the levels are:

- `0`: never score twofold repetions as draws. This is current behavior.
- `1`: score twofold repetitions when their first repetition is at root or in the tree, but not when it happened in history. This is the theoretically sound implementation, avoiding delusional evals, making this the preferred implementation.
- `2`: score twofold repetions as draws when they happen at depth 4 or deeper in the tree. This is equivalent to `level 1` in the most common 3-fold repetitions moving 1 piece each forward and backward, and much easier to implement
- `3`: score all twofold repetitions as draws. This isn't intended to be used for playing or analysis, only for testing purposes.

This PR isn't functional yet, and only identifies the positions in the code so far. The main problem with `level 1` is marking two-fold draws as non terminal when a move is played and therefore a tree node loses its property of being a two-fold with its first repetition at root or later.

I also want to thank @Mardak for his prior efforts on making a functional two-fold PR, his implementation was somewhere between `level 2` and `level 3` by making two-fold draws non-terminal only at depth 1.

### DIFF
--- a/src/chess/position.h
+++ b/src/chess/position.h
@@ -132,9 +132,7 @@ class PositionHistory {
   // Checks for any repetitions since the last time 50 move rule was reset.
   bool DidRepeatSinceLastZeroingMove() const;
 
-
  private:
-  int ComputeLastMoveRepetitions() const;
   int ComputeLastMoveRepetitions(int* cycle_length) const;
 
   std::vector<Position> positions_;
--- a/src/chess/position.h
+++ b/src/chess/position.h
@@ -54,7 +54,10 @@ class Position {
 
   // Someone outside that class knows better about repetitions, so they can
   // set it.
-  void SetRepetitions(int repetitions) { repetitions_ = repetitions; }
+  void SetRepetitions(int repetitions, int cycle_length) {
+    repetitions_ = repetitions;
+    cycle_length_ = cycle_length;
+  }
 
   // Number of ply with no captures and pawn moves.
   int GetRule50Ply() const { return rule50_ply_; }
--- a/src/chess/position.h
+++ b/src/chess/position.h
@@ -50,7 +50,7 @@ class Position {
   int GetRepetitions() const { return repetitions_; }
 
   // How many half-moves since the same position appeared in the game before.
-  int GetPliesSinceFirstRepetition() const { return cycle_length_; }
+  int GetPliesSincePrevRepetition() const { return cycle_length_; }
 
   // Someone outside that class knows better about repetitions, so they can
   // set it.
--- a/src/chess/position.h
+++ b/src/chess/position.h
@@ -49,6 +49,9 @@ class Position {
   // How many time the same position appeared in the game before.
   int GetRepetitions() const { return repetitions_; }
 
+  // How many half-moves since the same position appeared in the game before.
+  int GetPliesSinceFirstRepetition() const { return cycle_length_; }
+
   // Someone outside that class knows better about repetitions, so they can
   // set it.
   void SetRepetitions(int repetitions) { repetitions_ = repetitions; }
@@ -73,6 +76,8 @@ class Position {
   int rule50_ply_ = 0;
   // How many repetitions this position had before. For new positions it's 0.
   int repetitions_;
+  // How many half-moves since the position was repeated or 0.
+  int cycle_length_;
   // number of half-moves since beginning of the game.
   int ply_count_ = 0;
 };
@@ -124,11 +129,10 @@ class PositionHistory {
   // Checks for any repetitions since the last time 50 move rule was reset.
   bool DidRepeatSinceLastZeroingMove() const;
 
-  int ComputePliesSinceFirstRepetition() const;
-
 
  private:
   int ComputeLastMoveRepetitions() const;
+  int ComputeLastMoveRepetitions(int* cycle_length) const;
 
   std::vector<Position> positions_;
 };
--- a/src/chess/position.h
+++ b/src/chess/position.h
@@ -124,9 +124,11 @@ class PositionHistory {
   // Checks for any repetitions since the last time 50 move rule was reset.
   bool DidRepeatSinceLastZeroingMove() const;
 
+  int ComputePliesSinceFirstRepetition() const;
+
+
  private:
   int ComputeLastMoveRepetitions() const;
-  int ComputePliesSinceFirstRepetition() const;
 
   std::vector<Position> positions_;
 };
--- a/src/chess/position.h
+++ b/src/chess/position.h
@@ -126,6 +126,7 @@ class PositionHistory {
 
  private:
   int ComputeLastMoveRepetitions() const;
+  int ComputePliesSinceFirstRepetition() const;
 
   std::vector<Position> positions_;
 };
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -313,13 +313,23 @@ void Node::MakeTerminal(GameResult result, float plies_left, Terminal type) {
 
 void Node::MakeNotTerminal() {
   if (terminal_type_ == Terminal::TwoFold) {
-    /* under construction
-    for (auto node = this; node=node->parent_; node != ) {
+    // When reverting a terminal twofold repetition draw, this can happen
+    // in the tree. To keep consistency, we revert the effect of the visits
+    // to that terminal on all parent nodes and let PUCT revisit the nodes
+    // and fetch the evals without the twofold draw.
+    int depth = 0;
+    // Cache node's values as we reset them in the process. We could manually
+    // set wl and d, but if we want to reuse this for reverting other terminal
+    // nodes this is the way to go.
+    const auto wl = wl_;
+    const auto d = d_;
+    const auto m = m_;
+    for (Node node = this; node = node->GetParent(); node != nullptr ) {
       // Revert all visits on twofold terminal when making it non terminal.
-      node.RevertVisits(visits, wl, d, m);
-      // Best edge cache etc needs to be invalidated.
+      node.RevertTerminalVisits(wl, d, m + (float)depth, n);
+      depth++;
+      // If wl != 0, we would have to switch signs at each depth.
     }
-    */
     // Currently, only visits to the node itself are reset, making the tree
     // inconsistent.
     n_ = 0;
@@ -388,8 +398,34 @@ void Node::AdjustForTerminal(float v, float d, float m, int multivisit) {
   d_ += multivisit * d / n_;
   m_ += multivisit * m / n_;
   // Best child is potentially no longer valid. This shouldn't be needed since
-  // AjdustForTerminal is always called immediately after FinalizeScoreUpdate,
+  // AdjustForTerminal is always called immediately after FinalizeScoreUpdate,
   // but for safety in case that changes.
+  best_child_cached_ = nullptr;
+}
+
+void Node::RevertTerminalVisits(float v, float d, float m, int multivisit) {
+  // Compute new n_ first, as reducing a node to 0 visits is a special case.
+  const int n_new = n_ - multivisit;
+  if (n_new <= 0) {
+    if (parent_ != nullptr) {
+      // To keep consistency with FinalizeScoreUpdate() expanding a node again,
+      // we need to reduce the parent's visited policy.
+      parent_->visited_policy_ -= parent_->edges_[index_].GetP();
+    }
+    // If n_new == 0, reset all relevant values to 0.
+    wl_ = 0.0;
+    d_ = 1.0;
+    m_ = 0.0;
+    n_ = 0;
+  } else {
+    // Recompute Q and M.
+    wl_ -= multivisit * (v - wl_) / n_new;
+    d_ -= multivisit * (d - d_) / n_new;
+    m_ -= multivisit * (m - m_) / n_new;
+    // Decrement N.
+    n_ -= multivisit;
+  }
+  // Best child is potentially no longer valid.
   best_child_cached_ = nullptr;
 }
 
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -307,13 +307,11 @@ void Node::MakeTerminal(GameResult result, float plies_left, Terminal type) {
     d_ = 0.0f;
     // Terminal losses have no uncertainty and no reason for their U value to be
     // comparable to another non-loss choice. Force this by clearing the policy.
-    if (GetParent() != nullptr)
-            GetOwnEdge()->SetP(0.0f);
+    if (GetParent() != nullptr) GetOwnEdge()->SetP(0.0f);
   }
 }
 
 void Node::MakeNotTerminal() {
-  // SetBounds(GameResult::BLACK_WON, GameResult::WHITE_WON);
   terminal_type_ = Terminal::NonTerminal;
   n_ = 0;
 
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -312,6 +312,7 @@ void Node::MakeTerminal(GameResult result, float plies_left, Terminal type) {
 }
 
 void Node::MakeNotTerminal() {
+  SetBounds(GameResult::BLACK_WON, GameResult::WHITE_WON);
   terminal_type_ = Terminal::NonTerminal;
   n_ = 0;
 
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -307,7 +307,10 @@ void Node::MakeTerminal(GameResult result, float plies_left, Terminal type) {
     d_ = 0.0f;
     // Terminal losses have no uncertainty and no reason for their U value to be
     // comparable to another non-loss choice. Force this by clearing the policy.
-    if (GetParent() != nullptr) GetOwnEdge()->SetP(0.0f);
+    // However, don't set policy of twofold draws to zero as this only is a
+    // heuristic for more accurate eval
+    if (GetParent() != nullptr && terminal_type_ != Node::Terminal::TwoFold)
+            GetOwnEdge()->SetP(0.0f);
   }
 }
 
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -324,7 +324,7 @@ void Node::MakeNotTerminal() {
     const auto wl = wl_;
     const auto d = d_;
     const auto m = m_;
-    for (Node node = this; node = node->GetParent(); node != nullptr ) {
+    for (Node* node = this; node = node->GetParent(); node != nullptr ) {
       // Revert all visits on twofold terminal when making it non terminal.
       node.RevertTerminalVisits(wl, d, m + (float)depth, n);
       depth++;
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -326,6 +326,8 @@ void Node::MakeNotTerminal() {
     const auto m = m_;
     const auto terminal_visits = n_;
     for (Node* node = this; node != nullptr; node = node->GetParent()) {
+      // Logging stuff for debugging purposes
+      LOGFILE << "Reverted " << terminal_visits << " visits at depth " << depth;
       // Revert all visits on twofold terminal when making it non terminal.
       node->RevertTerminalVisits(wl, d, m + (float)depth, terminal_visits);
       depth++;
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -293,7 +293,7 @@ void Node::SortEdges() {
 }
 
 void Node::MakeTerminal(GameResult result, float plies_left, Terminal type) {
-  SetBounds(result, result);
+  if (type != Terminal::TwoFold) { SetBounds(result, result); }
   terminal_type_ = type;
   m_ = plies_left;
   if (result == GameResult::DRAW) {
@@ -313,7 +313,7 @@ void Node::MakeTerminal(GameResult result, float plies_left, Terminal type) {
 }
 
 void Node::MakeNotTerminal() {
-  SetBounds(GameResult::BLACK_WON, GameResult::WHITE_WON);
+  // SetBounds(GameResult::BLACK_WON, GameResult::WHITE_WON);
   terminal_type_ = Terminal::NonTerminal;
   n_ = 0;
 
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -293,7 +293,7 @@ void Node::SortEdges() {
 }
 
 void Node::MakeTerminal(GameResult result, float plies_left, Terminal type) {
-  if (type != Terminal::TwoFold) { SetBounds(result, result); }
+  if (type != Terminal::TwoFold) SetBounds(result, result);
   terminal_type_ = type;
   m_ = plies_left;
   if (result == GameResult::DRAW) {
@@ -312,8 +312,21 @@ void Node::MakeTerminal(GameResult result, float plies_left, Terminal type) {
 }
 
 void Node::MakeNotTerminal() {
+  if (terminal_type_ == Terminal::TwoFold) {
+    /* under construction
+    for (auto node = this; node=node->parent_; node != ) {
+      // Revert all visits on twofold terminal when making it non terminal.
+      node.RevertVisits(visits, wl, d, m);
+      // Best edge cache etc needs to be invalidated.
+    }
+    */
+    // Currently, only visits to the node itself are reset, making the tree
+    // inconsistent.
+    n_ = 0;
+  } else {
+    n_ = 0;
+  }
   terminal_type_ = Terminal::NonTerminal;
-  n_ = 0;
 
   // If we have edges, we've been extended (1 visit), so include children too.
   if (edges_) {
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -324,16 +324,15 @@ void Node::MakeNotTerminal() {
     const auto wl = wl_;
     const auto d = d_;
     const auto m = m_;
+    const auto terminal_visits = n_;
     for (Node* node = this; node = node->GetParent(); node != nullptr ) {
       // Revert all visits on twofold terminal when making it non terminal.
-      node.RevertTerminalVisits(wl, d, m + (float)depth, n);
+      node->RevertTerminalVisits(wl, d, m + (float)depth, terminal_visits);
       depth++;
       // If wl != 0, we would have to switch signs at each depth.
     }
-    // Currently, only visits to the node itself are reset, making the tree
-    // inconsistent.
-    n_ = 0;
   } else {
+    // Any other case, just setting n_ = 0 is sufficient, as parent is root.
     n_ = 0;
   }
   terminal_type_ = Terminal::NonTerminal;
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -307,9 +307,7 @@ void Node::MakeTerminal(GameResult result, float plies_left, Terminal type) {
     d_ = 0.0f;
     // Terminal losses have no uncertainty and no reason for their U value to be
     // comparable to another non-loss choice. Force this by clearing the policy.
-    // However, don't set policy of twofold draws to zero as this only is a
-    // heuristic for more accurate eval
-    if (GetParent() != nullptr && terminal_type_ != Node::Terminal::TwoFold)
+    if (GetParent() != nullptr)
             GetOwnEdge()->SetP(0.0f);
   }
 }
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -325,7 +325,7 @@ void Node::MakeNotTerminal() {
     const auto d = d_;
     const auto m = m_;
     const auto terminal_visits = n_;
-    for (Node* node = this; node = node->GetParent(); node != nullptr ) {
+    for (Node* node = this; node != nullptr; node = node->GetParent()) {
       // Revert all visits on twofold terminal when making it non terminal.
       node->RevertTerminalVisits(wl, d, m + (float)depth, terminal_visits);
       depth++;
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -327,9 +327,9 @@ void Node::MakeNotTerminal() {
     const auto terminal_visits = n_;
     for (Node* node = this; node != nullptr; node = node->GetParent()) {
       // Logging stuff for debugging purposes
-      LOGFILE << "Reverted " << terminal_visits << " visits at depth " << depth;
       // Revert all visits on twofold terminal when making it non terminal.
       node->RevertTerminalVisits(wl, d, m + (float)depth, terminal_visits);
+      LOGFILE << "Successfully everted " << terminal_visits << " visits at depth " << depth;
       depth++;
       // If wl != 0, we would have to switch signs at each depth.
     }
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -327,7 +327,7 @@ void Node::MakeNotTerminal() {
     const auto terminal_visits = n_;
     // Logging stuff for debugging purposes
     LOGFILE << "Attempting to revert terminal visits";
-    for (Node* node = this; node != nullptr; node = node->GetParent()) {
+/*    for (Node* node = this; node != nullptr; node = node->GetParent()) {
       // Logging stuff for debugging purposes
       LOGFILE << "Attempting to revert " << terminal_visits << " visits at depth " << depth;
       // Revert all visits on twofold terminal when making it non terminal.
@@ -335,8 +335,8 @@ void Node::MakeNotTerminal() {
       LOGFILE << "Successfully reverted " << terminal_visits << " visits at depth " << depth;
       depth++;
       // If wl != 0, we would have to switch signs at each depth.
-    }
-    // n_ = 0;
+    } */
+    n_ = 0;
   } else {
     // Any other case, just setting n_ = 0 is sufficient, as parent is root.
     n_ = 0;
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -326,8 +326,8 @@ void Node::MakeNotTerminal() {
     const auto m = m_;
     const auto terminal_visits = n_;
     // Logging stuff for debugging purposes
-    // LOGFILE << "Attempting to revert terminal visits";
-/*    for (Node* node = this; node != nullptr; node = node->GetParent()) {
+    LOGFILE << "Attempting to revert terminal visits";
+    for (Node* node = this; node != nullptr; node = node->GetParent()) {
       // Logging stuff for debugging purposes
       LOGFILE << "Attempting to revert " << terminal_visits << " visits at depth " << depth;
       // Revert all visits on twofold terminal when making it non terminal.
@@ -335,7 +335,7 @@ void Node::MakeNotTerminal() {
       LOGFILE << "Successfully reverted " << terminal_visits << " visits at depth " << depth;
       depth++;
       // If wl != 0, we would have to switch signs at each depth.
-    } */
+    }
     n_ = 0;
   } else {
     // Any other case, just setting n_ = 0 is sufficient, as parent is root.
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -327,7 +327,7 @@ void Node::MakeNotTerminal() {
     const auto terminal_visits = n_;
     // Logging stuff for debugging purposes
     LOGFILE << "Attempting to revert terminal visits";
-    for (Node* node = *this; node != nullptr; node = node->GetParent()) {
+/*    for (Node* node = this; node != nullptr; node = node->GetParent()) {
       // Logging stuff for debugging purposes
       LOGFILE << "Attempting to revert " << terminal_visits << " visits at depth " << depth;
       // Revert all visits on twofold terminal when making it non terminal.
@@ -335,7 +335,7 @@ void Node::MakeNotTerminal() {
       LOGFILE << "Successfully reverted " << terminal_visits << " visits at depth " << depth;
       depth++;
       // If wl != 0, we would have to switch signs at each depth.
-    }
+    } */
   } else {
     // Any other case, just setting n_ = 0 is sufficient, as parent is root.
     n_ = 0;
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -325,6 +325,8 @@ void Node::MakeNotTerminal() {
     const auto d = d_;
     const auto m = m_;
     const auto terminal_visits = n_;
+    // Logging stuff for debugging purposes
+    LOGFILE << "Attempting to revert terminal visits";
     for (Node* node = this; node != nullptr; node = node->GetParent()) {
       // Logging stuff for debugging purposes
       LOGFILE << "Attempting to revert " << terminal_visits << " visits at depth " << depth;
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -327,7 +327,7 @@ void Node::MakeNotTerminal() {
     const auto terminal_visits = n_;
     // Logging stuff for debugging purposes
     LOGFILE << "Attempting to revert terminal visits";
-    for (Node* node = this; node != nullptr; node = node->GetParent()) {
+    for (Node* node = *this; node != nullptr; node = node->GetParent()) {
       // Logging stuff for debugging purposes
       LOGFILE << "Attempting to revert " << terminal_visits << " visits at depth " << depth;
       // Revert all visits on twofold terminal when making it non terminal.
@@ -430,8 +430,7 @@ void Node::RevertTerminalVisits(float v, float d, float m, int multivisit) {
     n_ -= multivisit;
   }
   // Best child is potentially no longer valid.
-  // Deactivating temporarily
-  // best_child_cached_ = nullptr;
+  best_child_cached_ = nullptr;
 }
 
 void Node::UpdateBestChild(const Iterator& best_edge, int visits_allowed) {
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -401,9 +401,9 @@ void Node::RevertTerminalVisits(float v, float d, float m, int multivisit) {
     m_ -= multivisit * (m - m_) / n_new;
     // Decrement N.
     n_ -= multivisit;
-    // Best child is potentially no longer valid.
-    best_child_cached_ = nullptr;
   }
+  // Best child is potentially no longer valid.
+  best_child_cached_ = nullptr;
 }
 
 void Node::UpdateBestChild(const Iterator& best_edge, int visits_allowed) {
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -429,9 +429,9 @@ void Node::RevertTerminalVisits(float v, float d, float m, int multivisit) {
     m_ -= multivisit * (m - m_) / n_new;
     // Decrement N.
     n_ -= multivisit;
+    // Best child is potentially no longer valid.
+    best_child_cached_ = nullptr;
   }
-  // Best child is potentially no longer valid.
-  best_child_cached_ = nullptr;
 }
 
 void Node::UpdateBestChild(const Iterator& best_edge, int visits_allowed) {
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -313,29 +313,7 @@ void Node::MakeTerminal(GameResult result, float plies_left, Terminal type) {
 
 void Node::MakeNotTerminal() {
   if (terminal_type_ == Terminal::TwoFold) {
-    // When reverting a terminal twofold repetition draw, this can happen
-    // in the tree. To keep consistency, we revert the effect of the visits
-    // to that terminal on all parent nodes and let PUCT revisit the nodes
-    // and fetch the evals without the twofold draw.
-    int depth = 0;
-    // Cache node's values as we reset them in the process. We could manually
-    // set wl and d, but if we want to reuse this for reverting other terminal
-    // nodes this is the way to go.
-    const auto wl = wl_;
-    const auto d = d_;
-    const auto m = m_;
-    const auto terminal_visits = n_;
-    // Logging stuff for debugging purposes
-    LOGFILE << "Attempting to revert terminal visits";
-    for (Node* node = this; node != nullptr; node = node->GetParent()) {
-      // Logging stuff for debugging purposes
-      LOGFILE << "Attempting to revert " << terminal_visits << " visits at depth " << depth;
-      // Revert all visits on twofold terminal when making it non terminal.
-      node->RevertTerminalVisits(wl, d, m + (float)depth, terminal_visits);
-      LOGFILE << "Successfully reverted " << terminal_visits << " visits at depth " << depth;
-      depth++;
-      // If wl != 0, we would have to switch signs at each depth.
-    }
+    // Special handling of reverting terminal nodes happens in search.cc now.
     n_ = 0;
   } else {
     // Any other case, just setting n_ = 0 is sufficient, as parent is root.
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -336,6 +336,7 @@ void Node::MakeNotTerminal() {
       depth++;
       // If wl != 0, we would have to switch signs at each depth.
     } */
+    n_ = 0;
   } else {
     // Any other case, just setting n_ = 0 is sufficient, as parent is root.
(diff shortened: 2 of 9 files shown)


src/mcts/search.cc
```diff
@@ -1214,9 +1226,24 @@ void SearchWorker::ExtendNode(Node* node) {
...
       return;
+    } else if (repetitions == 1 && twofolddrawlevel > 0) {
+      if (twofolddrawlevel == 3) {
+        // always mark as draw
+        node->MakeTerminal(GameResult::DRAW, 0.0f, Node::Terminal::TwoFold);
+      } else if (twofolddrawlevel == 2) {
+        // only mark as draw if depth >= 4
+        node->MakeTerminal(GameResult::DRAW, 0.0f, Node::Terminal::TwoFold);
+      } else if (twofolddrawlevel == 1) {
+        // check whether first repetition happened at root or in the tree
```
Naphthalin, line 1243: This function doesn't exist yet, but should be a simple copy+paste from `position.cc:GetRepetitions()`

src/mcts/search.cc
```diff
@@ -1065,6 +1065,18 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend(
...
       return NodeToProcess::Collision(node, depth, collision_limit);
     }
+    // probably best place to check for two-fold draws consistently
+    if (node->terminal_type_ == Node::Terminal::TwoFold &&
+                    ( params_.GetTwoFoldDrawLevel() > 0 ) ) {
+      if (params_.GetTwoFoldDrawLevel() == 2 && depth <= 3) {
+        // Level 2: no two-fold draw at depth 3 or lower
+        node->MakeNotTerminal();
+      } else if (params_.GetTwoFoldDrawLevel() == 1) {
+        // Level 1: check whether first repetition was before root
```
Naphthalin, line 1075: This is the spot where continuous checks for terminal positions could happen, provided they aren't too costly.

Naphthalin:
Current status:

- level 0 works as intended (no change in behavior)
- level 1 definitely doesn't work as intended, but this is the last to fix anyway
- level 2 currently behaves like level 3, resetting terminal nodes doesn't work for some reason.
- level 3 successfully marks twofold repetitions as terminal, but the displayed PV is one ply too long and adds a wrong move after the supposed terminal position.

Testing position: `position fen r4rk1/1b6/p6Q/2p5/8/3p2P1/PP5P/6K1 w - - 0 29`

Naphthalin:
One thing I already realized: Setting policy to 0% for terminal nodes has some strange interaction with twofold repetition draws.

liamlim:
Policy is set to zero only for terminal loses (not for terminal draws and terminal wins). I am curious how it affects two fold draws. I can't think of any way how these two features could interact with each other

src/mcts/search.cc
```diff
@@ -1291,9 +1306,28 @@ void SearchWorker::ExtendNode(Node* node) {
...
+    const auto repetitions = history_.Last().GetRepetitions();
+    const auto twofolddrawlevel = params_.GetTwoFoldDrawLevel();
+    if (repetitions >= 2) {
       node->MakeTerminal(GameResult::DRAW);
       return;
+    } else if (repetitions == 1 && twofolddrawlevel > 0) {
+      if (twofolddrawlevel == 3) {
+        // always mark as draw
+        LOGFILE << "== marked level 3 twofold draw == depth: " << depth;
+        node->MakeTerminal(GameResult::DRAW, 0.0f, Node::Terminal::TwoFold);
```
Mardak, line 1319: Nod. There seems to be some questionable changes here that were working fine in #1197, e.g., https://github.com/LeelaChessZero/lc0/pull/1197/files#diff-83b7141f4fa67bcfdf95188467895cb5R1224

Naphthalin: Thanks for catching that -- this probably is the reason for the PV being one ply too long. Fixed that.

Naphthalin:
@liamlim you are right, I must have missed this. However, I definitely had the problem that it displayed `U: 0.0000 S: 0.0000` for a draw, so something else is fishy.

src/mcts/search.cc
Naphthalin: Thanks again, it now behaves as expected, and the correct PV is given.

Naphthalin:
General update on current status:

- level 1+2: calculation of `pliessincefirstrepetition` seems to be correct, twofold repetitions are only marked when the condition is met. Correct PV (min. 4 plies) is given when analyzing a position where the first repetition of a position occurs in the past.
- level 3: thanks to @Mardak catching the missing `return;` statements, this now works correctly, gives the right PV etc.

Naphthalin:
I realized that test positions which include forced moves aren't good for debugging. New test position from a training game: `R7/8/6k1/7p/6pP/4PrP1/4K3/8 w - - 2 50`

Naphthalin:
PR is functional now and ready for review. Before merging, debug logging should be removed.

src/mcts/search.cc
Naphthalin: It does exist now.

Naphthalin:
Example position from TCEC S18 DivP, game 47 (Lc0 - Stoofvlees), move 32. Lc0 spent `10:31` to finally play `32. Kh2` which leads to repeating positions by `32. Kh2 Qe5 33. Kg1 Qe3`. Running 703810 for 1M nodes with default values in `position fen 7k/p1p2Qb1/1p1r1np1/7p/8/1N2q2P/PP3RP1/5RK1 w - - 0 32` gives
```
info string b3a1  (437 ) N:   17259 (+ 0) (P:  3.91%) (WL:  0.04859) (D: 0.509) (M: 79.8) (Q:  0.04859) (U: 0.03051) (S:  0.07910) (V:  -.----)
info string f7g6  (1533) N:  376725 (+ 0) (P: 23.09%) (WL:  0.07086) (D: 0.517) (M: 83.7) (Q:  0.07086) (U: 0.00825) (S:  0.07911) (V:  -.----)
info string g1h2  (157 ) N:  581587 (+46) (P:  3.34%) (WL:  0.07833) (D: 0.530) (M: 81.4) (Q:  0.07833) (U: 0.00077) (S:  0.07910) (V:  -.----)
info string node  (  30) N: 1000030 (+48) (P: 100.0%) (WL:  0.06565) (D: 0.522) (M: 83.5) (Q:  0.06565) (V:  -.----)
```

Running the same position with this PR and `--two-fold-draw-level=1` results in
```
info string g1h2  (157 ) N:   73022 (+ 0) (P:  3.34%) (WL:  0.02504) (D: 0.934) (M: 12.7) (Q:  0.02504) (U: 0.00617) (S:  0.03121) (V:  -.----)
info string b3a1  (437 ) N:  341523 (+ 0) (P:  3.91%) (WL:  0.02966) (D: 0.570) (M: 99.5) (Q:  0.02966) (U: 0.00154) (S:  0.03120) (V:  -.----)
info string f7g6  (1533) N:  554827 (+46) (P: 23.09%) (WL:  0.02563) (D: 0.651) (M: 62.6) (Q:  0.02563) (U: 0.00560) (S:  0.03124) (V:  -.----)
info string node  (  30) N: 1000108 (+47) (P: 100.0%) (WL:  0.01777) (D: 0.638) (M: 73.7) (Q:  0.01777) (V:  -.----)
```

Naphthalin:
Second example position from TCEC S18 DivP, game 57 (Lc0 - rofChade), move 36. Leela spent `15:53` to play `36. Nf3` which leads to repeating positions by `36. Nf3 Bc6 37. Nd4 Bd7`
`r3q1k1/3b4/4pQp1/3pP1Br/2nN4/1RP5/p4PP1/R5K1 w - - 5 36`

Naphthalin:
Different example from Chad's gigabook used for T72 training. Position is a suicidal rook giving a perpetual check to force a stalemate. 
```2R5/p4pk1/6p1/3p3p/3Pp2P/p3r1PK/5r2/8 w - - 0 40
1. d4 Nf6 2. Nf3 g6 3. c4 Bg7 4. Nc3 d5 5. g3 O-O 6. Bg2 c6 7. cxd5 cxd5 8. O-O Nc6 9. Qb3 Na5 10. Qc2 Bf5 11. Qd1 Ne4 12. Nxe4 Bxe4 13. e3 Rc8 14. Bd2 Nc4 15. Bc3 Qb6 16. Nd2 Bxg2 17. Nxc4 Rxc4 18. Kxg2 Rfc8 19. Rc1 Qa6 20. a3 e5 21. Qf3 e4 22. Qg4 h5 23. Qd7 Qc6 24. Qxc6 R8xc6 25. h4 b5 26. Ra1 Bf8 27. a4 bxa4 28. Ra3 Bxa3 29. bxa3 Rxc3 30. Rb1 Rb3 31. Ra1 Rcc3 32. Rd1 Rxa3 33. Rb1 Rab3 34. Ra1 a3 35. Ra2 Rb2 36. Ra1 Rcb3 37. Rc1 Rxe3 38. Rc8+ Kg7 39. Kh3 Rxf2 *
```
703810 value head is hopeless, with defaults the position is evaluated as `-5.0` at least until a few million nodes. With `--two-fold-draw-level=1` the initial evaluation is identical, but at a few 100k nodes evals drop below -2.0 (exact numbers vary):
370k: -1.43
470k: -0.90
760k: -0.50
1.5M: -0.27
3M: -0.17
At this point, the PV consists of the white rook chasing the black king without any repetitions until they reach the top right corner where the king can't escape a perpetual check without capturing the rook.

While this is obviously infeasible to be learned at training conditions, it shows that small search improvements can make a big difference in Leela's evaluation here.

borg323:
What about something like this:
```diff
diff --git a/src/chess/position.cc b/src/chess/position.cc
index feaa43f..2b01496 100644
--- a/src/chess/position.cc
+++ b/src/chess/position.cc
@@ -91,10 +91,13 @@ void PositionHistory::Append(Move m) {
   //                has a bug in implementation of emplace_back, when
   //                reallocation happens. (it also reallocates Last())
   positions_.push_back(Position(Last(), m));
-  positions_.back().SetRepetitions(ComputeLastMoveRepetitions());
+  int cycle_length;
+  int repetitions = ComputeLastMoveRepetitions(&cycle_length);
+  positions_.back().SetRepetitions(repetitions, cycle_length);
 }
 
-int PositionHistory::ComputeLastMoveRepetitions() const {
+int PositionHistory::ComputeLastMoveRepetitions(int* cycle_length) const {
+  *cycle_length = 0;
   const auto& last = positions_.back();
   // TODO(crem) implement hash/cache based solution.
   if (last.GetRule50Ply() < 4) return 0;
@@ -102,6 +105,7 @@ int PositionHistory::ComputeLastMoveRepetitions() const {
   for (int idx = positions_.size() - 3; idx >= 0; idx -= 2) {
     const auto& pos = positions_[idx];
     if (pos.GetBoard() == last.GetBoard()) {
+      *cycle_length = positions_.size() - 1 - idx;
       return 1 + pos.GetRepetitions();
     }
     if (pos.GetRule50Ply() < 2) return 0;
@@ -109,21 +113,6 @@ int PositionHistory::ComputeLastMoveRepetitions() const {
   return 0;
 }
 
-int PositionHistory::ComputePliesSinceFirstRepetition() const {
-  const auto& last = positions_.back();
-  // copied TODO(crem) implement hash/cache based solution.
-  if (last.GetRule50Ply() < 4) return 0;
-
-  for (int idx = positions_.size() - 3; idx >= 0; idx -= 2) {
-    const auto& pos = positions_[idx];
-    if (pos.GetBoard() == last.GetBoard()) {
-      return positions_.size() - 1 - idx;
-    }
-    if (pos.GetRule50Ply() < 2) return 0;
-  }
-  return 0;
-}
-
 bool PositionHistory::DidRepeatSinceLastZeroingMove() const {
   for (auto iter = positions_.rbegin(), end = positions_.rend(); iter != end;
        ++iter) {
diff --git a/src/chess/position.h b/src/chess/position.h
index 496a753..7530fa7 100644
--- a/src/chess/position.h
+++ b/src/chess/position.h
@@ -49,9 +49,15 @@ class Position {
   // How many time the same position appeared in the game before.
   int GetRepetitions() const { return repetitions_; }
 
+  // How many half-moves since the same position appeared in the game before.
+  int GetPliesSinceFirstRepetition() const { return cycle_length_; }
+
   // Someone outside that class knows better about repetitions, so they can
   // set it.
-  void SetRepetitions(int repetitions) { repetitions_ = repetitions; }
+  void SetRepetitions(int repetitions, int cycle_length) {
+    repetitions_ = repetitions;
+    cycle_length_ = cycle_length;
+  }
 
   // Number of ply with no captures and pawn moves.
   int GetRule50Ply() const { return rule50_ply_; }
@@ -73,6 +79,8 @@ class Position {
   int rule50_ply_ = 0;
   // How many repetitions this position had before. For new positions it's 0.
   int repetitions_;
+  // How many half-moves since the position was repeated or 0.
+  int cycle_length_;
   // number of half-moves since beginning of the game.
   int ply_count_ = 0;
 };
@@ -124,11 +132,8 @@ class PositionHistory {
   // Checks for any repetitions since the last time 50 move rule was reset.
   bool DidRepeatSinceLastZeroingMove() const;
 
-  int ComputePliesSinceFirstRepetition() const;
-
-
  private:
-  int ComputeLastMoveRepetitions() const;
+  int ComputeLastMoveRepetitions(int* cycle_length) const;
 
   std::vector<Position> positions_;
 };
diff --git a/src/mcts/search.cc b/src/mcts/search.cc
index 756ee64..b08b0d9 100644
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -1323,7 +1323,7 @@ void SearchWorker::ExtendNode(Node* node, int depth) {
         // only mark as draw if depth of extended node is >= 4
         validtwofold = true;
       } else if (twofolddrawlevel == 1 && depth - 1 >= 4 && depth - 1 >=
-                 history_.ComputePliesSinceFirstRepetition()) {
+                 history_.Last().GetPliesSinceFirstRepetition()) {
         // check whether first repetition happened at root or in the tree
         // don't mark as draw if repetition happened in the game history
         validtwofold = true;
@@ -1331,7 +1331,7 @@ void SearchWorker::ExtendNode(Node* node, int depth) {
     }
     // if node turned out to be a valid twofold, mark it as Terminal::Twofold
     if (validtwofold) {
-      const auto cyclelength = history_.ComputePliesSinceFirstRepetition();
+      const auto cyclelength = history_.Last().GetPliesSinceFirstRepetition();
       // logging for debugging purpose
       LOGFILE << "== marked twofold draw == depth: " << depth - 1;
       LOGFILE << "== plies since first repetition: " << cyclelength;
```

src/mcts/search.cc
```diff
@@ -1291,9 +1307,38 @@ void SearchWorker::ExtendNode(Node* node) {
...
+      }
+    }
+    // if node turned out to be a valid twofold, mark it as Terminal::Twofold
+    if (validtwofold) {
+      const auto cyclelength = history_.ComputePliesSinceFirstRepetition();
+      // logging for debugging purpose
+      LOGFILE << "== marked twofold draw == depth: " << depth - 1;
+      LOGFILE << "== plies since first repetition: " << cyclelength;
+      // use plies since first repetition as moves left; exact if forced draw.
+      node->MakeTerminal(GameResult::DRAW, (float)cyclelength,
```
Mardak, line 1339: Clever stashing the value to avoid needing the position later. Will this have strange moves-left-effect behavior? Does this value want to be doubled assuming the best move to play from a 2-fold is to repeat the cycle into 3-fold? Otherwise a winning position might try to give a boost to this shorter-looking move?

Also, I don't think the cast is necessary.

src/mcts/search.cc
```diff
@@ -1291,9 +1307,38 @@ void SearchWorker::ExtendNode(Node* node) {
...
+        validtwofold = true;
+      } else if (twofolddrawlevel == 1 && depth - 1 >= 4 && depth - 1 >=
+                 history_.ComputePliesSinceFirstRepetition()) {
+        // check whether first repetition happened at root or in the tree
+        // don't mark as draw if repetition happened in the game history
+        validtwofold = true;
+      }
+    }
+    // if node turned out to be a valid twofold, mark it as Terminal::Twofold
+    if (validtwofold) {
```
Mardak, line 1333: You could just combine the conditions above:
```
if (rep == 1 && level > 0 &&
    (level == 1 && depth …) ||
    (level == 2 && depth >= 3)) {
  MakeTerminal();
  return;
}
```

src/mcts/search.cc
Naphthalin: If I understand your concern correctly: It actually is the correct `m_` assuming that repetition of the move is forced from here. The 2-fold is set as a terminal with `m_=cyclelength` (instead of 0 as other terminals). Also, MLH with Q scaling doesn't calculate any bonus for draws, unless we use `drawscore`.

src/mcts/search.cc
Naphthalin: Yes, very sensible suggestion. The current code organization is still from when I thought things were more complicated, I will reorganize it.

Naphthalin:
@borg323 do I understand it correctly that this optimization uses more memory, as the `cycle_length` is saved somewhere? But yes, this is exactly the possible performance optimization for not doing the same loop twice the current implementation should be measured against; also, possibly the `repetitions` field could simply be 0 for new positions, 2 for terminals, and `-cycle_length` for twofold repetitions if we want to save memory at the expense of code obfuscation.

borg323:
Yes, it needs an extra int in Position, but I don't think this is something to worry about. The performance gain is minor (if any) I just think it is cleaner this way. Still, this is something that can be done later.

Naphthalin:
Yes, I agree with a Get, Set and Compute split there even if only for style reasons.

Mardak:
> this is exactly the possible performance optimization for not doing the same loop twice the current implementation

To be clear, the loop runs 3 times for a level-1 2-fold whereas the suggestion in https://github.com/LeelaChessZero/lc0/pull/1375#issuecomment-659639892 only runs 1 time and also avoids copy/pasted code.

src/mcts/search.cc
```diff
@@ -1291,9 +1307,31 @@ void SearchWorker::ExtendNode(Node* node) {
       return;
     }
 
-    if (history_.Last().GetRepetitions() >= 2) {
+    const auto repetitions = history_.Last().GetRepetitions();
+    const auto twofolddrawlevel = params_.GetTwoFoldDrawLevel();
+    bool validtwofold = false;
```
liamlim, line 1312: this is no longer necessary, right?

src/mcts/node.h
```diff
@@ -165,6 +165,7 @@ class Node {
   // Returns whether the node is known to be draw/lose/win.
   bool IsTerminal() const { return terminal_type_ != Terminal::NonTerminal; }
   bool IsTbTerminal() const { return terminal_type_ == Terminal::Tablebase; }
+  bool IsTwofoldTerminal() const { return terminal_type_ == Terminal::TwoFold; }
```
liamlim, line 168: Isn't IsTwoFoldTerminal more consistent?

src/mcts/node.cc
```diff
@@ -293,7 +293,7 @@ void Node::SortEdges() {
 }
 
 void Node::MakeTerminal(GameResult result, float plies_left, Terminal type) {
-  SetBounds(result, result);
+  if (type != Terminal::TwoFold) { SetBounds(result, result); }
```
Mardak, line 296: clang-format indent style

src/mcts/params.h
```diff
@@ -53,6 +53,7 @@ class SearchParams {
   float GetCpuctFactor(bool at_root) const {
     return at_root ? kCpuctFactorAtRoot : kCpuctFactor;
   }
+  int GetTwoFoldDrawLevel() const {return options_.Get<int>(kTwoFoldDrawLevelId); }
```
Mardak, line 56: clang-format indent style

src/mcts/search.cc
```diff
@@ -1145,6 +1145,22 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend(
...
     }
+    // Probably best place to check for two-fold draws consistently.
+    // Depth starts with 1 at root, so real depth is depth - 1
+    if (node->IsTwofoldTerminal() && ( params_.GetTwoFoldDrawLevel() > 0 ) ) {
+      LOGFILE << "== encountered twofold draw == depth: " << depth - 1;
+      if (params_.GetTwoFoldDrawLevel() == 2 && depth - 1 <= 3) {
+        // Level 2: no two-fold draw at depth 3 or lower
+        LOGFILE << "== level 2 twofold draw reset ==";
+        node->MakeNotTerminal();
+      } else if (params_.GetTwoFoldDrawLevel() == 1 && depth - 1 < node->GetM()) {
```
Mardak, line 1156: clang-format 80 character width

src/mcts/search.cc
```diff
@@ -1145,6 +1145,22 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend(
       }
       return NodeToProcess::Collision(node, depth, collision_limit);
     }
+    // Probably best place to check for two-fold draws consistently.
+    // Depth starts with 1 at root, so real depth is depth - 1
```
Mardak, line 1149: Comment style is sentences with period

src/mcts/search.cc
```diff
@@ -1145,6 +1145,22 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend(
...
+    if (node->IsTwofoldTerminal() && ( params_.GetTwoFoldDrawLevel() > 0 ) ) {
```
Mardak: nit: parens around `>` expression aren't necessary and clang-format whitespace

src/mcts/search.cc
```diff
@@ -1145,6 +1145,22 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend(
...
+      LOGFILE << "== encountered twofold draw == depth: " << depth - 1;
+      if (params_.GetTwoFoldDrawLevel() == 2 && depth - 1 <= 3) {
+        // Level 2: no two-fold draw at depth 3 or lower
+        LOGFILE << "== level 2 twofold draw reset ==";
+        node->MakeNotTerminal();
+      } else if (params_.GetTwoFoldDrawLevel() == 1 && depth - 1 < node->GetM()) {
+        // Level 1: check whether first repetition was before root
+        // Length of repetition was stored in m_, so we don't need to calculate
+        LOGFILE << "== level 1 twofold draw reset ==";
+        node->MakeNotTerminal();
```
Mardak, line 1160: You can also combine the multiple branch bodies with a single condition as in the other code path. Also Tilps' comment about `MakeNotTerminal` not fixing up ancestors still applies as this happens for nodes not at root now: https://github.com/LeelaChessZero/lc0/pull/1197#discussion_r436212969

src/mcts/search.cc
```diff
@@ -1145,6 +1145,22 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend(
...
+      }
+
```
Mardak, line 1162: clang-format unnecessary whitespace

src/mcts/search.h
```diff
@@ -303,7 +303,7 @@ class SearchWorker {
   };
 
   NodeToProcess PickNodeToExtend(int collision_limit);
-  void ExtendNode(Node* node);
+  void ExtendNode(Node* node, int depth = 0);
```
Mardak, line 306: The default value isn't necessary.

src/mcts/search.cc
Naphthalin: Thanks, I obviously missed that.

src/mcts/node.h
Naphthalin: You are right. Corrected it.

src/mcts/node.h
```diff
@@ -165,6 +165,7 @@ class Node {
   // Returns whether the node is known to be draw/lose/win.
   bool IsTerminal() const { return terminal_type_ != Terminal::NonTerminal; }
   bool IsTbTerminal() const { return terminal_type_ == Terminal::Tablebase; }
+  bool IsTwoFoldTerminal() const { return terminal_type_ == Terminal::TwoFold; }
```
Mardak, line 168: Previously I believe crem suggested not adding the helper method yet for just a single use, so just inline it instead.

Naphthalin:
@Mardak Thanks for the detailed feedback. Will go through it one by one, and comment/ask if anything is unclear to me. Also thx for bringing up old memories with that CCC game, I appreciate it :)

#710 mentions a CCC4 game with a "suicidal rook" perpetual/stalemate. For reference, the pgn is
```
1. e4 c6 2. d4 d5 3. e5 Bf5 4. Be2 e6 5. Nf3 Bb4+ 6. c3 Be7 7. Be3 Nd7 8. O-O
Bg6 9. Nbd2 Nh6 10. Bxh6 gxh6 11. Nb3 f6 12. exf6 Bd6 13. Qd2 Qxf6 14. Qxh6
O-O-O 15. Rae1 Rhg8 16. Bd1 Rdf8 17. g3 Bf5 18. Qxf6 Nxf6 19. Nh4 Bh3 20. Ng2 h5
21. Re3 h4 22. Nxh4 Bxf1 23. Kxf1 Kd7 24. Kg2 Ng4 25. Bxg4 Rxg4 26. Nf3 b6
27. h3 Rg7 28. Nc1 Bf4 29. Re1 Bxc1 30. Rxc1 c5 31. g4 Kd6 32. Rc2 Rf4 33. Kg3
Re4 34. Rd2 Ke7 35. Ne5 Re1 36. h4 a5 37. h5 Rg1+ 38. Kf4 Rg8 39. f3 Kf6 40. Rh2
Kg7 41. Ng6 Kf6 42. Ne5 Kg7 43. Ng6 Rf1 44. Re2 Re8 45. Ne5 Rf8+ 46. Kg5 Rg1
47. Ng6 Rf5+ 48. Kh4 Rh1+ 49. Kg3 Rf6 50. Nh4 cxd4 51. cxd4 Rg1+ 52. Ng2 Rh6
53. Kh2 Rd1 54. Nf4 Rf6 55. Nxe6+ Kf7 56. Ng5+ Kg8 57. Kg3 Rg1+ 58. Kh2 Ra1
59. a3 Rd1 60. Kg3 Rg1+ 61. Kh3 b5 62. Re8+ Rf8 63. Re6 Rd1 64. h6 b4 65. axb4
axb4 66. Kh4 Rh1+ 67. Kg3 b3 68. Rb6 Rg1+ 69. Kh2 Ra1 70. Rxb3 Ra6 71. h7+ Kh8
72. Kg3 Raa8 73. f4 Rab8 74. Rc3 Rxb2 75. f5 Rb7 76. Kf4 Re7 77. Rc6 Re2 78. Rd6
Rf2+ 79. Kg3 Rf1 80. Rxd5 Rg1+ 81. Kf4 Rf1+ 82. Ke3 Re8+ 83. Re5 Rf8 84. Re6 Rd1
85. Rd6 Re8+ 86. Kf3 Rf1+ 87. Kg2 Rf4 88. Kg3 Rf1 89. Re6 Rb8 90. Rg6 Rb3+
91. Kg2 Rb2+ 92. Kxf1 *
```
and the game losing (well, drawing) move happens in position `1r5k/7P/4R3/5PN1/3P2P1/6K1/8/5r2 w - - 19 90` with `90. Rg6`. After this move, 703810 with `--two-fold-draw-level=1` already prefers the correct continuation `90. ... Rb3+` at 500 nodes, giving it D=0.13, at 1k nodes it is at D=0.52, 5k nodes D=0.74, and quickly rising. So it seems that enabling this in training could already be enough to learn this difficult type of perpetual check.

src/mcts/node.cc
Naphthalin: What does this mean, and what is the suggested change?

src/mcts/params.h
Naphthalin: What does this mean, and what is the suggested change?

src/mcts/search.cc
Naphthalin: Changed (I hope).

src/mcts/search.cc
Naphthalin: Combined the statements into a single one.

src/mcts/search.h
Naphthalin: Originally I didn't realize that there is only 1 place where `ExtendNode()` is called, which I modified.

src/mcts/node.h
Naphthalin: What do you mean by this?

Naphthalin:
I am yet to be convinced that reverting the visits on a terminal node when making the node non terminal really is the correct way of dealing with it; after all it is a search heuristic. It would be nice to have a consistent tree though w.r.t. visits and evals, so this probably is a general question, and not only for twofold draws.

src/mcts/params.h
liamlim: I guess it means that space before return is expected. At least I would expect a space there

src/mcts/node.h
borg323: In this case I believe having helpers for all is better for consistency.

Naphthalin: I agree, consistency here is preferred. I will mark this as resolved then.

src/mcts/node.cc
borg323: Although the style guide is not absolutely clear about this, at least for consistency either don't add braces here or put `SetBounds(result, result);` on the next line.

borg323:
We should start thinking about the command line (and uci) option(s) to use. I think this should be just a boolean flag `--two-fold-draws` choosing between current levels 0 and 1, but it seems some people think level 3 is also useful. In that case we can use either a 3 valued choice option, choosing between say `none`, `correct` and `excessive`, or a second boolean option.

Mardak:
Yeah, a Boolean flag will probably be best assuming performance isn't regressed compared to master when two fold is disabled and check with enabled too. This will require some cleanup of the code first.

Naphthalin:
I agree with that, and given that 

- the current level 1 is the theoretically correct solution,
- its implementation doesn't cost more than the simplified version,
- and current level 3 has a very specific usecase I doubt will apply to anyone,

I am d'accord with reducing it to a boolean flag between current level 0 / level 1. I will do the cleanup and remove the logging in the process. However, I will probably need some help with the implementation of reverting visits when resetting a node to NonTerminal, as we (according to Tilps) have to invalidate bestedge/... cache as well.

src/mcts/search.cc
```diff
@@ -1145,6 +1145,20 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend(
...
+          (params_.GetTwoFoldDrawLevel() == 1 && depth - 1 < node->GetM())) {
+        // Level 1: check whether first repetition was before root
+        // Length of repetition was stored in m_.
+        // TODO: Cleaner alternative to call position history?
+        LOGFILE << "== twofold draw reset ==";
+        node->MakeNotTerminal();
+      }
+    }
     // Either terminal or unexamined leaf node -- the end of this playout.
     if (node->IsTerminal() || !node->HasChildren()) {
```
Mardak, line 1163: If you notice a performance issue, you can try moving the two fold not terminal check under terminal check as a node can only be two fold if a terminal.

Naphthalin: The idea behind making this into a separate statement was that when a twofold terminal is made non terminal through tree reuse, we already want the first visit to it be processed as a "regular" visit on that node, so we mark it non terminal on the fly and continue then.

src/mcts/node.cc
Naphthalin: Removed the braces.

src/mcts/params.h
Naphthalin: Added the space resp. moved the return statement to the next line as it exceeded 80 characters.

Naphthalin:
Should the boolean `TwoFoldDraws` be included in the "cached parameter section" as it is accessed rather often?

src/mcts/search.cc
```diff
@@ -1291,9 +1302,20 @@ void SearchWorker::ExtendNode(Node* node) {
       return;
     }
 
-    if (history_.Last().GetRepetitions() >= 2) {
+    const auto repetitions = history_.Last().GetRepetitions();
+    const auto twofolddraws = params_.GetTwoFoldDraws();
```
borg323, line 1306: This is only accessed once in the `if`, better move it there.

src/chess/position.h
```diff
@@ -49,9 +49,15 @@ class Position {
   // How many time the same position appeared in the game before.
   int GetRepetitions() const { return repetitions_; }
 
+  // How many half-moves since the same position appeared in the game before.
+  int GetPliesSinceFirstRepetition() const { return cycle_length_; }
```
borg323, line 53: This is probably my fault, but I think `GetPliesSincePrevRepetition` is a more accurate name.

src/mcts/search.cc
Naphthalin: Agreed. Is it also possible to calculate `cyclelength` within the if statement and use it later? Right now, it is accessed twice for actual twofold draws.

src/chess/position.h
Naphthalin: Renamed it. Thanks :)

src/mcts/search.cc
```diff
@@ -1145,6 +1145,17 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend(
       }
       return NodeToProcess::Collision(node, depth, collision_limit);
     }
+    // Probably best place to check for two-fold draws consistently.
+    // Depth starts with 1 at root, so real depth is depth - 1.
+    if (node->IsTwoFoldTerminal() && params_.GetTwoFoldDraws()) {
+      if (depth - 1 < node->GetM()) {
```
Naphthalin, line 1151: @borg323 do you think it would be better to call `history_.Last().GetPliesSincePrevRepetition()` here?

src/mcts/search.cc
borg323: I hope the compiler will optimize it, but probably needs a `const` qualifier for that.

src/mcts/search.cc
Mardak: Yes, I realize the latter branch still wants to check if it's still a terminal, but this is one of the hottest loops in search, and checking if every node is a two fold involves loading the bitfield then masking and checking for the specific enum value whereas the terminal check is slightly faster because it's the common case default value 0. But might not need optimizing in the first place.

src/mcts/search.cc
borg323: Now that I understand it, I like it the way it is, but I admit it took some explaining to get there.

Mardak: There isn't history, which would need to regenerated to recompute the cycle length

src/mcts/search.cc
```diff
@@ -1291,9 +1302,20 @@ void SearchWorker::ExtendNode(Node* node) {
...
+    const auto repetitions = history_.Last().GetRepetitions();
+    // Mark two-fold repetitions as draws according to settings.
+    // Depth starts with 1 at root, so number of plies in PV is depth - 1.
+    if (repetitions >= 2) {
       node->MakeTerminal(GameResult::DRAW);
       return;
+    } else if (repetitions == 1 && params_.GetTwoFoldDraws() &&
+               depth - 1 >= 4 &&
+               depth - 1 >= history_.Last().GetPliesSincePrevRepetition()) {
+      const auto cyclelength = history_.Last().GetPliesSincePrevRepetition();
```
Mardak: At least clang seems to directly reuse this value now because it's just a cached value on the position instead of computed.

src/mcts/search.cc
borg323: The following seems marginally faster. Note that I also removed the unnecessary `params_.GetTwoFoldDraws()` call.
```diff
diff --git a/src/mcts/search.cc b/src/mcts/search.cc
index 224d080..6c0ea35 100644
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -1145,21 +1145,20 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend(
       }
       return NodeToProcess::Collision(node, depth, collision_limit);
     }
-    // Probably best place to check for two-fold draws consistently.
-    // Depth starts with 1 at root, so real depth is depth - 1.
-    if (node->IsTwoFoldTerminal() && params_.GetTwoFoldDraws()) {
-      if (depth - 1 < node->GetM()) {
+    // Either terminal or unexamined leaf node -- the end of this playout.
+    if (node->IsTerminal() || !node->HasChildren()) {
+      // Probably best place to check for two-fold draws consistently.
+      // Depth starts with 1 at root, so real depth is depth - 1.
+      if (node->IsTwoFoldTerminal() && depth - 1 < node->GetM()) {
         // Check whether first repetition was before root. If yes, remove
         // terminal status of node.
         // Length of repetition was stored in m_.
         // TODO: Cleaner alternative to call position history?
         node->MakeNotTerminal();
+      } else {
+        return NodeToProcess::Visit(node, depth);
       }
     }
-    // Either terminal or unexamined leaf node -- the end of this playout.
-    if (node->IsTerminal() || !node->HasChildren()) {
-      return NodeToProcess::Visit(node, depth);
-    }
     Node* possible_shortcut_child = node->GetCachedBestChild();
     if (possible_shortcut_child) {
       // Add two here to reverse the conservatism that goes into calculating the
```

src/mcts/search.cc
Naphthalin: OK, thanks for your answers. This seems to be the correct solution then, and we assume that `GetM()` will be available from now on anyway.

src/mcts/search.cc
Naphthalin: Thanks for checking, I appreciate it!

src/mcts/search.cc
```diff
@@ -1291,9 +1302,20 @@ void SearchWorker::ExtendNode(Node* node) {
...
 
-    if (history_.Last().GetRepetitions() >= 2) {
+    const auto repetitions = history_.Last().GetRepetitions();
+    // Mark two-fold repetitions as draws according to settings.
+    // Depth starts with 1 at root, so number of plies in PV is depth - 1.
+    if (repetitions >= 2) {
       node->MakeTerminal(GameResult::DRAW);
       return;
+    } else if (repetitions == 1 && params_.GetTwoFoldDraws() &&
+               depth - 1 >= 4 &&
```
borg323, line 1312: This is probably no longer helping, now that `GetPliesSincePrevRepetition()` is very lightweight.

Naphthalin: That is of course correct. I think my rationale behind keeping it was that once we try to implement the search as a DAG we have to fall back to something like this, but for one this isn't a reason to keep code confusing, and also I likely won't forget it anyway.
There is no situation where we expect `GetPliesSincePrevRepetition()` to fail here, right?

src/mcts/search.cc
Naphthalin: This seems to be the case according to @Mardak. Marking this as resolved then.

src/mcts/search.cc
Naphthalin: Yeah, I just blindly replaced all `Level` calls by the new boolean, not realizing that in this case, the existence of a twofold terminal implies that we in fact work with twofold terminals. Will implement these suggested changes.

src/mcts/search.cc
borg323: `GetPliesSincePrevRepetition()` is now always consistent with `GetRepetitions()`.

Naphthalin: Thanks for checking. Marking this as resolved.

src/mcts/search.cc
Naphthalin: I think I have to understand the logic again here. Does a move which isn't a twofold terminal draw anymore have children or not? I hope to be correct in what we expect here is that we can expand it in the same visit which then gets that node already its first visit and explores a child.

Naphthalin:
Ok, it seems not everything is working as intended right now. Running the position `7k/p1p2Qb1/1p1r1np1/7p/8/1N2q2P/PP3RP1/5RK1 w - - 0 32` in Nibbler, focusing `32. Kh2` and then playing it produces insane amounts of nps, but to no apparent child node. Any idea why?

Naphthalin:
Still doesn't work, will have to figure it out tomorrow.

src/mcts/search.cc
Naphthalin: I came to the conclusion that this reordering doesn't handle it correctly if a terminal twofold after being made non terminal has no edges (which it doesn't); logic should return that node for expanding, but it doesn't.

borg323: Right, this case was indeed wrong.

Naphthalin:
I can't reproduce the crashs from yesterday at all, things are working as intended -- with the only exception that while the visits are reverted in the tree, they aren't reverted in the search node budget.

src/mcts/node.cc
```diff
@@ -312,8 +312,14 @@ void Node::MakeTerminal(GameResult result, float plies_left, Terminal type) {
...
 void Node::MakeNotTerminal() {
+  if (terminal_type_ == Terminal::TwoFold) {
+    // Special handling of reverting terminal nodes happens in search.cc now.
+    n_ = 0;
+  } else {
+    // Any other case, just setting n_ = 0 is sufficient, as parent is root.
+    n_ = 0;
+  }
   terminal_type_ = Terminal::NonTerminal;
-  n_ = 0;
```
borg323, line 316: This change in not needed any more.

src/mcts/search.cc
borg323: Was this intentionally reverted?

src/mcts/node.cc
Naphthalin: Changed, thanks.

src/mcts/search.cc
Naphthalin: There were 2 points where I checked for `depth - 1 >= 4` before checking the actual condition, here and when making it non terminal after tree reuse. We removed the additional check there because it was 0 cost to do the direct check, but here I can't judge by myself whether calling history is more or less costly than the additional check.

src/mcts/search.cc
Naphthalin: We could still move it inside the `IsTerminal()` condition, and add another `return NodeToProcess::Visit(node, depth)` after checking for no children; not sure whether that's cheaper than current double checking.

borg323: I tried a different approach, moving the `!node->HasChildren()` in a separate if statement, as I think this is the cleaner alternative, but I didn't like the added complexity.

src/mcts/search.cc
borg323: In that case, moving it before `params_.GetTwoFoldDraws()` is probably an additional small gain.

src/mcts/search.cc
Naphthalin: Changed it the way you describe. It now fits the logic better: If a node is terminal, we check whether it should stay terminal, and either make it non terminal (twofold draw after playing into the variation and tree reuse) or return `NodeToProcess::Visit(node, depth)`. Anything which survives the statement isn't a terminal, and we continue with the regular checks.

src/mcts/search.cc
Naphthalin: I agree, especially given that I think `two-fold-draws` should be `true` per default. Changed the order.

src/mcts/search.cc
Naphthalin: In theory, a twofold terminal which just got reverted won't have children anyway, so we could return the `NodeToProcess` directly without checking for children. However, this case will be so rare that the additional code can't be considered to be an optimization, only a code obfuscation.

borg323:
The default for selfplay should be off, this can be done with a line like <https://github.com/LeelaChessZero/lc0/blob/master/src/selfplay/tournament.cc#L133> right after it.

src/mcts/node.cc
```diff
@@ -375,11 +375,37 @@ void Node::AdjustForTerminal(float v, float d, float m, int multivisit) {
...
+    n_ = 0;
+  } else {
+    // Recompute Q and M.
+    wl_ -= multivisit * (v - wl_) / n_new;
+    d_ -= multivisit * (d - d_) / n_new;
+    m_ -= multivisit * (m - m_) / n_new;
+    // Decrement N.
+    n_ -= multivisit;
+    // Best child is potentially no longer valid.
+    best_child_cached_ = nullptr;
```
Tilps, line 405: This should probably be done unconditionally, not only in the n_new > 0 case. Just for safety sake, I do see in practice that the if branch is only taken for a leaf in practice and hence there will be no best_child_cached_.

src/mcts/search.cc
```diff
@@ -1145,8 +1145,49 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend(
...
+        // Length of repetition was stored in m_.
+        int depth_counter = 0;
+        // Cache node's values as we reset them in the process. We could
+        // manually set wl and d, but if we want to reuse this for reverting
+        // other terminal nodes this is the way to go.
+        const auto wl = node->GetWL();
+        const auto d = node->GetD();
+        const auto m = node->GetM();
+        const auto terminal_visits = node->GetN();
+        LOGFILE << "== reverting " << terminal_visits << " visits to " <<
```
Tilps, line 1165: I strongly suggest removing this log line before submission.

src/mcts/search.cc
```diff
@@ -1145,8 +1145,49 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend(
...
+              "a twofold terminal at depth " << depth - 1 << " ==";
+        for (Node* node_to_revert = node; node_to_revert != nullptr;
+                        node_to_revert = node_to_revert->GetParent()) {
+          // Revert all visits on twofold draw when making it non terminal.
+          node_to_revert->RevertTerminalVisits(wl, d,
+                          m + (float)depth_counter, terminal_visits);
```
Tilps, line 1171: This looks like some pretty non-standard wrapping.  I suggest you run clang-format.

src/mcts/search.cc
```diff
@@ -1145,8 +1145,49 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend(
...
-    // Either terminal or unexamined leaf node -- the end of this playout.
-    if (node->IsTerminal() || !node->HasChildren()) {
+    // If terminal, we either found a twofold draw to be reverted, or
+    // reached the end of this playout.
+    if (node->IsTerminal()) {
+      // Probably best place to check for two-fold draws consistently.
+      // Depth starts with 1 at root, so real depth is depth - 1.
+      if (node->IsTwoFoldTerminal() && depth - 1 < node->GetM()) {
+        // Check whether first repetition was before root. If yes, remove
+        // terminal status of node and revert all visits in the tree.
```
Tilps, line 1155: This comment really should be before the if.

Also probably worth mentioning that first repetition for a two fold terminal can only be before root under tree reuse from a previous move.

src/mcts/node.cc
Naphthalin: Done.

src/mcts/search.cc
Naphthalin: Done.

src/mcts/search.cc
Naphthalin: Done.

src/mcts/search.cc
Naphthalin: Hopefully done.

src/mcts/search.cc
```diff
@@ -1162,13 +1163,11 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend(
...
         const auto terminal_visits = node->GetN();
-        LOGFILE << "== reverting " << terminal_visits << " visits to " <<
-              "a twofold terminal at depth " << depth - 1 << " ==";
         for (Node* node_to_revert = node; node_to_revert != nullptr;
                         node_to_revert = node_to_revert->GetParent()) {
           // Revert all visits on twofold draw when making it non terminal.
-          node_to_revert->RevertTerminalVisits(wl, d,
-                          m + (float)depth_counter, terminal_visits);
+          node_to_revert->RevertTerminalVisits(wl, d, m + (float)depth_counter,
+                                                terminal_visits);
```
Tilps, line 1170: This still looks wrong - I also think the for statement wrapping is non-standard too.

Naphthalin: Done.

PR #1376, opened 2020-06-28
Sort edges to reduce cpu load by allowing an early out in the inner pick loop.

Tilps:
This has 'obvious' improvements in cpu profiling - and small but measurable improvements (2-3%) with a 8 million node search of startpos T70 running on 2080TI on a fast intel cpu using windows PGO build.  8 million node search appears more gpu bound than cpu bound for most of the time.

FetchSingleNodeResult is definitely slower (maybe 30%), but the gains in PickNode are much better and clearly offset it in the profile traces I've gathered.  Maybe the sort could be optimized a bit more for small arrays - but it already seems a win.

NOTE:
This is not purely an nps gain. By changing the sort order of the edges it changes the order of tiebreaks in the pick inner loop.
I think the new mode is 'logically' better - but I'll try to run a before/after elo test just in case.
For reference - tie breaks are less about movegen order now.  They are more often based on policy, and in the case that policy is equal it is 'undefined' what order the tiebreak is since it depends on the implementation of std::sort.

### DIFF
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -283,6 +283,15 @@ bool Node::MakeSolid() {
   return true;
 }
 
+void Node::SortEdges() {
+  assert(edges_);
+  assert(!child_);
+  // Sorting on raw p_ is the same as sorting on GetP() as a side effect of
+  // the encoding, and its noticeably faster.
+  std::sort(edges_.get(), (edges_.get() + num_edges_),
+            [](const Edge& a, const Edge& b) { return a.p_ > b.p_; });
+}
+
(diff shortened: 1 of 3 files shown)


Tilps:
Fixed node regression check.
Score of lc0 dev local build vs lc0 newbaseline: 407 - 394 - 1199 [0.503]
Elo difference: 2.26 +/- 9.62

2000 of 2000 games finished.
Seems close to a no-op as expected.

oscardssmith:
What's the NPS difference?

Tilps:
It is going to depend on the situation a lot - I gave a measured number in the OP for one case.

PR #1377, opened 2020-06-28
New parameter to try and increase gpu batch fullness without incurring as much out of order risk.

Tilps:
This will need tuning at time control and a fully time control based elo check. Probably should tune in conjunction with MOOOEF and collision events.
I have no idea when 'I' will be able to give that a go. Maybe someone else will be able to help me out ;)

Note that setting it to 1 is not the same as the previous behavior! Previously collection would abort when out of order ran out even if the batch itself was tiny!  There is no preserving of previous behavior in any scenario that out of order was enabled before.
Setting to 1 is a no-op if out of order is disabled, like it is for training. So this should be 'safe' for training.

I do think the idea has merit though, hopefully testing agrees...

### DIFF
--- a/src/mcts/params.cc
+++ b/src/mcts/params.cc
@@ -177,6 +177,10 @@ const OptionId SearchParams::kMaxOutOfOrderEvalsId{
     "max-out-of-order-evals-factor", "MaxOutOfOrderEvalsFactor",
     "Maximum number of out of order evals during gathering of a batch is "
     "calculated by multiplying the maximum batch size by this number."};
+const OptionId SearchParams::kMaxNotOutOfOrderId{
+    "max-not-out-of-order-factor", "MaxNotOutOfOrderFactor",
+    "Maximum number of not out of order entries during gathering of a batch is "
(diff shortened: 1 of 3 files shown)


Tilps:
As a random example - with a MOOOEF of 2.4 and a position from sufi18 that is known to cause a large amount of small batches due to MOOOEF not being high enough, I got a 10% nps increase with T70 on my fast single core CPU and RTX 2080TI - gpu load also mostly looked similarly higher, when I used MNOOOF of 8.4 vs baseline master (72k nps to 79k nps over a 5 million node search).
I have no clue whether that came at a cost of elo strength :P (It didn't change the selected move for this position :P)

Naphthalin:
This is likely outdated/redundant due to the changes with multivisit collision scaling.

Tilps:
I think this concept is likely independent of collision scaling.

Naphthalin:
Ok, then I apparently misunderstood it. Nonetheless, are you planning on updating this and make it ready for testing?

PR #1381, opened 2020-06-30
try to generate moves in a better order

borg323:
This is just a small reordering of the code, to generate in the following order: rook moves, bishop moves, knight moves, castling moves, pawn moves and king moves. Hopefully this will make it easier to sort the moves in policy order. We can probably do a lot better here, @ddobbelaere this seems like something you may find interesting.

### DIFF
--- a/src/chess/board.cc
+++ b/src/chess/board.cc
@@ -428,145 +428,135 @@ void InitializeMagicBitboards() {
 MoveList ChessBoard::GeneratePseudolegalMoves() const {
   MoveList result;
   result.reserve(60);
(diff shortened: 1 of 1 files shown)


Tilps:
std::sort often uses an insertion sort for short lists like our edge list - so move gen order that has better correlation with policy order could have a performance improvement I guess.

borg323:
If it turns out to have decent correlation (untested so far), we can use a custom sort implementation.

Tilps:
I've heard that for people truly obsessed about small array sorting - you can hard code a sorting network per size - and they run much faster than std::sort.

Issue #1382, opened 2020-06-30
Limit of 64 search threads on Windows


cn4750:
Due to [processor group](https://docs.microsoft.com/en-us/windows/win32/procthread/processor-groups) limitations in Windows, we can't use more than 64 search threads.
Can we possibly leverage [Stockfish's code](https://github.com/official-stockfish/Stockfish/blob/ab5cd8340f2f7f8730aa7c77476edf4a98a166e4/src/misc.cpp#L423-L527) for this?

Tilps:
The limit can in practice be between 32 and 64 since windows attempts to balance the processor groups to reduce splitting of a socket across multiple groups.

borg323:
I think #1383 is ready for review - seems to work OK.

borg323:
Yes, but there is an issue with Alder Lake (or other hybrid core) systems. However this is not related to this, so closing.

PR #1383, opened 2020-07-01
try to allocate threads across windows processor groups

borg323:
This is a first effort for #1382. Needs testing and cleanup.

### DIFF
--- a/src/mcts/search.h
+++ b/src/mcts/search.h
@@ -42,10 +42,7 @@
 #include "syzygy/syzygy.h"
 #include "utils/logging.h"
 #include "utils/mutex.h"
-
-#ifdef _WIN32
-#include <windows.h>
-#endif
+#include "utils/numa.h"
 
 namespace lczero {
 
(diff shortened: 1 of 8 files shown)


cn4750:
Since you're gathering lots of useful CPU information, can we print some of it to screen so that it is easier to debug when people post logs?

src/mcts/search.h
```diff
@@ -204,12 +208,56 @@ class Search {
...
+      if (id == 0) {
+        CERR << "Group " << group_id << " has " << group_cores
+             << " core(s) and " << group_threads << " thread(s).";
+      }
+      // Allocate cores of each group in order, and distribute remaing threads
+      // to all groups.
+      if ((id < core_count && core_id < group_cores) ||
+          (id >= core_count && (id - core_count) % group_count == group_id)) {
+        affinity.Group = group_id;
+        affinity.Mask = (1ULL << group_threads) - 1;
```
Tilps, line 250: This is a bit magical if group_threads == 64 ... but I guess it works?

src/mcts/search.h
```diff
@@ -204,12 +208,56 @@ class Search {
...
  public:
-  SearchWorker(Search* search, const SearchParams& params)
+  SearchWorker(Search* search, const SearchParams& params, int id)
       : search_(search),
         history_(search_->played_history_),
         params_(params),
         moves_left_support_(search_->network_->GetCapabilities().moves_left !=
-                            pblczero::NetworkFormat::MOVES_LEFT_NONE) {}
+                            pblczero::NetworkFormat::MOVES_LEFT_NONE) {
+#if defined(_WIN64) && _WIN32_WINNT >= 0x0601
```
Tilps, line 217: Probably should extract this code somewhere else, since it'll be useful for backend threads as well.

src/mcts/search.h
borg323: No, you are right, this doesn't work for 64. Will use something like `-1ULL >> (64 - group_threads)`

src/neural/network_demux.cc
```diff
@@ -162,7 +163,8 @@ class DemuxingNetwork : public Network {
     }
   }
 
-  void Worker() {
+  void Worker(int id) {
+    Numa::BindThread(id);
```
borg323, line 167: Maybe this would be better as `id+1` to leave space for an active search thread on the first group?

Tilps: Sounds logical to me.

src/utils/numa.cc
```diff
@@ -0,0 +1,94 @@
...
+  GROUP_AFFINITY affinity = {};
+  for (int group_id = 0; group_id < group_count; group_id++) {
+    int group_threads = GetActiveProcessorCount(group_id);
+    int group_cores = group_threads / threads_per_core_;
+    // Allocate cores of each group in order, and distribute remaining threads
+    // to all groups.
+    if ((id < core_count && core_id < group_cores) ||
+        (id >= core_count && (id - core_count) % group_count == group_id)) {
+      affinity.Group = group_id;
+      affinity.Mask = -1ULL >> (64 - group_threads);
```
Tilps, line 82: or ~0ULL ?

src/neural/network_mux.cc
```diff
@@ -153,7 +154,8 @@ class MuxingNetwork : public Network {
     }
   }
 
-  void Worker(Network* network, const int max_batch) {
+  void Worker(Network* network, const int max_batch, int id) {
+    Numa::BindThread(id);
```
Tilps, line 158: Logic for demux being + 1 probably kind of applies here as well?

borg323: I'm still thinking about it.

borg323: OK

PR #1392, opened 2020-07-11
Fix msvc runtime library linking so we can use latest meson

borg323:


borg323:
Yes, we enforce meson >= 0.52 see <https://github.com/LeelaChessZero/lc0/blob/master/meson.build#L19>.
However a similar issue resurfaced with the latest meson (0.55 - released one day after this pr) so I won't merge this yet.

borg323:
The latest iteration sets the nvcc flag to match the `b_msvcrt` meson option, ensuring correct linking, and `b_msvcrt` is set to `mt` by default to link the runtime library statically, but can be overridden in the meson command line. This is done for the dnnl appveyor builds, since the libraries were already built with `/MD`.

Tilps (approved):

PR #1393, opened 2020-07-12
handle compressed book (and file open errors)

borg323:


src/chess/pgn.h
```diff
@@ -43,14 +45,31 @@ struct Opening {
...
 };
 
+inline bool GzGetLine(gzFile file, std::string& line) {
+  bool flag = false;
+  char s[200];
+  line.clear();
+  while (gzgets(file, s, sizeof(s))) {
+    flag = true;
+    line += s;
+    auto r = line.find_last_of('\n');
```
Tilps, line 55: Any issue if someone tries to use a windows line ending pgn file on windows?

borg323: I hope opening the file in "r" mode will take care of it - but a test would be nice.

borg323: A quick test shows both line encodings seem to work fine under windows and linux.

Tilps:
Still approved :)

PR #1401, opened 2020-07-18
remove selfplay interactive mode

borg323:
I don't think anyone is using this any more.

### DIFF
--- a/src/selfplay/loop.cc
+++ b/src/selfplay/loop.cc
@@ -36,9 +36,6 @@
 namespace lczero {
 
 namespace {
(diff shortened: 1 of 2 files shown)


ValZapod:
I use it ;)

borg323:
@ValZapod can you share some details about how you use it, and if there is anything we can add to improve it?

Naphthalin:
There wasn't any follow-up, so the clean-up can likely be done.

borg323:
Since there is at least one user I don't see the reason to merge it.

PR #1402, opened 2020-07-19
remove trailing carriage return from book lines

borg323:


src/chess/pgn.h
```diff
@@ -74,6 +74,7 @@ class PgnReader {
     bool in_comment = false;
     bool started = false;
     while (GzGetLine(file, line)) {
+      if (line.back() == '\r') line.pop_back();
```
Tilps, line 77: Isn't back() undefined if line is empty?

PR #1411, opened 2020-08-24
alternative eigen download and wrap replacement

borg323:


Tilps (approved):

Tilps:
In case this is still needed - approved.

borg323:
Yes, it is still needed, together with fairly new meson.

PR #1412, opened 2020-08-24
Install Eigen library into the Docker image used by CircleCi

DanielUranga:
This fixes CircleCI broken builds (reported by @Naphthalin )


Naphthalin:
Thanks a lot :)

borg323 (approved):
Thanks, this is a big help, although not a full solution to the problem: at some point appveyor will invalidate the cached files and it will also start failing, I'm trying to fix it with #1411 (until the "official" `eigen.wrap` becomes available), but it seems there are problems accessing the eigen gitlab repo from the CI servers.

Naphthalin (approved):

PR #1422, opened 2020-09-17
New cuda backend without cudnn

borg323:


ankan-ban:
great job @borg323. Not only this avoid dependency on cudnn library dlls, it's actually a tiny bit faster than cudnn backend on Titan RTX with 384x30 networks.

ankan-ban:
Looks like there is some bug affecting bigger batch sizes, this crashes for me:
`lc0 benchmark --backend=cuda --minibatch-size=1024`

borg323:
I also noticed some accuracy loss with `cude-fp16` that needs investigating.

borg323:
The crashes should be fixed now, and the accuracy looks similar to cudnn-fp16 in my tests.

ankan-ban (approved):
Looks good.

PR #1425, opened 2020-09-19
replace cudnn softmax with cpu version

borg323:
This is a follow-up to #1422. Needs performance testing whether it brings a small nps improvement.
1. The softmax of WDL values is probably not very well suited to a gpu implementation. 
2. @ankan-ban suggested doing the softmax after freeing the lock.


borg323:
This is not for merging without very careful benchmarking showing gains (if any).

Tilps (approved):

Tilps:
On the grounds that cudnn-fp16 softmax layer produces values that don't sum to 1 with any reliable accuracy (at least on cuda 11/cudnn 8.0.2 with my 2080 TI) - and that this code is what is used in the cuda-fp16 backend basically - approval as a bugfix rather than as a nps improvement.

borg323:
First time I fixed an issue months before finding it...

PR #1428, opened 2020-09-23
Residual block fusion winograd optimization

ankan-ban:
A relatively small/easy optimization to speed the custom winograd convolution path further.

In the original implementation of custom winograd path, we had three passes for each convolution:
1. input transform.
2. GEMM
3. output transform (fused with RELU, bias addition, and optionally SE operation)

Because the residual tower is made up of many such convolutions one after other, we can fuse the output transform of the first convolution with the input transform of the next convolution - doing them in the same kernel. So, after this optimization we have each convolution do:
<input is already in winograd transformed space>
1. GEMM
2. Output Transform, followed by SE/relu/bias, followed by Input transform (for next convolution)
For the second convolution in each block, we also need to store untransformed output (to use it as skip connection)

This optimization results in 5-15% speedup with 384x30 networks:
```
lc0 benchmark (nps)

GPU           network         baseline  optimized   perf gain
-------------------------------------------------- ------------
Titan RTX   32390   (256x20)  57757     61721        6.9 %
Titan RTX   sv-3010 (384x30)  17443     20084       15.1 %
A100        32390   (256x20)  107536    120719      12.3 %
A100        sv-3010 (384x30)  41785     48815       16.8 %
````

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -447,7 +447,7 @@ if get_option('build_backends')
       )
 	 files += cuda_files
     files += cuda_gen.process(cuda_files_nvcc_common)
-    files += cuda_gen.process(cuda_files_nvcc_fp16, extra_args: ['-arch=compute_70', '-code=sm_70'])
+    files += cuda_gen.process(cuda_files_nvcc_fp16, extra_args: ['-arch=compute_75', '-code=sm_80', '-code=sm_75', '-code=sm_86'])
     has_backends = true
   endif
   
--- a/meson.build
+++ b/meson.build
@@ -447,7 +447,7 @@ if get_option('build_backends')
       )
 	 files += cuda_files
     files += cuda_gen.process(cuda_files_nvcc_common)
-    files += cuda_gen.process(cuda_files_nvcc_fp16, extra_args: ['-arch=compute_75', '-code=sm_80', '-code=sm_75', '-code=sm_86'])
+    files += cuda_gen.process(cuda_files_nvcc_fp16, extra_args: ['-arch=compute_53'])
     has_backends = true
   endif
   
--- a/src/neural/cuda/layers.cc
+++ b/src/neural/cuda/layers.cc
@@ -1089,7 +1089,7 @@ void ResidualBlock<DataType>::Eval(
       transformed_input + scratch_size / (2 * sizeof(DataType));
 
   if (first_block_) {
-    InputTransform<DataType>(N, C, transformed_input, input);
+    InputTransform<DataType>(N, c_input_, transformed_input, input);
 
     cublasRowMajorMatrixMul(transformed_input, transformed_weights0_,
                             transformed_output, N * 4, C, c_input_, 36, cublas);
--- a/src/neural/cuda/layers.cc
+++ b/src/neural/cuda/layers.cc
@@ -904,12 +904,14 @@ FusedWinogradConvSELayer<DataType>::~FusedWinogradConvSELayer() {
 
 template <typename DataType>
 ResidualBlock<DataType>::ResidualBlock(
-    BaseLayer<DataType>* ip, int C, bool se, int se_k, bool use_gemm_ex)
+    BaseLayer<DataType>* ip, int C, bool se, int se_k, bool use_gemm_ex, bool first, bool last)
     : BaseLayer<DataType>(C, 8, 8, ip),
       c_input_(C),
       has_se_(se),
       se_k_(se_k),
-      use_gemm_ex_(use_gemm_ex) {
+      use_gemm_ex_(use_gemm_ex),
+      first_block_(first),
+      last_block_(last) {
   // Allocate memory for weights (filter tensor) and biases.
   const size_t weight_size = sizeof(DataType) * C * C * 3 * 3;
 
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: meson.build [hunk 1]
```diff
@@ -447,7 +447,7 @@ if get_option('build_backends')
       )
 	 files += cuda_files
     files += cuda_gen.process(cuda_files_nvcc_common)
-    files += cuda_gen.process(cuda_files_nvcc_fp16, extra_args: ['-arch=compute_70', '-code=sm_70'])
+    files += cuda_gen.process(cuda_files_nvcc_fp16, extra_args: ['-arch=compute_75', '-code=sm_80', '-code=sm_75', '-code=sm_86'])
```
cn4750, line 450: I don't see `sm_86` in the nvcc [documentation](https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html).
It is causing me issues since nvcc doesn't recognize it.

borg323, line 450: This fails to build with older cuda versions. I have a hack for it in <https://github.com/borg323/lc0/commit/8b359dae61ed228ab580ee4f6f2751efb5dfea97>, but increases binary size a lot, need to check if the target list can be reduced. Alternatively, maybe a simple `extra_args: '-arch=compute_53'`(or whatever) is enough (but want to test if there are any performance implications).

ankan-ban: sorry about that (that's supposedly the SM version for GA10x/RTX30xx GPUs). I have removed it and replaced it simply with  '-arch=compute_53' as suggested by borg323 which seems to work (tested on Turing and Ampere) without any perf degradation.

ankan-ban: Thanks! extra_args: '-arch=compute_53' seems to work without perf degradation.

cn4750:
Testing on my RTX 2070 Super seems to show a very tiny regression when using this patch on J92-190 (384x30):
```
lc0 benchmark

GPU    network           baseline  optimized  perf gain
--------------------------------------------------------
2070S  J92-190 (384x30)  11845     11723      -1.0 %
```
Also a regression on backendbench:
```
lc0 backendbench (to 80 mbs)

GPU    network           baseline  optimized  perf gain
--------------------------------------------------------
2070S  J92-190 (384x30)  6176      5476       -11.3 %
```

cn4750:
Additional more thorough testing reveals no significant change.
```
lc0 backendbench (to 80 mbs)

GPU    network           baseline  optimized  perf gain
--------------------------------------------------------
2070S  J92-190 (384x30)  6118      6117       0.0 %
```
Normal benchmark seems too noisy to measure accurately and is about even under more tests.

ankan-ban:
@cn4750, which version of cuda are you using? The optimisation is applicable only for the custom winograd path which gets enabled by default only worry cuda11 or later.

cn4750:
> 
> 
> @cn4750, which version of cuda are you using? The optimisation is applicable only for the custom winograd path which gets enabled by default only worry cuda11 or later.

```
       _
|   _ | |
|_ |_ |_| v0.27.0-dev+git.dirty built Sep 23 2020
Loading weights file from: C:\nets\J92-190
Creating backend [cudnn-auto]...
Switching to [cudnn-fp16]...
CUDA Runtime version: 11.0.0
Cudnn version: 7.6.5
Latest version of CUDA supported by the driver: 11.0.0
GPU: GeForce RTX 2070 SUPER
GPU memory: 8 Gb
GPU clock frequency: 1815 MHz
GPU compute capability: 7.5
```

cn4750:
*Actually* using properly built executables this time and I see proper gains for large nets:
```
lc0 benchmark (80 mbs)

GPU    network           baseline  optimized  perf gain
--------------------------------------------------------
2070S  J92-190 (384x30)  10325     11089       7.4 %
```
```
lc0 backendbench (80 mbs)

GPU    network           baseline  optimized  perf gain
--------------------------------------------------------
2070S  703810 (128x10)   68790     68909       0.2 %
2070S  J92-190 (384x30)   6176      6855      11.0 %
```

src/neural/cuda/layers.cc
```diff
@@ -900,6 +901,251 @@ FusedWinogradConvSELayer<DataType>::~FusedWinogradConvSELayer() {
...
+  //   - for last_block_, output is the final output of this block (untransformed)
+
+  // Split the scratch space into two parts - use first part for holding
+  // transformed input and second part for transformed output.
+  DataType* transformed_input = (DataType*)scratch;
+  DataType* transformed_output =
+      transformed_input + scratch_size / (2 * sizeof(DataType));
+
+  if (first_block_) {
+    InputTransform<DataType>(N, C, transformed_input, input);
```
borg323, line 1092: This should be `InputTransform<DataType>(N, c_input_, transformed_input, input);`

ankan-ban: Thanks. Will fix it. It works because right now the network has c_input_ == C for all the convolutions in residual tower, but it's better to not rely on this assumption.

src/neural/cuda/layers.cc
```diff
@@ -900,6 +901,251 @@ FusedWinogradConvSELayer<DataType>::~FusedWinogradConvSELayer() {
...
+  }
+}
+
+template <typename DataType>
+void ResidualBlock<DataType>::LoadWeights0(float* pfilter,
+                                           float* pBias,
+                                           void* scratch) {
+
+  const size_t weight_size = sizeof(float) * c_input_ * C * 3 * 3;
+  const size_t blas_size = sizeof(float) * C;
```
borg323, line 950: This should be `bias_size` (I know it is like that in other kernels as well).

src/neural/cuda/layers.cc
```diff
@@ -900,6 +901,251 @@ FusedWinogradConvSELayer<DataType>::~FusedWinogradConvSELayer() {
...
+
+  // Store untransformed weights in scratch.
+  const DataType* weights = (DataType*)scratch + weight_size + blas_size;
```
borg323, line 953: Why is `blas_size` added here?

borg323 (commented):
I commented on some minor stuff that I have seen in other kernels as well.

src/neural/cuda/layers.cc
ankan-ban: thanks. It was a typo that got copied over to all places. Fixed.

src/neural/cuda/layers.cc
ankan-ban: you are right - it's not needed. fixed.

borg323 (approved):

Tilps:
Does anyone have any non-regression confirmation results for this? A few thousand games without crash and equal elo at fixed node count (for example)?

lp200:
\_\_launch_bounds\_\_ (384) seems to be quite slow in some GPU.

PR #1434, opened 2020-09-30
fix unitialized variable

borg323:
Some cudnn backend crashes were reported because `use_res_block_winograd_fuse_opt_` is used uninitialized when `custom_winograd` is false (the default for small nets and RTX20x0 cards).

### DIFF
--- a/src/neural/cuda/network_cudnn.cc
+++ b/src/neural/cuda/network_cudnn.cc
@@ -285,13 +285,12 @@ class CudnnNetwork : public Network {
     // Winograd needs nchw tensor layout.
     if (use_custom_winograd_) nhwc_ = false;
 
+    use_res_block_winograd_fuse_opt_ = false;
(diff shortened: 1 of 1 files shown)


src/neural/cuda/network_cudnn.cc
```diff
@@ -873,7 +872,8 @@ class CudnnNetwork : public Network {
   bool use_custom_winograd_;  // Custom winograd convolution implementation for
                               // convolutions of the residual tower.
 
-  bool use_res_block_winograd_fuse_opt_;    // fuse operations inside the residual tower
+  bool use_res_block_winograd_fuse_opt_; // Fuse operations inside the residual
```
Tilps, line 875: I thought style guide was 2 spaces between ; and // for trailing line comments.

Tilps (approved):

PR #1438, opened 2020-10-05
Update Mac install instructions

### LINKED ISSUE #1433: On Mac, the ./build.sh doesn't finish and ends with 1 error generated.
Could you please change the installation instructions for Mac because of this error?

```
#include <gtest/gtest.h>
         ^~~~~~~~~~~~~~~
1 error generated.
[106/134] Compiling C++ ob.../src_chess_board_test.cc.
FAILED: chessboard_test.p/src_chess_board_test.cc.o 
c++ -Ichessboard_test.p -I. -I../.. -I../../src -flto -Xclang -fcolor-diagnostics -DNDEBUG -pipe -Wall -Winvalid-pch -Wnon-virtual-dtor -Wextra -Wpedantic -std=c++17 -O3 -Wthread-safety -march=native '-DBUILD_IDENTIFIER="git.unknown"' -DUSE_BLAS -DNO_PEXT -isystem../../third_party -MD -MQ chessboard_test.p/src_chess_board_test.cc.o -MF chessboard_test.p/src_chess_board_test.cc.o.d -o chessboard_test.p/src_chess_board_test.cc.o -c ../../src/chess/board_test.cc
../../src/chess/board_test.cc:21:10: fatal error: 'gtest/gtest.h' file not found
#include <gtest/gtest.h>
         ^~~~~~~~~~~~~~~
1 error generated.
[121/134] Linking target lc0
```

Thanks in advance!

gsobala:
Sort out gtest for Mac

Fixes #1433

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,12 +166,26 @@ Or.
  * Install Xcode command-line tools: ``xcode-select --install``
  * Install "XCode Developer Tools" through the app store. (First one on the list of Apps if searched.)
  * Associate the SDK headers in XCode with a command: export CPATH=\`xcrun --show-sdk-path\`/usr/include
+ 
+6. Optionally install googletest (aka gtest)
+
+```
+cd ~
+git clone https://github.com/google/googletest.git
+cd googletest
+mkdir install
+cd install
+cmake ..
+make
+make install
+```
 
 Now download the lc0 source, if you haven't already done so, following the instructions earlier in the page.
 
-6. Go to the lc0 directory.
-7. Run `./build.sh` (needs step 5)
-8. The resulting binary will be in build/release
+7. Go to the lc0 directory.
+8. Run `./build.sh` (needs step 5)
+ * If you did not install gtest in step 6 then run `./build.sh -Dgtest=false` instead.
+9. The resulting binary will be in build/release
 
 ### Raspberry Pi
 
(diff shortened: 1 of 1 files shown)


borg323 (commented):
Can you change the gtest dependency line in meson..build to check for the header as well?
Something like the following (untested):
```
gtest = dependency('gtest', required: false)
if not gtest.found() or not cc.has_header('gtest.h')
    gtest= subproject('gtest').get_variable('gtest_dep')
endif
```

gsobala:
That code snippet is not good. If gtest **is** preinstalled, the header and library paths successfully returned by pkg-config are not passed to the cc.has.header compile test (`/usr/local/..` by default), and so gtest.h is not found and the subproject is created unnecessarily. So it works, but fails to respect an existing googletest installation and builds a subproject anyway.

borg323:
Probably should be something like `cc.has_header('gtest.h', dependencies: gtest)`

gsobala:
Ok, that code works but on further testing this PR is probably not necessary. 

Having previously compiled and installed googletest from scratch on the system I then deleted it by clearing out headers, libs and .pc files in /usr/local. I found that the meson.build from HEAD of master no longer found a "prebuilt" gtest and lc0 compiled perfectly happily. I then checked using a vanilla install of Catalina in a VM and the same is true: lc0 compiles satisfactorily by default on vanilla OSX 15 without errors.

So I have no idea where my rogue partial gtest installation came from but we do not need to bother about this further. I would suggest however that we set gtest to default to false in meson_options.txt as the average user has no use for it, and this would cut down on support requests when it fails for some reason.

gsobala:
Alternatively, as in the modified PR, simply advise Mac users in the README to compile without gtest.

borg323 (approved):

PR #1441, opened 2020-10-07
check mkl_include dir exists

borg323:


GBeauregard:
Like the other times this check is used, this allows lc0 to be built on systems that have mkl installed from their package manager.

Tilps (approved):

PR #1447, opened 2020-10-11
modified caching

borg323:
This was originally titled "do position encoding only if not in cache" and the first comment below refer to this early stage. The latest version also modified the cache to only store P for legal moves in the order generated, allowing further simplifications.

**What this PR does in more detail** 
- `EncodePositionForNN()` is done when preparing the batch for the backend, so not if the position is already in the cache.
- The cache is storing the policy values compressed after softmax.
-  Only legal moves are cached and in move generator order, so there is no need to also cache the move index.
- ~~`GetV6TrainingData()` and related functions are moved to `selfplay/game.cc` with small modifications to handle the cached policy changes.~~ *(This is now in a new directory in master)*
- The search no longer knows about input encoding.


Tilps:
I have a different take on this.  The probability of the cache hit arriving between the existing cache check at start of method, and further down is quite low.  So instead I plan on removing the second check of the cache after encoding - to save the cache lookup cost. (I did this in 1483, but didn't do it in 1490 in order to minimize impact on existing flows for the initial merge.)

borg323:
This is just the first step in an experiment to eventually pass moves to the backend and do the encoding there.

borg323:
This is a larger speed gain than I expected and the next steps I'm considering are rather more invasive.

Naphthalin:
For me, this PR is a minor slowdown for 10b nets:

Benchmark on i7-4770 + RX Vega 56, dx12 backend

```
lc0_1447.exe benchmark -w weights\744181
===========================
Total time (ms) : 340466
Nodes searched  : 20977277
Nodes/second    : 61613
```

Comparison: 2nd benchmark from https://github.com/LeelaChessZero/lc0/pull/1488#issuecomment-867549057 because I simulated heating up the GPU
```
lc0_master benchmark -w 744181

===========================
Total time (ms) : 340491
Nodes searched  : 21740416
Nodes/second    : 63850
```

borg323:

PR #1448, opened 2020-10-13
Rescore piece count

jhorthos:
adding piece count filter to table base rescorer


borg323:
Anyone looking for the rescorer: <https://github.com/Tilps/lc0/tree/rescore_tb>

PR #1449, opened 2020-10-14
Reduce computation cost of first collision after a visit.

Tilps:
Will clean it up if it works.

Note that this transform is technically not safe if concurrent searchers is not the default value of 1. But otherwise, it should 'theoretically' be identical to previous behavior, just potentially less computation. (Assuming I didn't mess anything up :P)


Tilps:
Redundant after 1490.

PR #1451, opened 2020-10-15
use neon instructions on arm

borg323:


gsobala:
Succeeds on Armv8 (by not including the `-mpfu=neon` flag) but fails on armv7: `-mfpu=neon` included successfully for the compile stages but is also needed by the linker.

Tilps (approved):

PR #573, opened 2018-12-09
Let lc0 switch to another NN during the game. 

hans-ekbrand:
Let `lc0` switch to another NN during the game. The intended usecase is to combine a stronger general NN with and an NN which is better at endgames.

This patch exports five new UCI-options `SecondWeightsFile`, `SecondWeightsSwitchAt`, `SecondWeightsCpuct`, `SecondWeightsFpuReduction`, `SecondWeightsPolicySoftmaxTemp`

`SecondWeightsFile` defines a path to secondary weights file which substitute the primary weights file when `SecondWeightsSwitchAt`pieces remain on the board. `SecondWeightsCpuct`, `SecondWeightsFpuReduction`, `SecondWeightsPolicySoftmaxTemp` defines Cpuct, FpuReduction and PolicySoftmaxTemp to be used with the secondary NN.

As of now, this PR has been tested with lc0 T11258 vs lc0 T11258 + Ender, but the PR itself is NN neutral.

PR #1452, opened 2020-10-18
Let lc0 switch to another NN during the game (PR 573 rebased to current master)

hans-ekbrand:
This a a rewrite of [PR 573](https://github.com/LeelaChessZero/lc0/pull/573) against the current master.

Let `lc0` switch to another NN during the game. The intended use case is to combine a stronger general NN with and an NN which is better at endgames.

This patch exports five new UCI-options `SecondWeightsFile`, `SwitchAt`, `SWCpuct`, `SWFpuValue`, `SWPolicyTemperature`

`SecondWeightsFile` defines a path to secondary weights file which substitute the primary weights file when `SwitchAt`pieces remain on the board. `SWCpuct`, `SWFpuValue`, `SWPolicyTemperature` defines Cpuct, Fpu and PolicySoftmaxTemp to be used with the secondary NN.

This PR has been tested with Sergio Vieri's [128x10 net](https://www.comp.nus.edu.sg/~sergio-v/new/128x10-t60-2/128x10-t60-2-5300.pb.gz) against itself + [Medium Ender](https://github.com/dkappe/leela-chess-weights/releases/download/MediumEnder/me.pb.gz), but the PR itself is NN neutral.

To verify that the NN is switched, use the `LogFile` UCI directive and look for the string `Will switch to second NN`. For transparency, CPuct, FPU and PolicyTemperature is logged at each move, to clarify when the settings for the second NN is in effect.

### DIFF
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -81,6 +81,18 @@ MoveList StringsToMovelist(const std::vector<std::string>& moves,
   return result;
 }
 
+  // Stuff required for switching NN during play START
+  bool second_nn_already_loaded_ = false;
+  int k_pieces_left = 32;
+  std::string CPuct_for_primary_NN = "";
+  std::string FPU_for_primary_NN = "";
+  std::string PolicyTemperature_for_primary_NN = "";
+
+  // Purely informational vars, writing to them do not affect behaviour, they
+  // are just convenient to have when we want to tell about the current state.
+  std::string weights_currently_in_use;
+  // Stuff required for switching NN during play STOP
+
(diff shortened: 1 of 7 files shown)


jhorthos:
Excellent Hanse - thanks.

Naphthalin:
@borg323 what is the current status of your hybrid backend, and does it allow switching NNs based on specifyable conditions?

borg323:
No, but it isn't very hard to do - but it will only work with conditions that can be inferred from the NN input.

PR #1455, opened 2020-10-29
Implementing an analysis mode for Lc0 to allow forward/backward analysis without losing the tree.

Naphthalin:
This PR attempts to implement an analysis mode for Lc0 to allow forward/backward analysis without losing the tree. Note that this creates inconsistencies in the tree due to visits/results only being propagated until current root, and it will lead to wrong/distorted Q values when being used with plain PUCT, so use at your own risk or together with #963. 

To activate, use `--analysis-mode` or `"AnalysisMode": true`.


Naphthalin:
A fully working implementation can be found in my `betamcts-rents` branch at https://github.com/Naphthalin/lc0/tree/betamcts-rents, which fixes the tree inconsistencies (N, visited policy) and calculates correct evals. It also contains an experimental variant of the stochastic maximum entropy exploration strategy from #1418 with `--use-rents` for broader exploration.

Naphthalin:
Obsolete through #1543.

PR #1458, opened 2020-10-31
Improvement of backendbench assistant.

zz4032:
Current logic is to replace batchsize with highest nps, if new nps is 1+threshold times higher than previous best.
However, it is known that generally higher batchsize numbers perform worse in gameplay, e.g. a 4x higher batchsize with 1.05x higher nps shouldn't be accepted while a 1.10x higher batchsize with 1.04x higher nps is most probably better.
Now the acceptance rule is based on a linear function requiring +10% nps (compared to previous best) for each +100% of batchsize (to previous best).
Example:
```
batchsize ratio    nps ratio 
1.5                  1.05
2.0                  1.10
3.0                  1.20
```
Edit: Based on some short test matches I suspect that the default threshold can be increased from 0.1 to 0.2 (+15% nps with 2x batchsize was weaker in a test):
```
batchsize ratio    nps ratio 
1.5                  1.10
2.0                  1.20
3.0                  1.40
```

### DIFF
--- a/src/benchmark/backendbench.cc
+++ b/src/benchmark/backendbench.cc
@@ -75,7 +75,7 @@ void BackendBenchmark::Run() {
   options.Add<IntOption>(kMaxBatchSizeId, 1, 1024) = 256;
   options.Add<StringOption>(kFenId) = ChessBoard::kStartposFen;
   options.Add<BoolOption>(kClippyId) = false;
-  options.Add<FloatOption>(kClippyThresholdId, 0.0f, 1.0f) = 0.1f;
+  options.Add<FloatOption>(kClippyThresholdId, 0.0f, 1.0f) = 0.2f;
 
   if (!options.ProcessAllFlags()) return;
 
@@ -88,7 +88,7 @@ void BackendBenchmark::Run() {
     tree.ResetToPosition(option_dict.Get<std::string>(kFenId), {});
     const int batches = option_dict.Get<int>(kBatchesId);
 
-    int best = 0;
+    int best = 1;
     float best_nps = 0.0f;
     std::optional<std::chrono::time_point<std::chrono::steady_clock>> pending;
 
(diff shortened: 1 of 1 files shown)


src/benchmark/backendbench.cc
```diff
@@ -123,15 +117,15 @@ void BackendBenchmark::Run() {
...
-            (run == true || nps > best_nps * (1.0f + threshold))) {
-          best_nps = nps;
-          best = i;
-          run = true;
-          if (!pending) {
-            pending = std::chrono::steady_clock::now();
+
+        if (nps > best_nps) {
+          if (best==0 ||
+             ((i*1.0f/best*1.0f-1.0f)*threshold < nps/best_nps-1.0f)) {
```
borg323, line 123: You can rewrite this as `(i-best)*threshold*best_nps < (nps-best_nps)*best`.
Then you can initialize best to 1 to avoid the best==0 check and the nested `if`.

src/benchmark/backendbench.cc
```diff
@@ -79,8 +75,7 @@ void BackendBenchmark::Run() {
   options.Add<IntOption>(kMaxBatchSizeId, 1, 1024) = 256;
   options.Add<StringOption>(kFenId) = ChessBoard::kStartposFen;
   options.Add<BoolOption>(kClippyId) = false;
-  options.Add<FloatOption>(kClippyThresholdId, 0.0f, 1.0f) = 0.05f;
-  options.Add<FloatOption>(kClippyToleranceId, 0.0f, 1.0f) = 0.03f;
+  options.Add<FloatOption>(kClippyThresholdId, 0.0f, 1.0f) = 0.1f;
```
borg323, line 78: You suggestion to change the default to 0.2 seems reasonable,

zz4032: 0.2 would be my best bet for max. playing strength, favoring lower batch sizes. Can be changed later, if someone conducts more tests.

borg323 (approved):

PR #1462, opened 2020-11-25
Updated description of score-type Q

hans-ekbrand:
The manual says that if you request score-type Q, then you will get the internal value of Q multiplied by 100. 

```
--score-type=CHOICE
               What to display as score. Either centipawns (the UCI default), win percentage or
               Q (the actual internal score) multiplied by 100.
               [UCI: ScoreType  DEFAULT: centipawn  VALUES: centipawn,centipawn_with_drawscore,centipawn_2019,centipawn_2018,win_percentage,Q,W-L]
```

This is not correct, what you will get is the internal value of Q multiplied by 10000.

```
$ ./lc0 --backend=random --score-type=Q 
       _
|   _ | |
|_ |_ |_| v0.27.0-dev+git.34f4107 built Nov 25 2020
go nodes 1
Creating backend [random]...
info depth 1 seldepth 1 time 4 nodes 1 score cp -5752 tbhits 0 pv h2h4
bestmove h2h4
```

This patch updates the documentation to reflect the actual behaviour.

### DIFF
--- a/src/mcts/params.cc
+++ b/src/mcts/params.cc
@@ -214,7 +214,7 @@ const OptionId SearchParams::kPerPvCountersId{
 const OptionId SearchParams::kScoreTypeId{
     "score-type", "ScoreType",
     "What to display as score. Either centipawns (the UCI default), win "
-    "percentage or Q (the actual internal score) multiplied by 100."};
+    "percentage or Q (the actual internal score) multiplied by 10000."};
 const OptionId SearchParams::kHistoryFillId{
     "history-fill", "HistoryFill",
     "Neural network uses 7 previous board positions in addition to the current "


Naphthalin:
The description is actually correct: A GUI will indeed display "winrate / Q times 100", as the reported "cp" value is assumed to represent a decimal fraction `xx.xx` and will be displayed as such. If anything, what's confusing is that the usual eval `+1.25` is called a cp eval despite being a pawn score eval with cp accuracy.

borg323:
The description is certainly not clear so suggestions for rewording are welcome. Indeed the Q is displayed multiplied by 10000 (so that a GUI will display if with the correct scale) but also the win percentage is multiplied by 100 (for the same reason), so changing only the Q description would be then inconsistent. The description is for the command line help, so I think should refer to the raw output, so when changing this we can add a few words like `(typically GUIs display this divided by 100)`.

Naphthalin:
@borg323 what is "centipawn eval" referring to in your book? is `+1.26` or `126` the cp eval for you?
The UCI protocol (afaik) defines the cp field to display the cp value represented as an int, but cp 126 is supposed to be read as 1.26 per definition.

borg323:
The UCI "protocol" is unsurprisingly not very helpful, the only mention is:
>	* cp \<x\>
>			the score from the engine's point of view in centipawns.

My understanding is that GUIs display the score in *pawns*, so 126 is the centipawn value in your example

hans-ekbrand:
Good idea @borg323, I'll add "(typically GUIs display this divided by 100)".

PR #1465, opened 2020-11-28
simple backend

borg323:
This is a simple backend, based on the structure of the cudnn backend but using BLAS functions for the internals. The purpose is to make building backends for other frameworks easy in a step by step process, replacing layers one at a time. The main structure is in `network_simple.cc`, `layers.cc` and `layers.h`, the remaining files are only there to support the BLAS functions.
Starting with an earlier, more complicated, version of this, I managed to bring up a full backend in a few days of part time coding (together with some of the further simplifications now included).

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -487,7 +487,40 @@ if get_option('build_backends')
     subdir('src/neural/dx/shaders')
 
     has_backends = true
-  endif  
+  endif
+
+  build_simple = false
+  if get_option('simple')
+    if get_option('dnnl') and dnnl_lib.found()
+      add_project_arguments(['-DUSE_DNNL'], language : 'cpp')
+      includes += include_directories(get_option('dnnl_dir') + '/include')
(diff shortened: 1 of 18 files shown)


dje-dev:
Thanks again for this work. Two small issues I noticed, which possibly do not affect correctness.

1. Inconsistent code/comment
```
line 158 of winograd_convolution3.cc:
    cblas_sgemm(CblasColMajor,               // Row major format
```

2. Inconsistent size of buffer
```
line 196 of layers.cc
  size_t weight_size = sizeof(short) * used_size_ * 64;
  weights_ = (short*)malloc(weight_size);

versus:

line 203 of layers.cc:
  size_t weight_size = sizeof(short) * used_size_;
  memcpy(weights_, cpuWeight, weight_size);  
```

samoeltaodros:
لا اعرف كيفية تشغيل الخلفيه

OfekShochat:
Sorry for my broken Arabic, 
لصورة لا تعمل

PR #1468, opened 2020-12-05
v6 training format for issue 1308 and value repair

borg323:


src/selfplay/game.cc
```diff
@@ -217,6 +205,7 @@ void SelfPlayGame::Play(int white_threads, int black_threads, bool training,
         }
         if (edge.GetMove(tree_[idx]->IsBlackToMove()) == move) {
           cur_n = edge.GetN();
+          played_wl = edge.GetWL(best_eval.wl);
```
Tilps, line 208: The fall through value here should be fallback to parent wl rather than best?

Naphthalin: I think fallback to `best_wl` is fine in the application we are using it for, as it would assume that we picked the no temp move, and therefore never trigger a blunder detection.

Tilps: if the fall back is triggered, that means we've picked a move with 0 visits... 
Which should never happen (unless using a *positive* temp offset), so I'd rather go with being more consistent.

Naphthalin: Ok, I see & agree with that reasoning.

src/mcts/node.h
```diff
@@ -223,11 +223,17 @@ class Node {
   // in depth parameter, and returns true if it was indeed updated.
   bool UpdateFullDepth(uint16_t* depth);
 
-  V5TrainingData GetV5TrainingData(
+  struct Eval {
```
Tilps, line 226: Maybe elevate Eval to top level rather than being a inner class of Node.

src/selfplay/game.cc
```diff
@@ -206,17 +193,22 @@ void SelfPlayGame::Play(int white_threads, int black_threads, bool training,
...
+      auto node = tree_[idx]->GetCurrentHead();
+      for (auto edge : node->Edges()) {
         if (edge.GetN() > max_n) {
           max_n = edge.GetN();
         }
         if (edge.GetMove(tree_[idx]->IsBlackToMove()) == move) {
           cur_n = edge.GetN();
+          played_eval.wl = edge.GetWL(-node->GetWL());
+          played_eval.d = edge.GetD(node->GetD());
+          played_eval.ml = edge.GetM(node->GetM());
```
Tilps, line 211: Should be a +1 on the default here I think?

Tilps (commented):
I thought we were going to add some more flags?

borg323:
> I thought we were going to add some more flags?

Yes, this is why it is again a draft PR .

Naphthalin:
We will also need the corresponding update of `https://github.com/LeelaChessZero/lczero-training/blob/master/tf/chunkparser.py` which then reads v6 data and populates the `winner` not from `result` anymore but from `result_q` and `result_d`.

borg323:
I think it finally includes all the info suggested and then some.

src/chess/position.cc
```diff
@@ -74,7 +74,6 @@ GameResult PositionHistory::ComputeGameResult() const {
 
   if (!board.HasMatingMaterial()) return GameResult::DRAW;
   if (Last().GetRule50Ply() >= 100) return GameResult::DRAW;
-  if (Last().GetGamePly() >= 450) return GameResult::DRAW;
```
Tilps, line 77: I'm a huge fan of this change on its own ;)

src/neural/writer.h
```diff
@@ -49,21 +49,41 @@ struct V5TrainingData {
...
   uint8_t rule50_count;
-  // For input type 3 contains a bit field indicating the transform that was
-  // used and the original side to move info.
-  // Side to move is in the top bit, transform in the lower bits.
+  // Bitfield with the following allocation:
+  //  bit 7: side to move (input type 3)
+  //  bit 6: position marked for deletion by the rescorer (never set by lc0)
+  //  bit 5: game adjudicated (v6)
+  //  bit 4: max game length exceeded (v6)
+  //  bit 3: best_q is a proof (v6)
```
Tilps, line 57: I don't think bit 3 is implemented as I intended.  If best_q is 1.0 and this bit is set, then all good.
If best_q is 0 and this bit is set, we haven't ruled out that moves other than best_q are actually wins.
Similary if its -1 and this bit is set we haven't ruled out draws or wins for the other moves.

borg323: Should work as intended after the last commits.

src/selfplay/game.cc
```diff
@@ -243,6 +245,18 @@ void SelfPlayGame::Play(int white_threads, int black_threads, bool training,
     }
 
     if (training) {
+      bool best_is_proof = best_is_terminal;  // But check for better moves.
+      if (best_is_proof && best_eval.wl < 1) {
+        auto best =
+            (best_eval.wl == 0) ? GameResult::DRAW : GameResult::BLACK_WON;
+        auto upper = best;
+        for (const auto& edge : node->Edges()) {
+          upper = std::max(-edge.GetBounds().first, upper);
```
Tilps, line 254: I'm a bit confused about this negation.  bounds is an enum with range 1->3 for valid values.  Negating here means you are taking max between a valid gameresult and an invalid gameresult?

Tilps (approved):

PR #1470, opened 2020-12-15
Disable WDL lookups when playing from DTZ restricted search position.

Tilps:
I could make this an option, but I don't think the added complexity is worth it.  Yes we might not report mate score quite as frequently, but given our mate scores involving TB hits generally cause more confusion than not, and more realistic play is highly likely to occur with WDL disabled in search - the good should outweigh that tiny little detail.

Stockfish does this unconditionally as well - (further more disables WDL lookups if root position is a WDL loss or draw (even if DTZ is missing)- I guess to try and get some swindling in? - I don't think we can be bothered with that subtle detail).

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -52,7 +52,7 @@ const int kUciInfoMinimumFrequencyMs = 5000;
 MoveList MakeRootMoveFilter(const MoveList& searchmoves,
                             SyzygyTablebase* syzygy_tb,
                             const PositionHistory& history, bool fast_play,
-                            std::atomic<int>* tb_hits) {
+                            std::atomic<int>* tb_hits, bool* dtz_success) {
   // Search moves overrides tablebase.
   if (!searchmoves.empty()) return searchmoves;
   const auto& board = history.Last().GetBoard();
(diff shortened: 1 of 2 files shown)


src/mcts/search.cc
```diff
@@ -52,7 +52,7 @@ const int kUciInfoMinimumFrequencyMs = 5000;
 MoveList MakeRootMoveFilter(const MoveList& searchmoves,
                             SyzygyTablebase* syzygy_tb,
                             const PositionHistory& history, bool fast_play,
-                            std::atomic<int>* tb_hits) {
+                            std::atomic<int>* tb_hits, bool* dtz_success) {
   // Search moves overrides tablebase.
```
mooskagh, line 56: `assert (dtz_success);`?

src/mcts/search.h
```diff
@@ -174,6 +174,8 @@ class Search {
   const MoveList searchmoves_;
   const std::chrono::steady_clock::time_point start_time_;
   int64_t initial_visits_;
+  // dtz_position_ must be initialized before root_move_filter_.
+  bool dtz_position_ = false;
```
mooskagh, line 178: Does it mean that the root is dtz? Then it may make sense to reflect that in the variable name or at least a comment.

mooskagh (approved):

src/mcts/search.cc
Tilps: Done.  Also tb_hits since I'm here.

src/mcts/search.h
Tilps: Renamed to be clearer.

PR #1471, opened 2020-12-15
Make MLH default on.

Tilps:
Surely its time?

People can argue about exact thresholds and stuff, but I don't think any serious net wants it off anymore.


borg323 (approved):

kiudee:
The settings we currently use for TCEC (and other tournaments) could be a good starting point:
https://github.com/kiudee/lc0-tcec-submission/blob/master/gpu/LCZero_GPU.json

zz4032:
Btw, what are the guidelines for having Leela parameters hidden?
Because I think 99% of users will never touch MLH parameters except maybe MovesLeftThreshold to turn MLH on/off and 5 of the others can be hidden.
Same for all parameters ending with "AtRoot" but that's not related to this PR...

PR #1477, opened 2021-01-03
SF-like centipawn formula as distinct option.

zz4032:
Centipawn formula has been changed to a simpler one in April 2020 and doesn't correlate well to SF's CP evals especially in the lower evals. Discussion with correlation plots from TransInfinitum and eval data from Chad and me here:
https://github.com/LeelaChessZero/lc0/pull/1193#issuecomment-612854515
https://discord.com/channels/425419482568196106/539960268982059008/733434482128060516
Added formula in this PR is an updated version of `centipawn_2019` and has an even better fit in higher evals for T40 and later networks (helpful for correct TCEC adjudication), which is the reason why it was replaced (besides a more complicated inversion).
I don't understand the motivation behind the change to the currently used formula, for me correlation to most commonly used CPU-engine SF would be 1st priority and current formula is much worse here, but I'd like users to have the option of using one that correlates best. It would retain it's name `centipawn_SF` in the future, in case a better fit was found.

### DIFF
--- a/src/mcts/params.cc
+++ b/src/mcts/params.cc
@@ -321,6 +321,7 @@ void SearchParams::Populate(OptionsParser* options) {
                                          "centipawn_with_drawscore",
                                          "centipawn_2019",
                                          "centipawn_2018",
+                                         "centipawn_SF",
                                          "win_percentage",
                                          "Q",
                                          "W-L"};
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -243,6 +243,8 @@ void Search::SendUciInfo() REQUIRES(nodes_mutex_) {
       uci_info.score = 295 * wl / (1 - 0.976953126 * std::pow(wl, 14));
     } else if (score_type == "centipawn_2018") {
       uci_info.score = 290.680623072 * tan(1.548090806 * wl);
+    } else if (score_type == "centipawn_SF") {
+      uci_info.score = 295 * wl / (1 - 0.976953126 * std::pow(wl, 8));
     } else if (score_type == "win_percentage") {
       uci_info.score = wl * 5000 + 5000;
     } else if (score_type == "Q") {
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -244,7 +244,7 @@ void Search::SendUciInfo() REQUIRES(nodes_mutex_) {
     } else if (score_type == "centipawn_2018") {
       uci_info.score = 290.680623072 * tan(1.548090806 * wl);
     } else if (score_type == "centipawn_SF") {
-      uci_info.score = 295 * wl / (1 - 0.976953126 * std::pow(wl, 8));
+      uci_info.score = 317.6 * wl / (1 - 0.9751875 * std::pow(wl, 10));
     } else if (score_type == "win_percentage") {
       uci_info.score = wl * 5000 + 5000;
     } else if (score_type == "Q") {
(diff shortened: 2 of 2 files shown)


zz4032:
Correlation of LC0 v0.27.0 nJ94-100 to most recent SF-NNUE (games from lichess, Elo>2500, time control > 3min, random ply in each game).

Curve fitting in log space for cp values seems to work well.
Updated centipawn_SF formula:
`cp(q) = 317.6*q/(1-0.9751875*(q)^10)`
RMS: 0.621419
Current tan formula fit curve would be:
`cp(q) = 142.13*tan(1.5596936*q)`
RMS: 0.789902
cp(1) = 12800 as constraint for both formulas above.

Log cp scale:
plot_J94-100
Linear cp scale up to cp=500:
plot_J94-100_200

borg323:
Can you check with a recent net (after deblundering was turned on) that the approximation is still accurate?

zz4032:
Updated centipawn_SF formula:
`cp(q) = 357.2*q/(1-0.9720941*(q)^22)`
RMS: 0.590765 (lower is better)
Current tan formula fit curve would be:
`cp(q) = 142.15*tan(1.5596991*q)`
RMS: 0.880274

Visually the updated curve is very similar in the log cp space, especially at lower values. Not sure why there is a noticeable difference at extreme q values and therefore a much higher exponent in the formula. Maybe there is too much variation in data and more data points are required to reduce error.
plot_68724

Linear cp scale up to cp=500:
plot_68724_200

borg323:
It looks like the optimum conversion function would require an inflection point around q=0.5. Eyeballing it, seems like `357.2 * q / (1 + 1.4 * q^2 - 2.1 * q^4)` is a closer fit to the data points.

zz4032:
The constraint `cp(q=1) = 12800` that I used for curve fitting is not fulfilled by that function. cp(1) would only be 1191.
Fitted curve with constraint above for your proposal would be: `cp(q) = 341.8*q/(1-0.6988*q^6.667-0.2745*q^135.6)`, RMS: 0.574908.
plot_68724_new

yuzisee:
SF has moved to rescale so that `+100` centipaws intersects with `q=+0.5`
https://github.com/official-stockfish/Stockfish/pull/4216

Naphthalin:
@yuzisee 
> SF has moved to rescale so that `+100` centipaws intersects with `q=+0.5` [official-stockfish/Stockfish#4216](https://github.com/official-stockfish/Stockfish/pull/4216)

while that is not far from the truth, `q=+0.5` is not what SF `+100` centipawns means; it is `W=50%`. The difference is quite significant, as Leela will still give a decent chance for a black win even when white has a 50% winrate. Depending on the position, when W is at 50%, `q` is between `+0.35` and `+0.45`, which (with the centipawn formula) is about `60` to `80` centipawns.

> If the goal is correlating well to SF's CP evals, does that simplify the constraints in any way?

Short answer: this doesn't simplify anything, it just shows that it is impossible to reproduce the SF evals by converting `q` into a centipawn score, due to the broad WDL. I recommend to you to take a look at #1791, for two reasons: first, it allows sharpening the WDL to the point where W=50% is in fact nearly identical to `q=0.5`. More importantly (for this discussion at least) however is that it since a few days contains a new score type `WDL_mu` which adopts the same convention SF is using, and that alone is enough to reach the desired correspondence of evals without fitting anything at all.

Naphthalin:
@zz4032 Since the new `WDL_mu` eval from #1791 gives much better agreement to SF evals in a wide range of positions than all Q based cp functions we tested so far, do you still want to update this to find a Q based cp function which fits the "new" SF eval definition better?

zz4032:
This PR was not merged back when the correlation of the default cp curve to Stockfish's values was much worse. I don't think there is much sense to keep it open or updated today.

Naphthalin:
Right, makes sense. Closing this PR then.

yuzisee:
Does anyone know whether the next TCEC season plans to start displaying SF-style evals for Lc0 using the new WDL_mu setting?

Naphthalin:
@yuzisee already happening since (I think) DivP, same with CCC.

PR #1478, opened 2021-01-04
Multigather parallel

Tilps:
So I implemented an 'lc0' version of Ceres' parallel multiple gathering.
During 'benchmark' of a T70 net on my 2080TI I saw a position get over 370k NPS.  The final benchmark result went from 140k to 200k nps - most positions are completely gpu limited even with a T70 net on my machine before the changes, so that ~50% increase in nps is much larger in some positions than others.

There are a number of magic constants to do with how much work gets given to which threads, not to mention how many task threads to spin up is hard coded at '4' which is a number that pretty obviously doesn't make any sense...
The Numa assignment logic for these task threads is not really thought through.

My 1k nodes fixed elo test finally has come in as 'inconclusive' after fixing many bugs.  But I suspect there may still be some more...
```
Score of lc0testdev vs lc0testdev-Basline: 213 - 234 - 614 [0.490]
Elo difference: -6.88 +/- 13.56

1061 of 2000 games finished.
```
(Will update when I get to 2k)

I'll note that I have no idea whether draw score works correctly anymore - all depths could be 'off by one'.  This definitely needs another pass of careful checking.


Tilps:
32 bit build being broken is expected...

Results drifted outside of error bars over the final 1k games. Will continue to try and find bugs :)
```
Score of lc0testdev vs lc0testdev-Basline: 370 - 441 - 1189 [0.482]
Elo difference: -12.34 +/- 9.68

2000 of 2000 games finished.
```

Also fixed node benchmark testing suggests there is a crossover at about 100k nodes in terms of nps.  At 1k nodes the new code is only 70% of the speed of the old.  But it seemingly converges again back around 100 nodes.  I think this is potentially because I'm switching to using multiple threads too soon - further profiling that isn't just single position run to 10's of millions of nodes will probably be useful.

Tilps:
Closing as obsolete.

PR #1479, opened 2021-01-05
Multigather parallel spinlocks

Tilps:
This is a fork of #1478 which uses spin locks to significantly reduce latency of executing tasks.
At the cost of burning cpu a bit.


Tilps:
On both of the PRs, but only tested here - i've changed the default value of Max Collision Visits to 1000.  When testing with high max collision events to make things more like what we use in tournament conditions (and reduce quite different side effects of low max collision events in my branch) I found that the new branch was exiting based on max collision visits vastly more often than master does.  Getting more nodes at worse collision totals.  Using the 1k limit brings things more closely in to line with how the old code would gather trees.  As a slightly unexpected side effect changing to 1k also gave an NPS boost - so this branch is now faster than master in all node counts I've tested.  Only making significant breaks away at very large node counts.

1k fixed node tests under the conditions above makes it seem pretty even now.  I would suggest that max collision visits should be tuned at TC some more, maybe a bit lower is better for speed and elo.
Score of lc0testdev vs lc0testdev-Basline: 399 - 406 - 1195 [0.498]
Elo difference: -1.22 +/- 9.65

2000 of 2000 games finished.

Tilps:
Closing as obsolete.

PR #1483, opened 2021-01-08
multigather parallel take 4(?)

Tilps:
Using caching computation to perform out of order evaluations was just being too counter productive to multiple threads for processing picked nodes - too much percentage of time was being spent on spinlocks, making it less predictable that each thread would take approximately the same time to do its 'equal sized' chunk of the work.
So I made node_to_process conform to 'Computation' enough that I could make FetchSingleNode into a template which could be given a node_to_process or a caching computation.
This was enough of a refactor that I decided it was worth a separate branch (and correspondingly another draft PR). In my performance testing it sets new records at both small and large node counts.
I'll continue to make any more 'small' adjustments on this branch.


Tilps:
I think I'm actually starting to run out of optimizations worth making...
I think next step is to add some parameters for some of the things that are currently constant. (number of task threads per search thread, and the constants used for deciding when to split tree search, and how many child tasks should run for parallel picked node processing.  I'll do that tomorrow.
Most recent 1k nodes test (incomplete) was inconclusive after 1300 games.

Tilps:
Parameters added.
I've not found any noticeable improvements from tweaking them - variations in the noise of benchmarking make it hard to easily see.

However I did run into a lockup.  Debugger attached indicated that completed_tasks was 1 and task_count was 0. This suggests some kind of leakage of task completion from one round of task execution to another.  I have been unable to reproduce the problem, nor create a hand trace of the code that could have caused it. Someone else reports they have had a couple as well.

Tilps:
Still attempting to diagnose the lockup. I added some internal tracing and got one more sample with it there.  The additional tracing was not very useful, so I am trying again with 'more' tracing. But I am yet to trigger another failure with the additional internal tracing in place. (I'll note that the looping was in the picking section - but that multiple passes of picking and processing some with and some without any tasks being spawned had occurred since the start of the batch.)
I've also run under tsan for a couple of hours and the problems its identified so far are all pre-existing races before my code changes.

Tilps:
Found and fixed some memory corruption issues - which may or may not have had anything to do with the lockups.  Haven't had any lockups since, but its only been a few hours.

Tilps:
I'm reasonably confident that the last change has fixed the hang. Will continue to run testing for another 24 hours.

src/mcts/node.h
```diff
@@ -176,6 +179,15 @@ class Node {
   Bounds GetBounds() const { return {lower_bound_, upper_bound_}; }
   uint8_t GetNumEdges() const { return num_edges_; }
 
+  // Output must point to at least num_edges_ floats.
+  void CopyPolicy(float* output, int max_needed) const {
```
mooskagh, line 183: Doesn't seem to be used?

src/mcts/node.h
```diff
@@ -570,6 +553,108 @@ class Edge_Iterator : public EdgeAndNode {
...
+  return {*this, child_.get()};
+}
+inline Node_Iterator<false> Node::VisitedNodes() {
+  return {*this, child_.get()};
+}
+
+inline float Node::GetVisitedPolicy() const {
+  if (n_ == 0 || edges_ == nullptr || child_ == nullptr) return 0.0f;
+  float vp = 0.0f;
+  for (const auto& child : Edges()) {
```
mooskagh, line 645: Can that new shiny `VisitedNodesIterator` be used here?

src/mcts/params.cc
```diff
@@ -275,6 +275,25 @@ const OptionId SearchParams::kSolidTreeThresholdId{
     "solid-tree-threshold", "SolidTreeThreshold",
     "Only nodes with at least this number of visits will be considered for "
     "solidification for improved cache locality."};
+const OptionId SearchParams::kTaskWorkersPerSearchWorkerId{
+    "task-workers", "TaskWorkers",
+    "The number of task workers to use to help the search worker."};
+const OptionId SearchParams::kMinimWorkSizeForProcessingId{
```
mooskagh, line 281: Why is it `Minim` while others are `Minimum`? :)

src/mcts/search.cc
```diff
@@ -119,6 +119,26 @@ class MEvaluator {
     return m;
   }
 
+  float GetM(Node* child, float q) const {
+    if (!enabled_ || !parent_within_threshold_) return 0.0f;
+    const float child_m = child->GetM();
+    float m = std::clamp(m_slope_ * (child_m - parent_m_), -m_cap_, m_cap_);
+    // Microsoft compiler does not have a builtin for copysign and emits a
+    // library call which is too expensive for a hot path like this.
+#if defined(_MSC_VER)
```
mooskagh, line 128: May make sense to have an inline function in `util/fastmath.h` or create `util/misc.h` for such cases.

src/mcts/node.h
Tilps: Its used in PickNodesToExtendTask

src/mcts/search.cc
```diff
@@ -467,7 +496,7 @@ void Search::SendMovesStats() const REQUIRES(counters_mutex_) {
     LOGFILE << "=== Move stats:";
     for (const auto& line : move_stats) LOGFILE << line;
   }
-  for (auto edge : root_node_->Edges()) {
+  for (auto& edge : root_node_->Edges()) {
```
mooskagh, line 499: Hmmm it may make sense to make iterators return pointers rather than references to avoid such mistakes.

src/mcts/search.h
```diff
@@ -282,50 +301,143 @@ class SearchWorker {
...
+             int collision_limit)
+        : task_type(0),
+          start(node),
+          collision_limit(collision_limit),
+          base_depth(depth),
+          moves_to_base(base_moves) {}
+    PickTask(int start_idx, int end_idx)
+        : task_type(1), start_idx(start_idx), end_idx(end_idx) {}
+  };
+  std::mutex picking_tasks_mutex_;
```
mooskagh, line 430: With a mutex from utils/mutex.h we can have thread safety annotations. :)
But I think at this point it may make sense to switch to absl, it's pretty easy:

https://github.com/mooskagh/lc0/blob/lc2v3/subprojects/abseil.wrap
https://github.com/mooskagh/lc0/blob/lc2v3/meson.build#L545

src/mcts/search.h
```diff
@@ -282,50 +301,143 @@ class SearchWorker {
...
   PositionHistory history_;
   int number_out_of_order_ = 0;
   const SearchParams& params_;
   std::unique_ptr<Node> precached_node_;
   const bool moves_left_support_;
   IterationStats iteration_stats_;
   StoppersHints latest_time_manager_hints_;
+
+  struct PickTask {
+    int task_type;
```
mooskagh, line 406: enum?

src/mcts/search.h
```diff
@@ -282,50 +301,143 @@ class SearchWorker {
...
+
+    // For task type 0 - gathering.
```
mooskagh, line 408: There is `std::variant`, but it's quite a pain to use.

src/mcts/search.h
```diff
@@ -282,50 +301,143 @@ class SearchWorker {
...
+  // History is reset and extended by PrefetchIntoCache().
   PositionHistory history_;
   int number_out_of_order_ = 0;
   const SearchParams& params_;
   std::unique_ptr<Node> precached_node_;
   const bool moves_left_support_;
   IterationStats iteration_stats_;
   StoppersHints latest_time_manager_hints_;
+
+  struct PickTask {
```
mooskagh, line 405: That's quite a weird struct, constructor declared in the middle of variables, half of variables are with trailing underscores while the other half is not.

Ideally it would be nice to have it converted into a class (with all member variables private), but it's a good idea at least to tidy it up.

mooskagh (commented):
Done with the easy part - search.* and cache.* are still to do.

src/mcts/node.h
```diff
@@ -570,6 +553,108 @@ class Edge_Iterator : public EdgeAndNode {
...
+// This class has multiple hypostases within one class:
+// * Range (begin() and end() functions)
+// * Iterator (operator++() and operator*())
+// It's more customary to have those as two classes, but
+// creating zoo of classes and copying them around while iterating seems
+// excessive.
+//
+// All functions are not thread safe (must be externally synchronized).
+template <bool is_const>
+class Node_Iterator {
```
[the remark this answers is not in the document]
Tilps: Agreed I should rename this.

src/mcts/node.h
```diff
@@ -570,6 +553,108 @@ class Edge_Iterator : public EdgeAndNode {
...
+          break;
+        }
+      }
+      if (current_idx_ == total_count_) {
+        node_ptr_ = nullptr;
+      }
+    } else {
+      do {
+        node_ptr_ = node_ptr_->sibling_.get();
+        // If n started is 0, can jump direct to end.
```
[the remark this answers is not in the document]
Tilps: Is this a request to improve documentation or do you not know about the sorted edges optimization?
(Since edges are sorted by P, the edges can be partitioned into two sections NStarted > 0, and NStarted == 0) Anytime we need a new edge to start its the first from the NStarted == 0 section, since highest P wins.

src/mcts/params.cc
Tilps: Typos :)

src/mcts/search.cc
Tilps: true - this is a copy/paste of code immediately above.

src/mcts/search.h
Tilps: I chose std::mutex because I was using condition variable and didn't know about get_raw - will fix

src/mcts/search.h
Tilps: I'm thinking you may have misread this code a bit.  The constructors are at the end, there are no variables after the constructors.  What I will take as a point I had intended to cleanup was that I should move it as a whole up to be next to NodeToProcess rather than being here in the middle of the SearchWorker field declarations.

src/mcts/search.cc
```diff
@@ -1048,285 +1151,624 @@ void SearchWorker::GatherMinibatch() {
...
-      collision_limit =
-          std::min(collision_limit, estimated_visits_to_change_best);
-      assert(collision_limit >= 1);
-      second_best_edge.Reset();
-    }
+      const float cpuct = ComputeCpuct(params_, node->GetN(), is_root_node);
+      const float puct_mult =
+          cpuct * std::sqrt(std::max(node->GetChildrenVisits(), 1u));
+      int cache_filled_idx = -1;
+      while (cur_limit > 0) {
```
mooskagh, line 1562: When there are more than a bunch of visits to distribute, it's likely that using a heap will be more efficient.
For most of the leaf nodes probably not though.

Here is me trying to do this in a toy example: https://github.com/mooskagh/lc0/blob/lc2v3/src/lc2/search/nodes-worker.cc#L120

src/mcts/search.cc
```diff
@@ -1027,9 +1126,13 @@ void SearchWorker::InitializeIteration(
...
+  computation_->Reserve(params_.GetMiniBatchSize());
   minibatch_.clear();
+  minibatch_.reserve(2 * params_.GetMiniBatchSize());
 }
 
+#define USE_WORKERS 1
+
 // 2. Gather minibatch.
 // ~~~~~~~~~~~~~~~~~~~~
 void SearchWorker::GatherMinibatch() {
```
mooskagh, line 1138: Do I understand correctly that only one (SearchWorker)thread may be inside `GatherMinibatch()` at the same time?
If yes, how do we ensure that? it seems that the mutex is gone.
If no, how do we protect multiple gatherers from touching the same nodes?

mooskagh: Disregard this question, found lock in `PickNodesToExtend()`

src/mcts/search.cc
```diff
@@ -949,6 +979,74 @@ Search::~Search() {
...
+    PickTask* task = nullptr;
+    int id = 0;
+    {
+      int spins = 0;
+      while (true) {
+        int nta = next_task_available_.load(std::memory_order_acquire);
+        int tc = task_count_.load(std::memory_order_acquire);
+        if (nta < tc) {
+          int val = 0;
+          if (task_taker_.compare_exchange_weak(val, 1,
```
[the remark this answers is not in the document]
Tilps: I don't think atomic_flag is any more efficient than atomic_int - pretty sure they optimize to identical code... but I'll think about it.

src/mcts/search.cc
Tilps: heap scales better in worst case, but the average number of loops here is just too low in practice I think.  My cpu profiles show the vast majority of CPU cost is before this loop starts - incurred bringing everything into cache from memory. Inside this loop the actual inner loop is not the most expensive part anymore - and the rest of the logic doesn't improve with a heap.
The start of the inner loop is trivially cheap except for the first time we see a new edge - but that cost can't be avoided.
So this parts of the loop we potentially avoid with heap is now just reading from some smallish stack local arrays that stay hot and in the cache and running a couple of if statements to update best/second best.  Its quite cheap.
Rather than incurring the heap complexity I had considered pre-populating 'best' with 'second best' from the previous loop.  That way the if statements fire less often.  But as I said its so cheap its not been a priority for optimization.

src/utils/cache.h
```diff
@@ -322,4 +328,293 @@ class LruCacheLock {
...
+// A hash-keyed cache. Thread-safe. Takes ownership of all values, which are
+// deleted upon eviction; thus, using values stored requires pinning them, which
+// in turn requires Unpin()ing them after use. The use of HashKeyedCacheLock is
+// recommend to automate this element-memory management.
+// Unlike LRUCache, doesn't even consider trying to support LRU order.
+// Does not support delete.
+// Does not support replace! Inserts to existing elements are silently ignored.
+// FIFO eviction.
+// Assumes that eviction while pinned is rare enough to not need to optimize unpin for that case.
+template <class V>
```
mooskagh, line 340: It makes sense to put it into a separate `.h` file, or rename existing one to `cache-old-v2.h` like we did with https://github.com/LeelaChessZero/lc0/blob/master/src/utils/cache-old.h and keep this in cache.h.

src/utils/cache.h
```diff
@@ -322,4 +328,293 @@ class LruCacheLock {
...
+ private:
+  struct Entry {
+    Entry() {}
+    Entry(uint64_t key, std::unique_ptr<V> value)
+        : key(key), value(std::move(value)) {}
+    uint64_t key;
+    std::unique_ptr<V> value;
+    int pins = 0;
+    // 1 for populated.
+    int metadata = 0;
```
mooskagh, line 513: Deserves better name, and probably should be enum.

src/utils/cache.h
```diff
@@ -322,4 +328,293 @@ class LruCacheLock {
...
+        return true;
+      }
+      ++idx;
+      if (idx >= hash_.size()) idx -= hash_.size();
+    }
+    return false;
+  }
+
+  // Looks up and pins the element by key. Returns nullptr if not found.
+  // If found, brings the element to the head of the queue (makes it last to
```
mooskagh, line 403: Probably this part of the comment is not correct.

src/mcts/node.h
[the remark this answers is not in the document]
Tilps: Renamed in 1490.

src/mcts/node.h
```diff
@@ -240,6 +235,9 @@ class Node {
   ConstIterator Edges() const;
   Iterator Edges();
 
+  Node_Iterator<true> VisitedNodes() const;
```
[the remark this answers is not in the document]
Tilps: Done in 1490.

src/mcts/node.h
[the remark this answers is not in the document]
Tilps: I added half a paragraph :P - but done in 1490.

src/mcts/node.h
Tilps: In 1490 policy caching hasn't yet been removed, so this method doesn't exist - but otherwise yes it could.

src/mcts/params.cc
Tilps: Fixed in 1490.

src/mcts/search.h
Tilps: Fixed in 1490.

src/mcts/search.h
Tilps: Done in 1490.

src/mcts/search.h
```diff
@@ -282,50 +301,143 @@ class SearchWorker {
...
+          collision_limit(collision_limit),
+          base_depth(depth),
+          moves_to_base(base_moves) {}
+    PickTask(int start_idx, int end_idx)
+        : task_type(1), start_idx(start_idx), end_idx(end_idx) {}
+  };
+  std::mutex picking_tasks_mutex_;
+  std::vector<PickTask> picking_tasks_;
+  std::atomic<int> task_count_ = -1;
+  std::atomic<int> task_taker_ = 0;
```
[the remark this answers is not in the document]
Tilps: Renamed in 1490.

src/mcts/search.h
```diff
@@ -282,50 +301,143 @@ class SearchWorker {
...
+  std::atomic<int> next_task_available_ = 0;
```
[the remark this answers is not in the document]
Tilps: Renamed in 1490.

src/mcts/search.cc
```diff
@@ -949,6 +979,74 @@ Search::~Search() {
...
+        if (nta >= tc && exiting_) return;
+        task_added_.wait(lock);
+        // And refresh again now we're awake.
+        nta = next_task_available_.load(std::memory_order_acquire);
+        tc = task_count_.load(std::memory_order_acquire);
+        if (nta >= tc && exiting_) return;
+      }
+    }
+    if (task != nullptr) {
+      if (task->task_type == 0) {
```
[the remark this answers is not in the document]
Tilps: Done in 1490.

src/mcts/search.cc
```diff
@@ -1048,285 +1151,624 @@ void SearchWorker::GatherMinibatch() {
...
-namespace {
-void IncrementNInFlight(Node* node, Node* root, int amount) {
-  if (amount == 0) return;
-  while (true) {
-    node->IncrementNInFlight(amount);
-    if (node == root) break;
-    node = node->GetParent();
+void SearchWorker::PickNodesToExtend(int collision_limit) {
+  {
+    task_count_.store(0, std::memory_order_release);
```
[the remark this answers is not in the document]
Tilps: yes - they are different task batches.  I have extracted the reset logic to its own method in 1490.

src/mcts/search.cc
```diff
@@ -1048,285 +1151,624 @@ void SearchWorker::GatherMinibatch() {
...
+          Node* node = picked_node.node;
+          for (node = node->GetParent();
+               node != search_->root_node_->GetParent();
+               node = node->GetParent()) {
+            node->IncrementNInFlight(extra);
+          }
+        }
+        if (--collision_events_left <= 0) should_exit = true;
+        if ((collisions_left -= picked_node.multivisit) <= 0)
+          should_exit = true;
```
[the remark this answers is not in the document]
Tilps: Fixed.

src/mcts/search.cc
```diff
@@ -1048,285 +1151,624 @@ void SearchWorker::GatherMinibatch() {
...
-    // just start search of another node.
-    if (picked_node.IsCollision()) {
-      if (--collision_events_left <= 0) return;
-      if ((collisions_left -= picked_node.multivisit) <= 0) return;
-      if (search_->stop_.load(std::memory_order_acquire)) return;
-      continue;
+    int prev_size = static_cast<int>(minibatch_.size());
+    int new_start = prev_size;
+    PickNodesToExtend(
+        std::min(collision_events_left,
```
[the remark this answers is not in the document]
Tilps: Done, thanks.

src/mcts/search.cc
```diff
@@ -1048,285 +1151,624 @@ void SearchWorker::GatherMinibatch() {
...
+    task_added_.notify_all();
+  }
+  std::vector<Move> empty_movelist;
+  // This lock must be held until after the task_completed_ wait succeeds below.
+  // Since the tasks perform work which assumes they have the lock, even though
+  // actually this thread does.
+  SharedMutex::Lock lock(search_->nodes_mutex_);
+  PickNodesToExtendTask(search_->root_node_, 0, collision_limit, empty_movelist,
+                        &minibatch_, &main_workspace_);
+  {
```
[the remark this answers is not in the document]
Tilps: Yes - they are a remnant of when I used locks. Fixed in 1490.

src/mcts/search.cc
```diff
@@ -1048,285 +1151,624 @@ void SearchWorker::GatherMinibatch() {
...
 
-namespace {
-void IncrementNInFlight(Node* node, Node* root, int amount) {
-  if (amount == 0) return;
-  while (true) {
-    node->IncrementNInFlight(amount);
-    if (node == root) break;
-    node = node->GetParent();
+void SearchWorker::PickNodesToExtend(int collision_limit) {
+  {
```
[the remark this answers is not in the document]
Tilps: Done in 1490.

src/mcts/search.cc
```diff
@@ -1048,285 +1151,624 @@ void SearchWorker::GatherMinibatch() {
...
-      continue;
+    int prev_size = static_cast<int>(minibatch_.size());
+    int new_start = prev_size;
+    PickNodesToExtend(
+        std::min(collision_events_left,
+                 std::min(collisions_left,
+                          params_.GetMiniBatchSize() - minibatch_size)));
+    bool should_exit = false;
+    int non_collisions = 0;
+    for (int i = prev_size; i < minibatch_.size(); i++) {
```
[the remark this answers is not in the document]
Tilps: Fixed some of them ;) (in 1490)

src/mcts/search.cc
```diff
@@ -1048,285 +1151,624 @@ void SearchWorker::GatherMinibatch() {
...
 
-// Returns node and whether there's been a search collision on the node.
-SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend(
-    int collision_limit) {
-  // Starting from search_->root_node_, generate a playout, choosing a
-  // node at each level according to the MCTS formula. n_in_flight_ is
-  // incremented for each node in the playout (via TryStartScoreUpdate()).
+void SearchWorker::PickNodesToExtendTask(Node* node, int base_depth,
+                                         int collision_limit,
+                                         const std::vector<Move>& moves_to_base,
```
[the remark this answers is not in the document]
Tilps: moves to get from root to the base_depth...  I think its correct?

src/mcts/search.cc
```diff
@@ -1048,285 +1151,624 @@ void SearchWorker::GatherMinibatch() {
...
+                                         std::vector<NodeToProcess>* receiver,
+                                         TaskWorkspace* workspace) {
+  // TODO: Bring back pre-cached nodes created outside locks in a way that works
+  // with tasks.
+  // TODO: pre-reserve visits_to_perform for expected depth and likely maximum
+  // width. Maybe even do so outside of lock scope.
+  std::vector<std::unique_ptr<int[]>> visits_to_perform;
```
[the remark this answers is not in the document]
Tilps: I started with std::vector<std::vector<int>> - I switched to this when I decided to make them all fixed length and track the populated amount externally to vector in order to avoid paying clearing costs when I could just reset the tracked length.  I'll try std::array in 1490.

src/mcts/search.cc
```diff
@@ -1048,285 +1151,624 @@ void SearchWorker::GatherMinibatch() {
...
+  auto& vtp_buffer = workspace->vtp_buffer;
+  visits_to_perform.reserve(30);
+  std::vector<int> vtp_last_filled;
+  vtp_last_filled.reserve(30);
+  std::vector<int> current_path;
+  current_path.reserve(30);
+  std::vector<Move> moves_to_path = moves_to_base;
+  if (moves_to_path.capacity() < 30) {
```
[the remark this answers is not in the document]
Tilps: Yes that one is redundant.  I've got some others that look similar but are not... Fixed in 1490.

src/mcts/search.cc
```diff
@@ -1048,285 +1151,624 @@ void SearchWorker::GatherMinibatch() {
...
+    moves_to_path.reserve(30);
+  }
+  // Sometimes receiver is reused, othertimes not, so only jump start if small.
+  if (receiver->capacity() < 30) {
+    receiver->reserve(receiver->size() + 30);
+  }
+
+  // These 2 are 'filled pre-emptively'.
+  float current_pol[256];
```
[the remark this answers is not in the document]
Tilps: Done in 1490.

src/mcts/search.cc
```diff
@@ -1048,285 +1151,624 @@ void SearchWorker::GatherMinibatch() {
...
+  int max_limit = std::numeric_limits<int>::max();
+
+  current_path.push_back(-1);
+  while (current_path.size() > 0) {
+    // First prepare visits_to_perform.
+    if (current_path.back() == -1) {
+      // Need to do n visits, where n is either collision_limit, or comes from
+      // visits_to_perform for the current path.
+      int cur_limit = collision_limit;
+      if (current_path.size() - 1 > 0) {
```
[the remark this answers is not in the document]
Tilps: Fixed in 1490.

src/mcts/search.cc
```diff
@@ -1048,285 +1151,624 @@ void SearchWorker::GatherMinibatch() {
...
-          node_to_revert->RevertTerminalVisits(wl, d, m + (float)depth_counter,
-                                               terminal_visits);
-          depth_counter++;
-          // Even if original tree still exists, we don't want to revert more
-          // than until new root.
-          if (depth_counter > depth - 1) break;
-          // If wl != 0, we would have to switch signs at each depth.
+      // First check if node is terminal or not-expanded.  If either than create
+      // a collision of appropriate size and pop current_path.
+      if (!node->HasChildren() || node->GetN() == 0) {
```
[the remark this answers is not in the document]
Tilps: you were on to something here ;)
Fixed both in 1483 and 1490 as this was a bug.

src/utils/cache.h
Tilps: I've dropped this from 1490 for now - I'll revisit it in a future PR.

src/mcts/search.cc
```diff
@@ -1048,285 +1151,624 @@ void SearchWorker::GatherMinibatch() {
...
+      {
+        // Ensure nothing takes tasks yet
+        task_count_.store(0, std::memory_order_release);
+        next_task_available_.store(0, std::memory_order_release);
+        completed_tasks_.store(0, std::memory_order_release);
+        picking_tasks_.clear();
+        // Reserve because resizing breaks pointers held by the task threads.
+        // Reserve 100 - since we already did that in picking and why use a
+        // different size?
+        picking_tasks_.reserve(100);
```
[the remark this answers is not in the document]
Tilps: Consolidated to a single location instead in 1490.

src/mcts/search.h
```diff
@@ -282,50 +301,143 @@ class SearchWorker {
...
+        : task_type(0),
+          start(node),
+          collision_limit(collision_limit),
+          base_depth(depth),
+          moves_to_base(base_moves) {}
+    PickTask(int start_idx, int end_idx)
+        : task_type(1), start_idx(start_idx), end_idx(end_idx) {}
+  };
+  std::mutex picking_tasks_mutex_;
+  std::vector<PickTask> picking_tasks_;
```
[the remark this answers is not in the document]
Tilps: It makes the reset code a bit ugly compared to just calling vector.clear().  I'll keep it in mind, but I am not sold right now.

src/mcts/search.cc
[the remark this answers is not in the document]
Tilps: Actually no, it still needs extracting.

Tilps:
Closing in favour of already merged code and further future path.

PR #1485, opened 2021-01-09
replace floor in FastPow2 and rename to FastExp2

borg323:
@Tilps noticed that MSVC is not optimizing away the calls to `floor()` and after further investigation it seems that only gcc does.  This code replaces `floor(a)` with `static_cast<int32_t>(a)` for positive a and `static_cast<int32_t>(a - 1)` for negative a. While this is obviously wrong for negative integer a, the approximation used has zero error in this case (by design). 
Microbenchmarks with gcc show the resulting code to be considerably faster than the original, that already had the call to `floor()` optimized away.

### DIFF
--- a/src/utils/fastmath.h
+++ b/src/utils/fastmath.h
@@ -56,9 +56,17 @@ inline float FastLog2(const float a) {
 // integer and f the fractional part, f>=0. The constant k is used to tune the
 // approximation accuracy. In the final version some constants were slightly
 // modified for better accuracy with 32 bit floating point math.
-inline float FastPow2(const float a) {
-  if (a < -126) return 0.0;
-  int32_t exp = static_cast<int32_t>(floor(a));
+inline float FastExp2(const float a) {
+  int32_t exp;
+  if (a < 0) {
+    if (a < -126) return 0.0;
+    // Not all compilers optimize floor, so we use (a-1) here to round down.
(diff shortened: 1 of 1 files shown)


oscardssmith:
while we're changing this function, `out = 1.0f + out * (0.6602339f + 0.33976606f * out);` should probably be changed to `out = evalpoly(out, (1.0f, 0.6602339f, 0.33976606f))` which is cleaner and likely faster since it will use fma where available.

Tilps (approved):

borg323:
With gcc the speed improvement is enough to add an extra term to the polynomial and still be slightly ahead. The following reduces the error by an order of magnitude:
`out = 1.0f + out * (0.69571865f + out * (0.22518457f + 0.07909678 * out));`

ddobbelaere:
Cool trick (especially the negative integer case)!

borg323:
> Cool trick (especially the negative integer case)!

That was @Tilps' idea.

Naphthalin:
> 
> 
> With gcc the speed improvement is enough to add an extra term to the polynomial and still be slightly ahead. The following reduces the error by an order of magnitude:
> `out = 1.0f + out * (0.69571865f + out * (0.22518457f + 0.07909678 * out));`

I would highly appreciate the better accuracy of `FastExp` for my betaMCTS branch stuff if it doesn't cost much.

Issue #1486, opened 2021-01-10
TSAN found issues (2021-01-10).


Tilps:
1) GetBestChildrenNoTemperature doesn't check GetN() > 0 before accessing Edges() for input, or similarly before accessing IsTerminal for edges.
2) PopulateCommonIterationStats similarly doesn't check GetN() > 0.
3) GetTwoFoldDraws isn't cached at SearchParams construction time, but is accessed from multiple threads during search (and options dict is not technically thread safe).
4) std::localtime is used in FormatTime called from the logger, which can be called from arbitrary threads - but std::localtime is non-threadsafe.
5) FetchSingleNodeResult doesn't check node_to_process.IsCollision - which means it falls through to nn_requested and accesses the node despite it being a collision target.
6) PopulateCommonIterationStats writes 'now' to nps_start_time_, potentially from multiple threads simultaneously. (tsan registered two writes one of 8 bytes and one of 1 byte - but this is the only written shared state I can see, I thought a time_point was only 8 bytes...)
7) Selfplay game performs make move before destroying the search instance. (Thus pending shared collisions can be run against a deleted tree.)

mooskagh:
Can be closed?

Tilps:
yes.

PR #1487, opened 2021-01-11
Fix multiple tsan reported issues.

Tilps:
1) GetBestChildrenNoTemperature was not respecting the GetN test before accessing a node's details.
2) PopulateCommonIterationStats had similar issues.
3) GetTwoFoldDraws property was used by SearchWorkers without being cached.
4) Wrap std::localtime usage in a mutex.
5) Fix a pointless access of node WL/D/M for collisions.
6) Put nps_start_time_ behind counters mutex to avoid races.
7) Ensure search is disposed before performing the move during selfplay.

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -601,7 +601,8 @@ void Search::EnsureBestMoveKnown() REQUIRES(nodes_mutex_)
                            : GetBestChildNoTemperature(root_node_, 0);
   final_bestmove_ = bestmove_edge.GetMove(played_history_.IsBlackToMove());
 
-  if (bestmove_edge.GetN() > 0 && bestmove_edge.HasNode() && bestmove_edge.node()->HasChildren()) {
+  if (bestmove_edge.GetN() > 0 && bestmove_edge.HasNode() &&
+      bestmove_edge.node()->HasChildren()) {
     final_pondermove_ = GetBestChildNoTemperature(bestmove_edge.node(), 1)
                             .GetMove(!played_history_.IsBlackToMove());
   }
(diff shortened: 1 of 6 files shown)


Tilps:
Fixes everything found in #1486 I think...

src/mcts/search.cc
```diff
@@ -648,7 +653,10 @@ std::vector<EdgeAndNode> Search::GetBestChildrenNoTemperature(Node* parent,
           // This default isn't used as wl only checked for case edge is
           // terminal.
           const auto wl = edge.GetWL(0.0f);
-          if (!edge.IsTerminal() || !wl) return kNonTerminal;
+          // Not safe to access IsTerminal if GetN is 0.
+          if (edge.GetN() == 0 || !edge.IsTerminal() || !wl) {
```
borg323, line 657: Maybe the check should be added to `EdgeAndNode::IsTerminal()`?

Tilps: I 100% am onboard with that, but I'm not on board with having to validate all the existing uses of EdgeAndNode::IsTerminal to see which ones fall into code paths that are allowed to check IsTerminal... So I went with the 'safer' change.

src/mcts/search.cc
```diff
@@ -600,7 +601,8 @@ void Search::EnsureBestMoveKnown() REQUIRES(nodes_mutex_)
                            : GetBestChildNoTemperature(root_node_, 0);
   final_bestmove_ = bestmove_edge.GetMove(played_history_.IsBlackToMove());
 
-  if (bestmove_edge.HasNode() && bestmove_edge.node()->HasChildren()) {
+  if (bestmove_edge.GetN() > 0 && bestmove_edge.HasNode() &&
```
borg323, line 604: The `HasNode()` check is superfluous here (and should have been first).

borg323 (approved):
Approved with one comment.

src/mcts/search.cc
Tilps: I'll remove it.

PR #1489, opened 2021-01-11
Add SmartPruningStopper logic if 0 or 1 non losing moves remain.

Naphthalin:
Our current `SmartPruningStopper` logic checks stops search prematurely if there is only 1 legal move or a proven win. We also want to stop search if all moves are proven losses anyway and more importantly also when only 1 move isn't a proven loss. This PR attempts to do this by counting the number of proven losing moves, adding a `stats->num_losing_edges` field and comparing them with the total number of legal moves in `stopper.cc`.


Naphthalin:
Potential side effect of this: when being close to get mated, Leela will probably instamove but not displaying mate scores anymore as she will move before proving the loss for herself. But I think this is better than wasting time at actual only moves when the game isn't over yet, and if we want her to look for mates against herself we should consider to make it a special feature.

src/mcts/stoppers/stoppers.cc
```diff
@@ -196,6 +196,10 @@ bool SmartPruningStopper::ShouldStop(const IterationStats& stats,
     LOGFILE << "Only one possible move. Moving immediately.";
     return true;
   }
+  if (stats.edge_n.size() <= stats.num_losing_edges + 1) {
```
mooskagh, line 199: I believe `if (stats.edge_n.size() - stats.num_losing_edges <= 1)` would be a bit easier to read.

Naphthalin: I agree, I just wasn't exactly sure whether mixing `int` and `uint` could have a potential side effect there.

Tilps: static_cast<int>(stats.edge_n.size()) for clarity if making that change.

Tilps (approved):
Approved assuming that GetBestChildrenNoTemperature is bulletproof against ever returning a TB loss based on best N when there is an unproven option available.

src/mcts/stoppers/stoppers.cc
Naphthalin: I think the current way is easy enough to read, and is probably also shorter as we don't have to apply int casting.

PR #1490, opened 2021-01-13
An alternative multigather which skips a few details in order to be able to make it optional.

Tilps:
Specifically the node size isn't reduced.  Lots of extra code duplication.
I also dropped HashKeyCache from this branch since it wasn't practical to make it optional.

I can't really test how much benefit HashKeyCache is - since the cache isn't that much of a hotspot on my machine, although its definitely one on others...
The default values for MaxCollisionEvents/Visits are also unchanged in this PR - but the potential issues of leaving them as per their defaults still apply.

### DIFF
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -256,7 +256,6 @@ class Node {
   VisitedNode_Iterator<true> VisitedNodes() const;
   VisitedNode_Iterator<false> VisitedNodes();
 
-
   // Deletes all children.
   void ReleaseChildren();
 
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -646,7 +646,10 @@ class VisitedNode_Iterator {
     } else {
       do {
         node_ptr_ = node_ptr_->sibling_.get();
-        // If n started is 0, can jump direct to end.
+        // If n started is 0, can jump direct to end due to sorted policy
+        // ensuring that each time a new edge becomes best for the first time,
+        // it is always the first of the section at the end that has NStarted of
+        // 0.
         if (node_ptr_ != nullptr && node_ptr_->GetN() == 0 &&
             node_ptr_->GetNInFlight() == 0) {
           node_ptr_ = nullptr;
--- a/src/mcts/node.h
+++ b/src/mcts/node.h
@@ -120,6 +120,9 @@ class EdgeAndNode;
 template <bool is_const>
 class Edge_Iterator;
 
+template <bool is_const>
+class VisitedNode_Iterator;
+
 class Node {
  public:
   using Iterator = Edge_Iterator<false>;
@@ -175,6 +178,15 @@ class Node {
   typedef std::pair<GameResult, GameResult> Bounds;
   Bounds GetBounds() const { return {lower_bound_, upper_bound_}; }
   uint8_t GetNumEdges() const { return num_edges_; }
+
+  // Output must point to at least max_needed floats.
+  void CopyPolicy(float* output, int max_needed) const {
+    if (!edges_) return;
+    int loops = std::min(static_cast<int>(num_edges_), max_needed);
+    for (int i = 0; i < loops; i++) {
+      output[i] = edges_[i].GetP();
+    }
+  }
 
   // Makes the node terminal and sets it's score.
   void MakeTerminal(GameResult result, float plies_left = 0.0f,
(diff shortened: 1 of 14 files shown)


src/mcts/node.h
```diff
@@ -570,6 +591,98 @@ class Edge_Iterator : public EdgeAndNode {
...
+            node_ptr_->GetNInFlight() == 0) {
+          node_ptr_ = nullptr;
+          break;
+        }
+      } while (node_ptr_ != nullptr && node_ptr_->GetN() == 0);
+    }
+  }
+  Node* operator*() {
+    if (solid_)
+      return &(node_ptr_[current_idx_]);
```
mooskagh, line 666: Braces around the statement as it's in a separate line (same for `else`)

src/mcts/search.cc
```diff
@@ -123,6 +123,26 @@ class MEvaluator {
     return m;
   }
 
+  float GetM(Node* child, float q) const {
+    if (!enabled_ || !parent_within_threshold_) return 0.0f;
+    const float child_m = child->GetM();
+    float m = std::clamp(m_slope_ * (child_m - parent_m_), -m_cap_, m_cap_);
+    // Microsoft compiler does not have a builtin for copysign and emits a
+    // library call which is too expensive for a hot path like this.
+#if defined(_MSC_VER)
```
mooskagh, line 132: Would be nice to move that into `utils/misc.h` :)

src/mcts/search.h
```diff
@@ -213,6 +214,25 @@ class SearchWorker {
         moves_left_support_(search_->network_->GetCapabilities().moves_left !=
                             pblczero::NetworkFormat::MOVES_LEFT_NONE) {
     Numa::BindThread(id);
+    for (int i = 0; i < params.GetTaskWorkersPerSearchWorker(); i++) {
```
mooskagh, line 217: Maybe makes sense not to start threads if MultiGatherEnabled is false?

src/mcts/search.cc
```diff
@@ -1602,6 +2426,62 @@ void SearchWorker::FetchMinibatchResults() {
   }
 }
 
+template <typename Computation>
```
mooskagh, line 2429: Why template if it's only called with one type?

src/mcts/node.h
```diff
@@ -176,6 +179,15 @@ class Node {
   Bounds GetBounds() const { return {lower_bound_, upper_bound_}; }
   uint8_t GetNumEdges() const { return num_edges_; }
 
+  // Output must point to at least max_needed floats.
+  void CopyPolicy(float* output, int max_needed) const {
```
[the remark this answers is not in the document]
Tilps: Yes you are correct about style guide - have to say its inconsistent with memcpy where length of copy is passed after pointers, but will change.

src/mcts/node.h
```diff
@@ -176,6 +179,15 @@ class Node {
...
+    if (!edges_) return;
+    int loops = std::min(static_cast<int>(num_edges_), max_needed);
+    for (int i = 0; i < loops; i++) {
```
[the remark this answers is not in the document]
Tilps: fancy is the SSE intrinsic version I wrote but discarded because it didn't give any extra nps :P

src/mcts/node.h
Tilps: Hmm yes not sure how this one snuck in. (They usually happen when one liners get autoformatted by clang-format, but that doesn't seem to be the case here.)

src/mcts/search.h
Tilps: Maybe I should change SearchParams to force TaskWorkersPerSearchWorker to 0 if MultiGather is disabled.

src/mcts/search.cc
Tilps: Immediate follow up PRs intend to reduce code duplication - one of which will mean this will start being called from the non-multigather path - with a different type. (It was called with two types in 1483 where I copied the function from, I was just keeping this PR's influence into the existing path somewhat minimal for 'clarity of safety'.  The follow up PRs will be focused on the single duplications, making them easy to review without the noise of everything else this PR is adding.

src/mcts/node.h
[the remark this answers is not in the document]
Tilps: but yeah, I prefer the simple loop for this.

src/mcts/search.cc
Tilps: utils/fastmath.h - done.

mooskagh (approved):

PR #1492, opened 2021-01-18
build with alternative malloc

borg323:


Tilps (approved):
I read somewhere where for some compiler they also had to pass some CCFlags to disable 'builtin malloc' or somesuch. Hopefully that is legacy advise.

PR #1496, opened 2021-01-23
Change PickNodeToExtend to use EnsureNodeTwoFoldCorrectForDepth

Tilps:
Rather than duplicating code.
Cleanup post PR1490 merge.

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -1980,45 +1980,14 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend(
       }
       return NodeToProcess::Collision(node, depth, collision_limit);
     }
-    // If terminal, we either found a twofold draw to be reverted, or
-    // reached the end of this playout.
+
+    // Probably best place to check for two-fold draws consistently.
+    // Depth starts with 1 at root, so real depth is depth - 1.
+    EnsureNodeTwoFoldCorrectForDepth(node, depth - 1);
+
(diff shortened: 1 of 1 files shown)


src/mcts/search.cc
```diff
@@ -1980,45 +1980,14 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend(
       }
       return NodeToProcess::Collision(node, depth, collision_limit);
     }
-    // If terminal, we either found a twofold draw to be reverted, or
-    // reached the end of this playout.
+
+    // Probably best place to check for two-fold draws consistently.
+    // Depth starts with 1 at root, so real depth is depth - 1.
+    EnsureNodeTwoFoldCorrectForDepth(node, depth - 1);
```
borg323, line 1986: Any particular reason this was moved before the if?

Tilps: Because if its inside the if - we have to check IsTerminal again afterwards, since if we revert two fold it shouldn't count as a visit. (Previously that path fell through - but as this is a standalone function it isn't a straight forward transform.)
Moving it outside the if is safe because IsTwoFoldTerminal (checked inside the function) can only be true if IsTerminal is also true.

Tilps: (I also assume the compiler will inline/partial inline as appropriate.)

borg323 (approved):

PR #1497, opened 2021-01-23
Actively split pick tasks as early as possible.

Tilps:
Should reduce picking latency.

NPS tests not started yet - but fixed node test underway is 'fine' so far.

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -1576,35 +1576,6 @@ void SearchWorker::PickNodesToExtendTask(Node* node, int base_depth,
         // as its not handled on the path to it, since there isn't one.
         node->IncrementNInFlight(cur_limit);
       }
-      // !is_root_node only for clarity, it can never pass the second condition.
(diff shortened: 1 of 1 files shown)


Tilps:
Full benchmark results are not super conclusive but maybe 1% faster.  Startposition go 20M nodes was definitely in the noise.

It seems the 'correct' thing to do, I suspect that it'll make more of a difference on a slower cpu machine. Doesn't seem any real downsides.

Tilps:
Retested post merge - 1% improvement was maintained.

src/mcts/search.cc
```diff
@@ -1804,6 +1775,42 @@ void SearchWorker::PickNodesToExtendTask(Node* node, int base_depth,
         }
       }
       is_root_node = false;
+      // Actively do any splits now rather than waiting for potentially long tree walk to get there.
```
mooskagh, line 1778: Line too long. :-P

mooskagh (approved):

src/mcts/search.cc
Tilps: Fixed.

PR #1500, opened 2021-01-23
HashKeyedCache

Tilps:
Has less pointer chasing than the previous version.

This comes from PR1483.

On my machine I've gotten mixed results.
Simple 34 position 10 second a move benchmark (with appropriate command line args) was almost 2% faster.  But go nodes start position with two different cache sizes were both slightly slower. 
I think it should do better on 'tcec like machines' but I haven't benched there yet.

### DIFF
--- a/src/utils/cache.h
+++ b/src/utils/cache.h
@@ -31,6 +31,7 @@
 #include <cstring>
 #include <memory>
 #include <string>
+
 #include "utils/mutex.h"
 
 namespace lczero {
@@ -60,8 +61,8 @@ class HashKeyedCache {
     assert(allocated_ == 0);
   }
 
-  // Inserts the element under key @key with value @val.
-  // Puts element to front of the queue (makes it last to evict).
+  // Inserts the element under key @key with value @val. Unless the key is
+  // already in the cache.
   void Insert(uint64_t key, std::unique_ptr<V> val) {
     if (capacity_.load(std::memory_order_relaxed) == 0) return;
 
@@ -88,7 +89,7 @@ class HashKeyedCache {
     ShrinkToCapacity(capacity_);
   }
 
-  // Checks whether a key exists. Doesn't lock. Of course the next moment the
+  // Checks whether a key exists. Doesn't pin. Of course the next moment the
   // key may be evicted.
   bool ContainsKey(uint64_t key) {
     if (capacity_.load(std::memory_order_relaxed) == 0) return false;
@@ -107,8 +108,7 @@ class HashKeyedCache {
   }
 
   // Looks up and pins the element by key. Returns nullptr if not found.
-  // If found, brings the element to the head of the queue (makes it last to
-  // evict); furthermore, a call to Unpin must be made for each such element.
+  // If found, a call to Unpin must be made for each such element.
   // Use of HashedKeyCacheLock is recommended to automate this pin management.
   V* LookupAndPin(uint64_t key) {
     if (capacity_.load(std::memory_order_relaxed) == 0) return nullptr;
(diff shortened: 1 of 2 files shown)


Tilps:
I'll do some more testing, but the bugs would have clearly cause performance issues - so I'm marking this ready for review on the basis that I assume the performance is no longer 'mixed'.

Tilps:
My additional nps testing was still good gains, possibly better than before on benchmark. The startposition long think was in the noise.
On some profiles the cpu usage shows up lower with this patch in place. So I think I'm happy with it overall.

Tilps:
Synced in the other changes and retested yet again.
This time I also added a test case for rebuilding search tree for same search after truncating it in between.  Was 7% faster with the newcache implementation.  Start position tests other than that one continue to be 'in the noise, slightly lower'.  Benchmark was over 2% faster this time (closer to 2.5%).
Not a super cohesive story as to when its going to be faster (cache inserts/evicts, cache hits, or cache misses), but overall still looks good.

src/utils/cache.h
```diff
@@ -186,121 +207,99 @@ class LruCache {
...
         : key(key), value(std::move(value)) {}
-    K key;
+    uint64_t key;
     std::unique_ptr<V> value;
     int pins = 0;
-    Item* next_in_hash = nullptr;
-    Item* prev_in_queue = nullptr;
-    Item* next_in_queue = nullptr;
+    // 1 for populated.
+    int metadata = 0;
```
mooskagh, line 222: Why not `bool` or `enum`?
And probably something like `is_used` is better than `metadata`.

src/utils/cache.h
```diff
@@ -186,121 +207,99 @@ class LruCache {
...
-    iter->next_in_queue = lru_head_;
-    iter->prev_in_queue = nullptr;
-
-    if (lru_head_) {
-      lru_head_->prev_in_queue = iter;
-    }
-    lru_head_ = iter;
-    if (lru_tail_ == nullptr) {
-      lru_tail_ = iter;
+  void ShrinkToCapacity(int capacity) REQUIRES(mutex_) {
```
mooskagh, line 271: Maybe `EvictToCapacity`?
(I know it was in the original code, but now as I read it, Evict sounds clearer)

src/utils/cache.h
```diff
@@ -31,116 +31,132 @@
...
-    // Now lookup in active list.
-    auto hash = hasher_(key) % hash_.size();
-    for (Item* iter = hash_[hash]; iter; iter = iter->next_in_hash) {
-      if (key == iter->key && value == iter->value.get()) {
-        assert(iter->pins > 0);
-        --iter->pins;
+    // Now the main list.
+    size_t idx = key % hash_.size();
+    while (true) {
+      if (hash_[idx].metadata == 0) break;
```
mooskagh, line 152: `assert(hash_[idx].metadata == 1)` ?

src/utils/cache.h
Tilps: Yeah this evolved - it was a tri state and I was just being lazy.  Now a bool seems fine - will change.

src/utils/cache.h
Tilps: Once its a bool - that'll be implicitly covered by the assert(false) outside the loop.

src/utils/cache.h
Tilps: Done.

src/utils/cache.h
Tilps: Done.

mooskagh (approved):

PR #1501, opened 2021-01-23
Precache nodes per search worker.

Tilps:
Still in draft because:
1) Performance improvement seems small. (<1% on benchmark)
2) I'm precaching regardless of whether multigather is enabled.
3) The precaching size is hard coded at 4* minibatch - when it should probably be (1+moooef)*minibatch.  Possibly with a downscaling ratio of some kind when there is a non-zero number of task workers.  This detail only matters at super short time control where a search is likely to get less than 4*threads*minibatch total nodes - as otherwise it should always be worth it.

It also adds recycling of left over nodes from solid tree - but this seems a super tiny fraction in practice so it doesn't make much of a difference.

### DIFF
--- a/src/mcts/node.cc
+++ b/src/mcts/node.cc
@@ -241,7 +241,8 @@ std::string Node::DebugString() const {
   return oss.str();
 }
 
-bool Node::MakeSolid() {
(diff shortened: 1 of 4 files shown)


Tilps:
Retested after merge, performance improvement is still small - so I don't think it justifies the complexity at this point.

PR #1502, opened 2021-01-23
Better recycling to do with nodes to process.

Tilps:
Basic benchmark shows maybe 1-2% improvement.

There is a one liner in here which is potentially worth its own PR since its not really related. (Releasing the cache lock for ooo cache hit as soon as its processed inside processing task rather than waiting until later while the nodes lock is held.

Draft because:
1) moves_to_visit recycling in ExecuteOneIteration should be conditional on multi-gather (since otherwise we're just collecting empty arrays and never using them).
2) I'm still not really happy with the picking_results recycling.  Picking should recycle the instances used to communicate between task and main, rather than having to move the individual elements, potentially another small speed up over this, and maybe cleaner code.

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -1096,6 +1096,15 @@ void SearchWorker::ExecuteOneIteration() {
   // 7. Update the Search's status and progress information.
   UpdateCounters();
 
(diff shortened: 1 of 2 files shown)


Tilps:
Retested after merge and improvement is nothing now.  This will need further thought.

PR #1503, opened 2021-01-26
Early exit if backend is idle and there is work to do.

Tilps:
Inspired by investigations into `position fen 1r1qr1k1/6bp/1n4p1/2pPpp2/1n3P2/1QN5/1P1N2PP/R4RBK w - - 0 24`

In that position on my machine this change almost doubles NPS at one point during the search (a few million nodes in).
In my 'standard benchmark run' it may be worth a couple of percent.

I think this heuristic can probably be improved upon, but it seems a pretty good start.

Since its an early exit - so long as the nps stays high I think its a clear win (the early exit itself is potentially worth elo at fixed nodes - untested).

I am unclear if there are any scenarios where this will be a net loss of nps - but I haven't thought of any of signifiance so far.

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -1220,6 +1220,14 @@ void SearchWorker::GatherMinibatch2() {
     // If there's something to process without touching slow neural net, do it.
     if (minibatch_size > 0 && computation_->GetCacheMisses() == 0) return;
 
+    // If there is backend work to be done, and the backend is idle - exit
+    // immediately.
+    if (minibatch_size > 0 && computation_->GetCacheMisses() > 0 &&
+        search_->backend_waiting_counter_.load(std::memory_order_relaxed) ==
+            0) {
+      return;
+    }
+
     int new_start = static_cast<int>(minibatch_.size());
 
     PickNodesToExtend(
@@ -1342,12 +1350,6 @@ void SearchWorker::GatherMinibatch2() {
         if ((collisions_left -= picked_node.multivisit) <= 0) return;
         if (search_->stop_.load(std::memory_order_acquire)) return;
       }
-    }
-    // If only collisions and there is backend work to be done, and the backend is idle - exit immediately.
-    if (non_collisions == 0 && minibatch_size > 0 &&
-        computation_->GetCacheMisses() > 0 && 
-        search_->backend_waiting_counter_.load(std::memory_order_relaxed) == 0) {
-      return;
     }
   }
 }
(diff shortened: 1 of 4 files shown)


Tilps:
Worth mentioning this only works for multigather - but I've put the backend idle counter in unconditionally since it should be very cheap - and maybe a similar heuristic could be invented for non multigather if we wanted to bother doing so.

Tilps:
Umm this PR is 'broken' since I put the counter in the wrong spot.  Surprised that it still gives the nps improvement...

Tilps:
Having fixed the broken logic, I retested a more aggressive idea that had previously turned up bad and found it vastly superior.

Tilps:
I realized that this optimization while very cool, interacts very poorly if there is only one search worker. So I've disabled it in that case.
Some more thought suggests it probably isn't as good as it can be for users with multigpu using roundrobin or multiplexing backends since the number of concurrent tasks the backend can perform in that case exceeds 1. So it probably makes sense to cut batches short if the number of backend tasks pending is a number higher than 0.  This should still be better than nothing for that case - and very good for our normal one gpu users and demux backend users using the default 2 threads.
If we want to expand it multigpu roundrobin or multiplexing we'll need the backend to expose its 'minimal desired concurrent batch count'.  I'll leave that as an excercise for the reader for now though :P

Tilps:
Fixed nodes test:
```
Score of lc0testdev vs lc0testdev-Basline: 4314 - 3873 - 11813 [0.511]
...      lc0testdev playing White: 2776 - 1325 - 5899  [0.573] 10000
...      lc0testdev playing Black: 1538 - 2548 - 5914  [0.450] 10000
...      White vs Black: 5324 - 2863 - 11813  [0.562] 20000
Elo difference: 7.7 +/- 3.1, LOS: 100.0 %, DrawRatio: 59.1 %
20000 of 20000 games finished.
```

src/mcts/search.cc
```diff
@@ -1218,6 +1224,19 @@ void SearchWorker::GatherMinibatch2() {
...
+    // If there is backend work to be done, and the backend is idle - exit
+    // immediately.
+    // Only do this fancy work if there are multiple threads as otherwise we
+    // early exit from every batch since there is never another search thread to
+    // be keeping the backend busy. Which would mean that threads=1 has a
+    // massive nps drop.
+    if (multithreaded && minibatch_size > 0 &&
+        computation_->GetCacheMisses() > 0 &&
+        search_->backend_waiting_counter_.load(std::memory_order_relaxed) ==
+            0) {
```
borg323, line 1236: We can make one of the limits here configurable via an option, as a way to tune and disable the optimization.

borg323 (approved):

PR #1508, opened 2021-01-31
Better default value for minimum picking work and clean up some old comments

Tilps:
Both zz and I have found a speed up in benchmarking reducing this parameter to its minimum value.
Also clean up some comments that were wrong/obsolete that were bothering me.

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -1366,11 +1366,8 @@ void SearchWorker::GatherMinibatch2() {
 void SearchWorker::ProcessPickedTask(int start_idx, int end_idx,
                                      TaskWorkspace* workspace) {
   auto& history = workspace->history;
-  // This code runs multiple passes of work across the same input in order to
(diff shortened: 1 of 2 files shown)


src/mcts/search.cc
```diff
@@ -1366,11 +1366,8 @@ void SearchWorker::GatherMinibatch2() {
 void SearchWorker::ProcessPickedTask(int start_idx, int end_idx,
                                      TaskWorkspace* workspace) {
   auto& history = workspace->history;
-  // This code runs multiple passes of work across the same input in order to
-  // reduce taking/dropping mutexes in quick succession.
   history = search_->played_history_;
 
-  // First pass - Extend nodes.
```
kiudee, line 1373: The comments are now obsolete?

Tilps: yes, during a previous refactor all the other passes got removed from this function.

kiudee (approved):

Issue #1466, opened 2020-12-03
Evaluation Blindness to forced 3-fold repetition


Mr-Twave:
BUG REPORT 

https://www.chess.com/computer-chess-championship#event=ccc-blitz-championship-2020-semifinals&game=435

The link I posted for whatever reason now breaks. Here's the PGN.

```[Event "CCC Blitz Championship 2020: Semi-finals (5|5)"]
[Site "?"]
[Date "2020.12.02"]
[Round "1"]
[White "Lc0"]
[Black "Stockfish Classic"]
[Result "1/2-1/2"]
[ECO "A00"]
[GameDuration "00:13:46"]
[GameEndTime "2020-12-02T16:50:19.782 PST"]
[GameStartTime "2020-12-02T16:36:33.079 PST"]
[Opening "Mieses Opening"]
[PlyCount "90"]
[TimeControl "300+5"]

1. d3 e6 2. Be3 Bd6 3. Nc3 Nh6 4. Qd2 O-O 5. O-O-O Kh8 6. Kb1 Re8 7. Ka1 Bf8 8.
Qe1 Ng8 9. Bc1 Nf6 10. Nb1 Ng8 11. c3 f5 12. a3 d6 13. Nf3 Bd7 14. d4 Ba4 15.
Ng5 Qd7 16. Rd3 h6 17. e4 Nc6 18. Rh3 Na5 19. Nd2 Nf6 20. Bd3 c5 21. dxc5 dxc5
22. Bb1 Kg8 23. Ngf3 Nxe4 24. Nxe4 Nb3+ 25. Ka2 Qd5 26. Bxh6 Nd2+ 27. Ka1 Nb3+
28. Ka2 e5 29. Bg5 Nc1+ 30. Ka1 Nb3+ 31. Ka2 Nd4+ 32. c4 Qxc4+ 33. Ka1 Nb3+ 34.
Ka2 Nd2+ 35. Ka1 Nb3+ 36. Ka2 fxe4 37. Qxe4 Nc1+ 38. Ka1 Nb3+ 39. Ka2 Nd4+ 40.
Ka1 Bc2 41. Qh4 Nb3+ 42. Ka2 Nc1+ 43. Ka1 Nb3+ 44. Ka2 Nd4+ 45. Ka1 Nb3+ 1/2-1/2
```
Expected behavior: +0.00 evaluation, because there is a rather immediate 3-fold repetition that one side can force.
[...]

PR #1513, opened 2021-02-05
Introduce a switch for reverting twofold draw visits

Naphthalin:
When #1375 was merged, we decided to revert all visits to twofold draws if they were made into non terminals to keep the tree visit consistent. However, this introduced two problematic interactions:

1. In rare cases it's possible that the repetition line was the only non-losing line for one side, and reverting all visits to that line specifically will make first move of that line look much worse (or better) than it actually is.
2. In fast time controls, reverting the visits will lead to other root moves having more visits, therefore being selected accidentally. This can by made even more likely by an interaction with 1, and happened in #1466 

This PR introduces a switch whether to revert twofold visits when the twofold draw node is made non terminal; default behaviour is `--revert-two-fold-visits=false` for testing purposes, `=true` is equivalent to the current behaviour.


Naphthalin:
from Discord, thx to trre:

> decided to take a stab at it since it seems like `MakeNotTerminal(terminal_visits)` was doubling the visits the node should have. changed it to `MakeNotTerminal(0)` and kept the `if (!revert_two_fold_visits) break;` in there so nothing is reverted

trre123:
With changes listed above, this test was run at 6s+0.1s

>    # PLAYER           :  RATING  ERROR  POINTS  PLAYED   (%)  CFS(%)    W    D    L  D(%)
>    1 lc0-pr1513(?)    :     2.1   10.6   503.0    1000    50      65  133  740  127    74
>    2 lc0-master       :     0.0   ----   497.0    1000    50     ---  127  740  133    74

PR #1515, opened 2021-02-05
check uci move validity

borg323:
#1514 catches illegal moves in the search, but maybe doing this is also a good idea - the belt and suspenders approach.


borg323:
Shouldn't make a difference - in that issue it is the FEN that describes an "illegal" position.

Naphthalin:
In that case, is there any reason preventing this from being reviewed and merged?

PR #1524, opened 2021-02-13
Add a WLD transform to apply contempt.

Tilps:
Tuned specifically for T70 vs self at 800 nodes vs 1200.

20k games played with both sides contempt set to 0 - elo measured by self play was 75. (Note: this also formed my training data for the tuning of the 75 constants lots of systemic bias to be had here...)

Giving contempt=1.0 to the 1200 node player match is still under way, but very early result after 250 games is 115 elo.
Knowing my luck that will converge back down, but for now its promising.

### DIFF
--- a/src/mcts/params.cc
+++ b/src/mcts/params.cc
@@ -307,6 +307,9 @@ const OptionId SearchParams::kThreadIdlingThresholdId{
     "actively in the process of either sending data to the backend or waiting "
     "for data from the backend, assume that the backend is idle."};
 
+const OptionId SearchParams::kContemptId{
+"contempt", "Contempt", "Fraction of an approximately 75 elo delta WLD corrective transform to apply to WLD."};
+
 void SearchParams::Populate(OptionsParser* options) {
   // Here the uci optimized defaults" are set.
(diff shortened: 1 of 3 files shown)


Tilps:
Some convergence happening - but still 93 elo at almost 7000 games.  High confidence of superiority - causation maybe not estabilished :P

Tilps:
90 elo after 20k games.  I'll try retraining the parameters off of this data, which should make it more aggressive again - but stilll in a way which is consistent with its abilities vs such opponents. I expect that this could be applied iteratively with smaller and smaller gains each time, but I'm interested to see how much I can get from the first iteration.

Tilps:
Note: masterkni666 ran a test with equal node count opponents, and after 10k games was within error bars, slightly favourable to contempt.  Since it was only slightly favourable that suggests that the transformation isn't just better alligned with cpuct at the number of nodes tested.  Which is good news for causation liklihood being that the transform is better against weak opponents.

Naphthalin:
Closing this due to #1791 addressing the issue of this PR.

PR #1525, opened 2021-02-14
Alternative contempt WLD transform

Tilps:
This is trained off of games with contempt already enabled.  Training and test losses were not as good though - so perhaps I have done something wrong with the input data.

Early results are not as promising as early results were for the parameters in #1524 - but still within random walk distance easily. Will update when I finish 20k games.


Tilps:
Final Elo after 20k games 92.  A non-significant improvement over the old parameters.
Probably the gain I should have expected from this iterative improvement was always going to be less than error bars for common opponent at 20k games a pair. So not a terrible result.

Naphthalin:
Closing this due to #1791 addressing the issue of this PR.

PR #1526, opened 2021-02-14
Apply drift correction to q and d values in training data.

Tilps:


Tilps:
untested :P

Tilps:
Tested now - drift in d sometimes higher than I expected so I increased the warning threshold.
I added logging to every adjustment, and drift in d adjustment is by far the most common - usually for high W or high L.  All adjustments look valid given the input q and d values.
I didn't get any q or d range clipping events in the few hundred games I generated. (Not too surprising, D range exceeded usually only happens once in a thousand games or so.) Adjustments aren't all that common in training games - as adjudication usually stops the game before the evals can get extreme enough for a drift correction to be needed. Playing match games generated more events - but I only saw two kinds W or L high, and D high.  D high is rarer.

Tilps:
Based on a hint from borg I discovered that cudnn-fp16 (on my machine at least) performs a bad WDL softmax.  Switching to cuda-fp16 and the drift adjustments shrink down dramatically.  I've readjusted the eps back to 1e-6 instead of 2e-4 on these grounds - and we'll look to fix cudnn backend via #1425.

borg323 (approved):
Let's see if we get any reports of this triggering...

PR #1530, opened 2021-02-16
Ensure allowed piggybank time doesn't push total time limit past move overhead.

Tilps:


src/mcts/stoppers/smooth.cc
```diff
@@ -329,6 +329,11 @@ class SmoothTimeManager : public TimeManager {
       move_allocated_time_ms_ = std::max(static_cast<int64_t>(0LL),
                                          *time - params_.move_overhead_ms());
     }
+    if (allowed_piggybank_time_ms >
+        *time - params_.move_overhead_ms() - move_allocated_time_ms_) {
+      allowed_piggybank_time_ms =
+          *time - params_.move_overhead_ms() - move_allocated_time_ms_;
+    }
```
borg323, line 336: I think we need something similar for `time_to_piggybank_ms`

Tilps: I don't think I can reason what is the 'best' adjustment to do there.  Moving the clamping up onto the brutto variable before time_to_puggybank_ms is calculated almost seems to make sense, but its too much like a basic question of the design of smooth, while I'm just trying to fix the fact that its outputs aren't clamped.  If crem wants to do something specific there he can - if it makes this PR obsolete, that is also fine ;)

mooskagh (approved):

src/mcts/stoppers/smooth.cc
borg323: As a fix this pr seems sufficient, my question was indeed more about the overall design. Any improvements can wait for the next release when we will have more data.

PR #1535, opened 2021-02-23
Policy focus

jhorthos:
Add a field to the V6 training data record that measures how surprising the visits made during search are relative to the policies for a network. This information can be used to weight the training record during training (not yet implemented).

Things that need to be added include 
1) recording the KLD value in the appropriate place in the training record, implementing some offset so that no data (currently 0) and real KLD=0 are distinguishable.
2) changes to training Python code to use the information.


jhorthos:
Haha plus fix my garbage code.

jhorthos:
used an epsilon of +1.0 for recording KLD and modified the V6 record to store the value.

jhorthos:
too much Python, forgot the semi-colon :-)

src/mcts/node.cc
```diff
@@ -555,11 +555,48 @@ V6TrainingData Node::GetV6TrainingData(
...
+  }
+  float total = 0.0;
+  auto it = intermediate.begin();
   for (const auto& child : Edges()) {
-    result.probabilities[child.edge()->GetMove().as_nn_index(transform)] =
-        total_n > 0 ? child.GetN() / static_cast<float>(total_n) : 1;
+    auto nn_idx = child.edge()->GetMove().as_nn_index(transform);
+    float fracv = total_n > 0 ? child.GetN() / static_cast<float>(total_n) : 1;
+    if (fracv > 0 && nneval) {
+      float P = std::exp((*it - max_p) / temp);
```
Tilps, line 586: It isn't clear to me that temp should be applied in calculating the KLD.
The policy is trained to equal the visit distribution of training - without PST.  PST is used in training search, but that doesn't mean we should be measuring surprise of the network output vs its training target using it...

src/mcts/node.cc
```diff
@@ -555,11 +555,48 @@ V6TrainingData Node::GetV6TrainingData(
...
+      kld_sum += fracv * std::log(fracv / P);
+      total += P;
+      it++;
```
Tilps, line 589: 'it' should be advance unconditionally, not only if fracv is positive.

src/mcts/node.cc
borg323: Should we define a separate option for this?

src/mcts/node.cc
borg323: Argh!

src/mcts/node.cc
Tilps: A separate option seems much more sensible than the current option - but not sure if we'd use it....

src/mcts/node.cc
```diff
@@ -555,11 +555,48 @@ V6TrainingData Node::GetV6TrainingData(
...
+  auto it = intermediate.begin();
   for (const auto& child : Edges()) {
-    result.probabilities[child.edge()->GetMove().as_nn_index(transform)] =
-        total_n > 0 ? child.GetN() / static_cast<float>(total_n) : 1;
+    auto nn_idx = child.edge()->GetMove().as_nn_index(transform);
+    float fracv = total_n > 0 ? child.GetN() / static_cast<float>(total_n) : 1;
+    if (fracv > 0 && nneval) {
+      float P = std::exp((*it - max_p) / temp);
+      kld_sum += fracv * std::log(fracv / P);
+      total += P;
```
Tilps, line 588: Hrmm I feel like I need to do some math.  total looks like it serves the purpose of normalizing P values for the softmax temperature.  But if that is its purpose, it should sum over all P, not just the ones which got visits...
(But I can't rule out by eyeballing that the purpose of total is more complicated then that...)

src/mcts/node.cc
Naphthalin: In my opinion, the surprise should always be measured against the raw policies (PST=1.0) because that's what it is supposed to predict/be trained on. The purpose of PST in training is to counteract the sharpening of PUCT in RL in a specific way, giving a source of regularization.

src/mcts/node.cc
Naphthalin: At least for the standard formula used for KLD, terms with "real data" = 0 will contribute 0 to the KLD. Line 587 would need a `fracv == 0 ? 0 : fracv * log(fracv/p)` otherwise, but it produces the theoretically correct result this way as well.

Tilps: total gets subtracted off below as a log.  Given that currently total only contains P contributions where there are a visit - I'm still confused about whether the math works.

Tilps: err not subtracted - added - line 596. (I was thinking subtraction since for PST 1.0 it'll be negative.)

Naphthalin: You are absolutely right. P has to be added for unvisited moves as well, otherwise the sum=1 condition won't be met and the entropy will be calculated wrong.

borg323: I think I fixed it this time...

Tilps: think so :)

Tilps (approved):

src/mcts/node.cc
```diff
@@ -526,8 +526,8 @@ V6TrainingData Node::GetV6TrainingData(
     GameResult game_result, const PositionHistory& history,
     FillEmptyHistory fill_empty_history,
     pblczero::NetworkFormat::InputFormat input_format, Eval best_eval,
-    Eval played_eval, Eval orig_eval, bool best_is_proven, Move best_move,
-    Move played_move) const {
+    Eval played_eval, bool best_is_proven, Move best_move, Move played_move,
```
Naphthalin, line 529: Just to make sure I don't miss things: To me it looks as if this overwrites `orig_eval` which is needed for value focus; or am I missing something crucial here?

Naphthalin (changes requested):

src/mcts/node.cc
Tilps: orig_eval was calculated outside this function, its now calculated inside this function.

Naphthalin: Thanks, this clarifies things.

Naphthalin (approved):

src/mcts/node.cc
borg323: I can remove the temperature completely - I only added it when converting the original pr to get the policy from the cache, it was not there originally. If we have a need for it in the future it is easy to add back.

PR #1536, opened 2021-02-24
Fix log spam during infinite analysis

### LINKED ISSUE #1507: "Stopped search: Reached visits limit" floods logfile
_posted by user Tiranasta on Discord 2021-01-27 18:32_

> When using a log file and running with a RAM limit, if the node limit is reached in infinite analysis mode Leela starts outputting the "Stopped search: Reached visits limit" message to the log file over and over again at such a rapid rate that the log file can quickly reach tens of gigabytes in size before the user even notices there's a problem. In positions with a forced mate, the limit can be reached within moments of starting a fresh search.

[didn't try to reproduce it, just reporting]

DU-jdto:
Fixes "Stopped search: Reached visits limit" log file spam when the visits limit is reached in analysis mode.

In infinite analysis mode, once stop_ is true but "stop" hasn't yet been received from the GUI, there is no need to spam FireStopInternal() (since the watchdog thread wait_for is non-blocking at that point anyway), and none of the ShouldStop methods for stoppers that can be present at such times appear to have important side effects. This fix should therefore be safe.

Fixes #1507

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -531,7 +531,7 @@ void Search::MaybeTriggerStop(const IterationStats& stats,
   // Don't stop when the root node is not yet expanded.
   if (total_playouts_ + initial_visits_ == 0) return;
 
-  if (!stop_.load(std::memory_order_acquire) || !ok_to_respond_bestmove_) {
+  if (!stop_.load(std::memory_order_acquire)) {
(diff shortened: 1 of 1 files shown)


borg323 (approved):
Seems this check was no longer required after the stoppers refactoring.

PR #1537, opened 2021-02-27
Add n_in_flight limiting for small trees.

Tilps:
Used in combination with increasing default out of order limit and max collision limits aims to improve elo at match play without poor consequences at small fixed nodes.

Only implemented for multigather=true here, I kind of expect that we'll complete the multigather landing before this is ready.

Values for the new parameters come from a tune at fairly short TC (1"+0.1") with 2080TI using 703810. There are probably fairly broad options available that are similar. The 'max' is actually a no-op in the tuning scenario since its > threads*(minibatch_size+max_collision_visits).  The tune pretty much equally preferred all values in the deadspace above that limit, but value is selected for its pairing with the end parameter to create the slope.

Haven't started validation tests yet.

### DIFF
--- a/src/mcts/params.cc
+++ b/src/mcts/params.cc
@@ -306,6 +306,15 @@ const OptionId SearchParams::kThreadIdlingThresholdId{
     "If there are more than this number of search threads that are not "
     "actively in the process of either sending data to the backend or waiting "
     "for data from the backend, assume that the backend is idle."};
+const OptionId SearchParams::kBatchLimitStartId{
+    "batch-limit-start", "BatchLimitStart",
+    "Tree size where max n_in_flight allowed limit starts scaling up from 0."};
+const OptionId SearchParams::kBatchLimitEndId{
+    "batch-limit-end", "BatchLimitEnd",
+    "Tree size where n_in_flight allowed limit reaches max."};
+const OptionId SearchParams::kBatchLimitMaxId{
+    "batch-limit-max", "BatchLimitMax",
+    "Maximum allowed n_in_flight limit regardless of tree size."};
 
 void SearchParams::Populate(OptionsParser* options) {
   // Here the uci optimized defaults" are set.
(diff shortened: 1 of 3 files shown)


Tilps:
I'll note that this has training interactions - but they should strictly be an increase in early outs - maybe trading some need for increased parallelism to get maximal performance vs search quality.

Tilps:
Seems 'without poor consequences' might be overly hopeful.  Still better with than without.
Score of lc0 dev vs lc0: 214 - 554 - 1232 [0.415]
...      lc0 dev playing White: 148 - 186 - 666  [0.481] 1000
...      lc0 dev playing Black: 66 - 368 - 566  [0.349] 1000
...      White vs Black: 516 - 252 - 1232  [0.566] 2000
Elo difference: -59.6 +/- 9.3, LOS: 0.0 %, DrawRatio: 61.6 %
2000 of 2000 games finished.

Tilps:
Added a power curve.  Tuning the power curve got me two competing minima that were very different.  I explicitly selected the one which was going to be favourable to fixed nodes - simply changing the value of power from 1.0 to 2.5 without changing anything else... Fixed node regression is now quite a bit better.
Score of lc0 dev vs lc0: 312 - 432 - 1256 [0.470]
...      lc0 dev playing White: 229 - 138 - 633  [0.545] 1000
...      lc0 dev playing Black: 83 - 294 - 623  [0.395] 1000
...      White vs Black: 523 - 221 - 1256  [0.576] 2000
Elo difference: -20.9 +/- 9.3, LOS: 0.0 %, DrawRatio: 62.8 %
2000 of 2000 games finished.

While working on this I've had a few other ideas I want to try - but possibly this is getting close to acceptable... Although I do need to see whether how its low node nps characteristics are (Certainly given the 1.9k node tree size threshold this limit will have no practical influence in competition, meaning its effectively convergent on current competition submission parameters.)

Tilps:
At both time control and fixed nodes #1540 seems basically the same (if not better at TC) in terms of perf and the implementation is simpler. So I'm going to focus on that one.

PR #1539, opened 2021-02-28
implement button uci options

borg323:
Works by piggybacking on the functionality added in #850 to verify backend opts syntax, and adds a flag that is cleared when an option is set and set when read. This flag is used in `GetOnce()` to return true only once after the button is pressed.

~~There is no actual use for button options at the moment, it~~This is intended to be combined with #1455 to allow users to signal the search to clear old tree branches and free memory without destroying the current tree.

*Edit:* Adds a button to clear the tree before the next search.


src/utils/optionsdict.cc
```diff
@@ -77,6 +77,24 @@ bool OptionsDict::HasSubdict(const std::string& name) const {
   return subdicts_.find(name) != subdicts_.end();
 }
 
+template <>
+bool OptionsDict::GetOnce(const std::string& key) const {
+  for (const auto* alias : aliases_) {
+    const auto& dict = alias->TypeDict<bool>::dict();
+    auto iter = dict.find(key);
+    if (iter != dict.end()) {
+      if (iter->second.IsSet()) return false;
```
mooskagh, line 86: I'm not sure I understand how it works.  Where does the value reset to false after reading it the first time?

borg323: IIRC (it was some time ago) `IsSet()` return value depends on whether the last action was a `Set()` or a `Get()`. Any suggestions for a better implementation are appreciated.

Naphthalin:
While this is flying around for quite some time as a proof of concept, I believe there is a pretty strong use case for it now, namely for analysis with Contempt, and would love if it was included in an upcoming release.

Quick explanation of what I have in mind with it:
- main idea is to analyze the same position to the same node counts, but with different Contempt values
- changing Contempt with an existing tree causes incoherent behavior as the WDL of new nodes will be different, and existing nodes will keep their WDL
- clearing the tree without losing the NN cache solves this problem: we let search rebuild the tree but with WDL following the new distribution, which is a lot easier than the hypothetical solution of updating the tree manually

src/utils/optionsdict.h
```diff
@@ -100,7 +101,20 @@ class OptionId {
   const char short_flag_;
 };
 
+class Button {
+ public:
+  Button() { val = std::make_shared<bool>(false); }
+  Button(bool x) { val = std::make_shared<bool>(x); }
+  operator bool() { return *val; }
```
mooskagh, line 108: I'd do en explicit function `was_clicked()` or `test()` rather than `operator bool`.

Also, to save some lines, maybe `bool test_and_reset()` function.

mooskagh (approved):

mooskagh (approved):

Naphthalin:
I tested what I have in mind with the above sketched functionality from the command line, and it works like a charm. 

From Nibbler, I simply have to add a file `Contempt100` into the `scripts` folder with the following three lines
```
stop
setoption name Contempt value 100
setoption name ClearTree
```
(the `stop` command is sent automatically when using a script so the line could be left out as well, but is required for a custom solution) to change Contempt to 100, and then resume the search without having to reload the `engines.json` and engine. This makes experimenting with different Contempt values for analyzing an opening so much smoother, thanks a lot!

Maybe there even is a way to establish a "Change Contempt and resume analysis" feature directly in Nibbler using the button implementation, will talk to @rooklift about that.

Naphthalin (approved):

PR #1540, opened 2021-02-28
Limited mcv

Tilps:
Similar 1k nodes regression to limiting n_in_flight - after partial tuning I got the following result.
Score of lc0 vs lc0 dev: 438 - 303 - 1259 [0.534]
...      lc0 playing White: 273 - 103 - 624  [0.585] 1000
...      lc0 playing Black: 165 - 200 - 635  [0.482] 1000
...      White vs Black: 473 - 268 - 1259  [0.551] 2000
Elo difference: 23.5 +/- 9.2, LOS: 100.0 %, DrawRatio: 62.9 %
2000 of 2000 games finished.

Bonus here is that max collision visits preferred curvature is closer to straight line which potentially implies we can scale MCV=tree_size/4 to an almost arbitrary level.  Since this increases the limit over 1000 for even moderately sized trees - this potentially has better nps characteristics than our competition version. Also since this is a collision limit and not an n_in_flight limit, earlier batches should be a tiny bit faster.
Will run some time control tests of this and the other option, see if there is an obvious preference on that side.  This option is probably the default otherwise.

### DIFF
--- a/src/mcts/params.cc
+++ b/src/mcts/params.cc
@@ -306,6 +306,15 @@ const OptionId SearchParams::kThreadIdlingThresholdId{
     "If there are more than this number of search threads that are not "
     "actively in the process of either sending data to the backend or waiting "
     "for data from the backend, assume that the backend is idle."};
+const OptionId SearchParams::kBatchLimitStartId{
+    "batch-limit-start", "BatchLimitStart",
+    "Tree size where max n_in_flight allowed limit starts scaling up from 0."};
+const OptionId SearchParams::kBatchLimitEndId{
+    "batch-limit-end", "BatchLimitEnd",
+    "Tree size where n_in_flight allowed limit reaches max."};
+const OptionId SearchParams::kBatchLimitMaxId{
+    "batch-limit-max", "BatchLimitMax",
+    "Maximum allowed n_in_flight limit regardless of tree size."};
 
 void SearchParams::Populate(OptionsParser* options) {
   // Here the uci optimized defaults" are set.
@@ -340,10 +349,10 @@ void SearchParams::Populate(OptionsParser* options) {
   options->Add<FloatOption>(kFpuValueAtRootId, -100.0f, 100.0f) = 1.0f;
   options->Add<IntOption>(kCacheHistoryLengthId, 0, 7) = 0;
   options->Add<FloatOption>(kPolicySoftmaxTempId, 0.1f, 10.0f) = 1.359f;
-  options->Add<IntOption>(kMaxCollisionEventsId, 1, 65536) = 32;
-  options->Add<IntOption>(kMaxCollisionVisitsId, 1, 1000000) = 9999;
+  options->Add<IntOption>(kMaxCollisionEventsId, 1, 65536) = 917;
+  options->Add<IntOption>(kMaxCollisionVisitsId, 1, 1000000) = 1000;
   options->Add<BoolOption>(kOutOfOrderEvalId) = true;
-  options->Add<FloatOption>(kMaxOutOfOrderEvalsId, 0.0f, 100.0f) = 1.0f;
+  options->Add<FloatOption>(kMaxOutOfOrderEvalsId, 0.0f, 100.0f) = 2.4f;
   options->Add<BoolOption>(kStickyEndgamesId) = true;
   options->Add<BoolOption>(kSyzygyFastPlayId) = false;
   options->Add<IntOption>(kMultiPvId, 1, 500) = 1;
@@ -381,6 +390,9 @@ void SearchParams::Populate(OptionsParser* options) {
   options->Add<IntOption>(kMinimumWorkPerTaskForProcessingId, 1, 100000) = 8;
   options->Add<IntOption>(kIdlingMinimumWorkId, 0, 10000) = 0;
   options->Add<IntOption>(kThreadIdlingThresholdId, 0, 128) = 1;
+  options->Add<IntOption>(kBatchLimitStartId, 1, 10000) = 50;
+  options->Add<IntOption>(kBatchLimitEndId, 1, 10000) = 1893;
+  options->Add<IntOption>(kBatchLimitMaxId, 1, 10000) = 2622;
 
   options->HideOption(kNoiseEpsilonId);
   options->HideOption(kNoiseAlphaId);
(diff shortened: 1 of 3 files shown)


Tilps:
Some quick TC testing with 703810 on 2080TI vs SF13 single thread on my machine. All tests with multi-gather.
lc0 dev is this PR. lc0 is equivalent to current master (with MCV 32 to match MCE). lc0 aggressive is equivalent to current master but with 917/1000 MCE/MCV.
All tests aimed to stop at 1k nodes as full 2k was going to mean I couldn't fit a 4th test into my day. The last two ran over because I was not paying close attention.
Score of lc0 dev vs stockfish_13_win_x64_bmi2: 203 - 185 - 612 [0.509]
...      lc0 dev playing White: 139 - 67 - 294  [0.572] 500
...      lc0 dev playing Black: 64 - 118 - 318  [0.446] 500
...      White vs Black: 257 - 131 - 612  [0.563] 1000
Elo difference: 6.3 +/- 13.4, LOS: 82.0 %, DrawRatio: 61.2 %
1000 of 2000 games finished.

Score of lc0 vs stockfish_13_win_x64_bmi2: 176 - 263 - 609 [0.458]
...      lc0 playing White: 121 - 92 - 311  [0.528] 524
...      lc0 playing Black: 55 - 171 - 298  [0.389] 524
...      White vs Black: 292 - 147 - 609  [0.569] 1048
Elo difference: -28.9 +/- 13.6, LOS: 0.0 %, DrawRatio: 58.1 %
1048 of 2000 games finished.

Score of lc0 aggressive vs stockfish_13_win_x64_bmi2: 183 - 265 - 638 [0.462]
...      lc0 aggressive playing White: 127 - 86 - 330  [0.538] 543
...      lc0 aggressive playing Black: 56 - 179 - 308  [0.387] 543
...      White vs Black: 306 - 142 - 638  [0.576] 1086
Elo difference: -26.3 +/- 13.2, LOS: 0.0 %, DrawRatio: 58.7 %
1086 of 2000 games finished.

Tilps:
Fixed nodes test with a simple variant where without any additional tuning we take minimum of current formula with tree size as input, and current formula with estimated playouts remaining as input.  If my calculations are correct in the middle of search it should be regularly exceeding the prior default max collision visits (32) - and the average max collision visits limit should also be higher than default.  I haven't tested nps yet.  Despite the limit being higher most of the time, fixed nodes test came in as a win.

Score of lc0 dev vs lc0: 384 - 327 - 1289 [0.514]
...      lc0 dev playing White: 244 - 121 - 635  [0.561] 1000
...      lc0 dev playing Black: 140 - 206 - 654  [0.467] 1000
...      White vs Black: 450 - 261 - 1289  [0.547] 2000
Elo difference: 9.9 +/- 9.1, LOS: 98.4 %, DrawRatio: 64.5 %
2000 of 2000 games finished.

Tilps:
Hmm starting TC test with the new variant the nps is 'terrible' - something seems wrong :P

Tilps:
Found and fixed the bug, would not have affected the fixed node test above. - TC test running now - but for technical reasons I doubt it'll be much different to the first run above.  Our estimated playouts remaining is only updated when under time control after 200ms - and the time control I'm playing at means that almost never happens. (This also means smart pruning is almost never happening...)

Tilps:
Score of stockfish_13_win_x64_bmi2 vs lc0 dev: 401 - 389 - 1210 [0.503]
...      stockfish_13_win_x64_bmi2 playing White: 266 - 117 - 617  [0.575] 1000
...      stockfish_13_win_x64_bmi2 playing Black: 135 - 272 - 593  [0.431] 1000
...      White vs Black: 538 - 252 - 1210  [0.572] 2000
Elo difference: 2.1 +/- 9.6, LOS: 66.5 %, DrawRatio: 60.5 %
2000 of 2000 games finished.
Not quite as amazing as before but still combined with above this is like 24 elo +/- 16

Tilps:
Did an nps check for 1k nodes on bench.  There is a regression - its less than 10%.  Given the test above indicates maybe 10 elo stronger, I think a 10% regression in speed at fixed 1k nodes is fine. So I'm going to mark this ready for review.

Tilps:
Note that its possible to disable scaling (and have behavior like before) by setting MaxCollisionVisitsScalingEnd to 0. So anyone who wants the 10% nps back at fixed nodes (and lose the 10 elo) can just set MaxCollisionVisits to 32 and ScalingEnd to 0.

My next step is to see if there is a better tune for longer time controls than 1"+0.1"

brianprichardson:
I have looked in the log file and cannot find the the multi-gather status.

Tilps:
multi-gather isn't logged especially - its like most other command line flags in that respect...

src/mcts/search.cc
```diff
@@ -1216,11 +1216,41 @@ void SearchWorker::GatherMinibatch() {
     if (search_->stop_.load(std::memory_order_acquire)) return;
   }
 }
+
+namespace {
+int Mix(int high, int low, float ratio) {
```
mooskagh, line 1221: fyi there will be std::lerp soon: <https://en.cppreference.com/w/cpp/numeric/lerp>
But that's not exactly the same as here also rounding comes.

Also `low + ratio * (high-low)` may be more more readable (and a tiny bit faster because of one multiplication instead of two).

src/mcts/search.cc
```diff
@@ -1216,11 +1216,41 @@ void SearchWorker::GatherMinibatch() {
...
+  return static_cast<int>(std::round(high * ratio + low * (1.0f - ratio)));
+}
+int CalculateCollisionsLeft(int64_t nodes, const SearchParams& params) {
+  // End checked first
+  if (nodes >= params.GetMaxCollisionVisitsScalingEnd()) {
+    return params.GetMaxCollisionVisitsId();
```
mooskagh, line 1227: I think `GetMaxCollisionVisitsId()` should not have `Id` in its name. :)

src/mcts/params.cc
```diff
@@ -340,8 +350,12 @@ void SearchParams::Populate(OptionsParser* options) {
   options->Add<FloatOption>(kFpuValueAtRootId, -100.0f, 100.0f) = 1.0f;
   options->Add<IntOption>(kCacheHistoryLengthId, 0, 7) = 0;
   options->Add<FloatOption>(kPolicySoftmaxTempId, 0.1f, 10.0f) = 1.359f;
-  options->Add<IntOption>(kMaxCollisionEventsId, 1, 65536) = 32;
-  options->Add<IntOption>(kMaxCollisionVisitsId, 1, 1000000) = 32;
+  options->Add<IntOption>(kMaxCollisionEventsId, 1, 65536) = 917;
+  options->Add<IntOption>(kMaxCollisionVisitsId, 1, 100000000) = 80000;
+  options->Add<IntOption>(kMaxCollisionVisitsScalingStartId, 1, 100000) = 28;
+  options->Add<IntOption>(kMaxCollisionVisitsScalingEndId, 0, 100000000) = 145000;
```
mooskagh, line 356: Why can end be 0, but start cannot?
Also maybe add check that `start < end` after they are parsed in the constructor.

mooskagh (approved):

src/mcts/search.cc
Tilps: Yeah - pre-existing condition though :P

src/mcts/params.cc
Tilps: I deliberately allow end <= start to disable scaling entirely. In case anyone wants to go back to the old behavior. Passing an end value of 0 makes that super clear.

src/mcts/search.cc
Tilps: I'll make the change, on simplification grounds rather than readable grounds.

src/mcts/search.cc
Tilps: Fixed.

PR #1543, opened 2021-03-08
Analyse Mode + minimax/mcts hybrid

Naphthalin:
This is a partial rewrite of my betaMCTS PR #963, basically stripping off the parts which increase memory consumption (saving a parallel Q value, saving the relevance values instead of recalculating them) and using a slightly different formula. It also includes the AnalyseMode presented in #1455 and a LCB move ordering. **In total, this is supposed to be a lightweight Lc0 analysis implementation.**

To my best knowledge, the complete list of features is
- `UCI_AnalyseMode` supporting a forward/backward analysis style without losing the tree. **Unlike other engines, the Lc0 search tree allows instant access to sidelines etc without rebuilding them from cache.**
- the option to free memory by removing the lines not starting at the current root position, see (1) below
- recalculation of node stats with a `exp(Q / sharpness)` weight (see (2)), default update frequency 100. Necessary for `AnalyseMode`, deactived by `UpdateInterval: 0` to remove the small computation overhead.
- Lower Confidence Bound (LCB) move ordering both for VerboseMoveStats and for GetBestChildren; can be configured, and turned off by setting `LCBPercentile: 0`, see (3)
- Current defaults already use sensible parameter choices.
- All introduced features can be turned off completely by `--update-interval=0 --lcb-percentile=0.0 --no-analyse-mode` and should then differ from `lc0_master` only in a few trivial bool checks for params.

(1) Create a textfile [FreeMemory.txt](https://github.com/LeelaChessZero/lc0/files/6104593/FreeMemory.txt)
```
setoption name FreeMemory value true
go nodes 1
setoption name FreeMemory value false
```
and put it into the `Nibbler --> Engine --> Custom Scripts` folder. Monitor RAM on your own; `FreeMemory` completion can take up to 30s at least under Windows, but you can restart the analysis instantly without waiting. Needs the `Dev --> Allow arbitrary scripts` check introduced in Nibbler version 1.6.5, thanks @fohristiwhirl :)

(2) Sharpness is controlled by `RecalculateTemperature`, though `RecalculationSharpness` might be a better name. Suggestions welcome. Setting it to 0 will do plain averaging, but ignore the original eval of a node as it isn't saved anywhere in the tree except the eval itself.

(3) A log logistic distribution is assumed to approximate the unknown distribution of Q values at N nodes. The percentile is functionally equivalent to a sharpness parameters; 0.5 will produce Q orderung, values above 0.5 are allowed but will produce UCB values. Needs the Nibbler option `Dev --> Use VerboseMoveStats ordering` introduced in Nibbler 1.5.4, thanks again @fohristiwhirl


src/mcts/node.cc
```diff
@@ -252,8 +253,10 @@ Node* Node::CreateSingleChildNode(Move move) {
 void Node::CreateEdges(const MoveList& moves) {
   assert(!edges_);
```
Naphthalin, line 254: this assert probably needs to be removed, as it serves no purpose anymore.

src/mcts/search.cc
```diff
@@ -384,11 +385,14 @@ std::vector<std::string> Search::GetVerboseStats(Node* node) const {
   for (const auto& edge : node->Edges()) edges.push_back(edge);
 
   std::sort(edges.begin(), edges.end(),
-            [&fpu, &U_coeff, &draw_score](EdgeAndNode a, EdgeAndNode b) {
+            [&fpu, &U_coeff, &draw_score, &lcb_percentile](EdgeAndNode a,
```
Naphthalin, line 388: A couple of overfull lines, but no idea what the correct indentation is supposed to be.

src/mcts/search.cc
```diff
@@ -2431,6 +2441,12 @@ void SearchWorker::DoBackupUpdateSingleNode(
       m = n->GetM();
     }
     n->FinalizeScoreUpdate(v, d, m, node_to_process.multivisit);
+    if (params_.GetUpdateInterval() > 0 &&
+        (n->GetNStarted() + 1) % params_.GetUpdateInterval() == 0) {
```
Naphthalin, line 2445: the +1 here is to make sure that it takes a full interval until the first update.

src/mcts/search.cc
```diff
@@ -2022,7 +2032,7 @@ SearchWorker::NodeToProcess SearchWorker::PickNodeToExtend(
     EnsureNodeTwoFoldCorrectForDepth(node, depth - 1);
 
     // If terminal, we reached the end of this playout.
-    if (node->IsTerminal()) {
+    if (!is_root_node && node->IsTerminal()) {
```
Naphthalin, line 2035: This is needed because root can be a terminal now, and we don't want to make it non terminal. Might need to doublecheck the behaviour of navigating through proven mate lines.

fohristiwhirl:
As borg pointed out to me, UCI has a "button" option type, which seems more correct for this sort of "press to make some state change" feature.

borg323:
> UCI has a "button" option type, which seems more correct for this sort of "press to make some state change" feature.

I made #1539 for this

Naphthalin:
From the technical side, the PR consists of 3 separate changes.

1. `AnalyseMode`: This is implemented in `13e1814`, plus a one-line change in `4d2fe3b`
2. `Recalculation of node stats`: This is implemented in `89305a7..6143cb7` plus a one-line change in `6143cb7`
3. `LCB move ordering`: This is implemented in `fca5ffa..ca0d680`

Each of the 3 would function without the other two, but with questionable use.

Comments on design choices and technical aspects to keep in mind:
1. `FreeMemory` currently is a bit too convoluted, ultimately it should be implemented as a single command without the use of a Nibbler script to work in all UCI compliant GUIs. The `RAM limit MB` option doesn't work as it roughly estimates the tree size from the number of nodes at current root. Any automatism would likely need to monitor the actual RAM usage in the OS which is much harder and OS specific.
2. By design, the weights for a child needs to be lightweight, which means it has to be calculated from child and possibly parent node stats alone. The softmax over (possibly transformated, e.g. logit) child Q values with a temperature either constant or decreasing with parent N is the only possible choice which fulfills all necessary conditions; currently constant and configurable, could be replaced by some parent N scaling to actually converge to minimax after longer search. Note that the incremental updates don't reflect the exponential weighting as that would require a softmax calculation over all children. The only alternative would be to store the `relevance` as it is done in betaMCTS, but this increases `Node` size.
3. In general, a LCB implementation would have more than one parameter: The parameter(s) defining the shape and sharpness of the estimated distribution from Q and N, and the LCB percentile (e.g. 2 std devs). The distribution chosen here is a `tanh` (logit reshaped to `(-1, +1)`) transformed logistic distribution approximating a beta distribution also used in my betaMCTS approach, and possesses the curious property that for each pair `(percentile, sharpness)` there exists a `percentile'` for a fixed `sharpness'` which produces the same LCB value, so we only have 1 degree of freedom, and don't need a configurable sharpness of the distribution on top of the `LCBPercentile` parameter.

Naphthalin:
Benchmarks with 703810 on i7 4770, RX Vega 56, Win10 with dx12 build:
v0.27.0 is release versions, all others are AppVeyor.
```
lc0_v0.27.0
Total time (ms) : 340304
Nodes searched  : 15763250
Nodes/second    : 46321
```

```
lc0_master (PR1538 merge)
Total time (ms) : 340327
Nodes searched  : 16324392
Nodes/second    : 47967
```

```
lc0_pr1543 --lcb-percentile=0.0 --update-interval=0
Total time (ms) : 340357
Nodes searched  : 18437602
Nodes/second    : 54171
```

```
lc0_pr1543 --lcb-percentile=0.2 --update-interval=0
Total time (ms) : 340311
Nodes searched  : 19155510
Nodes/second    : 56288
```

PR #1544, opened 2021-03-11
blas backend softmax hardcoded

OfekShochat:


Corniel:
What is the gain here? (The implementation looks valid to me).

OfekShochat:
It is valid indeed. Firstly its cleaner imo and more fun to look at. If someone could test it, it would be amazing

PR #1545, opened 2021-03-15
gpu transpose kernel

OfekShochat:
create gpu transpose kernel


OfekShochat:
cudnn doesnt compile common kernels, which I put my kernel in. and it should not use it anyways. fixing

PR #1546, opened 2021-03-15
Backendbench assistant with TC-dependent output.

zz4032:
My batchsize test matches have shown a strong dependency of optimal Clippy threshold to average nodes/move (results on Discord: https://discord.com/channels/425419482568196106/530486338236055583/821040507077394483). Now the assistant shows 3 proposals for best batchsize for the most common use cases.
Example:
```  __
 /  \
 |  |    __________________________________________
 +  +   |                                          |
(@)(@) _| Recommended minibatch-size for this net: |
 |  |  \        1s/move   (Bullet):       68       |
 || |/  |       15s/move  (Rapid):       133       |
 || ||  |       3min/move (Tournament):  272       |
 |\_/|  |__________________________________________|
 \___/
```

NPS reported by backendbench are used for the calculation of expected nodes/move in games and 3 different thresholds are calculated to determine each of the 3 batchsizes. Threshold formula is also adjusted to account for the fact that real game NPS are higher than backendbench NPS.

Multi-GPU scenario is not considered here (even higher NPS in games compared to backendbench NPS on 1GPU), I'm not sure if this is necessary.


zz4032:
backendbench3
Function lacks data at higher nodes/move. I analyzed a few positions and defined one more data point at 200Knodes/move that I think leads to better batchsize selection. The PR status can be changed to "Draft", I'd like to run more position tests.

zz4032:
Screenshot from 2021-04-11 21-43-42
Position tests were inconclusive, instead I added another match result.
Trendline is now exponential and looks like the better choice. I added `+0.02` to make it converge at 0.02 instead of 0.0: `threshold = 0.16947 * exp(-4.1695e-6 * nodes/move) + 0.02`. I think the batchsize selection looks good now.

borg323 (approved):

PR #1547, opened 2021-03-18
Return lost check for whether root node is DTZ before doing WDL lookups in search.

Tilps:
I didn't merge the original PR including this change before doing the multigather branches. Multigather branches introduced a new version of this function in parallel.  Then following cleanup brought everything to use the new version, meaning the old version with the correct code had been lost.


borg323 (approved):
This was in #1470 but seems got lost in a merge.

PR #1550, opened 2021-03-22
Make multi-gather default.

Tilps:
Testing in training looks 'fine' - so I think its time to make it the default.

Forcing the task workers default to 0 for selfplay since we already have parallelism to chew up cpus.


Tilps:
Max collision visits adjustment will be temporary - once this is in I expect 1540 to land quickly to change it.

borg323 (approved):

PR #1553, opened 2021-03-29
A first step towards value repair

hans-ekbrand:
Implement value repair by triggering temp=0 if a large enough discrepancy between 1 node eval and best eval after search.

Currently, this patch hardcodes a threshold for surprise. If the absolute value of `best_eval.wl - orig_eval.wl` is greater than this threshold, temp=0 will be applied for the rest of the game.

Hopefully, this behaviour will in it self be good in that it should correct bad evaluations faster than if these positions were played out with temperature and deblunder, but the mechanism can also be combined with an opening book of positions that we know a priori will trigger this behaviour, in more concerted efforts at value repair.

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -539,7 +539,7 @@ void Search::MaybeTriggerStop(const IterationStats& stats,
   if (stop_.load(std::memory_order_acquire) && ok_to_respond_bestmove_ &&
       !bestmove_is_sent_) {
     SendUciInfo();
-    EnsureBestMoveKnown();
+    EnsureBestMoveKnown(false);
     SendMovesStats();
     BestMoveInfo info(final_bestmove_, final_pondermove_);
     uci_responder_->OutputBestMove(&info);
@@ -571,10 +571,10 @@ Eval Search::GetBestEval(Move* move, bool* is_terminal) const {
           best_edge.GetM(parent_m - 1) + 1};
 }
 
-std::pair<Move, Move> Search::GetBestMove() {
+std::pair<Move, Move> Search::GetBestMove(const bool force_temp_to_zero) {
   SharedMutex::Lock lock(nodes_mutex_);
   Mutex::Lock counters_lock(counters_mutex_);
-  EnsureBestMoveKnown();
+  EnsureBestMoveKnown(force_temp_to_zero);
   return {final_bestmove_, final_pondermove_};
 }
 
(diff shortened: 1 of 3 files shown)


src/mcts/search.cc
```diff
@@ -571,10 +571,10 @@ Eval Search::GetBestEval(Move* move, bool* is_terminal) const {
           best_edge.GetM(parent_m - 1) + 1};
 }
 
-std::pair<Move, Move> Search::GetBestMove() {
+std::pair<Move, Move> Search::GetBestMove(const bool force_temp_to_zero) {
```
borg323, line 574: I think `ignore_temp` is a better name.

src/mcts/search.cc
```diff
@@ -588,12 +588,12 @@ void Search::ResetBestMove() {
...
   bool old_sent = bestmove_is_sent_;
   bestmove_is_sent_ = false;
-  EnsureBestMoveKnown();
+  EnsureBestMoveKnown(false);
   bestmove_is_sent_ = old_sent;
 }
 
 // Computes the best move, maybe with temperature (according to the settings).
-void Search::EnsureBestMoveKnown() REQUIRES(nodes_mutex_)
+void Search::EnsureBestMoveKnown(const bool force_temp_to_zero) REQUIRES(nodes_mutex_)
```
borg323, line 596: Same here

src/selfplay/game.cc
```diff
@@ -111,6 +116,10 @@ void SelfPlayGame::Play(int white_threads, int black_threads, bool training,
       syzygy_tb_ = nullptr;
     }
   }
+
+  // Value repair. This is supposed local per game.
+  bool ignore_temp_because_of_value_repair=false;
```
borg323, line 121: Also `ignore_temp` seems enough here.

src/selfplay/game.cc
```diff
@@ -276,6 +285,24 @@ void SelfPlayGame::Play(int white_threads, int black_threads, bool training,
...
+      // check if position should trigger temp=0.
+
+      // This check depends on orig_eval.wl which is not available earlier,
+      // but it this means that the move selection (which is earlier)
+      // can have temperature. To fix that, move the whole block to
+      // come before move selection, But I didn't know the code well
+      // enough to judge if that would backfire.
+
+      if (!ignore_temp_because_of_value_repair) {
+	float surprise=abs(best_eval.wl - orig_eval.wl);
```
borg323, line 301: You need `std::fabs()` here.

src/mcts/search.cc
hans-ekbrand: Thanks for reviewing, I've adapted to your suggestions.

jhorthos:
It should also interact well with training-side value focus, since this won't change the representation of the positions in training, but it will improve the accuracy of game outcome.

masterkni666:
Do we always want temp to be set to 0 when it passes the threshold or do we only want a certain amount. Something that is adjustable at during training similar to value focus.

PR #1554, opened 2021-04-03
default promotion to knight was setting en-passant

borg323:


Tilps (approved):

Tilps:
It took me a while to see the problem.  I edited the PR title to make it a bit clearer.  Its not that enpassant is being left set from before - it is that enpassant was becoming set because the pawn was actually 'not promoted at all' and left as a pawn.  But because the logic for en-passant uses the back rank pawn info - we got the knight promotion 'by accident' because back-rank pawns were ignored.

PR #1556, opened 2021-04-29
cudnn switch to cuda for attention body nets

borg323:


src/neural/cuda/network_cudnn.cc
```diff
@@ -1078,6 +1078,17 @@ std::unique_ptr<Network> MakeCudnnNetwork(const std::optional<WeightsFile>& w,
...
+    // Check if we must switch to cuda.
+    if (weights.format().network_format().network() ==
+        pblczero::NetworkFormat::NETWORK_ATTENTIONBODY_WITH_HEADFORMAT) {
+      CERR << "Nework format not supported directly, switching backend.";
+      return NetworkFactory::Get()->Create(
+          "cuda" +
+              std::string(std::is_same<half, DataType>::value ? "-fp16" : ""),
+          weights, options);
+    }
+#endif
```
mooskagh, line 1091: Maybe add #else with the exception message that adds more context? ("Please download/use CUDA version/backend instead")

borg323: I decided to simplify things, now the cuda backend will always be available when cudnn is built.

mooskagh: Is it guaranteed that when libraries needed for the cudnn backend are present on the system, libraries for cuda are also there? (I guess so, likely cuda is subset of cudnn, just confirming)

If not, it makes sense to allow building just cudnn backend without cuda.

borg323: Yes, both backends use the same cuda libraries (cudart and cublas).
There is one thing I forgot to mention, if the cudnn lib is not found you will still get the cuda backend even if `plain_cuda=false`. I did it like that to simplify the logic in `meson.build` but I'm having second thoughts.

src/neural/cuda/network_cudnn.cc
```diff
@@ -1078,6 +1078,15 @@ std::unique_ptr<Network> MakeCudnnNetwork(const std::optional<WeightsFile>& w,
           pblczero::NetworkFormat::NETWORK_CLASSICAL_WITH_HEADFORMAT &&
       weights.format().network_format().network() !=
           pblczero::NetworkFormat::NETWORK_SE_WITH_HEADFORMAT) {
+    // Check if we must switch to cuda.
+    if (weights.format().network_format().network() ==
```
mooskagh, line 1082: Maybe move this `if` outside of the outer `if` as if the inner is true, outer will also be true?

Also, switch may be more readable (not sure), something like:

```cpp
switch (weights.format().network_format().network()) {
  case pblczero::NetworkFormat::NETWORK_CLASSICAL_WITH_HEADFORMAT:
  case pblczero::NetworkFormat::NETWORK_SE_WITH_HEADFORMAT:
      break;
  case pblczero::NetworkFormat::NETWORK_ATTENTIONBODY_WITH_HEADFORMAT:
      CERR << "Nework format not supported directly, switching backend.";
      return NetworkFactory::Get()->Create(
          "cuda" +
              std::string(std::is_same<half, DataType>::value ? "-fp16" : ""),
          weights, options);
   default:
      throw Exception("Network format " +
                    pblczero::NetworkFormat::NetworkStructure_Name(
                        weights.format().network_format().network()) +
                    " is not supported by CuDNN backend.");
}
```

borg323: I liked the switch version better.

src/neural/cuda/network_cudnn.cc
```diff
@@ -1074,14 +1074,21 @@ std::unique_ptr<Network> MakeCudnnNetwork(const std::optional<WeightsFile>& w,
...
-    throw Exception("Network format " +
-                    pblczero::NetworkFormat::NetworkStructure_Name(
-                        weights.format().network_format().network()) +
-                    " is not supported by CuDNN backend.");
+  switch (weights.format().network_format().network()) {
+    case pblczero::NetworkFormat::NETWORK_CLASSICAL_WITH_HEADFORMAT:
+    case pblczero::NetworkFormat::NETWORK_SE_WITH_HEADFORMAT:
+      break;
+    case pblczero::NetworkFormat::NETWORK_ATTENTIONBODY_WITH_HEADFORMAT:
+      CERR << "Nework format not supported directly, switching backend.";
```
mooskagh, line 1082: Minor nit:
Maybe explicitly write that that it's "not supported by **CuDNN**, switching to **CUDA**".

That way it's more clear to users what's going on.

mooskagh (approved):

PR #1558, opened 2021-05-11
Onednn backend

borg323:
This is the long awaited onednn backend (formerly dnnl, formerly mkl-dnn). To build pass `-Donednn=true -Ddnnl_dir=path/to/dnnl/library`  to meson. This works with both cpu and (intel) gpu. For gpu you will need to build the dnnl library yourself (or ask me for a dll).

For best performance use just one search thread when running on a cpu as a second search thread interferes with the onednn computing threads (onednn on a cpu uses all available cores by default).

There are several backend options, the most important are:

|option|values|default|comment|
|-|-|-|-|
|gpu|*empty* or integer|*empty*|select gpu to use, *empty* for cpu|
|winograd|*empty*, true or false|*empty*|Set to true to use Winograd 3x3 convolution on cpu, false for direct convolution, *empty* to let the library chose. Currently dnnl after v2.0.0 may get this wrong, and is not supported on all processors so may exit with a not very informative error.
|fp16|true or false|true on gpu, false on cpu| Use fp16 (or bf16 if you cpu supports it) for computation.
|batch|integer|64 for fp16, 32 for fp32|The minimum batch size to use, as the library prepares kernels on startup. Set to 0 for dynamic kernel recompilation on every batch (not recommended).
|steps|integer|2|Prepare kernels for this many multiples of batch size.
|threads|*empty* or integer|*empty*|Number of cpu threads to use, *empty* to let the library decide.


borg323:
This is dnnl 2.2.2 compiled with both gpu and cpu support: [dnnl.dll.zip](https://github.com/LeelaChessZero/lc0/files/6460854/dnnl.dll.zip)

Example command line for benchmark:
`./lc0 benchmark -w 703810.pb.gz --threads=1 --backend=onednn --backend-opts=gpu=0`

Tilps (approved):

aochoam:
@borg323 Result of the Sanity checking the dx12 driver.
       _
|   _ | |
|_ |_ |_| v0.28.2 built Dec 13 2021
Detected 4 core(s) and 8 thread(s) in 1 group(s).
Group 0 has 4 core(s) and 8 thread(s).
Found pb network file: C:\Arena\Engines\lc0_dx12/771479.pb.gz
Creating backend [check]...
Working backend set to dx12.
Reference backend set to eigen.
Creating backend [dx12]...
Creating backend [eigen]...
Using Eigen version 3.3.7
Eigen max batch size is 256.
Check mode: check only with relative tolerance 1.0e-04, absolute tolerance 5.0e-01.
Check rate: 100%.

Position: 1/1 rnbqkbnr/pppppppp/8/8/8/8/PPPPPPPP/RNBQKBNR w KQkq - 0 1
*** ERROR check failed for a batch of 1 policy incorrect (but value ok).
*** ERROR check failed for a batch of 19 policy incorrect (but value ok).
Benchmark time 6218 ms, 20 nodes, 21 nps, move b1c3
*** ERROR check failed for a batch of 20 policy incorrect (but value ok).
Benchmark time 6260 ms, 21 nodes, 21 nps, move b1a3
Benchmark time 10007 ms, 21 nodes, 4 nps, move b1a3
bestmove b1a3
*** ERROR check failed for a batch of 144 both value and policy incorrect.
*** ERROR check failed for a batch of 256 both value and policy incorrect.

===========================
Total time (ms) : 18655
Nodes searched  : 421
Nodes/second    : 23
Presione una tecla para continuar . . .

aochoam:
This version does not work for me.

PR #1561, opened 2021-05-14
support for windows mimalloc

borg323:
For windows, mimalloc seems to be a good replacement for the default allocator. Huge trees are destructed many times faster and I also see a measurable nps increase with the random backend.

To build you need to pass `-Dmalloc=mimalloc -Dmimalloc_libdir=c:\path\to\mimalloc\libraries` to meson, and to run you will need to copy `mimalloc-redirect.dll` and `mimalloc-override.dll`.

Thanks to @cn4750 for the suggestion.

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -116,7 +116,7 @@ option('malloc',
 option('mimalloc_libdir',
        type : 'string',
        value: '',
-       description: 'Library directory for malloc=mimallloc')
+       description: 'Library directory for malloc=mimalloc')
 
 option('popcnt',
        type: 'boolean',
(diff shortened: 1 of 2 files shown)


Tilps:
I assume there is no change in the existing support for custom malloc based on having moved the addition to deps earlier in the chain?

meson_options.txt
```diff
@@ -113,6 +113,11 @@ option('malloc',
        value: '',
        description: 'Use alternative memory allocator, e.g. tcmalloc/jemalloc')
 
+option('mimalloc_libdir',
+       type : 'string',
+       value: '',
+       description: 'Library directory for malloc=mimallloc')
```
Naphthalin, line 119: typo in mimallloc

borg323:
> I assume there is no change in the existing support for custom malloc based on having moved the addition to deps earlier in the chain?

tcmalloc works for me on linux. And BTW tcmalloc is also faster than the default (with the random backend)

Tilps (approved):
Approval assuming the typo mentioned is fixed.

PR #1566, opened 2021-05-29
illegal First rank pawns exception!

OfekShochat:
this commit has a test, and also the exception itself. pretty simple, when row is 0 or 8, dont allow pawns.
really the commits should be cleaned up. my master isnt clean..

### DIFF
--- a/src/chess/board.cc
+++ b/src/chess/board.cc
@@ -652,7 +652,7 @@ bool ChessBoard::ApplyMove(Move move) {
   }
 
   // Promotion.
-  if (move.promotion() != Move::Promotion::None) {
+  if (to_row == RANK_8 && pawns_.get(from)) {
     switch (move.promotion()) {
       case Move::Promotion::Rook:
         rooks_.set(to);
--- a/src/chess/board.cc
+++ b/src/chess/board.cc
@@ -1,22 +1,17 @@
 /*
   This file is part of Leela Chess Zero.
   Copyright (C) 2018-2019 The LCZero Authors
-
   Leela Chess is free software: you can redistribute it and/or modify
   it under the terms of the GNU General Public License as published by
   the Free Software Foundation, either version 3 of the License, or
   (at your option) any later version.
-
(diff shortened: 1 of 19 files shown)


src/chess/board.cc
```diff
@@ -1030,6 +1025,8 @@ void ChessBoard::SetFromFen(std::string fen, int* rule50_ply, int* moves) {
       rooks_.set(row, col);
       bishops_.set(row, col);
     } else if (c == 'P' || c == 'p') {
+      if (row == 8 || row == 0)
+        throw std::out_of_range("pawns at the first/eighth rank are forbidden.");
```
borg323, line 1029: The exception should be in the same form as the others around it.

src/chess/board_test.cc
```diff
@@ -53,6 +51,16 @@ TEST(BoardSquare, BoardSquare) {
   }
 }
 
+TEST(ChessBoard, IllegalFirstRankPawns) {
+  ChessBoard board;
+  try {
+    board.SetFromFen("nqrbkrnr/bnnbnbnn/8/8/8/8/NNNBPNBN/QNRPKPQQ w - - 0 1");
+    FAIL() << "expected pawns at the first/eighth rank are forbidden.";
+  } catch (std::out_of_range const &err) {
+    EXPECT_EQ(err.what(), std::string("pawns at the first/eighth rank are forbidden."));
```
borg323, line 60: Don't check for the exact error, just that it fails.

src/chess/board.cc
```diff
@@ -1,22 +1,17 @@
 /*
   This file is part of Leela Chess Zero.
   Copyright (C) 2018-2019 The LCZero Authors
-
```
borg323, line 4: Lots of unrelated white space changes.

src/chess/board_test.cc
OfekShochat: How do I do that?

borg323 (approved):

PR #1567, opened 2021-05-30
Use NHCW layout for fused winograd residual block

ankan-ban:
- Allows better memory access pattern.
 - As the optimization is currently limited to networks with filter size <= 384, add __launch_bounds__ compiler hint to allow it use max possible no of registers to support upto 384 channels.
 - Get rid of residual block path from cudnn backend (people who are able to use that with cudnn backend should be already running the new cuda backend).

With lc0 benchmark running on A100, recent T60 network, minibatch-size=96:
Baseline: 54152
Optimized: 55320

### DIFF
--- a/src/neural/cuda/layers.cc
+++ b/src/neural/cuda/layers.cc
@@ -677,7 +677,7 @@ PolicyMapLayer<DataType>::~PolicyMapLayer() {
 template <typename DataType>
 FusedWinogradConvSELayer<DataType>::FusedWinogradConvSELayer(
     BaseLayer<DataType>* ip, int C, int H, int W, int Cin, bool relu, bool bias,
-    bool skip_add, bool se, int se_k, bool use_gemm_ex)
+    bool skip_add, bool se, int se_k, bool use_gemm_ex, bool op_nhcw)
     : BaseLayer<DataType>(C, H, W, ip, false),
       c_input_(Cin),
       use_relu_(relu),
@@ -685,7 +685,8 @@ FusedWinogradConvSELayer<DataType>::FusedWinogradConvSELayer(
       skip_add_(skip_add),
       has_se_(se),
       se_k_(se_k),
-      use_gemm_ex_(use_gemm_ex) {
+      use_gemm_ex_(use_gemm_ex),
+      op_nhcw_(op_nhcw) {
   // Allocate memory for weights (filter tensor) and biases.
   const size_t weight_size = sizeof(DataType) * c_input_ * C * 3 * 3;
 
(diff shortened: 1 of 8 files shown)


ankan-ban:
>however I get huge speedup (~50%) with "lc0 benchmark" or also when actually running lc0! 
I see the same speedup in other recent appveyor builds too (without this PR) - which means the the search (or some default params) likely have changed on latest Master branch compared to what it was on 0.27.0 release. Sorry for the false hope of 50% improvement (edited the original comment to avoid confusion).

ankan-ban:
Pushed another optimization (to make the first fully connected layer of SE run faster). Now I have:
Baseline: 54152
Optimized: 57154

borg323:
I see the `res_block_fusing` option is removed from the cudnn backend. In that case, shouldn't `custom_winograd` be disabled for filters>384?

ankan-ban:
> I see the `res_block_fusing` option is removed from the cudnn backend. In that case, shouldn't `custom_winograd` be disabled for filters>384?

The winograd transform kernels used by custom_winograd path without res-block fusing are different (separate kernels for input and output transform) but it's likely that those also run into register limit with bigger filter counts and don't run. I haven't tested. If we plan to support bigger networks, maybe it's important to make the fast path work for them. I will think about rewriting the transform kernels so that we don't run into this register limit.

src/neural/cuda/layers.cc
```diff
@@ -836,24 +837,29 @@ void FusedWinogradConvSELayer<DataType>::Eval(
...
-        nullptr, nullptr);
-  else if (!has_se_ && use_relu_ && use_bias_ && skip_add_)
-    OutputTransform<DataType, false, true, true, true>(
+  else if (!has_se_ && use_relu_ && use_bias_ && !skip_add_) {
+    if (op_nhcw_)
+      OutputTransform<DataType, false, true, true, false, false, true>(
+          N, C, 0, output, transformed_output, nullptr, biases_, nullptr,
+          nullptr, nullptr, nullptr);
+    else
+      OutputTransform<DataType, false, true, true, false, false, false>(
```
borg323, line 854: This looks like it can be simplified to `OutputTransform<DataType, false, true, true, false, false, op_nhcw_>(`

borg323:
What about doing the following (against master)?
```diff
--- a/src/neural/cuda/network_cudnn.cc
+++ b/src/neural/cuda/network_cudnn.cc
@@ -375,7 +375,14 @@ class CudnnNetwork : public Network {
     // 2. Build the network, and copy the weights to GPU memory.
 
     // Input.
-    {
+    if (use_custom_winograd_) {
+      auto inputConv = std::make_unique<FusedWinogradConvSELayer<DataType>>(
+          nullptr, kNumFilters, 8, 8, kNumInputPlanes, true, true, false, false,
+          0, use_gemm_ex, use_res_block_winograd_fuse_opt_);
+      inputConv->LoadWeights(&weights.input.weights[0],
+                             &weights.input.biases[0], scratch_mem_);
+      network_.emplace_back(std::move(inputConv));
+    } else {
       auto inputConv = std::make_unique<ConvLayer<DataType>>(
           nhwc_, kNumFilters, 8, 8, 3, kNumInputPlanes, true, true);
       inputConv->LoadWeights(&weights.input.weights[0],
```

src/neural/cuda/layers.cc
ankan-ban: The problem with this (and all other template params) is that they need to be compile time constants, so passing the op_nhcw_ variable as template param doesn't work. I think it would have been lot cleaner if we made the class itself a template on all those variables (has_se, use_relu_, use_bias_, skip_add_ and op_nhcw_).

ankan-ban:
> What about doing the following (against master)?
> 
> ```diff
> --- a/src/neural/cuda/network_cudnn.cc
> +++ b/src/neural/cuda/network_cudnn.cc
> @@ -375,7 +375,14 @@ class CudnnNetwork : public Network {
>      // 2. Build the network, and copy the weights to GPU memory.
>  
>      // Input.
> -    {
> +    if (use_custom_winograd_) {
> +      auto inputConv = std::make_unique<FusedWinogradConvSELayer<DataType>>(
> +          nullptr, kNumFilters, 8, 8, kNumInputPlanes, true, true, false, false,
> +          0, use_gemm_ex, use_res_block_winograd_fuse_opt_);
> +      inputConv->LoadWeights(&weights.input.weights[0],
> +                             &weights.input.biases[0], scratch_mem_);
> +      network_.emplace_back(std::move(inputConv));
> +    } else {
>        auto inputConv = std::make_unique<ConvLayer<DataType>>(
>            nhwc_, kNumFilters, 8, 8, 3, kNumInputPlanes, true, true);
>        inputConv->LoadWeights(&weights.input.weights[0],
> ```

Yes, I think that should work. Let me test it out.

borg323 (approved):

PR #1568, opened 2021-05-31
Multi-stream Optimization for CUDA backend

ankan-ban:
- Create cuda stream and memory needed to run the network for each InputsOutputs object.
- This allows concurrent execution of multiple parallel Eval requests. 
- Note that this PR also include kernel opts from PR #1567 

Some performance nos (for lc0 benchmark running on A100, recent T60 network, minibatch-size=96 all other params default)
```
Baseline       : 54152
Kernel opts    : 57154
+ multi-stream : 67439
+ roundrobin   : 67249 
+ t 3 ( w/ RR) : 70419
  t 4 ( w/ RR) : 72333
  t 4 ( w/Mux) : 73442
```

### DIFF
--- a/src/neural/cuda/common_kernels.cu
+++ b/src/neural/cuda/common_kernels.cu
@@ -597,25 +597,45 @@ template void PolicyMap<half>(int N, half* output, const half* input,
 template void FilterTransform<float>(int N, int C, float* transformedFilter,
                                      const float* filter);
 
-template void InputTransform<float>(int N, int C, float* transformed_input,
-                                    const float* input);
+template void InputTransform<float, true>(int N, int C,
+                                          float* transformed_input,
+                                          const float* input);
 
-template void OutputTransform<float, true, true, true, true>(
+template void InputTransform<float, false>(int N, int C,
+                                           float* transformed_input,
+                                           const float* input);
+
+template void OutputTransform<float, true, true, true, true, false, false>(
     int N, int C, int se_K, float* output, const float* input,
     const float* skip, const float* bias, const float* w1, const float* b1,
     const float* w2, const float* b2);
 
-template void OutputTransform<float, false, true, true, false>(
+template void OutputTransform<float, false, true, true, true, false, false>(
     int N, int C, int se_K, float* output, const float* input,
     const float* skip, const float* bias, const float* w1, const float* b1,
     const float* w2, const float* b2);
 
-template void OutputTransform<float, false, true, true, true>(
+template void OutputTransform<float, true, true, true, true, true, false>(
     int N, int C, int se_K, float* output, const float* input,
     const float* skip, const float* bias, const float* w1, const float* b1,
     const float* w2, const float* b2);
 
-template void OutputTransform<float, false, false, true, false>(
+template void OutputTransform<float, false, true, true, true, true, false>(
+    int N, int C, int se_K, float* output, const float* input,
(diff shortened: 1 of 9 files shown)


src/neural/cuda/inputs_outputs.h
```diff
@@ -65,13 +66,38 @@ struct InputsOutputs {
...
+      multi_stream_ = true;
+      ReportCUDAErrors(cudaStreamCreate(&stream_));
+      ReportCUDAErrors(cudaMalloc(&scratch_mem_, scratch_size));
+      for (auto& mem : tensor_mem_) {
+        ReportCUDAErrors(cudaMalloc(&mem, tensor_mem_size));
+        ReportCUDAErrors(cudaMemsetAsync(mem, 0, tensor_mem_size, stream_));
+      }
+      ReportCUBLASErrors(cublasCreate(&cublas_));
+      ReportCUBLASErrors(cublasSetMathMode(cublas_, CUBLAS_TENSOR_OP_MATH));
+      ReportCUBLASErrors(cublasSetStream(cublas_, stream_));
```
Tilps, line 81: Does this assume that InputOutputs is constructed on the same thread that evaluation is called?  I didn't think that was guaranteed with some backend combinations?

Tilps: Similar but different question.  cudaSetDevice doesn't appear to necessarily have been called for the thread that constructs InputsOutputs - is there potential that this change breaks multi-gpu?

ankan-ban: if cudaSetDevice isn't called on the thread that constructs InputsOutputs we already would have a problem with multi-gpu as we were allocating some memory there before this change too.

ankan-ban:
As excepted the perf benefit on relatively slower GPUs is much less. 

On Titan RTX, lc0 benchmark running with  recent T60 network, minibatch-size=64 all other params default:
```
baseline                   : 21279
kernel opts                : 21353
kernel opts + multi-stream : 22286
```

src/neural/cuda/inputs_outputs.h
Tilps: I was getting lost in the diff - I've found where cudaSetDevice is called correctly now - so I'm happy now :)

src/neural/cuda/kernels.h
```diff
@@ -32,11 +32,11 @@ namespace cudnn_backend {
 // activation (relu, tanh or sigmoid).
 template <typename T>
 void addVectors(T* c, T* a, T* b, int size, int asize, int bsize, bool relu,
-                bool use_tanh, bool use_sigmoid);
+                bool use_tanh, bool use_sigmoid, cudaStream_t stream = 0);
```
Tilps, line 35: Consider removing the default value for stream here.

src/neural/cuda/common_kernels.cu
```diff
@@ -324,12 +324,12 @@ __global__ void expandPlanes_kernel_Fp16_NCHW(half* output,
 }
 
 void expandPlanes_Fp16_NCHW(half* output, const uint64_t* masks,
-                            const float* values, int n) {
+                            const float* values, int n, cudaStream_t stream = 0) {
```
Tilps, line 327: Consider removing the default value for stream here.
It seems a risky item to have a default on, and almost every single other function which takes a stream doesn't use a default.

Tilps (approved):

borg323 (approved):

PR #1581, opened 2021-06-16
change default winograd on gtx16x0

masterkni666:
recent changes in PR1567 seems to cause GTX16x0 to crash when using custom_winograd
may need to add extra code to throw exception when set to true


borg323:
Specifically, `res_block_fusing=true` is the case crashing here. However, `custom_winograd=false` was tested to give better performance, so it is the suggested default for those gpus.

ankan-ban (approved):
The change looks good if disabling custom_winograd improves performance.

I will try to figure out why it crashes on GTX 16xx and see if we can make it faster there - but that may take some time (need to get access to a system with GTX 16xx).

PR #1584, opened 2021-06-22
revise legacy time manager

borg323:


src/mcts/stoppers/legacy.cc
```diff
@@ -113,16 +113,15 @@ std::unique_ptr<SearchStopper> LegacyTimeManager::GetStopper(
...
+  auto total_moves_time = std::max(0.0f, *time + increment * (movestogo - 1) -
+                                             move_overhead_ - time_spared_ms_);
 
   // If there is time spared from previous searches, the `time_to_squander` part
   // of it will be used immediately, remove that from planning.
   int time_to_squander = 0;
   if (time_spared_ms_ > 0) {
     time_to_squander = time_spared_ms_ * spend_saved_time_;
     time_spared_ms_ -= time_to_squander;
-    total_moves_time -= time_to_squander;
```
kiudee, line 125: Shouldn’t `time_to_squander` then be added to this_move_time?

borg323: Yes, it was already there. With the default value of `immediate-use` (that is 1) the behavior of the code up to this point is the same.

src/mcts/stoppers/legacy.cc
```diff
@@ -143,10 +142,10 @@ std::unique_ptr<SearchStopper> LegacyTimeManager::GetStopper(
   constexpr int kSmartPruningToleranceMs = 200;
   if (slowmover_ < 1.0 ||
       this_move_time * slowmover_ > kSmartPruningToleranceMs) {
-    this_move_time *= slowmover_;
     // If time is planned to be overused because of slowmover, remove excess
     // of that time from spared time.
     time_spared_ms_ -= this_move_time * (slowmover_ - 1);
```
kiudee, line 147: If `time_spared_ms_` becomes negative, this is then added to total_moves_time.
Or is this prevented somewhere?

borg323: I think it is safe, but I see a way to improve it.

src/mcts/stoppers/legacy.cc
```diff
@@ -113,13 +113,14 @@ std::unique_ptr<SearchStopper> LegacyTimeManager::GetStopper(
...
-                                             move_overhead_ - time_spared_ms_);
+  auto total_moves_time =
+      std::max(0.0f, *time + increment * (movestogo - 1) - move_overhead_);
 
   // If there is time spared from previous searches, the `time_to_squander` part
   // of it will be used immediately, remove that from planning.
   int time_to_squander = 0;
   if (time_spared_ms_ > 0) {
+    total_moves_time = std::max(0.0f, total_moves_time - time_spared_ms_);
     time_to_squander = time_spared_ms_ * spend_saved_time_;
```
kiudee, line 124: Looks good to me now.
So `time_spared_ms_` can be negative for a while as a kind of "time debt" until enough time was saved to get it positive again.

src/mcts/stoppers/legacy.cc
mooskagh: And if `slowmover_` is < 1.0, it will add time, is that intended?

borg323: This is how I understood the intention of the code: To adjust `time_spared_ms_` according to what `slowmover` saves/uses for the next move. Whether this is the best way is an open question.

mooskagh (approved):
Cannot say I'm 100% sure it works correctly now, but at least it should be more correct than before. :)

zz4032:
A small remark: variable name `midpoint-move` here is called `midpoint` in smooth TM.

Tilps:
# PLAYER    :  RATING  ERROR  POINTS  PLAYED   (%)  CFS(%)    W     D    L  D(%)
   1 lc0_b     :      31     10   542.5    1000  54.2      84  154   777   69  77.7
   2 lc0_a     :      24     10   533.0    1000  53.3     100  147   772   81  77.2
   3 sf        :       0   ----   924.5    2000  46.2     ---  150  1549  301  77.5

White advantage = 67.64 +/- 3.85
Draw rate (equal opponents) = 87.33 % +/- 1.06

LTC (60+0.6 on V100 with T60) test of this patch on fleet.  probability of regression is reasonably low :P

PR #1590, opened 2021-06-27
FPU experiment

mooskagh:
This is my attempt to get rid of visited policy by trying to predict the Q from parent's Q and ratio between edge's P and max edge's P of the same node.

In a scientific sample of 10 games of fixed 10000 nodes looks negative (all draws, but eval was drifting towards the losing side).

The FPU function is hand-fitted and is like this:

image

:)


mooskagh:
Updated to this, still doesn't look like it's any stronger.
image

PR #1596, opened 2021-06-30
Quickfix for smooth TM nps misprediction.

### LINKED ISSUE #1585: Lc0 instamoved (TCEC 21 G44 vs KomodoDragon)
In TCEC 21 DivP game 44 (vs KomodoDragon), Lc0 instamoved moves 14..22, while having as few nodes as 2.5M.

Probably because it underestimated nps.

More data follows.

mooskagh:
Fixes #1585.

There are surely more to fix, including some simple and obvious things, but in this PR I'd really like only to fix the most urgent issue of instamoving due to nps misprediction.

What it PR does:
1. If at any moment, the current avg move nps (move nodes / move time, the thing that is displayed in `uci`) is higher than predicted, just assign that nps to whatever time manager uses.
2. If we are longer than `nps-update-period` (default is 20 seconds) in the move, also use "displayed move nps" as time manager nps.
3. Otherwise linearly decrease nps at the rate so that `nps-update-period` after the move start it lands exactly at current move's nps.

The graph of nps estimation for 14 seconds nps-update-period:
image

### DIFF
--- a/src/mcts/stoppers/smooth.cc
+++ b/src/mcts/stoppers/smooth.cc
@@ -178,13 +178,14 @@ float LinearDecay(float cur_value, float target_value,
   if (time_since_last_update_ms <= 0) return cur_value;
   // If window is large enough, we consider
   if (time_since_reliable_ms >= window_size_ms) return target_value;
-  // How long there was a reliable nps in the window, in ms.
-  const float time_reliable = time_since_reliable_ms - window_size_ms;
+  // time_since_reliable_ms during the previous call of this function.
+  const float prev_time_since_reliable_ms =
+      time_since_reliable_ms - time_since_last_update_ms;
   // The same, but as fraction of window size.
-  const float reliable_frac = time_reliable / window_size_ms;
+  const float prev_reliable_frac = prev_time_since_reliable_ms / window_size_ms;
   // What was the value when it was reliable.
-  const float reliable_value =
-      (reliable_frac * target_value - cur_value) / (reliable_frac - 1);
+  const float reliable_value = (prev_reliable_frac * target_value - cur_value) /
+                               (prev_reliable_frac - 1);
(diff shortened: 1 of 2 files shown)


src/mcts/stoppers/smooth.cc
```diff
@@ -154,6 +154,48 @@ float ExponentialDecay(float from, float to, float step, float value) {
...
+float LinearDecay(float cur_value, float target_value,
+                  float time_since_reliable_ms, float time_since_last_update_ms,
+                  float window_size_ms) {
+  // If target is higher than cur value, just return target without any
+  // smoothing. That makes sense for nps, but for other averages we should look
+  // a bit closer.
+  if (cur_value <= target_value) return target_value;
+  // If no time passed since last update, return previous value.
+  if (time_since_last_update_ms <= 0) return cur_value;
+  // If window is large enough, we consider
```
borg323, line 179: Something seems to be missing in the comment.

src/mcts/stoppers/smooth.cc
```diff
@@ -154,6 +154,48 @@ float ExponentialDecay(float from, float to, float step, float value) {
...
+//    the case of nps, end of the previous move is considered reliable point).
+// @time_since_last_update_ms -- how long ago was this function last called.
+// @window_size_ms -- in what time should the output converge to @target_value.
+float LinearDecay(float cur_value, float target_value,
+                  float time_since_reliable_ms, float time_since_last_update_ms,
+                  float window_size_ms) {
+  // If target is higher than cur value, just return target without any
+  // smoothing. That makes sense for nps, but for other averages we should look
+  // a bit closer.
+  if (cur_value <= target_value) return target_value;
```
borg323, line 176: Maybe this would be clearer outside the function?

borg323 (approved):
My comments are minor and can be addressed later.

PR #1600, opened 2021-07-03
Evil hack to focus search.

Tilps:
Creating PR just so I can try testing it.
I've only tried a short search from start position - this could cause crashes and race conditions for all I know :P


Tilps:
Tested on fleet - was in the noise, marginally positive.  So probably doesn't crash :P

borg323:
I made it configurable with the hidden option `--policy-trim` that defaults to 255 so it is a no-op (but can be tuned). The tests mentioned above were with the equivalent of `--policy-trim=20`, while for ceres it is `--policy-trim=63`.

Tilps:
Maybe worth making a note in the config option comment that this 'can totally break things' when set to low numbers.
Like invalid mate proofs.

PR #1601, opened 2021-07-04
Implement 'micro' neural network for smart pruning.

Tilps:
Values chosen based on training data generated from games played at 300+3 on a V100.
Possibly will require tuning for each TC...

On the training data (which is generated in a way which may make it unfair to SPF)
an SPF of 1.3 gets precision of 0.989 and recall of 0.62.
This tiny formula (with appropriate threshold, I just selected 2.1 as a guess in this actual CL since I didn't have enough data to know the actual value) on the same data set can get a recall of 0.81 for that same precision.


Tilps:
Definitely bad at 30+0.3

Tilps:
and at 300+3 - curious.
I think maybe I need to change the training strategy. Or maybe I screwed up the calculation for the threshold.

PR #1604, opened 2021-07-13
Record movestats for pgn

Tilps:
Alternative hacks to PR1599, that work off a pgn of positions rather than recording stats while playing games.

### DIFF
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -27,6 +27,11 @@
 
 #include "engine.h"
 
+#include <chess/pgn.h>
+#include <utils/random.h>
+
+
+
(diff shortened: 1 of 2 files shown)


Naphthalin:
This seems to be very related to the #1682 PR by borg? Is this PR still needed then, given that we want training data under training conditions?

Tilps:
This is for generating hyperpruning training data - not lc0 training data.

Naphthalin:
Ah, I see -- in that case, `demo` is probably the more appropriate label, as this is relevant to any sort of attempt at having a non-handcrafted pruning strategy.

PR #1613, opened 2021-08-02
Add ability to limit in-flight throughout search tree based on scaling limit.

Tilps:
Doesn't appear to work reliably yet.  I think I've probably missed some corner cases.

The NPS impact of the random parameters I made the default isn't as bad I was expecting... when it doesn't just completely break.

This also should be disabled by default in selfplay for training purposes.


Tilps:
The current default parameters here are 90% nps on a 1k node bench - and +45 elo(+/-20) - with some random small network I happen to have around.
Currently running a tune to see if I can find something clearly good at at least short TC.

The 'hack' clearing shared collisions after the sleep in UpdateCounters is not great. I think I have a better alternative (more proactive clearing to let second thread make progress), but so far it seems that the condition the hack is aiming to fix is kind of rare, so 'optimizing' probably isn't a huge win.

PR #1626, opened 2021-08-21
Increment time added to Alphazero TM.

zz4032:


src/mcts/stoppers/alphazero.cc
```diff
@@ -36,33 +36,40 @@ class AlphazeroTimeManager : public TimeManager {
...
   const bool is_black = position.IsBlackToMove();
   const std::optional<int64_t>& time = (is_black ? params.btime : params.wtime);
+  const std::optional<int64_t>& time_inc = (is_black ? params.binc : params.winc);
   // If no time limit is given, don't stop on this condition.
   if (params.infinite || params.ponder || !time) return nullptr;
 
   auto total_moves_time = *time - move_overhead_;
 
-  float this_move_time = total_moves_time * (alphazerotimepct_ / 100.0f);
+  float this_move_time = (total_moves_time - *time_inc) * (alphazerotimepct_ / 100.0f)
```
borg323, line 67: I'm not sure what is the best way to handle this. Subtracting `time_inc` may result in negative `this_move_time`. My initial thinking was that just using `total_moves_time` like in the first iteration is OK, and it is safe as we guard against exceeding the time limit.

zz4032: I noticed that `wtime` is always main+increment time added (at least in Cutechess). So `this_move_time` should never become negative.

borg323: At least make it `std::max(0, total_moves_time - *time_inc)`. I'm not sure all uci hosts do it this way.

zz4032:
Time usage graph of PR1626 and master. Higher percentage for increment time leads to less unused time at the end of the game.
time

src/mcts/stoppers/alphazero.cc
```diff
@@ -70,6 +70,12 @@ std::unique_ptr<SearchStopper> AlphazeroTimeManager::GetStopper(
       (alphazerotimepct_ / 100.0f) + *increment *
       (alphazeroincrementpct_ / 100.0f);
 
+  // If increment time is added only after a move has been made,
```
zz4032, line 73: I realized that `*increment * (alphazeroincrementpct_ / 100.0f)` can still get higher than available time.
Now the calculation looks safe. If increment is added before the move is made (e.g. in Cutechess), the code works without precautions as in bbfa544. If increment is added after, 1) `*increment` or 2) `*increment * (alphazeroincrementpct_ / 100.0f)` can be higher than available time and this is now covered with (1c21588) and this commit.

zz4032:
I realized that `*increment * (alphazeroincrementpct_ / 100.0f)` can still get higher than available time.
Now the calculation looks safe. If increment is added before the move is made (e.g. in Cutechess), the code works without precautions as in bbfa544. If increment is added after, 1) `*increment` or 2) `*increment * (alphazeroincrementpct_ / 100.0f)` can be higher than available time and this is now covered with (1c21588) and this commit.

src/mcts/stoppers/alphazero.cc
```diff
@@ -36,33 +36,49 @@ class AlphazeroTimeManager : public TimeManager {
   AlphazeroTimeManager(int64_t move_overhead, const OptionsDict& params)
       : move_overhead_(move_overhead),
         alphazerotimepct_(
-            params.GetOrDefault<float>("alphazero-time-pct", 12.0f)) {
+            params.GetOrDefault<float>("alphazero-time-pct", 12.0f)),
+        alphazeroincrementpct_(
+            params.GetOrDefault<float>("alphazero-increment-pct", 95.0f)) {
     if (alphazerotimepct_ < 0.0f || alphazerotimepct_ > 100.0f)
       throw Exception("alphazero-time-pct value to be in range [0.0, 100.0]");
+    if (alphazeroincrementpct_ < 0.0f || alphazeroincrementpct_ > 100.0f)
```
mooskagh, line 44: There should be `{` `}` around the body of the if statement if it doesn't fit one line. (also in the line above)

src/mcts/stoppers/alphazero.cc
```diff
@@ -36,33 +36,49 @@ class AlphazeroTimeManager : public TimeManager {
...
   const float alphazerotimepct_;
+  const float alphazeroincrementpct_;
 };
 
 std::unique_ptr<SearchStopper> AlphazeroTimeManager::GetStopper(
     const GoParams& params, const NodeTree& tree) {
   const Position& position = tree.HeadPosition();
   const bool is_black = position.IsBlackToMove();
   const std::optional<int64_t>& time = (is_black ? params.btime : params.wtime);
+  const std::optional<int64_t>& inc = (is_black ? params.binc : params.winc);
```
mooskagh, line 62: No need it to be a reference, temporary object will be created anyway for std::optional (also in the line above)

src/mcts/stoppers/alphazero.cc
```diff
@@ -36,33 +36,49 @@ class AlphazeroTimeManager : public TimeManager {
...
+  const int increment = inc ? std::max(int64_t(0), *inc) : 0;
```
mooskagh, line 63: Possibly`std::max(0LL, inc.value_or(0))` is nicer.

src/mcts/stoppers/alphazero.cc
```diff
@@ -36,33 +36,49 @@ class AlphazeroTimeManager : public TimeManager {
...
 
-  float this_move_time = total_moves_time * (alphazerotimepct_ / 100.0f);
+  float this_move_time = std::max<int64_t>(0, total_moves_time - increment) *
+      (alphazerotimepct_ / 100.0f) + increment *
+      (alphazeroincrementpct_ / 100.0f);
 
-  LOGFILE << "Budgeted time for the move: " << this_move_time << "ms"
-          << "Remaining time " << *time << "ms(-" << move_overhead_
-          << "ms overhead)";
+  // If increment time is added only after a move has been made,
```
mooskagh, line 73: Can it really happen? It's `total_move_time - increment` above.

zz4032: Only in the first move.
In official FIDE rules it says: `increment: 6.1. An amount of time (from 2 to 60 seconds) added from the start before each move for the player.` Also in blitz games of world class players you can see 03:02 on the clock before the start of the game with 3+2 time control.
But here some say it should be added after a move: https://old.chesstempo.com/chess-forum/general_chess_discussion/question_about_fisher_clock-t2042.0.html Maybe it's the old FIDE rules.

mooskagh: Indeed, I didn't think about the first move.
I think in human (and WCCC) championships it's indeed added before the first move, but in computer ones it's after.

It can also happen in werd TCs like "1 hour for first 40 moves, than 15 seconds increment per move".

src/mcts/stoppers/alphazero.cc
zz4032: `inc.value_or(0)` is not sufficient?

mooskagh: Probably it is sufficient. I'm not sure whether there is such a thing as negative increment.
Negative base time exists though, and smoth TM happily crashes on it.

mooskagh (approved):

PR #1633, opened 2021-08-27
Alphazero TM increment adaptation.

seniorincognito:


seniorincognito:
Explanation of why this was made and what it does:
Currently the A0 percentage use to calculate the budged time to make a move is fixed. This results in what I see as a big problem when the initial time and the increment proportions are different from what it was tuned for.

To show an example that make its obvious, lets say you have 2 time controls, 60' +5'' (tuned time control) and 30' + 30'' (not tuned), if we use the same percentage for both, the second time control is going to have half the budged time than first due to A0 percentage only takes in account bank time but not projected time, even tho both time control are going to converge (in this case in about 72 moves).

Under the hypothesis that because those time control converge you would want to 30' 30'' to spend about the same time than 60' 60'' which is the tuned time control, I do this calculation: alphazerotimepct_ * (tuned_initial_time_ / initial_time_sec_) * (expected_game_time_ / expected_tuned_game_time_) which gives a new percentage that would result in about A0% being doubled because in this way is not only taken in account the initial time, but also the projected time (i do in 75 moves that is more or less the TCEC Superfinal move average at the time i did this.

I gave the option to decay the % to the initial one, which is untested whether its desirable or not, and the way it decays could be better also.

In addition, but not very consequential, it adds a percentage of the increment to assigned as budget time:
  // Using part of the increment based on the difference between initial total time and current total time.
  alphazeroincrementpct_ = (1 - std::min<float>(1, std::max<float>(0,(total_moves_time - *increment))/initial_time_)) * 100.0f;

Less time left more increment used. On my testing this required some moveoverhead (20-50ms) to avoid flagging because the budgeted time would approach very closely to the total time left.

Lastly I wanna add that for sure there is better ways to do this, but I believe this is a good improvement over the default and provides an idea to improve on.

Naphthalin:
Reopening this for testing and comparing.

PR #1638, opened 2021-09-06
More correctly clear some fields when doing a v5 upgrade.

Tilps:


borg323 (approved):
Probably won't apply after #1637 is merged.

Tilps:
one extra line in between is magic ;)

PR #1639, opened 2021-09-10
high/low node split and low node sharing through the cache

borg323:
The objective is to expand the low node to include information that will allow better handling of transpositions.
The first stage of this with most of the required refactoring is in #1447.

**What is new in this PR**
- Introduces `struct LowNode` and the reference counted `SharedLowNodePtr`.
- The `LowNode` is shared between cache and nodes. On a hit the cache just returns a `SharedLowNodePtr`.
- Since `LowNode` also stores the moves, there is no move generation on cache hits, just a move validity check.
- When adding Dirichlet noise a copy of the `LowNode` is made (that is not cached).


borg323:

PR #1640, opened 2021-09-19
move GetV6TrainingData() and related files to new directory

borg323:
This introduces a new class `V6TrainingDataArray` that together with `V6TrainingData`, `TrainingDataReader` and `TrainingDataWriter` are now in the `trainingdata` directory. 

*Original description* (was "move GetV6TrainingData() to game.cc"):

Note that there are some changes:
```
  V6TrainingData GetV6TrainingData(
      GameResult result, const PositionHistory& history,
      FillEmptyHistory fill_empty_history,
      pblczero::NetworkFormat::InputFormat input_format, Eval best_eval,
      Eval played_eval, bool best_is_proven, Move best_move, Move played_move,
      const NNCacheLock& nneval) const;
```
becomes
```
  V6TrainingData GetV6TrainingData(
      const NodeTree& tree, pblczero::NetworkFormat::InputFormat input_format,
      Eval best_eval, Eval played_eval, bool best_is_proven, Move best_move,
      Move played_move) const;
```
with the game result hardcoded to `GameResult::UNDECIDED` and the rest obtained from the tree and search.


Tilps:
Alternative to consider - extracting it to be a public method in somewhere more central.  I prefer this new signature to its previous home - but I have in the past had custom branches which created training data out of uci mode, and there was someone talking about generating training data by walking through a pgn and doing a search for each position - which might not want to be in self play game.

src/trainingdata/trainingdata.h
```diff
@@ -1,6 +1,6 @@
 /*
```
Tilps, line 1: I'm not sure we really want git to have marked trainingdata.h as a fork of writer.h

src/trainingdata/writer.h
```diff
@@ -0,0 +1,61 @@
+/*
```
Tilps, line 1: Seems git failed to auto recognize this as a move?

src/trainingdata/trainingdata.h
Tilps: Is it possible that the files in neural being moved to trainingdata were not actually deleted?

src/trainingdata/trainingdata.h
```diff
@@ -90,28 +87,28 @@ static_assert(sizeof(V6TrainingData) == 8356, "Wrong struct size");
...
-
-  // Flushes file and closes it.
-  void Finalize();
-
-  // Gets full filename of the file written.
-  std::string GetFileName() const { return filename_; }
+  V6TrainingDataArray(FillEmptyHistory white_fill_empty_history,
+                      FillEmptyHistory black_fill_empty_history,
+                      pblczero::NetworkFormat::InputFormat white_input_format,
+                      pblczero::NetworkFormat::InputFormat black_input_format)
```
Tilps, line 95: I would suggest only taking in one input_format - makes it clearer why we're checking and asserting that they are the same back near the call site.

src/trainingdata/trainingdata.h
borg323: I just did a straight `git mv`, but git detects that `trainingdata.h` has more in common with the original `writer.h` than the new one. I can do the move as a separate commit, but I expect it will still show it like that in the next step.

src/selfplay/game.cc
```diff
@@ -79,7 +77,15 @@ SelfPlayGame::SelfPlayGame(PlayerOptions white, PlayerOptions black,
                            bool shared_tree, const Opening& opening)
     : options_{white, black},
       chess960_{white.uci_options->Get<bool>(kUciChess960) ||
-                black.uci_options->Get<bool>(kUciChess960)} {
+                black.uci_options->Get<bool>(kUciChess960)},
+      training_data_(SearchParams(white.uci_options).GetHistoryFill(),
+                     SearchParams(black.uci_options).GetHistoryFill(),
+                     white.network->GetCapabilities().input_format,
+                     black.network->GetCapabilities().input_format) {
+  if (white.network->GetCapabilities().input_format !=
```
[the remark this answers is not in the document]
borg323: I moved the check into `Play()`, where we have the `training` flag.

Tilps (approved):

PR #1645, opened 2021-09-28
do policy softmax for valid moves in check backend

borg323:


Tilps:
Generally looks fine assuming it works...
Just wondering if you were deliberately delaying legal move gen to evaluate time rather than input add time - because it appears board isn't useful other than for generating legal moves, and generating legal moves potentially ends up happening multiple times?

borg323:
> Just wondering if you were deliberately delaying legal move gen to evaluate time rather than input add time - because it appears board isn't useful other than for generating legal moves, and generating legal moves potentially ends up happening multiple times?
No reason, I just didn't think it through...

Tilps (approved):

PR #1660, opened 2021-11-24
Add support for INPUT_STATIC_SQUARES

Tilps:
untested, entirely possible that the masks are created in the wrong order. - depends on lczero-common PR18

### DIFF
--- a/src/neural/encoder.h
+++ b/src/neural/encoder.h
@@ -46,6 +46,7 @@ int TransformForPosition(pblczero::NetworkFormat::InputFormat input_format,
 // Encodes the last position in history for the neural network request.
 InputPlanes EncodePositionForNN(
     pblczero::NetworkFormat::InputFormat input_format,
+    pblczero::NetworkFormat::InputStaticFormat input_static_format,
     const PositionHistory& history, int history_planes,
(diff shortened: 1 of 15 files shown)


src/neural/encoder.h
```diff
@@ -46,6 +46,7 @@ int TransformForPosition(pblczero::NetworkFormat::InputFormat input_format,
 // Encodes the last position in history for the neural network request.
 InputPlanes EncodePositionForNN(
     pblczero::NetworkFormat::InputFormat input_format,
```
borg323, line 48: Why not pass the `NetworkCapabilities` structure directly?

borg323 (commented):
A simple check for unsupported `input_static_format` is enough for all other backends, to be fixed in a later PR.

src/neural/encoder.h
Tilps: A similar thought crossed my mind.  Also passing NetworkFormat to NetworkCapabilities instead of the 3 format properties.

Tilps:
I think other backends should be easy enough to fix so I probably won't disable them.

Tilps:
Step 1 is actually finding some evidence of elo gain with a normal network rather than my hybrid coatnet...

PR #1668, opened 2021-12-12
Update version.inc

von-dev:
changed version name to 0.28.1 from 0.29.0-dev


borg323:
Thanks, but this is not the way we do it: Only "dev" versions are in the master branch, the releases are in "release/x.xx" branches, like "release/0.28". You were probably confused because I made a mistake and made the v0.28.1 release from the master branch. This release is now deleted and replaced by v0.28.2.

von-dev:
That makes sense.

PR #1673, opened 2021-12-16
Alternative caching modification

borg323:
This is a simpler version of #1447.


borg323:
Converted to draft since I though of another alternative, so don't want this reviewed yet.

borg323:

PR #1675, opened 2021-12-20
WAR for cublas issue on TU11x GPUs with version CUDA 11.0+

ankan-ban:
- set CUBLAS_PEDANTIC_MATH to avoid cublas making use of tensor math on TU11x GPUs.
- From documentation:
"Starting with cuBLAS version 11.0.0, the library will automatically make use of Tensor Core capabilities wherever possible, unless they are explicitly disabled by selecting pedantic compute modes in cuBLAS (see cublasSetMathMode(), cublasMath_t)."
- Unfortunately there is a bug in cublas which enables tensor math even on TU11x GPUs which is emulated in software is orders of magnitude slower, so this PR is a workaround to disable tensor math to make cublas pick 2xFP16 math kernels on TU11x.

### DIFF
--- a/src/neural/cuda/cuda_common.h
+++ b/src/neural/cuda/cuda_common.h
@@ -37,6 +37,10 @@
 typedef void* cudnnHandle_t;
 #endif
 
+#if CUBLAS_VER_MAJOR < 11
+#define CUBLAS_PEDANTIC_MATH CUBLAS_DEFAULT_MATH
+#endif
+
 namespace lczero {
 namespace cudnn_backend {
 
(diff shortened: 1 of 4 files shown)


borg323:
I added the following to fix building with cuda versions before 11.
```diff
--- a/src/neural/cuda/cuda_common.h
+++ b/src/neural/cuda/cuda_common.h
@@ -37,6 +37,10 @@
 typedef void* cudnnHandle_t;
 #endif
 
+#if CUBLAS_VER_MAJOR < 11
+#define CUBLAS_PEDANTIC_MATH CUBLAS_DEFAULT_MATH
+#endif
+
 namespace lczero {
 namespace cudnn_backend {
 
```

borg323 (approved):

PR #1676, opened 2021-12-23
Add cuda backend support for bottleneck nets

Tilps:
No fancy fusing...

Have to add
conv3 = 4;
conv4 = 5; 
to Residual in net.proto (to match the same change required for PR186 in training)

*only* cuda backend for right now... 
It compiles, but otherwise completely untested right this second... I'll be testing it soon.

### DIFF
--- a/src/neural/cuda/network_cuda.cc
+++ b/src/neural/cuda/network_cuda.cc
@@ -216,6 +216,12 @@ class CudaNetwork : public Network {
     if (weights.residual[0].has_se) {
       has_se_ = true;
     }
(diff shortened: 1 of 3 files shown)


Tilps:
Policy output doesn't look crazy with my very early net, so it probably isn't too wrong...
However backendbench results show the bottleneck 192x15 (which thus has 96 filter 3x3 convolutions) being slightly slower than my baseline 192x15. Assumedly some fusing could probably push bottleneck in front, but it isn't trivially obvious that we could get a large enough speed gain to make it a clear winner...

PR #1677, opened 2021-12-24
Add support to cuda backend for channel output of first convolution in residual being different size.

Tilps:
Only cuda backend for now. The cudnn change is just so it compiled, doesn't actually add support.


Tilps:
Now also cudnn backend probably... Also supports a different number of outputs for the first conv in convolutional policy head.

PR #1678, opened 2021-12-26
Optimized Res-block Fusion without SE

ankan-ban:
* Add specialized kernel to handle res-block fusion (i.e, OutputTansform -> relu/skip-add -> InputTransform) when SE isn't used. This kernel uses a fixed thread-block size (64), and so it doesn't run out of register file or shared memory when dealing with networks with larger no of filters. 
* Right now this doesn't provide any noticeable speed improvement for existing networks (even 40bx512f networks like from T78 run on Ampere GPUs). Barely 1-3%.
* Also modified the res-block fusion checks to enable fusing in more cases (i.e, for alternate layers that don't have SE even for SE networks). This allows the res block fusion to get half the benefit on Turing GPUs even for 512 filter networks. E.g: about 12-14% speed improvement.

Some performance nos with this PR:
```
RTX 2060, backend bench, 512x40 network:
-----------------------------------------
Baseline: Benchmark batch size 80 with inference average time 34.3924ms - throughput 2326.1 nps.
Opt:      Benchmark batch size 80 with inference average time 30.1432ms - throughput 2654 nps.

Baseline: Benchmark batch size 256 with inference average time 101.979ms - throughput 2510.33 nps.
Opt:      Benchmark batch size 256 with inference average time 90.4115ms - throughput 2831.5 nps.

RTX 2060, benchmark (with minibatch size of 80):
------------------------------------------------
baseline: Nodes/second    : 3812
opt:      Nodes/second    : 4347

A100, backend bench (512x40 network):
-------------------------------------
Baseline: Benchmark batch size 96 with inference average time 6.27982ms - throughput 15287.1 nps.
Opt:      Benchmark batch size 96 with inference average time 6.24666ms - throughput 15368.2 nps.

Baseline: Benchmark batch size 192 with inference average time 11.9449ms - throughput 16073.9 nps.
Opt:      Benchmark batch size 192 with inference average time 11.7005ms - throughput 16409.6 nps.

Baseline: Benchmark batch size 256 with inference average time 18.0344ms - throughput 14195.1 nps.
Opt:      Benchmark batch size 256 with inference average time 17.4163ms - throughput 14698.9 nps.

A100, benchmark (default minibatch size)
----------------------------------------
Baseline: Nodes/second    : 32951
Opt:      Nodes/second    : 33702

Titan RTX, benchmark (default minibatch size)
---------------------------------------------
Nodes/second    : 9893
Nodes/second    : 11078
```


src/neural/cuda/common_kernels.cu
```diff
@@ -553,16 +553,22 @@ void OutputInputTransform(int N, int C, int se_K, T* output, const T* input,
...
+    dim3 grid_dim(DivUp(C, kOpInpTransformBlockSize), N, 1);
+    OutputTransform_relu_InputTransform_kernel<float, relu, use_bias, use_skip>
+        <<<grid_dim, kOpInpTransformBlockSize, 0, stream>>>(N, C, output, input,
+                                                            (float*)skip, bias);
+  } else if (C > kMaxResBlockFusingChannels) {
+    throw Exception(
+        "res block fusing opt not supported for the given data type and no "
+        "of filters\n");
   } else {
     OutputTransform_SE_relu_InputTransform_kernel<float, use_se, relu, use_bias,
```
borg323, line 566: This kernel will always be called with use_se=true, so some additional simplification is possible.

ankan-ban: Thanks. I have adjust the change appropriately (after resolving conflicts). This also fixes the perf regression caused by the MISH change (due to running out of registers and local memory spills for use_se==false case). I also got rid of 8x8 skipInp array for the use_se==true kernel to reduce register pressure a bit (which also fixes the random functional failure that we were seeing after the activation change on some GPUs).

src/neural/cuda/fp16_kernels.cu
```diff
@@ -434,17 +434,24 @@ void OutputInputTransform(int N, int C, int se_K, T* output, const T* input,
...
+                                                            (half*)skip, bias);
+  } else if (C > kMaxResBlockFusingChannels) {
     // Use special kernel with reduced register pressure - only works on Ampere,
     // and only for fp16.
     if (C <= kMaxResBlockFusingSeKFp16Ampere) {
       cudaFuncSetAttribute(
           OutputInputTransformKernel_fp16_shmem_board<use_se, activation,
                                                       use_bias, use_skip>,
-          cudaFuncAttributeMaxDynamicSharedMemorySize, 72 * 1024);
+          cudaFuncAttributeMaxDynamicSharedMemorySize,
```
Tilps, line 450: This path is only ever called with use_se true now right?

meson.build
```diff
@@ -490,7 +490,7 @@ if get_option('build_backends')
     cuda_cc = get_option('cc_cuda') # Unfortunately option cuda_cc is reserved.
     nvcc_extra_args = []
     if cuda_cc != ''
-      nvcc_extra_args = ['-arch=compute_' + cuda_cc, '-code=sm_' + cuda_cc]
+      nvcc_extra_args = ['--use_fast_math', '-arch=compute_' + cuda_cc, '-code=sm_' + cuda_cc]
```
Tilps, line 493: Note: we had a report that this PR is faster without this change on the dev-public channel.

Tilps (approved):

src/neural/cuda/fp16_kernels.cu
ankan-ban: correct. Removed the use_se template param.

meson.build
ankan-ban: that's strange. At my end I am not seeing any perf difference. I have reverted this part to avoid confusion/corner cases.

src/neural/cuda/fp16_kernels.cu
```diff
@@ -248,105 +248,97 @@ void OutputInputTransformKernel_fp16_shmem_board(
...
+    for (int x = 0; x < 8; x++) {
+      if (use_bias) boardRow[x] += b;
+      S += (float)boardRow[x];
     }
+    if (use_bias) copyAs<uint4>(&BOARD(y, 0), &boardRow);
   }
 
-  if (use_se) {
     __shared__ float shared_data[kMaxResBlockFusingSeKFp16Ampere];
-    float avg = S / 64;
```
Tilps, line 267: I think the line above is still indented when it should have shifted across with everything else.

ankan-ban: sorry about that. Fixed.

PR #1683, opened 2021-12-31
Add support for cuda graphs and async copy of policy output

ankan-ban:
- Async policy output allows running the policy output copy from GPU memory to system memory run in parallel with value and mlh head computation - which is about 2-3% of network execution time.
- Using CUDA graphs helps with reducing the overhead of kernel launches.
- The two optimizations combined is worth about 5% speedup in benchmark with 384fx30b network on A100 (more improvement in backendbench, but it doesn't translate to full benchmark speedup because multi_stream opt was already achieving parallelism by running multiple inferences in parallel).
- Here are some detailed performance nos on A100:
```
Backend Bench (multi_stream enabled):
=====================================
384fx30b:
baseline: Benchmark batch size 96 with inference average time 3.43217ms - throughput 27970.6 nps.
opt:      Benchmark batch size 96 with inference average time 3.16205ms - throughput 30360.1 nps.
baseline: Benchmark batch size 192 with inference average time 6.52237ms - throughput 29437.2 nps
opt:      Benchmark batch size 192 with inference average time 6.28403ms - throughput 30553.6 nps.

512fx40b:
baseline: Benchmark batch size 96 with inference average time 6.23764ms - throughput 15390.4 nps.
opt:      Benchmark batch size 96 with inference average time 5.94518ms - throughput 16147.5 nps.
baseline: Benchmark batch size 192 with inference average time 11.9312ms - throughput 16092.2 nps.
opt:      Benchmark batch size 192 with inference average time 11.5244ms - throughput 16660.3 nps.


Benchmark (with multi_stream enabled, default params, num_positions=1):
=======================================================================
384fx30b:
baseline: Nodes/second    : 63883
opt:      Nodes/second    : 66878

512fx40b:
baseline: Nodes/second    : 34284
opt:      Nodes/second    : 35525
```

### DIFF
--- a/src/neural/cuda/network_cuda.cc
+++ b/src/neural/cuda/network_cuda.cc
@@ -435,7 +435,7 @@ class CudaNetwork : public Network {
 
     DataType* tensor_mem[3];
     void* scratch_mem;
-    cudaStream_t stream;
+    cudaStream_t stream, stream_copy;
     cublasHandle_t cublas;
     if (multi_stream_) {
       // We use tensor and scratch memory from InputOutputs (so that multiple
@@ -443,11 +443,13 @@ class CudaNetwork : public Network {
       for (int i = 0; i < 3; i++) tensor_mem[i] = (DataType*)io->tensor_mem_[i];
       scratch_mem = io->scratch_mem_;
       stream = io->stream_;
+      stream_copy = io->stream_copy_;
       cublas = io->cublas_;
     } else {
       for (int i = 0; i < 3; i++) tensor_mem[i] = tensor_mem_[i];
       scratch_mem = scratch_mem_;
       stream = 0;           // default stream
+      stream_copy = 0;
       cublas = cublas_;
     }
 
@@ -530,11 +532,21 @@ class CudaNetwork : public Network {
       }
     }
 
+    if (stream_copy != stream) {
+      // signal policy_ready_event_ to indicate that the copy stream can start
+      // copying policy output
+      ReportCUDAErrors(cudaEventRecord(io->policy_ready_event_, stream));
+
+      // make the copy stream wait for the above event
+      ReportCUDAErrors(
+          cudaStreamWaitEvent(stream_copy, io->policy_ready_event_, 0));
+    }
+
     // Copy policy output from device memory to host memory.
     ReportCUDAErrors(cudaMemcpyAsync(
         io->op_policy_mem_, io->op_policy_mem_gpu_,
                         sizeof(float) * kNumOutputPolicy * batchSize,
-                        cudaMemcpyDeviceToHost, stream));
+                        cudaMemcpyDeviceToHost, stream_copy));
 
     // value head
     network_[l++]->Eval(batchSize, tensor_mem[0], tensor_mem[2], nullptr,
@@ -597,6 +609,7 @@ class CudaNetwork : public Network {
     }
 
     if (multi_stream_) {
+      ReportCUDAErrors(cudaStreamSynchronize(stream_copy));
       ReportCUDAErrors(cudaStreamSynchronize(stream));
     } else {
       ReportCUDAErrors(cudaDeviceSynchronize());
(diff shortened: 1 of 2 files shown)


src/neural/cuda/network_cuda.cc
```diff
@@ -435,165 +438,214 @@ class CudaNetwork : public Network {
...
     }
 
-    bool fp16 = std::is_same<half, DataType>::value;
-    if (fp16) {
-      expandPlanes_Fp16_NCHW((half*)(tensor_mem[0]), ipDataMasks,
+    if (!graph_created) {
+
+      if (use_cuda_graphs)
+        ReportCUDAErrors(cudaStreamBeginCapture(
+            stream, cudaStreamCaptureModeThreadLocal));
```
borg323, line 467: The second argument for `cudaStreamBeginCapture()` was added in cuda 10.1 (while we still use cuda 10.0 for the cudnn builds). Does it make sense to make this conditional? We can update the cuda version used for the builds, but I'm not confident there are no downsides for older gpu support.

src/neural/cuda/network_cuda.cc
```diff
@@ -435,165 +438,220 @@ class CudaNetwork : public Network {
...
+      cudaStreamCaptureMode captureMode = (cudaStreamCaptureMode)0;
+#if CUDART_VERSION >= 11000
+      // We need to use cudaStreamCaptureModeThreadLocal to make sure only the
+      // calls made by the current thread are captured. The cuda-build uses cudnn
+      // 11+, so it should work. This compile time check is only for fixing the
+      // cudnn build that uses old cuda runtime version.
+      captureMode = cudaStreamCaptureModeThreadLocal;
+#endif
+      if (use_cuda_graphs)
+        ReportCUDAErrors(cudaStreamBeginCapture(stream, captureMode));
```
borg323, line 473: This still won't work, and from the above comment it seems graphs can't work with cuda < 10.1. In that case just revert the last commit and we'll make cuda 10.1 the minimum supported version. If people complain we can disable the graph code for earlier versions.

dje-dev:
With Ceres I found I had to limit the number of distinct batch sizes created because each graph consume GPU memory. With T60, 39 different graph sizes are created, increasing the GPU memory usage to 4262mb from the baseline of 3251mb. Even with the reduced count, some users with lesser hardware reported errors and had to use an option to disable use of the feature.
Were these tests run on the SXM/PCI and 40mb/80mb variants of the A100?

ankan-ban:
> With Ceres I found I had to limit the number of distinct batch sizes created because each graph consume GPU memory. With T60, 39 different graph sizes are created, increasing the GPU memory usage to 4262mb from the baseline of 3251mb. Even with the reduced count, some users with lesser hardware reported errors and had to use an option to disable use of the feature. Were these tests run on the SXM/PCI and 40mb/80mb variants of the A100?

I didn't check the memory usage - but it's a good point. I will check at my end and try creating lesser no of distinct graphs (maybe also worth sharing graphs across streams). The tests were run on 40GB PCIe version of A100.

borg323:
I think this PR should be split into two, asynchronous policy copy should be easier to test and merge.

dje-dev:
Another thing to consider is if/when the backend is enhanced to get a legal moves mask then only the relevant values can be copied from the GPU (reducing bytes transferred by a factor of 50 or more) and then asynchronous copy doesn't save enough time to compensate for the added complexity. Ceres has a trivial kernel which applies the mask on the GPU to distill down to only (up to ) 96 legal moves which are then copied off.
```
__global__ void copyMaskedMovesKernel(half* inPolicies, short* inMovesIndices, float* outPoliciesMasked, int size) 
{
  int itemIndex = threadIdx.x + blockDim.x * blockIdx.x;

  int offsetPolicies = 1858 * itemIndex;
  int offsetMoves = 96 * itemIndex;
  int offsetPoliciesOut = 96 * itemIndex;

  if (itemIndex < size)
  {
    for (int i=0;i<96;i++)
    {
      outPoliciesMasked[offsetPoliciesOut + i] = inPolicies[offsetPolicies + inMovesIndices[offsetMoves + i]];
    }
  }

  }
```

PR #1685, opened 2022-01-16
fix cuda se block threading

borg323:


ankan-ban:
The fix looks good to me. Thanks!

ankan-ban (approved):

PR #1686, opened 2022-01-19
Alphazero TM revised

trre123:
One downside of current alphazero time management is that it normally never has the ability to use the full time available. For instance if you had a time control of 60s+1s and chose a percentage of 5%, then it would end every game with no less than 20 seconds on the clock, which is a significant amount of time to be left unused at that tc.

The goal with this pr was to make a version of a0 tm that increases the percentage of time available every move until it eventually reaches 100%. Also hoping that it would better scale from vstc to vltc, and thus be easier to tune.

The a0_revised params have been tuned at vstc and the test results indicate good scaling to ltc.

lc0 @ 6s+0.1s
```
   # PLAYER            :  RATING  ERROR  POINTS  PLAYED   (%)  CFS(%)    W     D    L  D(%)
   1 sf                :     0.0   ----  1313.5    2500    53      91  672  1283  545    51
   2 lc0-a0_revised    :    -9.0   13.2   609.0    1250    49      96  281   656  313    52
   3 lc0-legacy        :   -26.8   14.1   577.5    1250    46     ---  264   627  359    50
```

lc0 @ 120s+2s
```
   # PLAYER            :  RATING  ERROR  POINTS  PLAYED   (%)  CFS(%)    W    D    L  D(%)
   1 lc0-a0_revised    :    12.3   15.4   103.5     200    52      94   13  181    6    91
   2 lc0-legacy        :     0.0   ----    96.5     200    48     ---    6  181   13    91
```

Note also that I first tried a log scale for increasing the a0 pct to 100, but it didnt perform as well as the flat increase per ply that was settled on here.

### DIFF
--- a/src/mcts/stoppers/alphazero.cc
+++ b/src/mcts/stoppers/alphazero.cc
@@ -35,20 +35,24 @@ class AlphazeroTimeManager : public TimeManager {
  public:
   AlphazeroTimeManager(int64_t move_overhead, const OptionsDict& params)
       : move_overhead_(move_overhead),
-        timepct_(params.GetOrDefault<float>("time-pct", 4.23f)),
-        maxtimeply_(params.GetOrDefault<float>("max-time-ply", 235.0f)) {
-    if (timepct_ <= 0.0f || timepct_ > 100.0f)
-      throw Exception("time-pct value to be in range [0.0, 100.0]");
-    if (maxtimeply_ < 1.0f || maxtimeply_ > 1000.0f)
-      throw Exception("max-time-ply value to be in range [1.0, 1000.0]");
+        minpct_(params.GetOrDefault<float>("min-pct", 2.52f)),
+        timemult_(params.GetOrDefault<float>("time-mult", 2.17f)),
+        plymult_(params.GetOrDefault<float>("ply-mult", 2.72f)) {
+    if (minpct_ <= 0.0f || minpct_ > 10.0f)
+      throw Exception("min-pct value to be in range [0.0, 10.0]");
+    if (timemult_ <= 0.0f || timemult_ > 10.0f)
+      throw Exception("time-mult value to be in range [0.0, 10.0]");
+    if (plymult_ <= 0.0f || plymult_ > 10.0f)
+      throw Exception("ply-mult value to be in range [0.0, 10.0]");
   }
   std::unique_ptr<SearchStopper> GetStopper(const GoParams& params,
                                             const NodeTree& tree) override;
 
  private:
   const int64_t move_overhead_;
-  const float timepct_;
-  const float maxtimeply_;
+  const float minpct_;
+  const float timemult_;
+  const float plymult_;
 };
 
 std::unique_ptr<SearchStopper> AlphazeroTimeManager::GetStopper(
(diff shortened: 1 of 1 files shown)


zz4032:
Has `lc0-legacy` been tuned in a similar way as the PR version?

trre123:
> 

No since I assume legacy tm has already been tuned?

trre123:
> Has `lc0-legacy` been tuned in a similar way as the PR version?

No since I assume legacy tm has already been tuned?

trre123:
> Has `lc0-legacy` been tuned in a similar way as the PR version?

@zz4032 No since I assume legacy tm has already been tuned?

zz4032:
Ok, for some reason I thought `lc0-legacy` above is the default alphazero TM. I think a match vs. tuned alphazero TM would allow a better comparison since I guess this PR is supposed to replace the current alphazero TM.
How would this PR work with no increment TC?

trre123:
> Ok, for some reason I thought `lc0-legacy` above is the default alphazero TM. I think a match vs. tuned alphazero TM would allow a better comparison since I guess this PR is supposed to replace the current alphazero TM. How would this PR work with no increment TC?

@zz4032 The reason i chose to run this PR vs legacy tm is because we know the legacy tm params scale well from vstc to vltc, whereas i believe kiudee attempted an a0 tm tune at vstc but it was shown to scale very poorly to ltc, hence he retuned at much longer tc to come up with the current values. 

How about I run an additional gauntlet @ vstc with no increment? It can be sf vs a0_revised & a0_current (using kiudee tune). I believe the kiudee a0 tune params are alphazero-time-pct=23.0629, smart-pruning-factor=3.6141 ?

zz4032:
I think that's a good idea and those parameter values should be close to optimal. But wouldn't the PR lose on time without increment after ply 235?

trre123:
@zz4032 Yea it definitely would, though it might not matter much in terms of elo because very few games should make it to that ply count. I will test and see if a special case is needed where I cap max % for no-increment matches.

trre123:
@zz4032 

Reworked the formula and tuned it vs sf @ vstc. Ran tests vs original a0 tm using multiple types of time control (no-increment, increment-only, vstc @ standard ratio, ltc @ sufi ratio). The no-increment match was stopped early because the original a0 tm did so bad.

**6s+0.1s**
```
   # PLAYER        :  RATING  ERROR  POINTS  PLAYED   (%)  CFS(%)    W    D    L  D(%)
   1 sf            :     0.0   ----   794.0    1500    53      86  382  824  294    55
   2 lc0-pr1686    :    -9.4   16.8   365.0     750    49      97  160  410  180    55
   3 lc0-master    :   -31.9   16.8   341.0     750    45     ---  134  414  202    55
```

**0s+0.2s**
```
   # PLAYER        :  RATING  ERROR  POINTS  PLAYED   (%)  CFS(%)    W    D    L  D(%)
   1 lc0-pr1686    :    18.4   15.3   263.0     500    53      99   81  364   55    73
   2 lc0-master    :     0.0   ----   237.0     500    47     ---   55  364   81    73
```

**12s+0s**
```
   # PLAYER        :  RATING  ERROR  POINTS  PLAYED   (%)  CFS(%)    W    D    L  D(%)
   1 lc0-pr1686    :   163.6   33.9   143.5     200    72     100   94   99    7    50
   2 lc0-master    :     0.0   ----    56.5     200    28     ---    7   99   94    50
```

**144s+0.2s**
```
   # PLAYER        :  RATING  ERROR  POINTS  PLAYED   (%)  CFS(%)    W    D    L  D(%)
   1 lc0-pr1686    :    15.5   15.1   130.5     250    52      98   21  219   10    88
   2 lc0-master    :     0.0   ----   119.5     250    48     ---   10  219   21    88
```

Full test details in discord - https://discord.com/channels/425419482568196106/530486338236055583/936026184821637231

trre123:
Despite the good results so far, Im marking this as draft until I can do more testing/tuning. The `min-pct` is too low IMO, and it could be because I use a 3-ply opening book which may require less thinking time on the first move than a longer opening book.

trre123:
After some more testing, learned that tc's with 0 increment dont want the % to grow at all, whereas tc's with an increment >= total time want the % to grow to 100 right away. So I simplified the formula further and I think it scales well from one extreme to the other. I thought about using a multiplier for ```min-pct``` based on timeratio also, but then we wouldnt be able to use the original form of a0 tm if desired. 

One downside of having gameply in the calculation is that using a FEN might allow for inaccurate calculation since it can have a long opening while being considered 0 ply. Guess there is no easy way around that though.

Updated params with values from new tune at tc below spf threshold, and test results show the values scale well above the threshold too. To me this indicates no tuning of spf is needed. Results also show a better scaling to different tc ratios.

**18s+0.3s**
```
   # PLAYER        :  RATING  ERROR  POINTS  PLAYED   (%)  CFS(%)    W    D    L  D(%)
   1 lc0-pr1686    :    18.3   12.2   263.0     500    53     100   56  414   30    83
   2 lc0-master    :     0.0   ----   237.0     500    47     ---   30  414   56    83
```
**60s+0s**
```
   # PLAYER        :  RATING  ERROR  POINTS  PLAYED   (%)  CFS(%)    W    D    L  D(%)
   1 lc0-pr1686    :    49.6   20.1   114.0     200    57     100   31  166    3    83
   2 lc0-master    :     0.0   ----    86.0     200    43     ---    3  166   31    83
```

PR #1688, opened 2022-01-30
Attention policy

ankan-ban:
Work in progress changes for Attention policy head in the cuda backend.
 * Add support for attention weights in LegacyWeights class
 * Hook up cuda backend to allocate GPU memory for various weights used by the attention head, and copy weights to GPU memory.
 * First WIP implementation of the NN evaluation
   - Square embedding
   - Encoder blocks with MHA/FFN and Layer normalizations.
   - Policy self-attention.

Remaining items: 
- [X] promotion_logits 
- [X] ApplyAttentionPolicyMap
- [x]  FP32 support
- [X] Testing and debugging


src/neural/cuda/common_kernels.cu
```diff
@@ -566,6 +576,190 @@ void OutputInputTransform(int N, int C, int se_K, T* output, const T* input,
...
+  if ((c & 0x1F) == 0) atomicAdd(&sum, s);
+
+  __syncthreads();
+
+  float mean = sum / C;
+  float d = x - mean;
+  float d_sq = d * d;
+
+  s = warpReduce(d_sq);
+  if ((c & 0x1F) == 0) atomicAdd(&sum, s);
```
Tilps, line 657: I think you need a reset of sum and a syncThreads after calculating the mean. Otherwise sum / C will be mean+variance. If mean and variance are similar scales you could just subtract the mean off the total to recover the variance - but I don't know how safe that is numeric stability wise in our use cases.

src/neural/cuda/network_cuda.cc
```diff
@@ -490,7 +516,17 @@ class CudaNetwork : public Network {
     }
 
     // Policy head.
-    if (conv_policy_) {
+    if (attn_policy_) {
+      network_[l++]->Eval(batchSize, tensor_mem[0], tensor_mem[2],
```
Tilps, line 520: The attention policy head assumes these inputs of tensor_mem are at least as big as scratch_mem.  But the logic only currently guarantees that if res block fusing is enabled.  That check should be changed to also include attn_policy_.  Not sure but maybe also need to ensure that size is at least twice as big as the outputs of this layer...

src/neural/cuda/common_kernels.cu
ankan-ban: yeah, very stupid of me to miss this! Better to keep two variables than adding more sync's.

src/neural/cuda/network_cuda.cc
```diff
@@ -490,7 +516,17 @@ class CudaNetwork : public Network {
...
+                          tensor_mem[1], scratch_mem, scratch_size_, nullptr,
+                          cublas,
+                          stream);  // Entire Attention policy head except for the policy map
+
+      network_[l++]->Eval(batchSize, (DataType*)opPol, tensor_mem[0], nullptr,
```
Tilps, line 525: This calls policy map with direct output to opPol on fp16 - which is not what we do in the other policy output branches.
I replaced this with
      if (fp16) {
        network_[l++]->Eval(batchSize, tensor_mem[1], tensor_mem[0], nullptr,
                            scratch_mem, scratch_size_, nullptr, cublas,
                            stream);  // policy map layer
        copyTypeConverted(opPol, (half*)(tensor_mem[1]),
                          batchSize * kNumOutputPolicy,
                          stream);  // POLICY output
      } else {
        network_[l++]->Eval(batchSize, (DataType*)opPol, tensor_mem[0], nullptr,
                            scratch_mem, scratch_size_, nullptr, cublas,
                            stream);  // policy map layer  // POLICY output
      }
and got 'better' output.

src/neural/cuda/layers.cc
```diff
@@ -1225,6 +1141,424 @@ ResidualBlock<DataType>::~ResidualBlock() {
...
+
+template <typename DataType>
+AttentionPolicyHead<DataType>::AttentionPolicyHead(BaseLayer<DataType>* ip,
+                                                   const LegacyWeights& weights,
+                                                   void* scratch)
+    : BaseLayer<DataType>(ip->GetC(), 8, 8, ip) {
+  embedding_op_size_ = weights.ip_pol_b.size();
+  wq_op_size_ = weights.ip2_pol_b.size();
+  wk_op_size_ = weights.ip3_pol_b.size();
+  ppo_op_size_ = weights.ip4_pol_b.size();
```
Tilps, line 1167: This is unused - but also ppo has no bias so the unused value calculated here is technically wrong.

src/neural/cuda/layers.cc
```diff
@@ -1225,6 +1141,424 @@ ResidualBlock<DataType>::~ResidualBlock() {
...
+                               64 * policy_d_model_,     /*strideB*/
+                               &beta, 
+                               output /*C*/,  // output (policy_attn_logits)
+                               CUDA_R_16F, 
+                               64 /*LDC*/, 
+                               64 * 64 /*strideC*/,
+                               N, CUDA_R_16F, CUBLAS_GEMM_DEFAULT);
+  }
+
+  // Compute promotion_logits in a single kernel (and put the result just after policy_attn_logits to get concat for free)
```
Tilps, line 1511: This logic seems wrong - you don't get the concat for free since policy map is expecting an input in format N *(64*64+8*24), not N * 64* 64 + N*8*24)
I think you need to set strideC to 64*64+8*24 and change the ComputePromotionLogits implementation to interleave the output as appropriate.
As it stands this will work for N=1 only I think...

Tilps:
[patch.txt](https://github.com/LeelaChessZero/lc0/files/8103552/patch.txt)
Here is some additional patches to get rid of some of the hardcoded size assumptions in the current code.

Tilps:
That patch isn't very clean since its off of something not quite the same as this branch.  Maybe a better diff to read through...
https://github.com/ankan-ban/lc0/compare/attention-policy..Tilps:attention - note my branch has some debugging hacks and stuff.

src/neural/cuda/layers.cc
```diff
@@ -1225,6 +1141,425 @@ ResidualBlock<DataType>::~ResidualBlock() {
...
+  half* scratch4 = output + scratch_size / (2 * sizeof(half));
+
+  // half alpha = float2half_rn(1.0f), beta = float2half_rn(0.0f);
+  const __half_raw one_h{0x3C00};
+  const __half_raw zero_h{0};
+  half alpha = one_h;
+  half beta = zero_h;
+
+
+  fp16NCHWtoNHWC(scratch1, input, N, 64, N, 64, 8, 8);
```
Tilps, line 1275: The 64's in this line should be input_->GetC()

src/neural/cuda/layers.cc
```diff
@@ -1225,6 +1141,425 @@ ResidualBlock<DataType>::~ResidualBlock() {
...
+      cudaMemcpy(scratch, &cpu_src[0], cpu_src.size() * sizeof(float), cudaMemcpyHostToDevice));
+  copyTypeConverted((DataType*)(*gpu_dest), (float*)scratch,
+                    (int)cpu_src.size(), 0);
+}
+
+template <typename DataType>
+AttentionPolicyHead<DataType>::AttentionPolicyHead(BaseLayer<DataType>* ip,
+                                                   const LegacyWeights& weights,
+                                                   void* scratch)
+    : BaseLayer<DataType>(ip->GetC(), 8, 8, ip) {
```
Tilps, line 1163: ip->GetC() isn't correct here ... its a bit ugly - but technically 67 would work?

Tilps: maybe better to set H and W to 1 and instead set C to 64*64+24*8

src/neural/cuda/layers.cc
```diff
@@ -1225,6 +1187,431 @@ ResidualBlock<DataType>::~ResidualBlock() {
...
+AttentionPolicyHead<DataType>::EncoderWeights::EncoderWeights(
+    const LegacyWeights::EncoderLayer& cpu_weights, void* scratch) {
+  mha_q_size_ = cpu_weights.mha.q_b.size();
+  mha_k_size_ = cpu_weights.mha.k_b.size();
+  mha_v_size_ = cpu_weights.mha.v_b.size();
+  mha_dense_size_ = cpu_weights.mha.dense_b.size();
+  ffn_dense1_size_ = cpu_weights.ffn.dense1_b.size();
+  ffn_dense2_size_ = cpu_weights.ffn.dense2_b.size();
+
+  // debug!
```
masterkni6, line 1244: still have these debug statements

masterkni6, line 1244: still have these debug statements

src/neural/cuda/layers.cc
```diff
@@ -1225,6 +1187,431 @@ ResidualBlock<DataType>::~ResidualBlock() {
...
+  printf("\nsize of ln1 betas/gammas: %d, %d\n",
+         (int)cpu_weights.ln1_betas.size(), (int)cpu_weights.ln1_gammas.size());
+  printf("\nsize of ln2 betas/gammas: %d, %d\n",
+         (int)cpu_weights.ln2_betas.size(), (int)cpu_weights.ln2_gammas.size());
+  printf("\nsize of weight ffn.dense1_b/w: %d, %d\n",
+         (int)cpu_weights.ffn.dense1_b.size(),
+         (int)cpu_weights.ffn.dense1_w.size());
+  printf("\nsize of weight ffn.dense2_b/w: %d, %d\n",
+         (int)cpu_weights.ffn.dense2_b.size(),
+         (int)cpu_weights.ffn.dense2_w.size());
```
masterkni6, line 1263: these debugs are still here (third attempt to highlight the code correctly)

ankan-ban: thanks. Removed them.

src/neural/cuda/layers.cc
ankan-ban: fixed, thanks.

src/neural/cuda/layers.cc
ankan-ban: thanks. Fixed.

src/neural/cuda/layers.cc
ankan-ban: updated to new net.proto, and removed them.

src/neural/cuda/network_cuda.cc
ankan-ban: added some checks when allocating these allocations. Hopefully should work for most networks otherwise the backend will exit with error.

src/neural/cuda/network_cuda.cc
```diff
@@ -416,8 +416,15 @@ class CudaNetwork : public Network {
       maxSize = std::max(maxSize, layer->GetOutputSize(max_batch_size_));
     }
 
-    if (use_res_block_winograd_fuse_opt_ && scratch_size_ > maxSize)
+
+    if ((attn_policy_ || use_res_block_winograd_fuse_opt_) && (scratch_size_ > maxSize)) {
+      // When attention is used, we split the tensor memory allocations into two
+      // halves and use them as scratch space to hold intermediate data
+      if (attn_policy_ && (scratch_size_ < maxSize * 2)) {
```
Tilps, line 423: another thing to consider is that attn_policy head isn't a single input output layer where intermediate steps have same size or less compared to input or output.  Especially with encoder blocks it is normal for inner steps to have larger size.  We may need to override GetMaxSize for attention policy head to determine the largest intermediate layer rather than just calculating the output size.

src/neural/cuda/layers.cc
```diff
@@ -1225,6 +1187,409 @@ ResidualBlock<DataType>::~ResidualBlock() {
...
+unsigned short float_to_half(const float x) {
+  const unsigned int b = (*(unsigned int*)&x) + 0x00001000;
+  const unsigned int e = (b & 0x7F800000) >> 23;
+  const unsigned int m = b & 0x007FFFFF;
+  return (b & 0x80000000) >> 16 |
+         (e > 112) * ((((e - 112) << 10) & 0x7C00) | m >> 13) |
+         ((e < 113) & (e > 101)) *
+             ((((0x007FF000 + m) >> (125 - e)) + 1) >> 1) |
+         (e > 143) * 0x7FFF;
+}
```
borg323, line 1280: Probably not important, but this gets round-to-even and infinity wrong. In dx12 have x86 code with a bit accurate fallback that can be moved to utils.

src/neural/cuda/layers.cc
```diff
@@ -1225,6 +1189,408 @@ ResidualBlock<DataType>::~ResidualBlock() {
...
+    const int num_outputs = policy_d_model_;
+    const int batch = N * 64;
+    cublasXgemm(cublas, CUBLAS_OP_T, CUBLAS_OP_N, num_outputs, batch,
+                num_inputs, 1.0f, ip3_pol_w_, num_inputs, scratch0, num_inputs,
+                0.0f, scratch2, num_outputs);
+    addVectors(scratch2, ip3_pol_b_, scratch2, num_outputs * batch, num_outputs,
+               num_outputs * batch, NONE, stream);
+  }
+
+  // dk = tf.math.sqrt(tf.cast(tf.shape(keys)[-1], self.model_dtype))  #
```
Tilps, line 1531: I think auto format kind of messed up this comment a bit.

Tilps (approved):

src/neural/cuda/network_cuda.cc
ankan-ban: hopefully fixed with latest commit. Can you please take a look.

src/neural/cuda/layers.cc
ankan-ban: fixed

src/neural/cuda/network_cuda.cc
```diff
@@ -511,10 +541,14 @@ class CudaNetwork : public Network {
 
     // Policy head.
     if (attn_policy_) {
+#if 0
```
Tilps, line 544: Remove this debugging?

src/neural/cuda/network_cuda.cc
```diff
@@ -48,6 +48,35 @@ using namespace cudnn_backend;
...
+  }
+
+  const size_t encoder_heads = weights.pol_encoder_head_count;
+
+  size_t size = N * 64 *
+                std::max(std::max(embedding_op_size, encoder_dff),
+                         std::max(policy_d_model, encoder_d_model));
+
+  // size of matmul_qk matrix = encoder_heads_ * Batch * 64 * 64
+  const size_t matmul_qk_size = encoder_heads * N * 64 * 64;
```
Tilps, line 75: The final matmul isn't covered here, but it is covered by the layer itself - so I think that should be fine?

ankan-ban:
I think the change is ready for review/merge. This is just the first implementation without too much thought on performance.

Right now for 256x20 networks there is a significant perf hit due to the policy head - e.g: with 3 encoder layer net I got from Arcturai, on RTX 3080 I see, backendbench nps for batch size of 256:
```
Without evaluating policy head : 35666 nps
With attention policy head     : 26293 nps
```
I think it should be possible to cut down the cost of the policy head to about half with optimizations. 
Some of the things I plan to try out:
1. Merge MHA q,k,v matrix multiplications into a single gemm (batched). Basically run them in parallel in the same kernel launch.
2. Get rid of the loop on encoder_heads_ - merging them into a single gemm/kernel. Again the idea is to run those operations in parallel in the same kernel launch.
3. Get rid of the bias addition/activation kernels by trying to fuse them with previous gemm (plan to switch to cutlass gemm instead of cublas).
4. Try to fuse the softmax and layer norm operations too with the previous layers (gemm, followed by bias add followed by these ops all in single fused kernel).

I will be trying out these optimizations in separate PRs.

src/neural/cuda/network_cuda.cc
```diff
@@ -48,6 +48,35 @@ using namespace cudnn_backend;
 template <typename DataType>
 class CudaNetwork;
 
+static size_t getMaxAttentionHeadSize(const LegacyWeights& weights, int N) {
```
Tilps, line 51: don't need to solve right now - but I could imagine that the 'cudnn' implementation of attention head reuses the attention head block as it stands for cuda - and so this code may need to be moved to a shared location.

src/neural/cuda/network_cuda.cc
ankan-ban: you are right. I think it's better to add it to avoid scratch size calculation go wrong in some corner cases. Updated.

src/neural/cuda/network_cuda.cc
ankan-ban: yeah.. my bad.

src/neural/cuda/network_cuda.cc
ankan-ban: yes, I think it can same code can be supported in cudnn backend with minimal modifications. I will handle it in separate PR. (are people still using cudnn backend over cuda?)

Tilps: there are old gpu users where cuda is less of an option due to memory usage of the winograd transformations ... and maybe some 1650 type gpus where cudnn is faster than cuda? I don't keep track of all the details...

borg323: The 1650 may be fine after the workaround in #1675 (approved but not yet merged to avoid conflicts with this one), but all other GTX cards seem to benefit from cudnn.

PR #1694, opened 2022-02-09
Calculate original evalulation if it is not available in cache

nokturnusmf:
Currently, if the root position is not found in the cache, the original evaluation is not stored into the training data. However, it can easily be calculated.

Due to numerical accuracy, the values are not completely identical. Typically Q is incorrect by at most 1e-5, and D by 0.001.

M could also be recalculated like this, but I found drift could be much more severe. Since this is a fallback and the original M value doesn't seem to be used anywhere anyway, I decided to leave it as NaN.

### DIFF
--- a/src/trainingdata/trainingdata.cc
+++ b/src/trainingdata/trainingdata.cc
@@ -74,6 +74,41 @@ void DriftCorrect(float* q, float* d) {
     }
   }
 }
+
+Eval CalculateOriginalEval(const Node* node) {
+  Eval result;
+
+  int n = node->GetN();
+  result.wl = -node->GetWL() * n;
+  result.d = node->GetD() * n;
+
+  for (auto& edge : node->Edges()) {
+    int n = edge.GetN();
+    result.wl -= edge.GetWL(0) * n;
+    result.d -= edge.GetD(0) * n;
+  }
+
+  // Checks to compensate for drift.
+  if (result.wl > 1.0f) {
+    result.wl = 1.0f;
+  } else if (result.wl < -1.0f) {
+    result.wl = -1.0f;
+  }
+
+  if (result.d > 1.0f) {
+    result.d = 1.0f;
+  } else if (result.d < 0.0f) {
+    result.d = 0.0f;
+  }
+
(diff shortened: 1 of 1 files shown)


borg323:
Can you give details about the drift in M you observed? Ideally, a relative error plot under training conditions would be very helpful (also for Q and D if you do it).

nokturnusmf:
I generated 100 games using the latest T77 and all training parameters (with the exception of opening book). For each position with a cached evaluation available, I saved the original and recalculated Q, D, and M values. In total there were 10845 such positions. Results as CSV [here](https://github.com/LeelaChessZero/lc0/files/8049262/Results.csv). Columns are True Q, Calculated Q, True D, Calculated D, True M, Calculated M.


#### Absolute Error in Q:
chart
Maximum was 0.000124753.

#### Absolute Error in D:
chart
Maximum was 0.00135136.

#### Absolute Error in M:
chart
Maximum was 0.48584 (although this corresponds to a relative error of around 0.3%).
The maximum relative error was 1.36719%.

#### Worst Case
As a worst case, I generated another 10 games with a fixed 4k nodes per move, for a total of 1643 positions. The maximum errors were:
Q: 0.000712335.
D: 0.00850949.
M: 2.84717.

#### Comparison of Errors in Q and D
Finally, I also generated a quick comparison between errors in Q and D. There doesn't appear to be any systematic bias or correlation.
chart

borg323:
Thanks for the data. I made some plots of absolute error as a function of the Q/D/M value:
Q
Q

D
D

M
M

Those look acceptable to me but I would like a second opinion.

borg323:
I made a new data set of 1068 actual missing evaluations (from 1715 T75 selfplay games): [data.csv](https://github.com/LeelaChessZero/lc0/files/8087366/data.csv)

Here are the same plots of absolute error as a function of the Q/D/M value:

Q
q

D
d
 
M
m

borg323:
@nokturnusmf can you add M estimation to the code?
Also I think we should add a flag to indicate when the values are estimated. The `dummy` field in the training data can be used for that, rename it to something like `info2` and reserve a bit (say bit 7) for this.

PR #1695, opened 2022-02-12
MishNet backend support for cuda

Tilps:
Note that since I don't actually have a flag in the proto file yet, this requires changing a constant in network_cuda.cc and recompiling to switch between.

Cudnn backend not supported properly yet...

### DIFF
--- a/src/neural/cuda/common_kernels.cu
+++ b/src/neural/cuda/common_kernels.cu
@@ -71,7 +71,6 @@ __global__ void addVectors_kernel(T* c, T* a, T* b, int size, int asize,
       } else {
         cVal = cVal - 2 * __fdividef(cVal, n + 2);
       }
-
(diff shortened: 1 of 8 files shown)


Tilps:
Cudnn should work now as well. Masterkni6 performed some 1node tests which suggest its in good shape.

Tilps:
Needs lczero-common PR20. Which adds the following to NetworkFormat.

  // Activation used everywhere except head outputs or otherwise specified.
  enum DefaultActivation {
    DEFAULT_ACTIVATION_RELU = 0;
    DEFAULT_ACTIVATION_MISH = 1;
  }
  optional DefaultActivation default_activation = 7;

Tilps:
Obsolete by PR1699 which handles the merge better.

PR #1698, opened 2022-02-21
save trainingdata from undecided games

borg323:
For game result the final evaluation of the game is used - this means the rescorer will delete them so no overall change to training. Currently this is only for games exceeding the 450 ply limit, and we already have a flag to identify those if we want to change behavior at some point, e.g. allow the ones that end in a TB position.

### DIFF
--- a/src/selfplay/tournament.cc
+++ b/src/selfplay/tournament.cc
@@ -352,41 +352,39 @@ void SelfPlayTournament::PlayOneGame(int game_number) {
   auto player2_threads = player_options_[1][color_idx[1]].Get<int>(kThreadsId);
   game.Play(player1_threads, player2_threads, kTraining, syzygy_tb_.get(),
             enable_resign);
-  
-  // If game was aborted, it's still undecided.
-  if (game.GetGameResult() != GameResult::UNDECIDED) {
-    // Game callback.
-    GameInfo game_info;
-    game_info.game_result = game.GetGameResult();
-    game_info.is_black = player1_black;
-    game_info.game_id = game_number;
-    game_info.initial_fen = opening.start_fen;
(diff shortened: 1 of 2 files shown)


src/trainingdata/trainingdata.cc
```diff
@@ -105,6 +105,11 @@ void V6TrainingDataArray::Write(TrainingDataWriter* writer, GameResult result,
     if (adjudicated && result == GameResult::UNDECIDED) {
       chunk.invariance_info |= 1u << 4;  // Max game length exceeded.
     }
+    if (result == GameResult::UNDECIDED) {
+      // If the game is undecided use the final evaluation as result.
+      chunk.result_q = training_data_.back().best_q;
```
Tilps, line 110: This needs to alternate?

borg323: Indeed it does

src/selfplay/tournament.cc
```diff
@@ -352,41 +352,39 @@ void SelfPlayTournament::PlayOneGame(int game_number) {
   auto player2_threads = player_options_[1][color_idx[1]].Get<int>(kThreadsId);
   game.Play(player1_threads, player2_threads, kTraining, syzygy_tb_.get(),
             enable_resign);
-  
-  // If game was aborted, it's still undecided.
-  if (game.GetGameResult() != GameResult::UNDECIDED) {
```
Tilps, line 357: Is there any potential we should handle a case where the game is 'actually' aborted like the old comment?

borg323: Something like `if (!abort_ || game.GetGameResult() != GameResult::UNDECIDED) {`?
Or just `if (!abort_) {`?

PR #1703, opened 2022-03-01
hide eigen warnings

borg323:


mooskagh (approved):
Oh, I was missing that `as_system()` in so many personal projects..

PR #1705, opened 2022-03-03
Optimizations for attention policy head

ankan-ban:
1. Do keys, queries, values matrix multiplications in parallel in the encoder layers (as batched GEMM).
2. Do keys, queries matrix multiplication for policy head in parallel.
3. Add kernel to do batched bias addition. Optimize it to reach close to full memory bandwidth utilization (AddVectors kernel was not written with performance in mind).
4. Optimize LayerNorm and Softmax kernels to reach close to full memory bandwidth utilization.

About 7% improvement in total end to end network evaluation time on RTX 3080 for the 20b Attention head network I got from Arcturai (20b_43-swa-2000000.pb.gz):
Before: Benchmark batch size 256 with inference average time 9.68394ms - throughput 26435.5 nps.
After: Benchmark batch size 256 with inference average time 9.07259ms - throughput 28216.8 nps.

### DIFF
--- a/src/neural/cuda/fp16_kernels.cu
+++ b/src/neural/cuda/fp16_kernels.cu
@@ -443,7 +443,7 @@ void OutputInputTransform(int N, int C, int se_K, T* output, const T* input,
           72 * C * sizeof(half));
       OutputInputTransformKernel_fp16_shmem_board<activation, use_bias,
                                                   use_skip>
-          <<<N, C, kMaxResBlockFusingSeFp16AmpereSmem, stream>>>(
+          <<<N, C, 72 * C * sizeof(half), stream>>>(
               N, C, se_K, (half*)output, (const half*)input, (half*)skip,
               (half*)bias, (half*)w1, (half*)b1, (half*)w2, (half*)b2);
     } else {
--- a/src/neural/cuda/layers.cc
+++ b/src/neural/cuda/layers.cc
@@ -1349,6 +1349,27 @@ AttentionPolicyHead<DataType>::AttentionPolicyHead(BaseLayer<DataType>* ip,
   allocAndUpload<DataType>(&ip3_pol_w_, weights.ip3_pol_w, scratch);
   allocAndUpload<DataType>(&ip3_pol_b_, weights.ip3_pol_b, scratch);
 
+  // big allocation to hold wq and wk weights one after the other
+  {
+    size_t elements = weights.ip2_pol_w.size();
+    assert(elements == weights.ip3_pol_w.size());
+
+    size_t size = elements * sizeof(DataType) * 2;
+    ReportCUDAErrors(cudaMalloc(&wqk_w_, size));
+    ReportCUDAErrors(
+        cudaMemcpy(wqk_w_, ip2_pol_w_, size / 2, cudaMemcpyDeviceToDevice));
+    ReportCUDAErrors(cudaMemcpy(wqk_w_ + elements, ip3_pol_w_, size / 2,
+                                cudaMemcpyDeviceToDevice));
+
+    elements = weights.ip2_pol_b.size();
+    size = elements * sizeof(DataType) * 2;
+    ReportCUDAErrors(cudaMalloc(&wqk_b_, size));
+    ReportCUDAErrors(
+        cudaMemcpy(wqk_b_, ip2_pol_b_, size / 2, cudaMemcpyDeviceToDevice));
+    ReportCUDAErrors(cudaMemcpy(wqk_b_ + elements, ip3_pol_b_, size / 2,
+                                cudaMemcpyDeviceToDevice));
+  }
+
   allocAndUpload<DataType>(&ip4_pol_w_, weights.ip4_pol_w, scratch);
 
   for (const auto& enc : weights.pol_encoder) {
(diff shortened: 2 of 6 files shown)


src/neural/cuda/fp16_kernels.cu
```diff
@@ -443,7 +443,7 @@ void OutputInputTransform(int N, int C, int se_K, T* output, const T* input,
           72 * C * sizeof(half));
       OutputInputTransformKernel_fp16_shmem_board<activation, use_bias,
                                                   use_skip>
-          <<<N, C, kMaxResBlockFusingSeFp16AmpereSmem, stream>>>(
+          <<<N, C, 72 * C * sizeof(half), stream>>>(
```
Tilps, line 446: Stray diff?

src/neural/cuda/network_cuda.cc
```diff
@@ -76,7 +76,10 @@ static size_t getMaxAttentionHeadSize(const LegacyWeights& weights, int N) {
   const size_t output_size = N * (64 * 64 + 8 * 24);
 
   size = std::max(size, std::max(matmul_qk_size, output_size));
-  return size;
+
+  // We process the k, q, and v matrix multiplications in parallel using single allocation
+  // so need 3x of this.
+  return size * 3;
```
Tilps, line 82: This seems pessimistic - shouldn't you multiply encoder_d_model by 3 instead and policy_d_model by 2?

src/neural/cuda/network_cuda.cc
```diff
@@ -280,10 +283,9 @@ class CudaNetwork : public Network {
     scratch_size_ = std::max(scratch_size_, 2 * transformed_tensor_size);
 
     // Attention policy head may need more memory
-    // (We also split the allocations into two parts, so need 2x)
     const size_t attentionSize =
         getMaxAttentionHeadSize(weights, max_batch_size_);
-    scratch_size_ = std::max(scratch_size_, 2 * attentionSize);
+    scratch_size_ = std::max(scratch_size_, attentionSize);
```
Tilps, line 288: hrmm not sure about this. Allocations are still split?

src/neural/cuda/fp16_kernels.cu
ankan-ban: No, it's actually a fix that I forgot to include with my last change (otherwise nets bigger than 384 filters and smaller than 512 filters would fail here).

src/neural/cuda/network_cuda.cc
ankan-ban: yeah, will make it more accurate.

src/neural/cuda/network_cuda.cc
ankan-ban: I added "size * 3" in getMaxAttentionHeadSize() which should take care of this part (It's not very obvious, but the code doesn't actually use second part of the allocation when using them for batched gemms)

Tilps: If getMaxAttentionHeadSize uses multipliers for each phase as appropriate - depending on whether they use the bottom half only or top as well as well as the actual size after batching - then yes it makes sense to drop the 2 here.

ankan-ban: I have simplified intermediate memory allocations a bit (also reduced from 5 scratch allocations to 4). Hopefully it's accurate now.

src/neural/cuda/network_cuda.cc
```diff
@@ -68,14 +68,17 @@ static size_t getMaxAttentionHeadSize(const LegacyWeights& weights, int N) {
...
   size_t size = N * 64 *
-                std::max(std::max(embedding_op_size, encoder_dff),
-                         std::max(policy_d_model, encoder_d_model));
+      std::max(std::max(embedding_op_size, encoder_dff), policy_d_model);
 
   // size of matmul_qk matrix = encoder_heads_ * Batch * 64 * 64
   const size_t matmul_qk_size = encoder_heads * N * 64 * 64;
   const size_t output_size = N * (64 * 64 + 8 * 24);
-
   size = std::max(size, std::max(matmul_qk_size, output_size));
```
Tilps, line 76: output_size goes in a single allocation right? - so since we're going to multiply by 2 down below, maybe it could be halved? (Or maybe just don't bother :P)

src/neural/cuda/layers.cc
```diff
@@ -1677,7 +1691,7 @@ void AttentionPolicyHead<DataType>::Eval(
   DataType* promotion_logits = output + 64 * 64;
 
   ComputePromotionLogits<DataType>(N, policy_d_model_, promotion_logits,
-                                   scratch2, ip4_pol_w_, output, stream);
+                                   scratch0, ip4_pol_w_, output, stream);
```
Tilps, line 1694: There was a bug report that promotions are broken with this PR.  Looking at the code I think that scratch0 here should actually be wk.

Tilps: Kovax tried out this fix and it seemed to work.

ankan-ban: sorry about that! Yeah, I can confirm that was unintentional.

Tilps (approved):

PR #1708, opened 2022-03-11
Add mish support to blas backend.

Tilps:
Probably a performance regression without 'whole program optimization' since I didn't define Activate in the header.
Outputs for eigen look sane, I haven't tried other 'blas' implementations.


borg323:
This is superseded by #1709 - closing

PR #1713, opened 2022-03-22
WIP attention body changes

ankan-ban:
Need to finalize the kPosEncoding array before merge.


ankan-ban:
The change works with Arcturai's attention body T02 network. It doesn't work yet with T12 and T13 yet because they use different kPosEncoding array (different shape too).

ankan-ban:
Removed the hack/workaround for the training bug with ply count plane. The PR should now work only for networks that have the bug fixed (and use the policy encoding currently hardcoded in the backend). Right now the only known network that works is `T12-swa-2000000-resave.pb.gz`

Kovax007:
There is some corruption in batches higher than 1:
Score of lc0_onnx T12 vs  lc0_cuda T12: 13 - 0 - 1 [0.964]
...      lc0_onnx T12 playing White: 7 - 0 - 0  [1.000] 7
...      lc0_onnx T12 playing Black: 6 - 0 - 1  [0.929] 7
...      White vs Black: 7 - 6 - 1  [0.536] 14
Elo difference: 572.5 +/- nan, LOS: 100.0 %, DrawRatio: 7.1 %
The effect dampends on newer gpu's

borg323:
Superseded by #1821

PR #1718, opened 2022-04-02
fix ordering of multiPV lines.

hans-ekbrand:
When multiPV is above 1, several lines are send by `SendUciInfo()`, and they are order by how good Leela thinks they are, so that the first line has the move that will be played. When chess GUI:s like cutechess display these UCI-info lines, they treat them atomistically, and every subsequent line pushes the previous line down the display, which make the most important line drop to the bottom of the list. To work around this problem, this patch reverses the order of the `uci_infos` vector before it is handled over to `OutputThinkingInfo()`, if multiPV is above 1. See the two screen-shots that illustrate the effect.

This is without the patch
bar

This is how it looks with the patch applied
foo

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -290,7 +290,12 @@ void Search::SendUciInfo() REQUIRES(nodes_mutex_) REQUIRES(counters_mutex_) {
   if (current_best_edge_ && !edges.empty()) {
     last_outputted_info_edge_ = current_best_edge_.edge();
   }
-
+  // Cutechess treats each UCI-info line atomically, and if we send
+  // multiple lines we have to send the best line last for it to stay
+  // on top. This only matters if multipv is set (above one).
+  if (max_pv > 1) {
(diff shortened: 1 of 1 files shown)


mooskagh:
Could you check what other chess engines (e.g. stockfish) do with multipv on?

If they output it in reverse order (that's likely), then it's our bug, and it's not really cutechess specific.
Otherwise, I think it would make sense to fix it on the cutechess side instead.

Naphthalin:
I think this is yet another issue which ideally will be solved/modified easily with #1734, as move ordering/PV calculation is one of the parts.

tissatussa:
I remember you should set MPV to 1 in CuteChess. It's not useful to set it higher. CuteChess just shows the best move at each depth. If you want to see MPV lines, use a tool like SCID. In general an engine performs best when setting MPV 1, although i like to see multi-PV lines also, but not in CuteChess.

PR #1724, opened 2022-04-05
make use of IsValidCoord method

WWL-MikeRoberts:
no reason to duplicate logic


mooskagh (approved):
It looks like `IsValidCoord` is not used anywhere.
Alternative to this change would be to kill `IsValidCoord`. (up to you what you like more)

PR #1726, opened 2022-04-12
Speed optimizations for the blas backend.

borg323:
Big update: Using SGEMM instead of hardcoded multiplication is a big gain for attention policy, >6% for T79.
~~This~~The rest is worth about 3% for 10b nets, less for larger ones. There are two parts:
1. Optimization of the ispc Winograd transform code.
2. Fusion of the bias addition with the SE unit, where it can be done at the channel level. (So each channel bias is added twice, once in the global pooling and once together with the SE bias addition (but scaled), so 2 additions and 1 multiplication per channel instead of 64 additions and the corresponding memory accesses.

### DIFF
--- a/src/neural/blas/network_blas.cc
+++ b/src/neural/blas/network_blas.cc
@@ -271,17 +271,14 @@ void BlasComputation<use_eigen>::ComputeBlocking() {
                         conv2.weights.data(), conv_out);
 
       if (residual.has_se) {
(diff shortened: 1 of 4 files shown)


Tilps (approved):
As per out of band, the winograd transform changes don't really seem reviewable - if the results are the same then its fine.

PR #1731, opened 2022-04-15
Functionality for choosing a move randomly weighted by the move probability outputs.

danegraphics:
I added this for the intent of allowing Maia nets to play with randomized variety when `randompyp` is set to `true`.


danegraphics:
Note on line 819 in search.cc.
`if (roll <= 0.0) return edge; // In case of floating point nonsense.`

I just realized that this line doesn't exactly solve the problem I intended it to solve. Due to floating point subtraction, there is a possibility that `roll` ends up being larger than the last move checked (assuming it gets that far). If that is the case, then the function will end without a move being selected.

I'm not entirely certain what the best solution to this issue is. But I will take this PR back and find a better way to solve this.

PR #1733, opened 2022-04-16
Adding Human TM

TalvinJacobs:
A slight modification to smooth but should allow for more human-like TM when using weak networks at low node counts.


Naphthalin:
Can you explain a bit more what you changed, and why you consider it to be "human"? Which TCs (increment, total time) did you test?

danegraphics:
I too am curious about what changes you've made.

PR #1740, opened 2022-04-22
first try at dynamic Root Q based drawscore adjustment.

hans-ekbrand:
This patch adjusts draw score for side to move and for opponent based the current on Root Q.

If Root Q is negative, side to move has the upperhand, and the even draw score is reduced by 1/100 of the new parameter `DrawScoreDelta`, and the odd draw score is increased.
If Root Q is positive, side to move is worse and the even draw score is increased and the odd draw sore is decreased.


Naphthalin:
From what I see, this is an attempt at using drawscore for contempt in a more general WDL transformation, which is already done automatically in #1791. Can you maybe comment on that?

PR #1743, opened 2022-04-28
onednn backend support for mish and attention policy

borg323:
This is not yet fully optimized, but works well enough. 
Note that it needs a new `dnnl.dll` to work, and the latest one is uploaded on appveyor, compiled with cpu and gpu support. To use the igpu you add `--backend-opts=gpu=0` to the command line. Similarly, the latest `dnnl.dll` versions seem to have trouble selecting Winograd convolution automatically, so `--backend-opts=winograd=true` is worth trying on a cpu.
Also for nets larger than T74 it seems a single search thread is better than two, at least on my test system.


Tilps (approved):
Seems fine.  I assume you've tested it thoroughly with both new and old networks.

borg323:
I'm still testing it, so not merging yet.

borg323:
Found a possible issue, so making it draft while I investigate.

borg323:
Apparently some onednn primitives don't like offset memory descriptors on the gpu - also something was going wrong in the last part of promotion offsets calculation on the gpu, so I moved this bit to the cpu side.

Tilps (approved):

PR #1744, opened 2022-05-01
Scale cpuct by depth

hans-ekbrand:
Explore wider at low depth and narrower at high depth, by letting `cpuct_init` vary linearly so that at depth = 5 it is the user given parameter value, at depth 0 it scales up the value by 1.07581 and for depth = 30 cpuct is scaled by ~ 0.8.


AlexisOlson:
Where does the depth value come from?

PR #1746, opened 2022-05-03
first version of easy interface

borg323:
Looks for "simple" in the executable name and offers a vastly simplified user interface. Most options are hidden, ~~but there are two differences: A new `Contempt` option replaces the draw score options and~~*and* the legacy time manager is made default with only `Slowmover` exposed as a new option. I'm sure there are cleaner ways to do this, so suggestions are welcome.


zz4032:
Contempt 300 (Elo) would result in `DrawScoreSidetomove=-0.698` and `DrawScoreOpponent=0.698`, correct?
That should probably be corrected by scaling, tuning results I've seen in Discord suggest much lower values (e.g. -0.18 and 0.25 for 365 Elo advantage): https://discord.com/channels/425419482568196106/766262364949315614/827503684396908621

borg323:
The formula should give 0.5 for 200 Elo, so without doing the math your result seems correct. This was just a first attempt, any alternative suggestions are welcome. However at this stage I'm more interested in ideas for a cleaner implementation.

mooskagh (approved):

PR #1747, opened 2022-05-04
cuda blas backward compatibility

borg323:
@ankan-ban is there a reason to use `cublasGemmStridedBatchedEx()` here?


ankan-ban:
I remember seeing some performance difference on RTX GPUs using the "Ex" vs the non-Ex versions (the Ex version was faster). Not sure if that bug is still present.

borg323:
Switched it to check CC at runtime like in other uses of `cublasGemmStridedBatchedEx()`. Should fix #1820.

borg323:
Reported to work with a GTX660.

PR #1752, opened 2022-05-09
cudnn attention policy

borg323:
Just a straight port of the cuda code. At a minimum, some refactoring is required (`getMaxAttentionHeadSize()`).

~~Completely untested~~ ~~Very lightly tested.~~ ~~T78 doesn't work but T80 *seems* fine, which hints at insufficient scratch memory allocation.~~ The issue seems to have been the tensor memory size allocated, not the scratch size.


Tilps (approved):

Naphthalin:
Is "very lightly tested" still a correct description, or can this be merged?

borg323:
There have been some independent tests, so this is probably OK, but I need to verify the size allocation after ankan's fix.

PR #1755, opened 2022-05-21
Return command name as UCI response after successful parameter setting.

### LINKED ISSUE #1598: Confirm parameter setting with `info string` uci response
Currently we don't return anything on successful `setparam` set, and return `error` non-standart response in error.

It would be more visible in GUI if we used `info string <text>` in both cases.

ErdoganSeref:
Fixed #1598. But still unsure what to return as a confirmation.

### DIFF
--- a/src/chess/uciloop.cc
+++ b/src/chess/uciloop.cc
@@ -139,6 +139,9 @@ void UciLoop::RunLoop() {
       // Ignore empty line.
       if (command.first.empty()) continue;
       if (!DispatchCommand(command.first, command.second)) break;
+      else {
+        SendResponse(std::string("executed ") + command.first);
+      }
     } catch (Exception& ex) {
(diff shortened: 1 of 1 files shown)


src/chess/uciloop.cc
```diff
@@ -139,6 +139,9 @@ void UciLoop::RunLoop() {
       // Ignore empty line.
       if (command.first.empty()) continue;
       if (!DispatchCommand(command.first, command.second)) break;
+      else {
+        SendResponse(std::string("executed ") + command.first);
```
mooskagh, line 143: It doesn't do what was required, what was needed is `info string` command.

Also the main thing to fix is error part, not successful execution.

For the success part, we (potentially) only need it for setting variables, not for all UCI commands.

mooskagh (changes requested):

PR #1761, opened 2022-07-15
speed up position and history

borg323:
Removes `them_board_` from `Position` as it is rarely used. This gives a small but measurable search speedup. Also  starts checking for repetitions 4 ply before the current position instead of 2.
*Note:* The python interface part is completely untested.


src/chess/position.h
```diff
@@ -64,20 +64,12 @@ class Position {
 
   // Gets board from the point of view of player to move.
   const ChessBoard& GetBoard() const { return us_board_; }
-  // Gets board from the point of view of opponent.
-  const ChessBoard& GetThemBoard() const { return them_board_; }
-  // Gets board from the point of view of the white player.
-  const ChessBoard& GetWhiteBoard() const {
-    return us_board_.flipped() ? them_board_ : us_board_;
```
mooskagh, line 71: (Out of scope of this PR, please ignore, just thoughts)

I believe flipped() should be called black_to_move().

Also, either it should be moved to Position, or even better, all fields from Position would be moved to Board, and Position would be removed.

mooskagh (approved):

PR #1762, opened 2022-07-17
In smooth TM, use piggybank when it seems like the bestmove can be overtaken.

### LINKED ISSUE #1582: Smooth time manager should use piggybank then the eval for the current bestmove rapidly drops
In TCEC 2021 DivP game 10 (vs Stockfish), Lc0 found the move 27.Bxf8 which initially had a very high eval. Later during the search, Lc0 found a tactical refutation so the eval started to drop rapidly. 

However, Lc0 still made the move before Q dropped low enough for this move not to be the best, making possibly blunder move.

In such cases, it makes sense to trigger piggybank time usage.

Similar behaviour may be useful when some of the moves that are currently not the best, start to rapidly raise eval.

mooskagh:
Should help against this:
image

Not tested, likely also fixes #1582.

### DIFF
--- a/src/mcts/stoppers/smooth.cc
+++ b/src/mcts/stoppers/smooth.cc
@@ -163,11 +163,11 @@ Params::Params(const OptionsDict& params, int64_t move_overhead)
       max_piggybank_moves_(
           params.GetOrDefault<float>("max-piggybank-moves", 36.5f)),
       trend_nps_update_period_ms_(
-          params.GetOrDefault<int>("trend-nps-update-period-ms", 200)),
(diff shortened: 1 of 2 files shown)


mooskagh:
Example where it worked.

image

mooskagh:
Looking at times where this logic triggered at WCCC, it didn't help, gave up too early. 

Example from WCSC R4 game vs Ginkgo:

image

mooskagh:
image
Example where it did help!

Naphthalin:
Do we plan on using/maintaining the smooth time manager any time soon?

borg323 (approved):

PR #1261, opened 2020-05-02
Python bindings for Lc0 backend.

mooskagh:
Building:

1. Run meson.py with -Dpython_bindings
2. After the build the file `backends.cpython-38-x86_64-linux-gnu.so` will appear in the build directory (in Windows, the extension should be .pyd).
3. Put this file in the following directory structure:
```
(root)
|
+-- lczero/
    |
    +-- __init__.py
    +-- backends.cpython-38-x86_64-linux-gnu.so
```
4. cd into (root) and start python.
5. Do `import lczero.backends` and it should all work!

PR #1772, opened 2022-09-05
Version detection in meson.build for new editions of Python

patrik-ha:
I just went through the [walkthrough for building the Python-bindings](https://github.com/LeelaChessZero/lc0/pull/1261), and found that my versions of Python above 3.10 are mistaken for being older than 3.7. Using `.version_compare()` instead of direct string comparison fixes this.

I'm guessing that this is pretty niche, but I was able to complete the build process (at least for the bindings) with meson using 3.10.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -683,7 +683,7 @@ endif
 if get_option('python_bindings')
   pymod = import('python')
   python = pymod.find_installation('python3')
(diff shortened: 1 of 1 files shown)


mooskagh (approved):
Thanks for catching this!

PR #1783, opened 2022-11-02
Add attention policy and mish support to onnx backend.

borg323:


src/neural/onnx/converter.cc
```diff
@@ -122,6 +146,28 @@ std::unique_ptr<OnnxConst> Converter::GetWeghtsConverter(
...
+  auto flow = builder->Softplus(name + "/softplus", input);
+  flow = builder->Tanh(name + "/tanh", flow);
+  return builder->Mul(name, flow, input);
+}
+
+std::string Converter::MakeActivation(OnnxBuilder* builder,
+                                      const std::string& input,
+                                      const std::string& name,
+                                      ActivationFunction activation) {
+  if (activation == RELU) {
```
mooskagh, line 160: Sorry for arbitrary comment, but I suspect `switch` statement would look better here than chained `if`.

borg323: Indeed, looks a lot better.

src/neural/onnx/converter.cc
```diff
@@ -186,12 +236,121 @@ void Converter::AddStdInitializers(OnnxBuilder* builder) {
...
+  auto Q = builder->Transpose(name + "/mha/Q/transpose", flow, {0, 2, 1, 3});
+
+  flow = builder->MatMul(
+      name + "/mha/K/w", encoder_in,
+      *GetWeghtsConverter(layer.mha.k_w, {embedding_size, d_model}, {1, 0}));
+  flow = builder->Add(name + "/mha/K/b", flow,
+                      *GetWeghtsConverter(layer.mha.k_b, {d_model}));
+
+  flow = builder->Reshape(name + "/mha/K/reshape", flow, mha_shape);
+
```
mooskagh, line 267: Just a common impression, this function looks too sparse. :)
With so many empty lines.

src/neural/onnx/converter.cc
```diff
@@ -186,12 +236,121 @@ void Converter::AddStdInitializers(OnnxBuilder* builder) {
...
+  auto mha_shape =
+      builder->AddInitializer("/const" + name + "/mha/shape",
+                              Int64OnnxConst({-1, 64, heads, depth}, {4}));
+
+  auto flow = builder->MatMul(
+      name + "/mha/Q/w", encoder_in,
+      *GetWeghtsConverter(layer.mha.q_w, {embedding_size, d_model}, {1, 0}));
+  flow = builder->Add(name + "/mha/Q/b", flow,
+                      *GetWeghtsConverter(layer.mha.q_b, {d_model}));
```
mooskagh, line 254: Hmmm I'm not sure that `operator*` of temporary object will extent it's lifetime...

This surely doesn't work:
```
const X& f(const X& x) { return x; }
const auto& t = f( X() );    // X() will be destroyed immideately.
```

This surely works:
```
const auto& t = std::make_unique<X>();   // std::unique_ptr lifetime is extended
```

About this I'm not sure, will research tomorrow:
```
const auto& t = *std::make_unique<X>(); 
```

mooskagh (commented):
Here's some comments, but I'll continue tomorrow.

src/neural/onnx/converter.cc
borg323: I'm not sure I understand the concern - we only need the temporary to be valid until the end of the function call, no?

src/neural/onnx/builder.cc
```diff
@@ -205,6 +233,23 @@ std::string OnnxBuilder::Reshape(const std::string& name,
   return out;
 }
 
+std::string OnnxBuilder::Transpose(const std::string& name,
```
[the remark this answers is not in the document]
borg323: I agree, the alternative does not look better.

src/neural/onnx/converter.cc
mooskagh: Yes we do, I wasn't sure that the nested temporary object would do it, but apparently it would, so we are good here.

src/neural/onnx/converter.cc
```diff
@@ -186,12 +236,121 @@ void Converter::AddStdInitializers(OnnxBuilder* builder) {
...
+      name + "/ln2", flow,
+      *GetWeghtsConverter(layer.ln2_gammas, {embedding_size}),
+      *GetWeghtsConverter(layer.ln2_betas, {embedding_size}), 1);
+
+  return flow;
+}
+
 namespace {
-std::vector<int> MakePolicyMap() {
+std::vector<int> MakePolicyMap(const short* map, int size) {
```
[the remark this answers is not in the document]
borg323: That would require also changing `kConvPolicyMap` and the changes would be in all backends - I don't think it is worth it.

src/neural/onnx/converter.cc
```diff
@@ -186,12 +236,121 @@ void Converter::AddStdInitializers(OnnxBuilder* builder) {
...
+  flow = builder->Add(name + "/mha/V/b", flow,
+                      *GetWeghtsConverter(layer.mha.v_b, {d_model}));
+
+  flow = builder->Reshape(name + "/mha/V/reshape", flow, mha_shape);
+
+  auto V = builder->Transpose(name + "/mha/V/transpose", flow, {0, 2, 1, 3});
+
+  flow = builder->MatMul(name + "/mha/QK/matmul", Q, K);
+
+  const OnnxConst& scale = GetDataType() == pblczero::TensorProto::FLOAT16
```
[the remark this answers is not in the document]
borg323: Done. I was also worried about this one.

src/neural/onnx/converter.cc
```diff
@@ -201,7 +360,133 @@ std::vector<int> MakePolicyMap() {
 void Converter::MakePolicyHead(pblczero::OnnxModel* onnx, OnnxBuilder* builder,
                                const std::string& input,
                                const LegacyWeights& weights) {
-  if (!weights.policy1.weights.empty()) {
+  if (src_.format().network_format().policy() ==
+      pblczero::NetworkFormat::POLICY_ATTENTION) {
```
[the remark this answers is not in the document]
borg323: Done

src/neural/onnx/converter.cc
```diff
@@ -186,12 +236,121 @@ void Converter::AddStdInitializers(OnnxBuilder* builder) {
...
+  const int d_model = layer.mha.q_b.size();
+  const int depth = d_model / heads;
+
+  auto mha_shape =
+      builder->AddInitializer("/const" + name + "/mha/shape",
+                              Int64OnnxConst({-1, 64, heads, depth}, {4}));
+
+  auto flow = builder->MatMul(
+      name + "/mha/Q/w", encoder_in,
+      *GetWeghtsConverter(layer.mha.q_w, {embedding_size, d_model}, {1, 0}));
```
uwuplant, line 255: looks unrelated to PR, but typo needs fixing

src/neural/onnx/converter.cc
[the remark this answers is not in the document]
mooskagh: I think it would be worth it, but can be done some time later when we touch this part of code next time.

mooskagh (approved):

PR #1786, opened 2022-11-07
improvements for the pgn file reader

borg323:
This is extracted from action replay as it looks more generally useful. Main features:
* Fix for first move from black size
* Errors give line number
* Configurable pgn variation reading: can be disabled (together with syntax checks) or limit number of history plies
* PGN annotations ignored
* Tracks game result
* Ignores missing * at end of move list
* Some additional error checking

I considered adding error recovery, but decided against it. Given the use cases, I think the book should be error free before starting game generation.


src/chess/pgn.h
```diff
@@ -75,31 +77,29 @@ class PgnReader {
     bool in_comment = false;
     bool started = false;
     while (GzGetLine(file, line)) {
+      line_no_++;
```
mooskagh, line 80: As far as I can see, it's possible to call `AddPgnFile` multiple times with different files. In this case, I believe `line_no_` should be reset on every call.

Also, as the class keeps the state anyway, maybe it's possible to split it into more fuctions (it kind of recembles recursive descent parsing anyway)?
And also move implementations into .cc file?

src/chess/pgn.h
```diff
@@ -75,31 +77,29 @@ class PgnReader {
...
       // Check if we have a UTF-8 BOM. If so, just ignore it.
       // Only supposed to exist in the first line, but should not matter.
       if (line.substr(0,3) == "\xEF\xBB\xBF") line = line.substr(3);
       if (!line.empty() && line.back() == '\r') line.pop_back();
       // TODO: support line breaks in tags to ensure they are properly ignored.
-      if (line.empty() || line[0] == '[') {
+      if ((line.empty() || line[0] == '[') && !in_comment) {
         if (started) {
-          Flush();
+          Flush("*");
```
mooskagh, line 88: Probably `started_=false` makes sense to move to `Flush()`?
(also maybe rename `Flush()` to something like `AppendCurrentGame()` or at least `Finalize()`? Especially now that it takes a std::string parameter, it's not just flushing buffer.

src/chess/pgn.h
```diff
@@ -123,12 +123,78 @@ class PgnReader {
...
+          iss >> word;
+        } else {
+          word.swap(tmp_word);
+        }
+        if (word[0] == '(') {
+          if (cur_game_.size() == 0) {
+            CERR << "Variation with no move on line " << line_no_ << "!!";
+            throw Exception("Invalid variation.");
+          }
+          variation_stack_.push_back(cur_game_);
```
mooskagh, line 139: It's totally optional!

but it would really be easier with a recusive function that parses variation, rather than keeping the stack and other state explicitly.

Instead of the outer loop reading tokens, have two functions that can be called from anywhere:
```
std::string PeekToken();
std::string NextToken(); // Reads next and returns it.
void UndoToken(const std::string&); // Not sure if needed.
```

Or even better, not `std::string` tokens as done by std::iostream, but [tokens from the standard](http://www.saremba.de/chessgml/standards/pgn/pgn-complete.htm#c7)

(also I can see that the parser relies on newlines -- actually standard doesn't tell newlines are significant (except as the final character of a single-line comments)).

mooskagh (commented):
I'm not sure what is the best way to proceed.. PGN format is certainly easy enough so that writing it as a simple recursive descent would be simpler and cleaner than updating the current approach.  
On the other hand, the current code already works.

So I'd suggest to rewrite it properly, but I'm also totally fine with updating the current code. (tell me what you think)

I also certainly remember that I wrote pgn parser some time ago, but don't remember the context what I did it for..

PR #1789, opened 2022-11-10
Alternative P value scale method

PowerMove72:
Instead of fixed scale method 1 / total, we can customize the P scaled values shape. A p_shape value of , for example, 0.9 force Lc0 to check more moves. Same nodes/sec but less depth. Conversely, a p_shape value of, for example, 1.5 forces lc0 to focus on a few moves.
Pshape

Alternative formula is Pow(Current P  /  Max P,  P_shape) where P_shape must be a positive value.
From my tests, the better values are in the range 0.5-1.5.

### DIFF
--- a/src/mcts/params.cc
+++ b/src/mcts/params.cc
@@ -56,6 +56,8 @@ FillEmptyHistory EncodeHistoryFill(std::string history_fill) {
 }
 
 }  // namespace
+const OptionId SearchParams::KPshapeId{"P-shape", "P shape",
+                                       "Adjust P shape."};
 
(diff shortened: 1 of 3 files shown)


Naphthalin:
Maybe I misunderstand what you are going for, but what your `pow(pol,shape)` is doing is basically exactly what `PolicyTemperature` is already doing (with the exception of the sum=1), and the sum of the policy can effectively be rescaled through `cpuct`. This therefore adds a parameter but no functionality over the parameters we already have.

Naphthalin:
Any further comment, especially on whether the added functionality isn't already covered by policy temperature?

PR #1791, opened 2022-11-13
WDL Conversion for more realistic WDL and contempt

Naphthalin:
The current contempt implementation is based on drawscore, which works to some extent but is ultimately flawed because the WDL distribution (especially the draw rate) coming from training games doesn't take the Elo difference into account, and ultimately the stronger side is avoiding not only short draws, but also positions Leela can likely hold against herself.

Similarly, in high level matches D is usually too low (and the winning chances of the disadvantaged side absurdly high), which again makes the WDL inaccurate, as DivP opponents won't make the same frequent inaccuracies as training game Leela.

Third, Leela would be much more interesting for opening game preparation if she would suggest different lines in the opening depending on the Elo difference between black and white and the time control (and playing level) dependent expected draw rate. Current drawscore contempt absolutely fails at this, as it basically never influences the opening choice.

This PR adds a general WDL conversion which hopefully serves both as an accuracy rescaling and as an expected inaccuracy based contempt. I will explain the general mechanics and the derivation of the theoretical model in a separate post.


Naphthalin:
### Explaining the general mechanics

The formulas used for WDL conversion aren't as random as they look, and instead are the result of an underlying mathematical model. The current parametrization reflects this inner model, but it would probably much more accessible if one could enter playing level and Elo difference, or simply white and black Elo.

#### The mathematical model

We model a chess game (or, really any other game where there is a theoretically best move) as players taking turns with playing inaccuracies drawn from an inaccuracy distribution. The current position gets an "objective" value, from which a random walk is started, and if at some point the value is above (or below) a certain value, the game is scored as a win (loss); if it ultimately ends up between these values, it's a draw. The WDL distribution at any point should predict the expected outcomes of the three results based on the skill level of both players.

##### Assumptions for simplification:

- the separation values between win, draw and loss are irrelevant, as everything can be rescaled linearly to two arbitrary values. We therefore adopt `+1` and `-1` as the two bounds.
- the random walk is assumed to take place in an arbitrary one-dimensional space, without any assumptions how it relates to cp evals
- the expected outcome is represented as a continuous distribution over true values, and WDL refers to the area above +1, in [-1,+1], and below -1
- in principle, one would need to know the distribution of inaccuracies by both players in all phases of the game, and neither player can play better than the objective truth. However, when fixing the shape of the distribution to any distribution described through first and second moment alone (like normal or logistic), we only need the respective moments of both players' inaccuracy distribution, and ignore the fact that they are supposed to be cut off at 0. Since the logistic distribution has a nice explicit cdf and is also used in the Elo calculation, that's a natural choice.
- the respective inaccuracy distribution is applied either on every move, or at least equally often for white and black.
- Lc0's self-play RL training also applies such a reasonable distribution, assuming deblunder is active and working as intended; rescaling the WDL thus ultimately means relating the target precision to the training precision.

##### Derivation of the formulas:

The areas above +1 and below -1 of a logistic curve with mean `µ` and scale `s`are simple expressions, and can be inverted to get simple equations for `2/s` and `2µ/s` from `W` and `L`, which then give the mean and variance of the "true" logistic distribution.

Applying a random walk with `n` steps for both sides results in the mean and variance
`µ = µ_0 - n * E(sidetomove) + n * E(opponent)`
`var = n * var(sidetomove) + n * var(opponent)`

Since we know mean and variance of all 4 involved players (white/black, reference/target players), we can approximate `n` from the WDL distribution given by the NN (as that reflects the statistical outcomes of training games post deblunder), and calculate the new `µ` and `var` for the target precisions. This is then used to calculate the scale `s`, and the cdf of the transformed logistic distribution is evaluated at +1 and -1 to again get W and L values.

However, in this transformation not all 8 parameters are neded. In fact, as training is between equal opponents anyway, scaling E and var by a constant factor only changes the internal value of `n`, and only the difference in the mean inaccuracy and the sum of the variances is required anyway, this can be reduced to just 2 parameters, the ratio of (sum of variances) between match and training, and the inaccuracy difference, thus the two parameters `ratio` and `diff`.

src/mcts/search.cc
```diff
@@ -2086,8 +2086,39 @@ void SearchWorker::FetchSingleNodeResult(NodeToProcess* node_to_process,
...
+      node_to_process->v = w_new - l_new;
+      node_to_process->d = d_new;
+    } else {
+      node_to_process->v = v;
+      node_to_process->d = d;
+    }
+  } else {
+    node_to_process->v = v;
+    node_to_process->d = d;
+  }
```
Etcaqab, line 2121: I think that this new code should be a function, instead of an inlined block.

src/mcts/params.cc
```diff
@@ -382,10 +392,15 @@ void SearchParams::Populate(OptionsParser* options) {
...
   options->Add<IntOption>(kMaxConcurrentSearchersId, 0, 128) = 1;
+  std::vector<std::string> perspective = {"auto", "white", "black"};
+  options->Add<ChoiceOption>(kPerspectiveId, perspective) = "auto";
   options->Add<IntOption>(kDrawScoreSidetomoveId, -100, 100) = 0;
   options->Add<IntOption>(kDrawScoreOpponentId, -100, 100) = 0;
   options->Add<IntOption>(kDrawScoreWhiteId, -100, 100) = 0;
   options->Add<IntOption>(kDrawScoreBlackId, -100, 100) = 0;
+  options->Add<FloatOption>(kWDLRescaleRatioId, 1e-6f, 1e6f) = 1.0f;
+  options->Add<FloatOption>(kWDLRescaleDiffId, -100.0f, 100.0f) = 0.0f;
+  options->Add<FloatOption>(kNpsLimitId, 0.0f, 1e6f) = 0.0f;
```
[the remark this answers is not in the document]
Naphthalin: What do you mean? Did you by any chance accidentally look at one of the early commits where I made a copy+paste error?

src/mcts/search.cc
Naphthalin: I definitely agree! However, I am not yet worried about the optimization and code style, and only about the functionality. Also, I was wondering whether it is a good idea to do the conversion here, which is also executed on NN cache hits, instead of doing it in a way that the converted values are cached.

src/mcts/params.cc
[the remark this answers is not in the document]
Naphthalin: you're right, I completely missed that one. Thanks a lot! Resolved now.

src/mcts/search.cc
Etcaqab: I would like to see this conversion pushed "deeper" so that search is getting already adjusted `NNEval`s. This should also make merge with DAG easier.

Naphthalin: What do you consider "deeper"? Right now, it is as deep as it can get without touching the NN output itself; whenever a new node is created, it gets the (unaltered) WDL from the NN (either by an actual NN GPU eval or by a cache hit), and applies the conversion. Do you suggest to alter the NN WDL output, so we don't have to calculate the conversion multiple times on cache hits?

Etcaqab: Yes, I suggest to do the WDL adjustment before or on cache insertion so that we don't have to do that (multiple times) while fetching NN evals in search.

Naphthalin: I have thought about it a bit more, and I don't overly like storing it in the NN cache, as I can easily imagine restarting search with different parameters without clearing the cache. However, the conversion formula now really is lightweight, equivalent to calculating the PUCT score for maybe 5 nodes, but only performed once per addition of a new node. If there is any measureable CPU delay at all and (with DAG) the ratio of `(new nodes requesting NN evals) / (cache insertions)` is much bigger than 1, it is worth reconsidering, though that would likely mean touching multiple backends, and somehow transferring search parameters there.

src/mcts/search.cc
```diff
@@ -235,8 +262,18 @@ void Search::SendUciInfo() REQUIRES(nodes_mutex_) REQUIRES(counters_mutex_) {
...
-    const auto wl = edge.GetWL(default_wl);
-    const auto floatD = edge.GetD(default_d);
+    auto wl = edge.GetWL(default_wl);
+    auto floatD = edge.GetD(default_d);
+    auto wl_internal = wl;
+    auto d_internal = floatD;
+    // Only the diff effect is inverted, so we only need to call if diff != 0.
+    if (params_.GetWDLRescaleDiff() != 0) {
+      auto sign = ((params_.GetPerspective() == "auto") ||
+                   ((params_.GetPerspective() == "white") ^
```
mooskagh, line 272: btw XOR for boolean values is `!=` rather than `^`. :)

src/mcts/search.cc
```diff
@@ -258,10 +297,12 @@ void Search::SendUciInfo() REQUIRES(nodes_mutex_) REQUIRES(counters_mutex_) {
...
-    auto w =
-        std::max(0, static_cast<int>(std::round(500.0 * (1.0 + wl - floatD))));
-    auto l =
-        std::max(0, static_cast<int>(std::round(500.0 * (1.0 - wl - floatD))));
+    auto w = std::max(
+        0,
+        static_cast<int>(std::round(500.0 * (1.0 + wl_internal - d_internal))));
+    auto l = std::max(
+        0,
+        static_cast<int>(std::round(500.0 * (1.0 - wl_internal - d_internal))));
```
Naphthalin, line 305: this was done automatically by clang-format, please check whether that's as intended.

Naphthalin: I realize that this is my own code change anyway, so it's either "shorten the variable names" or "accept how it looks" I guess.

src/mcts/params.cc
```diff
@@ -375,17 +412,27 @@ void SearchParams::Populate(OptionsParser* options) {
   options->Add<ChoiceOption>(kHistoryFillId, history_fill_opt) = "fen_only";
   options->Add<FloatOption>(kMovesLeftMaxEffectId, 0.0f, 1.0f) = 0.0345f;
   options->Add<FloatOption>(kMovesLeftThresholdId, 0.0f, 1.0f) = 0.0f;
-  options->Add<FloatOption>(kMovesLeftSlopeId, 0.0f, 1.0f) = 0.0027f;
+  options->Add<FloatOption>(kMovesLeftSlopeId, 0.0f, 1.0f) = 0.00f;
```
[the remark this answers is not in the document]
Naphthalin: Yes -- unfortunately, the way our moves left head works relies on the value head not providing conflicting information. When using rather extreme contempt and MLH at the same time, the stronger side will often pick the worst move because it has the lowest M.
I'm aware of this not being ideal, and setting the default slope to 0.00 is mainly for people testing out the PR at defaults. @borg323 suggested a check at the end of `params.cc` which throws a warning/expection when both `mlh-slope` and `wdl-rescale-diff` are different from 0, and/or sets slope to 0 automatically in that case.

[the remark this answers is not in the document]
OfekShochat: Maybe instead have something to change the default when contempt != 0?

src/mcts/search.cc
```diff
@@ -2092,8 +2137,23 @@ void SearchWorker::FetchSingleNodeResult(NodeToProcess* node_to_process,
...
   // First the value...
-  node_to_process->v = -computation.GetQVal(idx_in_computation);
-  node_to_process->d = computation.GetDVal(idx_in_computation);
+  auto v = -computation.GetQVal(idx_in_computation);
+  auto d = computation.GetDVal(idx_in_computation);
+  // Check whether root moves are from the set perspective.
+  if (params_.GetPerspective() != "none") {
+    bool root_stm = (params_.GetPerspective() == "auto")
+                        ? true
+                        : ((params_.GetPerspective() == "black") ==
```
mooskagh, line 2146: Here the check is on the path which is quite important for performance (unlike the code that outputs info and is only called once per few second).

So it's better not to compare strings here in the inner loop (and not fetch the parameters from the parameters dict which is also not fast), but instead cache the  values in the SearchParams class as enum or set of bools (similarly to how we do with other performance critical params).

It would also be useful to factor out this check into a separate function, and it can even be inside SearchParams, something like `GetWdlPerspective(bool is_black_to_move)`.

Naphthalin: Do you have a reference for this? Obviously the options could internally be represented by the (unique) first character, or is this also too expensive?

src/mcts/search.cc
```diff
@@ -235,8 +261,22 @@ void Search::SendUciInfo() REQUIRES(nodes_mutex_) REQUIRES(counters_mutex_) {
     ++multipv;
     uci_infos.emplace_back(common_info);
     auto& uci_info = uci_infos.back();
-    const auto wl = edge.GetWL(default_wl);
-    const auto floatD = edge.GetD(default_d);
+    auto wl = edge.GetWL(default_wl);
+    auto floatD = edge.GetD(default_d);
```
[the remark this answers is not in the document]
Naphthalin: I fully agree with both of your points. It's probably still slightly confusing because the different uses of `wl` and `d` in the following code (choice between internal and external) require the distinction?

src/mcts/search.cc
```diff
@@ -194,6 +194,32 @@ void ApplyDirichletNoise(Node* node, float eps, double alpha) {
 }
 }  // namespace
 
+namespace {
+// WDL conversion formula based on random walk model.
+inline void WDLRescale(float& v, float& d, float wdl_rescale_ratio,
```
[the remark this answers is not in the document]
Naphthalin: The function is used both ways, and only in the output related call the original values are saved. I am open for either suggestion.

src/mcts/params.cc
```diff
@@ -269,6 +275,37 @@ const OptionId SearchParams::kDrawScoreWhiteId{
...
     "Adjustment, added to a draw score of a black player."};
+const OptionId SearchParams::kWDLRescaleRatioId{
+    "wdl-rescale-ratio", "WDLRescaleRatio",
+    "Rescales the logistic WDL scale by the given ratio."};
+const OptionId SearchParams::kWDLRescaleDiffId{
+    "wdl-rescale-diff", "WDLRescaleDiff",
+    "Shifts the logistic WDL mean by diff in white's favor."};
+const OptionId SearchParams::kWDLContemptId{
+    "wdl-contempt", "WDLContempt",
+    "The simulated rating advantage for the WDL conversion."};
```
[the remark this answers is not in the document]
Naphthalin: The name of this (and the description) is basically waiting for TCEC's `UCI_RatingAdv` extension to become an official part of the UCI protocol, and is intended to have that exact meaning.

src/mcts/params.cc
```diff
@@ -269,6 +275,37 @@ const OptionId SearchParams::kDrawScoreWhiteId{
...
+const OptionId SearchParams::kWDLContemptAttenuationId{
+    "wdl-contempt-attenuation", "WDLContemptAttenuation",
+    "This scales the given Elo advantage used for contempt."};
```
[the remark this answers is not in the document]
Naphthalin: Basic idea: real Elo difference is given through `WDLContempt`; the attenuation factor is applied to turn this into a "simulated Elo difference", as it might make sense to "underestimate" the actual difference for optimal play. For analysis of actual games, this should probably be set to 1, while the tunes for match strength were consistently using 0.5 ... 0.6

In general, contempt 200 with attenuation 0.5 and contempt 100 with attenuation 1.0 will behave exactly the same, but since we expect `UCI_RatingAdv` to be set automatically, having this scaling factor is important.

src/mcts/params.cc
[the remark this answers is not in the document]
Naphthalin: There actually is a solution for turning mlh back on: slightly change the interpretation of mlh-threshold and remove the discontinuity caused by it. We likely should give a warning if contempt is set but mlh-threshold is at the default 0.0.

src/mcts/params.cc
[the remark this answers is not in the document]
mooskagh: Maybe this? (generated by ChatGPT)

```
Scales Elo advantage used for contempt. It converts real Elo difference to simulated Elo
difference to optimize play. For analyzing actual games, use 1.0, and for match strength,
use 0.5-0.6.
```

What parameters do, should be more or less clear from the help description, and now it's not.
If it's not clear, people will come and ask, and then we'll have to explain it anyway. So it makes sense to write some description now. Even if it will be a bit lengthy (8-10 lines).

src/mcts/search.cc
mooskagh: The fetching parameters from `options_` dictionary (stored inside `params_`) is slow, that's why a copy of parameters that are referenced, is cached in `SearchParams` itself.  The string comparison itself is not slow, but it's still nice (and easy) to avoid.

What I mean is to convert string to enum and cache, similarly to what is done for the `GetHistoryFill()`. With enum it will also be possible to use switch() which may (or may not, because of required `break;`s) look nicer.

src/mcts/search.cc
```diff
@@ -436,10 +490,13 @@ std::vector<std::string> Search::GetVerboseStats(Node* node) const {
         up = -up;
         std::swap(lo, up);
       }
-      *oss << (lo == up                                                ? "(T) "
-               : lo == GameResult::DRAW && up == GameResult::WHITE_WON ? "(W) "
-               : lo == GameResult::BLACK_WON && up == GameResult::DRAW ? "(L) "
-                                                                       : "");
+      *oss << (lo == up
```
mooskagh, line 493: Likely chain of `if .. *oss << "(T)"; else if .. else if .. else` will look nicer.

src/mcts/search.cc
mooskagh: Looks good to me as it is.

src/mcts/params.cc
```diff
@@ -487,6 +553,74 @@ SearchParams::SearchParams(const OptionsDict& options)
           options.Get<int>(kMaxCollisionVisitsScalingEndId)),
       kMaxCollisionVisitsScalingPower(
           options.Get<float>(kMaxCollisionVisitsScalingPowerId)) {
+  // Calculate ratio and diff for WDL conversion from the contempt settings,
+  // unless they are changed from their default values explicitly.
+  kWDLRescaleRatio = options.Get<float>(kWDLRescaleRatioId);
```
mooskagh, line 558: Move this function body to a separate function, that way it will be possible to keep kWDLRescaleRatio and kWDLRescaleDiff const.

It may even make sense to group all the related WDL params in a struct, and make the entire struct a member variable of SearchParams.  Something like:

=== In params.h:
```
class SearchParams {
public:
   struct WDLRescaleParams {
       float ratio;
       float diff;
       ...
   };
   const WDLRescaleParams& GetWDLRescaleParams() const { return kWdlRescaleParams; }
private:
   ...
   const WDLRescaleParams kWdlRescaleParams;
}
```

=== In params.cc
```
namespace {
   WDLRescaleParams BuildWDLRescaleParams(const OptionsDict& options) {
      WDLRescaleParams pars;
      /// extract
      return pars;
   }
}  // namespace

...
SearchParams::SearchParams(const OptionsDict& options)
   ...
   kWdlRescaleParams(BuildWDLRescaleParams(options)) {}
```

src/mcts/params.cc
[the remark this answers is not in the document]
Naphthalin: I believe there is an issue with some GUIs ignoring any `UCI_...` named params in config files, even if they aren't set by the GUI. I'm pretty sure this happened recently both at CCC and with Plutie's own tests.

src/mcts/params.cc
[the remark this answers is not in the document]
Naphthalin: Since the last update, attenuation doesn't scale the Elo difference itself anymore, but the effective "eval shift" caused by contempt. Thanks for the suggestet descriptions, I'll make it more specific and accurate.

src/mcts/params.cc
Naphthalin: This sounds like a very sensible suggestion indeed. I might however require some assistance with the exact implementation, especially with where to put the formulas.

src/mcts/search.cc
Naphthalin: Makes sense; do you want me to update this within the scope of this PR, or either of us to create a separate PR for it?

src/mcts/search.cc
Naphthalin: Thanks, will change it accordingly to how `GetHistoryFill()` is handled.

src/mcts/params.cc
```diff
@@ -257,6 +258,11 @@ const OptionId SearchParams::kMaxConcurrentSearchersId{
     "max-concurrent-searchers", "MaxConcurrentSearchers",
     "If not 0, at most this many search workers can be gathering minibatches "
     "at once."};
+const OptionId SearchParams::kPerspectiveId{
+    "perspective", "Perspective",
```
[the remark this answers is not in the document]
Naphthalin: I agree. Will change `auto` to `sidetomove`, and `Perspective` to `WDLContemptPerspective` to make it clear that (at least for now) it's only used in the WDL contempt context.

Naphthalin:
I realized that enough has happened since I created this PR and wrote down the first ideas that an update post is due.

## Overview over what has happened

- while the internal parameters `diff` and `ratio` are sufficient for fully describing the model, their actual values feel arbitrary and counterintuitive. They are related to the Elo difference and the expected overall draw rate though, so a generalized Elo model was added to this PR to allow easier to use settings for contempt by specifying the Elo difference and the target draw rate, including support for the `UCI_RatingAdv` TCEC UCI extension.
- the Elo contempt model is pretty accurate at all levels and draw rates, however it can't properly deal with big Elo differences (like it happened in TCEC Swiss4) due to fixed target draw rate being wrong. Therefore, a second contempt model was added, where instead of the target draw rate, a (rapid) CalibrationElo is now used with an internal draw rate adjustment model. This Elo contempt model is probably slightly less accurate than the first one, but easy to set up and use, and it works with high Elo differences as well.
- some Leela nets (like BT2) have rather "dirty" WDL, compared to the "clean" WDL (meaning that in decisive positions, there are really small remaining chances estimated for the weaker side) of RL nets like T80. This created an issue in interaction with contempt, which was fixed by limiting the WDL derived width of the logistic distribution by `1.4`.
- finally, a few days ago I added a new score type `WDL_mu` which exposes the `mu` calculated in the `WDLRescale()` function as a centipawn output, which has the nice property that it follows the "+1.00 means 50% W" convention Stockfish https://github.com/official-stockfish/Stockfish/pull/4216 and other engines have adopted.

Naphthalin:
The PR in its current form uses 3 sets of parameters for making the functionality actually usable. If more than one set is changed from the default values at the same time, the one with the highest priority is used.

### Contempt implementation 1
Advantages: Easiest to use, works for human games and engines alike and up to several hundreds Elo difference, only 1 parameter with rather clear meaning
Disadvantages: Less precise in several aspects, especially for Elo <2400


### Contempt implementation 2
Advantages: Accurate representation of the Elo formula at smaller Elo differences
Disadvantages: Doesn't work well with high Elo differences, needs 2 parameters which are both slightly obscure to use


### Contempt implementation 3
Advantages: Directly manipulates the 2 internally used params, doesn't care about the Elo system or anything.
Disadvantages: Both parameters and their effects are rather unintuitive. Better leave their calculation to Leela.

Naphthalin:
## Examples on how to use the PR

### Use the PR for opening research, to seek out/avoid sharp lines

Independent from the approach taken here, there has been interest in how to make better use of Lc0's WDL to judge the sharpness of positions, and use that for opening selection/preparation, see [here](https://www.chess-journal.com/evaluatingSharpness1.html) and [here](https://www.chess-journal.com/evaluatingSharpness2.html).

As indicated in the 2nd post, we're evaluating the sharpness of a position from the WDL, and the contempt implementation in this PR simply applies an eval bonus/malus for sharp positions, scaled with the Elo difference.

To quote the post,
> One thing I often struggle with is finding a line with which I’m comfortable against a rare gambit. Usually the “best” line according to the engine is also very sharp and the opponent probably knows at least something in this line. So it would be nice if the sharpness scores could guide you to a specific line and help to find a quieter line that still leaves you with a good position. 

This PR is (partially) intended for automating exactly that. First, let's take a look at the position after `1. d4 Nf6 2. c4 e6`, and assume that white is actively looking for complications, at the risk of getting an objectively slightly worse position. Reasons could be that they're in a must win game (while black is happy with a draw), or that they're simply the stronger player. We're using Nibbler with `lc0_PR1791` and the following settings from this PR (I copied this from the "options" section in my `engines.json` in Nibbler; remove the `--> comments` if you intend to copy these settings).

```
"WeightsFile": "810026.lc0",       --> a recent T80 net
"WDLDrawRateReference": 0.58,      --> the internal draw rate from T80
"ContemptPerspective": "white",
"WDLCalibrationElo": 2700,         --> using higher CalibrationElo qualitatively gives the same results, but requires slightly higher Elo differences (contempt)
"WDLEvalObjectivity": 1.0,         --> we will use both 0.0 and 1.0, and compare what we can learn from the eval
"ScoreType": "WDL_mu",             --> this allows an easy comparison of SF and Lc0 eval. Note that the Lc0 search optimizes the expected score `q`, so it's possible that the most searched move doesn't have the highest eval.
"MovesLeftThreshold": 0.95,        --> in general, we want to favour shorter wins, or prolong losses. However, Lc0's MLH estimations are without contempt, so they can work against us, so we set the MLH effect to 0 unless the game is decided.
"Contempt": "100",                 --> we're using 0, 100 and 200 for comparison, so it's a 2700 player with white against a simulated 2700, 2600, 2500 player with black.
```

`Contempt: "0"`
As expected, `3. ... Nxg4` is the preferred move.
grafik

`Contempt: "100"`
While `3. ... Nxg4` still has the highest eval, black wants to avoid complications, and prefers the less sharp `3. ... d5`
grafik

`Contempt: "300"`, `WDLCalibrationElo: 2300` (had to use rather extreme settings as h6 is a lot worse than d5 and Nxf4)
If black wants to avoid complications even more, Leela starts to consider `3. ... h6` as well.
grafik

Now, let's take a look at the position one move earlier.

`Contempt: "0"` (100 gives basically the same picture, with Nc3 being a bit higher due to lower draw rate)
Unsurprisingly, the 3 considered moves are also the 3 top choices in human theory, even if at a different order.
grafik

`Contempt: "200"`
Suddenly, a different move emerges at the top: `3. g4` -- objectively, this gives black a small advantage, however the simulated WDL at +200 Elo shows a high reward (4% higher win chance) at a moderate risk (1% higher loss chance).
grafik

Note that the estimated WDL in the screenshots depends on the contempt settings, despite the objective eval. Writing this paragraph made me realize though that this is unintuitive, so ignore the WDL for now :)

yuzisee:
@Naphthalin 

> Contempt: "0"
> As expected, 3. ... Nxg4 is the preferred move.

and

> Contempt: "100"
> While 3. ... Nxg4 still has the highest eval, black wants to avoid complications, and prefers the less sharp 3. ... d5

are showing the same screenshot twice (was that intended?)

Naphthalin:
Definitely not intended, and certainly wrong. I'll recreate whatever of the two is missing and update the post. Thanks for notifying me :)

src/mcts/search.cc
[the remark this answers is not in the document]
Naphthalin: I simplified the "internal d" away a few commits ago, and renamed the wdl output variables to `wdl_w` etc. I consider this resolved now.

src/mcts/params.cc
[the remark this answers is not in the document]
Naphthalin: Elo is mentioned in the description now.

src/mcts/search.cc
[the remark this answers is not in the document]
Naphthalin: `v/d_internal` was simplified away, so `v` and `d` are always replaced by `RescaleWDL(...)`, original values aren't needed anymore. Should they still be turned into pointers as you initially suggested?

src/mcts/params.cc
```diff
@@ -55,6 +58,104 @@ FillEmptyHistory EncodeHistoryFill(std::string history_fill) {
...
+  for (auto& entry : StrSplit(contempt_str, ",")) {
+    auto parts = StrSplit(entry, "=");
+    if (parts.size() == 1) {
+      try {
+        contempt = std::stof(parts[0]);
+      } catch (std::exception& e) {
+        throw Exception("Invalid default contempt: " + entry);
+      }
+    } else if (parts.size() == 2) {
+      if (std::search(name.begin(), name.end(), parts[0].begin(),
```
mooskagh, line 81: This is to check that parts[x] is a substring of name, right? (case insensitive).
(are you sure that you need a substring rather than a full match)

mooskagh (approved):

mooskagh (approved):

src/mcts/params.cc
borg323: IIRC the original reasoning was to catch something like "Dragon" and "Komodo Dragon".

src/mcts/search.cc
```diff
@@ -259,20 +311,31 @@ void Search::SendUciInfo() REQUIRES(nodes_mutex_) REQUIRES(counters_mutex_) {
       uci_info.score = q * 10000;
     } else if (score_type == "W-L") {
       uci_info.score = wl * 10000;
+    } else if (score_type == "WDL_mu") {
+      // Reports the WDL mu value whenever it is reasonable, and defaults to
+      // centipawn otherwise.
+      const float centipawn_fallback_threshold = 0.99f;
```
Naphthalin, line 317: Threshold currently has the fallback if W or L drops below 0.5%, which seems to be a bit soon. Should probably be changed to 0.996 or so, which would make it act if W or L drop below 0.2%

yuzisee:
If using Lc0 in analysis mode on an already-completed armageddon game between two players, is there a combination of values to `WDLEvalObjectivity` and `Contempt` and `ContemptPerspective` that could be used to determine "who is winning" at any given point in the game?

Naphthalin:
## Examples on how to use the PR, pt 3
### Armageddon games

Armageddon games are usually played under the following conditions:
- between human GMs (so, around 2700)
- in rapid/blitz time control (slightly faster TC than the `WDLCalibrationElo` assumes)
- with white having a significant time advantage (i.e. black is modeled as maybe 50-100 Elo weaker)

This could be for example achieved by the following settings:
```
"WDLDrawRateReference": 0.58, --> for T80
"WDLCalibrationElo": 2700,
"ContemptPerspective": "white",
"Contempt": "100", --> actual white advantage + TC dependent white bonus
"WDLContemptAttenuation": 1.0,
"WDLEvalObjectivity": 0.0, --> highly relevant for contempt: we want to see the internally used WDL
"DrawScoreWhite": -100,
"DrawScoreBlack": 100,
"ScoreType": "centipawn_with_drawscore" or  "WDL_mu"```

note: a position with equal chances for white and black will show as approx. 0.00 when using `centipawn_with_drawscore`, or +1.00 when using `WDL_mu`. Alternatively there is also the scoretype `Q` which directly shows the expected score, from +100 (white win) to -100 (draw or black win).

I used these settings to kibitz the KGA Armageddon game between Hikaru and Magnus, and from a "practical chances" perspective the KGA didn't seem like a too bad choice for an Armageddon game :) Just curious, is this also the game you were interested in @yuzisee ?

PS: In the near future, the `drawscore` implementation will be simplified and only require 1 setting instead of 2 for Armageddon.

yuzisee:
> Just curious, is this also the game you were interested in

How did you know 🙂

PR #1792, opened 2022-11-15
directml onnx-dml backend and onnx fp16 support

borg323:
The directml onnxruntime provider works (several orders of magnitude) better with a fixed batch side, so this adds a mechanism to create several onnxruntime sessions with different batch sizes and breaks the input batch into suitable blocks. This is controlled by the `batch` and `steps` backend options, creating sessions for batch sizes from `batch` to `steps * batch`. The options default to 16 and 8 respectively for onnx-dml, -1 (variable batch size) and 1 for the other onnx backends.
This PR also adds fp16 support to the onnx backends, default on for onnx-cuda and onnx-dml.

### DIFF
--- a/src/neural/onnx/adapters.cc
+++ b/src/neural/onnx/adapters.cc
@@ -31,6 +31,22 @@ Program grant you additional permission to convey the resulting work.
 #include "utils/transpose.h"
 
 namespace lczero {
+namespace {
+template <class T>
+std::string TransposeAndReturnRaw(const std::vector<int>& dims,
+                                  std::vector<int> order,
+                                  const std::vector<T>& from) {
+  if (order.empty()) {
+    return {reinterpret_cast<const char*>(from.data()),
+            reinterpret_cast<const char*>(from.data() + from.size())};
+  } else {
+    std::vector<T> dst(from.size());
+    TransposeTensor(dims, order, from, &dst[0]);
+    return {reinterpret_cast<const char*>(dst.data()),
+            reinterpret_cast<const char*>(dst.data() + dst.size())};
+  }
+}
+}
 
 FloatOnnxWeightsAdapter::FloatOnnxWeightsAdapter(
     const std::vector<float>& weights, std::initializer_list<int> dims,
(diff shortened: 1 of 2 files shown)


src/neural/onnx/adapters.cc
```diff
@@ -57,4 +58,22 @@ std::string FloatOnnxWeightsAdapter::GetRawData() const {
...
-}  // namespace lczero
\ No newline at end of file
+pblczero::TensorProto::DataType Float16OnnxWeightsAdapter::GetDataType() const {
+  return pblczero::TensorProto::FLOAT16;
+}
+
+std::string Float16OnnxWeightsAdapter::GetRawData() const {
+  std::vector<uint16_t> fp16(weights_.size());
+  std::transform(weights_.begin(), weights_.end(), fp16.begin(), FP32toFP16);
+  if (order_.empty()) {
```
mooskagh, line 68: Totally optional, but as the code repeats now, maybe add something like this function to transpose.h (or even here in adapters.cc in anonymous namespace):

```cpp
template <class T>
std::string TransposeAndReturnRaw(const std::vector<int>& dims, std::vector<int> order,
                     const std::vector<T>& from) {
  if (order.empty()) {
    return {reinterpret_cast<const char*>(from.data()),
            reinterpret_cast<const char*>(from.data() + weights_.size())};
  } else {
    std::vector<T> dst(weights_.size());
    TransposeTensor(dims, order, weights, &dst[0]);
    return {reinterpret_cast<const char*>(dst.data()),
            reinterpret_cast<const char*>(dst.data() + dst.size())};
  }
}
```

src/neural/onnx/network_onnx.cc
```diff
@@ -98,74 +111,191 @@ class OnnxNetwork : public Network {
...
+             FP16toFP32(data[sample * 3 + 2]);
+    } else {
+      return data[sample * 3 + 0] - data[sample * 3 + 2];
+    }
   } else {
-    const auto& data =
-        output_tensors_[network_->value_head_].GetTensorData<float>();
-    return data[sample];
+    const auto& data = output_tensors_data_[network_->value_head_];
+    if (std::is_same<Ort::Float16_t, DataType>::value) {
```
mooskagh, line 169: This construction is repeated several times in the file, consider having a separate function for it.
Maybe even not a template one, but just override. Something like:

```cpp
float AsFloat(float x) { return x; }
float AsFloat(Ort::Float16_t x} { return FP16toFP32(x); }
```

mooskagh (approved):

PR #1794, opened 2022-11-18
onnx backend option to handle tf exported models

borg323:


src/neural/onnx/network_onnx.cc
```diff
@@ -193,15 +219,19 @@ Ort::Value OnnxComputation<DataType>::PrepareInputs(int start, int batch_size) {
   input_tensor_data_.clear();
   input_tensor_data_.resize(batch_size * kInputPlanes * 8 * 8);
   auto iter = input_tensor_data_.data();
+  int idx = 0;
   int end = std::min(start + batch_size, static_cast<int>(raw_input_.size()));
   for (int i = start; i < end; i++) {
```
ChinChangYang, line 224: Shouldn't the `idx` be reset below this line?
```
for (int i = start; i < end; i++) {
  int idx = 0;
```

borg323: Yes, that was the original version, messed it up in a marge later on.

borg323:
This is now superseded by `onnx2leela` options.

PR #1796, opened 2022-11-21
add fp16 support to onnx backend

borg323:
This is excerpted from #1792 for easier review.


borg323:
Since #1792 is now under review, there is no point reviewing this as well.

borg323:
#1792 is now merged.

PR #1797, opened 2022-11-23
only use thread binding with blas backend

borg323:
Apparently binding threads to a windows processor group is a bad idea if lc0 is used simultaneously with SF for analysis: both engines try to use the same cores.
Since thread binding is only useful for more than 32 threads, in most cases it may be better to let windows handle the allocation.
The only real use case is with the blas backend, so here thread binding is made conditional on `USE_BLAS`. This does not include the eigen backend, but I think this is acceptable as it is mainly intended as a reference implementation.

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -36,7 +36,6 @@
 #include "utils/commandline.h"
 #include "utils/esc_codes.h"
 #include "utils/logging.h"
-#include "utils/numa.h"
 #include "version.h"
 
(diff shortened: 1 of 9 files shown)


mooskagh:
Would it make sense to call `Numa::Init()` from blas backend creation instead?

(the Numa::Init function can also keep the static flag to make sure it's not initialized twice)

mooskagh (approved):

PR #1801, opened 2022-11-29
cross build for arm windows

borg323:


ArminiusTux:
Official WinARM64 support would be appreciated.

PR #1805, opened 2022-12-04
support processors with hybrid cores under windows

borg323:


mooskagh:
Is it still relevant? If so, I think it should be rebased as the same code was recently moved around in master.

PR #1806, opened 2022-12-07
onnx attention body

borg323:
~~The attention body part isn't working yet.~~ ~~Seems to work fine now.~~ Has been extensively tested and works fine.


src/neural/onnx/converter.cc
```diff
@@ -329,7 +411,13 @@ std::string Converter::MakeEncoderLayer(
                                           {embedding_size, dff_size}, {1, 0}));
   flow = builder->Add(name + "/ffn/dense1/b", flow,
                       *GetWeghtsConverter(layer.ffn.dense1_b, {dff_size}));
-  flow = MakeActivation(builder, flow, name + "/ffn/dense1", activation);
+  if (layer.mha.has_smolgen) {
```
almaudoh, line 414: There's also a separate `square_relu_ffn: true` configuration parameter in the yaml file for this (which I also missed out in the proto update). `has_smolgen` can be used as a proxy for now though until all of that is settled.

src/neural/onnx/converter.cc
```diff
@@ -203,6 +230,10 @@ std::string Converter::MakeSqueezeAndExcite(
     const std::string& input, const std::string& name) {
   const int se_filters = se_unit.b1.size();
 
+  for (static bool flag = false; !flag; flag = true) {
```
mooskagh, line 233: Is the intention to have common initializer for several blocks?

I think it won't work e.g. if network will be changed during the game (we support this), or if in selfplay player1 and player2 use different ONNX networks (that happens during match games).
Maybe add the (non-static) flag as the Converter member variable instead?

mooskagh (approved):

PR #1807, opened 2022-12-07
Attention body for metal backend

almaudoh:


borg323 (approved):
Can you run it though clang-format before committing?

PR #1811, opened 2022-12-10
update onednn version

borg323:
This one will also fix the onednn package build issue as the zip used includes some files that were missing in the previous one.


Naphthalin (approved):
Didn't check functionality, only looked for consistency to avoid accidental typos.

PR #1819, opened 2022-12-18
use FILE_FLAG_RANDOM_ACCESS in syzygy

borg323:


mooskagh (approved):

borg323:
There is some alarming documentation from Microsoft regarding this, like <https://learn.microsoft.com/en-GB/troubleshoot/windows-server/application-management/operating-system-performance-degrades> and <https://learn.microsoft.com/en-us/windows-server/administration/performance-tuning/subsystem/cache-memory-management/troubleshoot#system-file-cache-contains-memory-mapped-files>, so I would feel a lot more comfortable with some testing before merging.

mooskagh:
Oh indeed it looks like it never unmaps regions of a file if they are open with FILE_FLAG_RANDOM_ACCESS? Then it may be a subtle issue, happening only after reading many distinct parts of a file... Worth testing indeed.

mstembera:
Just FYI this is what SF uses https://github.com/official-stockfish/Stockfish/blob/3f7fb5ac1d58e1c90db063053e9f913b9df79994/src/syzygy/tbprobe.cpp#L243C81-L243C81 so at least from that point of view it's tested :)

borg323:
> this is what SF uses

We know, but the SF commit history doesn't show any testing, and the problem (if it really exists) may be complex to reproduce.

PR #1822, opened 2022-12-25
Fix softmax cuda

almaudoh:
Fix nan overflow on cuda softmax

### DIFF
--- a/src/neural/cuda/winograd_helper.inc
+++ b/src/neural/cuda/winograd_helper.inc
@@ -419,6 +419,24 @@ __device__ __forceinline__ float warpReduce(float x) {
   return x;
 }
 
+// fast max reduction for the warp
+__device__ __forceinline__ float warpMax(float x) {
+#pragma unroll
+  for (int mask = 16; mask > 0; mask >>= 1)
+    x = max(x, __shfl_xor_sync(0xFFFFFFFF, x, mask));
+
+  return x;
+}
+
+// atomic max implementation for floats
+__device__ __forceinline__ float atomicMaxFloat (float * addr, float val) {
+    float max;
+    max = (val >= 0) ? __int_as_float(atomicMax((int *)addr, __float_as_int(val))) :
+         __uint_as_float(atomicMin((unsigned int *)addr, __float_as_uint(val)));
+
+    return max;
+}
+
 // Helper fuction to do vector loads/stores
 template <typename T>
 __device__ __forceinline__ void copyAs(void* dst, const void* src) {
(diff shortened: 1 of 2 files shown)


src/neural/cuda/winograd_helper.inc
```diff
@@ -419,6 +419,24 @@ __device__ __forceinline__ float warpReduce(float x) {
...
+  for (int mask = 16; mask > 0; mask >>= 1)
+    x = max(x, __shfl_xor_sync(0xFFFFFFFF, x, mask));
+
+  return x;
+}
+
+// atomic max implementation for floats
+__device__ __forceinline__ float atomicMaxFloat (float * addr, float val) {
+    float max;
+    max = (val >= 0) ? __int_as_float(atomicMax((int *)addr, __float_as_int(val))) :
```
Tilps, line 434: As mentioned in chat, stackoverflow suggests val >= 0 probably should be !signbit(val) to ensure a val of negative 0 doesn't get handled incorrectly as being smaller than existing negative values.

src/neural/cuda/winograd_helper.inc
```diff
@@ -419,6 +419,24 @@ __device__ __forceinline__ float warpReduce(float x) {
...
+__device__ __forceinline__ float warpMax(float x) {
+#pragma unroll
+  for (int mask = 16; mask > 0; mask >>= 1)
+    x = max(x, __shfl_xor_sync(0xFFFFFFFF, x, mask));
+
+  return x;
+}
+
+// atomic max implementation for floats
+__device__ __forceinline__ float atomicMaxFloat (float * addr, float val) {
```
Tilps, line 432: Stray space after atomicMaxFloat.  Also float * addr should be float* addr
Casts below should be int* and unsigned int* similarly (no space between type name and the * character)

Tilps (approved):
Approved assuming style changes are applied.

Tilps:
Plutie did some perf testing - strangely showed up marginally faster than master even on backendbench... In any case, probably not a significant performance regression for T78.  Waiting for confirmation of non-regression of fixed node elo.

Tilps:
I ran policy_tournament of T78 vs T80 both with and without patch.  1822 was better, but probably not significantly. (0.14 elo difference after 20k games each...)
Masterkni6 has found this patch fixes policy output for some networks trained with more policy encoders.

Tilps:
Side note: this patch should also fix Nan from underflow (if all inputs are strongly negative, the old cuda code could produce 0/0 for all softmax outputs.)

PR #1824, opened 2022-12-27
inline fp16 conversions

borg323:
Uses the intel fp16c intrisics on all binaries by default except openblas. Should only make a difference in onnx-dml (and onednn with #1756).


src/neural/onnx/network_onnx.cc
```diff
@@ -196,12 +196,11 @@ Ort::Value OnnxComputation<DataType>::PrepareInputs(int start, int batch_size) {
   int end = std::min(start + batch_size, static_cast<int>(raw_input_.size()));
   for (int i = start; i < end; i++) {
     for (const auto& plane : raw_input_[i]) {
+      DataType value = std::is_same<Ort::Float16_t, DataType>::value
+                           ? FP32toFP16(plane.value)
+                           : plane.value;
       for (auto bit : IterateBits(plane.mask)) {
-        if (std::is_same<Ort::Float16_t, DataType>::value) {
```
mooskagh, line 200: btw if the previous version, `if constexpr` would ensure it's evaluated at the compile time (it worked without it too, though).

src/utils/fp16_utils.cc
```diff
@@ -25,24 +25,14 @@
...
-#define NO_F16C
-#endif
-
 namespace lczero {
 
-uint16_t FP32toFP16(float f32) {
 #if defined(NO_POPCNT) || defined(NO_F16C) || \
     (defined(__GNUC__) && !defined(__F16C__))
+uint16_t FP32toFP16(float f32) {
   unsigned int x;
```
mooskagh, line 36: Maybe put those implementations into the header too?

src/utils/fp16_utils.h
```diff
@@ -25,9 +25,41 @@
...
 
+#if defined(NO_POPCNT) || defined(NO_F16C) || \
+    (defined(__GNUC__) && !defined(__F16C__))
+
 uint16_t FP32toFP16(float f32);
 float FP16toFP32(uint16_t f16);
 
+#else
+
+static inline uint16_t FP32toFP16(float f32) {
```
mooskagh, line 50: There's no really need for `static`.

src/neural/onnx/network_onnx.cc
borg323: I did this for the `FP32toFP16()` call, I'm not sure MSVC can figure out it can be pulled out of the loop.

src/utils/fp16_utils.cc
borg323: Seemed like a lot, but then again we don't use it in too many places.

mooskagh (approved):

PR #1840, opened 2023-01-30
fix syzygy loading

borg323:
If you change the syzygy path to something invalid in order to disable tb use and later back to the original path to use the tb again it doesn't work.

### DIFF
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -134,7 +134,7 @@ void EngineController::UpdateFromUciOptions() {
       syzygy_tb_ = nullptr;
     }
     tb_paths_ = tb_paths;
-  } else if (tb_paths.empty() && syzygy_tb_) {
+  } else if (tb_paths.empty()) {
     syzygy_tb_ = nullptr;
     tb_paths_.clear();
   }
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -132,9 +132,8 @@ void EngineController::UpdateFromUciOptions() {
     if (!syzygy_tb_->init(tb_paths)) {
       CERR << "Failed to load Syzygy tablebases!";
       syzygy_tb_ = nullptr;
-    } else {
-      tb_paths_ = tb_paths;
     }
+    tb_paths_ = tb_paths;
   }
 
   // Network.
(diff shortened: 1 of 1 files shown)


src/engine.cc
```diff
@@ -132,9 +132,8 @@ void EngineController::UpdateFromUciOptions() {
     if (!syzygy_tb_->init(tb_paths)) {
       CERR << "Failed to load Syzygy tablebases!";
       syzygy_tb_ = nullptr;
-    } else {
-      tb_paths_ = tb_paths;
     }
+    tb_paths_ = tb_paths;
   }
```
Tilps, line 137: Should there be an else block here as well where syzygy_tb_ is cleared to nullptr and tb_paths_ is updated back to empty?

src/engine.cc
```diff
@@ -132,9 +132,11 @@ void EngineController::UpdateFromUciOptions() {
     if (!syzygy_tb_->init(tb_paths)) {
       CERR << "Failed to load Syzygy tablebases!";
       syzygy_tb_ = nullptr;
-    } else {
-      tb_paths_ = tb_paths;
     }
+    tb_paths_ = tb_paths;
+  } else if (tb_paths.empty() && syzygy_tb_) {
+    syzygy_tb_ = nullptr;
+    tb_paths_.clear();
```
Tilps, line 139: I think there is still a case - a failed syzygy load will mean syzygy_tb_ is null but tb_paths_ is not already empty.  And we end up with a stale tb_paths_ value...

Tilps (approved):

src/engine.cc
borg323: Since `tb_paths_` is only used to check whether the UCI option changed there is no problem, but if we clear it the next `go` command will try lo load the tb again.

Tilps: this sequence.
set tb_path  to invalid directory.
set tb_path to empty
set tb_path to same invalid directory (which is now fixed) - the first if branch is skipped because tb_paths_ still contains the old invalid directory value.  Thus tb doesn't load.

borg323: So remove `&& syzygy_tb_` from the `else if`? Seems reasonable.

Tilps: Sounds good.

PR #1842, opened 2023-02-08
BLAS performance improvement for attention policy parts.

borg323:
Speeds up T78 policy head with blas, which is not very noticeable on the full net evaluation, but should come in handy in the future.


mooskagh (approved):
Looks good, although not that I understand fully what's going on there.

PR #1844, opened 2023-02-10
Change mlh threshold effect to create a smooth transition

Naphthalin:
Back when we implemented MLH, a mlh threshold was introduced to allow the deactivation of mlh if the game isn't decided yet. However, it was done in a way that creates a discontinuity at the threshold even without using a `constant_`, so we never actually use the mlh threshold for its intended purpose. This minor change removes the discontinuity by using a transformation on `q` which has the same value for `q = +-1`, is continuous, and 0 between `-q_threshold` and `q_threshold`.

While this change should have been done years ago, the main motivation to do it now is to allow contempt working together with mlh.


src/mcts/search.cc
```diff
@@ -111,6 +111,11 @@ class MEvaluator {
     const float child_m = child.GetM(parent_m_);
     float m = std::clamp(m_slope_ * (child_m - parent_m_), -m_cap_, m_cap_);
     m *= FastSign(-q);
+    if (q_threshold_ > 0.0f && q_threshold_ < 1.0f) {
```
mooskagh, line 114: Could you add a comment what `GetM` function does?

I though it returns moves left, but then it returns 0 if |q| < q_thredshol, so it's not clear what it's supposed to do.

Naphthalin: It returns the "moves left head effect" term which gets added to the PUCT score, so that `PUCT = Q + U + M`. This PR doesn't change the meaning of `GetM()`, but your comment makes clear that it was already confusing before this PR, so a comment (even if it has nothing to do with this PR) would certainly justified.

Naphthalin:
This figure illustrates the difference between old mlh effect and mlh effect of this PR with `threshold=0.8`, `mlh-slope=0.0027`:

src/mcts/search.cc
mooskagh: Something like this before the function declaration.
```
// Returns adjustment to Q based on Q and M value of a node.
```
Could you add this in this PR?

Also, `GetM()` sounds like a wrong name for this function. Maybe rename to `GetQAdjustment()` ?
(it's more or less clear that it's based on M because it inside the MEvaluator class).

Naphthalin: Settled at `GetMUtility()` because it best describes both where it comes from (M) and what it is (a utility).

src/mcts/search.cc
```diff
@@ -106,26 +106,36 @@ class MEvaluator {
...
+      // This allows a smooth M effect with higher q thresholds, which is
+      // necessary for using MLH together with contempt.
+      q = std::max(0.0f, (std::abs(q) - q_threshold_)) / (1.0f - q_threshold_);
+    }
     m *= a_constant_ + a_linear_ * std::abs(q) + a_square_ * q * q;
     return m;
   }
 
-  float GetM(Node* child, float q) const {
+  float GetMUtility(Node* child, float q) const {
```
Naphthalin, line 123: I don't really like that this code is copypasted; probably we should make this one the actual one, and turn the other into a wrapper function.

src/mcts/search.cc
```diff
@@ -106,26 +106,36 @@ class MEvaluator {
...
     if (!enabled_ || !parent_within_threshold_) return 0.0f;
     const float child_m = child->GetM();
```
Naphthalin, line 125: only difference in the old code: this call is without `parent_m_` as a default value, which I think is unintended. Obviously works as intended currently, but should be decided whether this is wanted or unwanted when making one of the functions a wrapper of the other.

src/mcts/search.cc
borg323: There are only two calls of the other version and one of this. Just keeping this variation and adjusting the callers seems cleaner.

Naphthalin: Sounds like the better choice for sure. Can you check whether it makes a difference for the DAG compatibility which of the two we keep?

borg323: The exact same patch applies to both master and dag-update and seems to work fine.

borg323 (approved):

PR #1849, opened 2023-02-13
separate build system control for f16c instructions

borg323:
Apparently there are systems with RTX gpus and older processors without native fp16 conversion instructions, so I made the cuda/cudnn builds to default to the fallback conversion code.


mooskagh:
There's no easy way to have both versions within one binary, right?  (otherwise could be a parameter, or even a separate backend which would toggle some bool template parameter)

mooskagh (approved):

borg323:
The cuda just uses it to convert the `alpha` and `beta` parameters for some gemm calls, so nothing performance critical. Everything else should be fine with the current hardcoded values, so not worth the trouble to make it run time selectable.

PR #1850, opened 2023-02-18
Fix ninja detection on some platforms.

ddobbelaere:
This fixes a build failure encountered on an Ubuntu Github cloud runner, where the meson logs contained the line `!meson_ci!/ci_include "/tmp/tmp.57HLIN4Fd8/lc0/build/release/build.ninja` that followed `Found ninja-1.11.1.git.kitware.jobserver-1 at /home/runner/work/lczero-onnx/lczero-onnx/build/venv/bin/ninja`. Here is a snippet of the log:

```
Found ninja-1.11.1.git.kitware.jobserver-1 at /home/runner/work/lczero-onnx/lczero-onnx/build/venv/bin/ninja
detecting CPU family based on trial='x86_64'
detecting CPU family based on trial='x86_64'
detecting CPU family based on trial='x86_64'
detecting CPU family based on trial='x86_64'
detecting CPU family based on trial='x86_64'
detecting CPU family based on trial='x86_64'
detecting CPU family based on trial='x86_64'
detecting CPU family based on trial='x86_64'
detecting CPU family based on trial='x86_64'
detecting CPU family based on trial='x86_64'
detecting CPU family based on trial='x86_64'
detecting CPU family based on trial='x86_64'
!meson_ci!/ci_include "/tmp/tmp.57HLIN4Fd8/lc0/build/release/build.ninja"
detecting CPU family based on trial='x86_64'
detecting CPU family based on trial='x86_64'
```

This led to `${NINJA}` being an empty string, and hence, nothing was built in the end.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -30,7 +30,7 @@ fi
 
 cd ${BUILDDIR}
 
-NINJA=$(awk '/ninja/ {ninja=$4} END {print ninja}' meson-logs/meson-log.txt)
+NINJA=$(awk '/Found.*ninja/ {ninja=$4} END {print ninja}' meson-logs/meson-log.txt)
 
 if [ -n "${INSTALL_PREFIX}" ]
 then
--- a/build.sh
+++ b/build.sh
@@ -17,7 +17,8 @@ esac
 
 BUILDDIR=build/${BUILDTYPE}
 
-MESON=$(PATH="${PATH}:${HOME}/.local/bin" command -v meson)
+MESON=$(PATH="${PATH}:${HOME}/.local/bin" command -v meson || true)
+MESON=${MESON:?"Could not find meson. Is it installed and in PATH?"}
 
 if [ -f "${BUILDDIR}/build.ninja" ]
 then
--- a/build.sh
+++ b/build.sh
@@ -1,7 +1,5 @@
 #!/usr/bin/env bash
 
-pushd "$(dirname "$0")"
-
 set -e
 
 case $1 in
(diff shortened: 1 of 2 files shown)


eli-schwartz:
Hmm, why are you parsing the build log for ninja? Meson literally just looks for `ninja` in PATH (under two names: the C implementation of ninja is `samu`), so it's probably safe to simply do that yourself too. It will also respect $NINJA in the environment.

So you can just check for $NINJA, or check if either `ninja` or `samu` exist. But probably checking for `${NINJA:-ninja}` is enough.

ddobbelaere:
@eli-schwartz Good question. I also didn't know but here is the answer: https://github.com/LeelaChessZero/lc0/pull/591

ddobbelaere:
The annoying question is, does it break when the path to ninja contains spaces? It probably does.

eli-schwartz:
Are there really systems where this is still `ninja-build`? :eyes: I would have thought that is really just legacy code...

Still, good point. That's 3 things to check. You can check it yourself or you can parse the ninja logfile for that information. ;)

> The annoying question is, does it break when the path to ninja contains spaces? It probably does.

Meson will print that as:
```
Found ninja-X.Y.Z at '/path/to/directory with spaces/ninja'
```
or
```
Found ninja-X.Y.Z at '/path/to/directory with spaces'"'"' and quotes/ninja'
```

This is shell-escaped.

ddobbelaere:
Hmm, but then awk will still see the spaces as delimiters I guess.

How about a simple `BUILDER=$(command -v ninja | command -v ninja-build | command -v samu)`? If none of the three is found, we bail out because of `set -e`.

eli-schwartz:
That seems fine, although you may want to only do so when $NINJA isn't set in the environment.

ddobbelaere:
> That seems fine, although you may want to only do so when $NINJA isn't set in the environment.

Is this actually a common thing to do, I wonder, checking for `${NINJA}` first?

ddobbelaere:
I think the current version is way better (and follows meson docs): just use `meson compile` (available since v0.55) and `meson install`.

build.sh
```diff
@@ -1,7 +1,5 @@
 #!/usr/bin/env bash
 
-pushd "$(dirname "$0")"
```
borg323, line 3: Is there a reason to remove this? See #865.

ddobbelaere: No was probably too hasty after removing `cd "${BUILDDIR}"`. I'll add them again.

build.sh
```diff
@@ -16,27 +14,14 @@ esac
...
 then
-  ${NINJA} install
+  "${MESON}" configure "${BUILDDIR}" -Dbuildtype="${BUILDTYPE}" -Dprefix="${INSTALL_PREFIX:-/usr/local}" "$@"
 else
-  ${NINJA}
+  "${MESON}" "${BUILDDIR}" --buildtype "${BUILDTYPE}" --prefix "${INSTALL_PREFIX:-/usr/local}" "$@"
 fi
 
-popd
+"${MESON}" compile -C "${BUILDDIR}"
```
borg323, line 26: Can you think of a simple way to check meson version? If not then I suppose we will have to increase the minimum version in `meson.build` to 0.55.

build.sh
ddobbelaere: I have added a `cd` command.

build.sh
ddobbelaere: According to https://mesonbuild.com/Commands.html#compile, 0.54 is enough, so I've bumped the requirement.

eli-schwartz: `meson --version`? :)

Doing natural version sorting from a shell script is a bit fiddly, but if you assume GNU coreutils then pipe both version numbers to `sort -V` and then check whether the first (older version) line is the current version of Meson or the minimum one you need.

ddobbelaere: Ubuntu 20.04 ships 0.53 it seems (18.04 ships 0.45 so not supported anymore). Should we really support these rather old versions if `pip install meson` does the trick on all those systems?

borg323: I have asked on discord whether this would cause an issue for anyone there.

borg323 (approved):

PR #1851, opened 2023-02-21
Onnx blas ab

almaudoh:
Implementation of attentionbody for blas / eigen / mkl backends.


almaudoh:
Closed. Superceded by #1852

PR #1852, opened 2023-02-21
Attention body support for blas / eigen / mkl

almaudoh:
Implementation of attentionbody for blas / eigen / mkl backends.


src/neural/loader.cc
```diff
@@ -108,6 +108,20 @@ void FixOlderWeightsFile(WeightsFile* file) {
   using nf = pblczero::NetworkFormat;
   auto network_format = file->format().network_format().network();
   const auto has_network_format = file->format().has_network_format();
+  // Hack for old encoding compatibility. REMOVE BEFORE RELEASE.
```
mooskagh: What is exactly the reason of this hack, in particular,

- What is the harm of keeping it after the release?
- Why is it fine to have in master before the release?

If there's indeed a reason to keep it in master before the release but remove after, maybe add a version check that throws the exception (or even better, some kind of static assert), so that we don't forget about it.

src/neural/shared/activation.h
```diff
@@ -21,19 +21,20 @@
...
-  NONE = pblczero::NetworkFormat::ActivationFunction::NONE,
-  RELU = pblczero::NetworkFormat::ActivationFunction::RELU,
-  TANH = pblczero::NetworkFormat::ActivationFunction::TANH,
-  SIGMOID = pblczero::NetworkFormat::ActivationFunction::SIGMOID,
-  SELU = pblczero::NetworkFormat::ActivationFunction::SELU,
-  MISH = pblczero::NetworkFormat::ActivationFunction::MISH,
-  SWISH = pblczero::NetworkFormat::ActivationFunction::SWISH,
-  RELU_2 = pblczero::NetworkFormat::ActivationFunction::RELU_2,
-  SOFTMAX = pblczero::NetworkFormat::ActivationFunction::SOFTMAX
+  DEFAULT = 0,
```
mooskagh: Maybe prefix them all with ACTIVATION or make them `enum class` instead of `enum`?
Because names like `DEFAULT` and `NONE` are quite generic and exposing them right into the `lczero` namespace without any class is not that nice.

borg323: Those were already merged with #1858 but we can obviously revise them as needed. The names were like that before, and this is only included by backend code.

src/neural/loader.cc
borg323: This was also merged with #1858, it is for compatibility with experimental attention body/smolgen nets that were trained before the net.proto update, in particular many BT2 nets and some distilled ones that people still use. As newer nets come out there should be less need for this so it can be removed, but for now it is a common issue. There is no harm keeping it indefinitely, but as those are unofficial nets it didn't seem worthwhile.

src/neural/shared/activation.h
borg323: I made https://github.com/LeelaChessZero/lc0/pull/1860 to discuss this.

mooskagh (approved):
Looks good to me (although I didn't check for issues from NN perspective)

PR #1855, opened 2023-03-04
Change calculation of binomial coefficients in syzygy

cj5716:
Reduce by 1 for loop


borg323:
Can you explain a bit more what this does? For easier maintenance we prefer the tb code to stay as close as possible to the cfish one (https://github.com/syzygy1/Cfish/blob/master/src/tbprobe.c) but note https://github.com/syzygy1/Cfish/issues/205?

cj5716:
@borg323 
this takes advantage of the pascal's triangle in the calculation of binomial coefficients as n+1Cr+1 = nCr+1 + nCr
e.g. 5c2 = 4c2 + 4c3

PR #1861, opened 2023-03-29
permanent fix for old ab nets

borg323:


mooskagh:
Maybe detect this case in `describenet` command too and print some message, like "Using a workaround for a legacy old-style ab net"?

For the case we'd want to keep this code in the release at all. But if there are no drawbacks, like misdetecting other networks with this code, then why not keep it.

borg323:
Sure we can keep it, shouldn't be an issue - and if it is we can always drop it later.  I marked this as ready for review to move it forward.

Regarding `describenet`, currently it only sees the net after it is updated. Maybe we should change this in order to be able to accurately describe every network file, even very old ones.

mooskagh (approved):

PR #1862, opened 2023-03-29
Optimize blas memory buffer use.

borg323:
This PR:
- Reduces number of temporary buffers used.
- Reuses the same buffers in subsequent computations.
- Dynamically adjusts the buffer size when needed.
- Stops swapping pointers.
- Removes a few unnecessary memory copies.

The result is a small but consistent speed-up.


SunnyWar:
I love these kinds of improvements.

src/neural/blas/network_blas.cc
```diff
@@ -134,13 +147,21 @@ class BlasComputation : public NetworkComputation {
   ActivationFunction ffn_activation_;
   bool attn_policy_;
   bool attn_body_;
+
+  static std::mutex buffers_lock_;
```
mooskagh, line 151: Why making them static rather than the member variable of `BlasNetwork`?
With static, it will not work if e.g. we create two instances of BlasNetwork (happens during selfplay, for example).

mooskagh (approved):

PR #1864, opened 2023-04-11
Rename cuda activations

almaudoh:
This PR simplifies #1821. Activation function renames in cuda files and merging with ActivationFunction enum in shared activation.h file.


borg323:
This is no longer needed as #1821 was merged as is.

PR #1868, opened 2023-04-18
Update describenet for new net archs

almaudoh:
Update describenet to properly handle attention body and attention policy nets.


borg323 (commented):
Can you also add default activation and whether SELU is used in attention policy?

src/lc0ctl/describenet.cc
```diff
@@ -129,28 +143,67 @@ void ShowNetworkWeightsInfo(const pblczero::Net& weights) {
...
-  COUT << Justify("Filters")
-       << w.input().weights().params().size() / 2 / 112 / 9;
-  COUT << Justify("Policy") << (w.has_policy1() ? "Convolution" : "Dense");
-  if (!w.has_policy1()) {
-    int policy_channels = w.policy().biases().params().size() / 2;
-    if (policy_channels == 0) {
-      policy_channels = w.policy().bn_means().params().size() / 2;
+
+  const auto attn_body = w.encoder_size() > 0;
+  if (attn_body) {
```
mooskagh, line 148: Optional nit: this function is becoming long, and it's easy to split it, maybe do that?

src/lc0ctl/describenet.cc
```diff
@@ -129,28 +143,67 @@ void ShowNetworkWeightsInfo(const pblczero::Net& weights) {
...
+           << w.pol_encoder(0).mha().q_b().params().size() / 2;
+      COUT << Justify("Policy encoder DFF")
+           << w.pol_encoder(0).ffn().dense1_b().params().size() / 2;
+    }
+    COUT << Justify("Policy DModel") << w.ip2_pol_b().params().size() / 2;
+    if (attn_body) {
+      COUT << Justify("Policy activation")
+           << NetworkFormat::ActivationFunction_Name(
+                  NetworkFormat::ACTIVATION_DEFAULT);
+      COUT << Justify("Policy FFN activation")
```
borg323, line 184: The "Policy FFN activation" should only be printed if there are policy encoders. Probably cleaner to move "Policy activation" before the `if (format.policy() == NetworkFormat::POLICY_ATTENTION)`.

src/lc0ctl/describenet.cc
almaudoh: Done

borg323 (approved):

PR #1869, opened 2023-04-19
add a mac build to circleci

borg323:
This PR does several things:
1. Adds mac builds to circleci, generating a universal (x86_64 and arm64) macos binary.
2. Removes clang linux builds (as the mac uses clang) and stops storing ubuntu 18.04 binaries that nobody uses.
3. Fixes non-cross arm builds with ispc.


cross-files/aarch64-darwin
```diff
@@ -0,0 +1,27 @@
...
+ld = 'ld'
+
+[built-in options]
+c_args = ['-arch', 'arm64']
+cpp_args = ['-arch', 'arm64']
+objc_args = ['-arch', 'arm64']
+objcpp_args = ['-arch', 'arm64']
+c_link_args = ['-arch', 'arm64']
+cpp_link_args = ['-arch', 'arm64']
+objc_link__args = ['-arch', 'arm64']
```
almaudoh, line 26: Seems like the double underscore might be a typo.

borg323: Indeed, so this is not needed - I'll leave it in for future reference.

gsobala:
Meson fails for me as `host_machine.cpu_family() == 'arm'` evaluates as false on Macbook Pro - needs aarch64

See <https://mesonbuild.com/Reference-tables.html#cpu-families>

borg323:
> Meson fails for me as `host_machine.cpu_family() == 'arm'` evaluates as false on Macbook Pro - needs aarch64
> 
> See https://mesonbuild.com/Reference-tables.html#cpu-families
Fixed

mooskagh (approved):

PR #1885, opened 2023-05-10
simplify cuda architectures used

borg323:
Newer versions of nvcc have a `-arch=all-major` option that compiles code for all `sm_?0` architectures. Since we still need to support earlier nvcc versions, this PR does the same in a roundabout way for fp16 code, and should considerably speed up compilation.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -500,22 +500,25 @@ if get_option('build_backends')
     )
 
     # Handling of fp16 cuda code.
-    nvcc_sm_list = ['80', '75', '86', '70', '89', '90']
+    nvcc_arch = '-arch=compute_70'
+    nvcc_sm_list = ['sm_70', 'sm_80', 'sm_90']
     if host_machine.system() != 'windows'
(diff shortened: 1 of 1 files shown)


meson.build
```diff
@@ -500,22 +500,25 @@ if get_option('build_backends')
...
-      nvcc_sm_list += ['60']
+      nvcc_arch = '-arch=compute_60'
+      nvcc_sm_list += ['sm_60']
       if ['arm', 'aarch64'].contains(host_machine.cpu_family())
-        # Add Jetson versions to the list.
-        message('Jetson support enabled.')
-        nvcc_sm_list += ['72', '62', '53', '87']
+        message('Compiling for Jetson.')
+        nvcc_arch = '-arch=compute_53'
+        nvcc_sm_list = ['sm_53', 'sm_62', 'sm_72']
```
Naphthalin, line 511: What happened to the 87?

cn4750 (approved):

PR #1891, opened 2023-05-28
Modify legacy TM to allow a bonus for first move if we expect shallow openings

Naphthalin:
adds a first-move-bonus to legacy time manager to accompany book-ply-bonus for shallow openings especially in FRC and DFRC.

Reasoning: Due to high % of tree reuse and saved time from smart pruning, LTC games at TCEC without spending more time on the first move have nearly linearly increasing node counts for the first 8-12 moves (see #1493). When playing chess from regular startpos, this isn't an issue due to policy head agreeing with value head. However, we can't rely on policy in FRC and DFRC, so having an option which does the same without book exits would be required. Based on the 8-12 move observation which is (nearly) made equal for 12+ ply exits with `book-ply-bonus=0.4`, I recommend using `first-move-bonus=5.0`.

### DIFF
--- a/src/mcts/stoppers/legacy.cc
+++ b/src/mcts/stoppers/legacy.cc
@@ -72,7 +72,7 @@ class LegacyTimeManager : public TimeManager {
             params.GetOrDefault<float>("midpoint-move", 51.5f)),
         time_curve_steepness_(params.GetOrDefault<float>("steepness", 7.0f)),
         spend_saved_time_(params.GetOrDefault<float>("immediate-use", 1.0f)),
-        first_move_bonus_(params.GetOrDefault<float>("first-move-bonus", 0.0f)),
+        first_move_bonus_(params.GetOrDefault<float>("first-move-bonus", 1.8f)),
         book_ply_bonus_(params.GetOrDefault<float>("book-ply-bonus", 0.25f)) {}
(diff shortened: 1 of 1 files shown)


src/mcts/stoppers/legacy.cc
```diff
@@ -72,6 +72,7 @@ class LegacyTimeManager : public TimeManager {
             params.GetOrDefault<float>("midpoint-move", 51.5f)),
         time_curve_steepness_(params.GetOrDefault<float>("steepness", 7.0f)),
         spend_saved_time_(params.GetOrDefault<float>("immediate-use", 1.0f)),
+        first_move_bonus_(params.GetOrDefault<float>("first-move-bonus", 0.0f)),
```
borg323, line 75: I think the default here should be between 1 and 2, to give some boost in very short openings. In #645 we settled to the equivalent of 0.6, but that was probably too conservative.

Naphthalin: Yeah, probably a good idea to set the default to something like that. 1.8 would be exactly the difference between bookplybonus 0.25 (default) and 0.4 (what we use).

borg323: Sounds good.

borg323 (approved):

PR #1894, opened 2023-06-04
Drawscore simplification for Armageddon

Naphthalin:
Originally, the drawscore implementation #1066 was intended for both contempt/humbleness and Armageddon support. Since then it turned out however that
1. the parametrization doesn't even span all possible combinations
2. the effect on opening selection etc is very miniscule
3. we were using symmetric settings between white and black anyway
As we have a working contempt implementation now with #1791, we can reduce drawscore to the one remaining use case, where Armageddon is one extreme case, and we only need 1 parameter for that instead of 4.

### DIFF
--- a/src/mcts/params.cc
+++ b/src/mcts/params.cc
@@ -360,18 +360,10 @@ const OptionId SearchParams::kMaxConcurrentSearchersId{
     "max-concurrent-searchers", "MaxConcurrentSearchers",
     "If not 0, at most this many search workers can be gathering minibatches "
     "at once."};
-const OptionId SearchParams::kDrawScoreSidetomoveId{
-    "draw-score-sidetomove", "DrawScoreSideToMove",
-    "Score of a drawn game, as seen by a player making the move."};
-const OptionId SearchParams::kDrawScoreOpponentId{
-    "draw-score-opponent", "DrawScoreOpponent",
-    "Score of a drawn game, as seen by the opponent."};
-const OptionId SearchParams::kDrawScoreWhiteId{
-    "draw-score-white", "DrawScoreWhite",
-    "Adjustment, added to a draw score of a white player."};
-const OptionId SearchParams::kDrawScoreBlackId{
-    "draw-score-black", "DrawScoreBlack",
-    "Adjustment, added to a draw score of a black player."};
+const OptionId SearchParams::kDrawScoreId{
+    "draw-score", "DrawScore",
+    "Adjustment of the draw score from white's perspective. For Armageddon, "
+    "use -100."};
 const OptionId SearchParams::kContemptPerspectiveId{
     "contempt-perspective", "ContemptPerspective",
     "Affects the way asymmetric WDL parameters are applied. Default is "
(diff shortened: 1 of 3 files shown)


src/mcts/params.cc
```diff
@@ -538,10 +530,7 @@ void SearchParams::Populate(OptionsParser* options) {
       -0.6521f;
   options->Add<BoolOption>(kDisplayCacheUsageId) = false;
   options->Add<IntOption>(kMaxConcurrentSearchersId, 0, 128) = 1;
-  options->Add<IntOption>(kDrawScoreSidetomoveId, -100, 100) = 0;
-  options->Add<IntOption>(kDrawScoreOpponentId, -100, 100) = 0;
-  options->Add<IntOption>(kDrawScoreWhiteId, -100, 100) = 0;
-  options->Add<IntOption>(kDrawScoreBlackId, -100, 100) = 0;
+  options->Add<IntOption>(kDrawScoreId, -100, 100) = 0;
```
borg323, line 533: Since we are overhauling this, why not make it a float option?

src/mcts/params.cc
```diff
@@ -360,18 +360,10 @@ const OptionId SearchParams::kMaxConcurrentSearchersId{
...
-const OptionId SearchParams::kDrawScoreWhiteId{
-    "draw-score-white", "DrawScoreWhite",
-    "Adjustment, added to a draw score of a white player."};
-const OptionId SearchParams::kDrawScoreBlackId{
-    "draw-score-black", "DrawScoreBlack",
-    "Adjustment, added to a draw score of a black player."};
+const OptionId SearchParams::kDrawScoreId{
+    "draw-score", "DrawScore",
+    "Adjustment of the draw score from white's perspective. For Armageddon, "
+    "use -100."};
```
borg323, line 366: I remember some discussion that somewhat more modest values may be better. Maybe this needs to be reworded to hint at this?

src/mcts/params.cc
Naphthalin: Fully agree, and already considered that. Main question is whether the -100...100 still makes sense then, or whether we should switch to -1.0 ... 1.0

src/mcts/params.cc
Naphthalin: The discussion I remember is without WDL adjustment (softening/contempt), where the issue is that from startpos, optimizing pure W is a serious/desperate uphill battle due to W starting out at 25%. Ideally we could come up with reasonable testing ideas to determine whether a more moderate value gives better/more realistic performance, but my intuition is that this was only due to pure WDL having so low white chances to begin with.

src/mcts/params.cc
borg323: Only [-1.0, 1.0] makes sense as a range to me, I hadn't considered anything else. I went back to find the justification for going to integer values, and it was that it looked better in guis, which I suppose is reasonable for 4 values that need to be set together.

src/mcts/params.cc
borg323: Isn't this still true for default settings? I which case maybe we could `add a minimal suggestion for other setting to change? Something like "... and for best results adjust XXX and YYY for the opponent."

Naphthalin: Yep, you're right. Maybe rewording this into `value -1 gives Armageddon scoring` is better to make it less of a "this is sufficient for Armageddon performance" comment? Ultimately, I think we should deliver example configs with the Lc0 releases, and Armageddon mode would be a prime example for that.

borg323 (approved):

src/mcts/params.cc
```diff
@@ -360,18 +360,10 @@ const OptionId SearchParams::kMaxConcurrentSearchersId{
...
-const OptionId SearchParams::kDrawScoreWhiteId{
-    "draw-score-white", "DrawScoreWhite",
-    "Adjustment, added to a draw score of a white player."};
-const OptionId SearchParams::kDrawScoreBlackId{
-    "draw-score-black", "DrawScoreBlack",
-    "Adjustment, added to a draw score of a black player."};
+const OptionId SearchParams::kDrawScoreId{
+    "draw-score", "DrawScore",
+    "Adjustment of the draw score from white's perspective. Value 0 gives "
+    "standard scoring, value -1 gives Armageddon scoring."};
```
yuzisee, line 366: If there's going to be "value X for Armageddon" type examples here, should https://github.com/LeelaChessZero/lc0/blob/78155977f745aafac7042a4ac7797096dd2a2656/src/mcts/params.cc#L397-L401 get something similar?

Naphthalin: In general I'd prefer to not mix the different PR topics.

About your question: There will be a longer blog post for the next release, explaining the features, the parameters and their meaning in much more detail than a few lines of code (which people using a GUI won't even see) could possibly transport.

borg323 (approved):

PR #1890, opened 2023-05-25
Use back-off spin to wait for available searcher

pdeng6:
This change is to improve performance in high core count scenario, and reduce CPU utilization.

**Environment:**
Hardware:
Intel Icelake 2 sockets/80 physical cores/160 logical cores platform

OS:
CentOS Stream9 system

Benchmark:
./lc0 benchmark  -b eigen --threads=$num_threads -w b30e742bcfd905815e0e7dbd4e1bafb41ade748f85d006b8e28758f1a3107ae3 --num-positions=34--movetime=-1 --nodes=26000

**Problems**:
1. Spin wastes CPU cycles, especially when other function/application also need CPU.
2. Even no other function/application asks CPU, high contention spin downgrades the performance
example:
When num_threads=128, pending_searchers_ spin waiting takes ~18% cpu cycles.
When num_threads=160, this spin takes ~88% cpu cycles, and score goes down to ~50% of num_threads=128

**Solution in this Change**:
Exponential back-off spin with sleep to reduce contention on pending_searchers_, and release CPU in when it is not available for a long time.
By this change:
1. CPU utilization can be reduced as num_threads from 64 to 160.
2. Score generally keeps flat before 128, while, improves from 128 to 160.

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/pdeng6/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/pdeng6/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">

</head>

<body link="#0563C1" vlink="#954F72">

opt/base -1 | 8 | 16 | 32 | 64 | 128 | 160
-- | -- | -- | -- | -- | -- | --
CPU   utilization(count) change | 0% | 0% | -1% | -3% | -13% | -36%
Performance improvement | 0% | 1% | -1% | 1% | 0% | 67%

</body>

</html>

PR #1895, opened 2023-06-05
rename contempt perspective to contempt mode

borg323:
This renames the `ContemptPerspective` uci option to `ContemptMode` (and the command line option) with new option names:
* `play`
* `white_side_analysis`
* `black_side_analysis`
* `disable`

In contrast to the current version, `disable` only forces `Contempt` to zero, to keep `WDL_mu` score type and WDL sharpening working.

### DIFF
--- a/src/mcts/search.cc
+++ b/src/mcts/search.cc
@@ -40,9 +40,9 @@
 #include "mcts/node.h"
 #include "neural/cache.h"
 #include "neural/encoder.h"
-#include "utils/spinhelper.h"
 #include "utils/fastmath.h"
 #include "utils/random.h"
+#include "utils/spinhelper.h"
 
 namespace lczero {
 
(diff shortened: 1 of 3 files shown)


src/mcts/search.cc
```diff
@@ -282,13 +282,13 @@ void Search::SendUciInfo() REQUIRES(nodes_mutex_) REQUIRES(counters_mutex_) {
...
     float mu_uci = 0.0f;
     // Only the diff effect is inverted, so we only need to call if diff != 0.
-    if (params_.GetContemptPerspective() != ContemptPerspective::NONE) {
-      auto sign =
-          ((params_.GetContemptPerspective() == ContemptPerspective::STM) ||
-           ((params_.GetContemptPerspective() == ContemptPerspective::BLACK) ==
-            played_history_.IsBlackToMove()))
-              ? 1.0f
-              : -1.0f;
+    if (params_.GetWDLRescaleDiff() != 0.0f) {
```
Naphthalin, line 285: This check will cause `WDL_mu` to fail whenever contempt=0 is set, irrespective of the ContemptMode set.

borg323: I looked at the comment above (`// Only the diff effect is inverted, so we only need to call if diff != 0.`), and a brief look through the code seemed to confirm it, obviously I was just confused. Should I remove the comment as well?

Naphthalin: Yes, comment predates the `mu_uci` population, so it is confusing (and wrong) by now.

src/mcts/search.cc
```diff
@@ -40,9 +40,9 @@
 #include "mcts/node.h"
 #include "neural/cache.h"
 #include "neural/encoder.h"
-#include "utils/spinhelper.h"
 #include "utils/fastmath.h"
 #include "utils/random.h"
+#include "utils/spinhelper.h"
```
Naphthalin, line 45: Just checking that this is intentional here, as it isn't related to contempt.

borg323: There were some minor style issues in #1890 that I didn't notice when reviewing, but clang-format fixed.

PR #1898, opened 2023-06-30
handle play contempt mode with infinite search

borg323:
Makes sure play contempt mode is never used inside the search, it is switched to white/black depending on the first move side, same in normal searches, flipped when pondering or disabled in infinite search. For infinite search there is a warning if contempt would have an effect.
In addition there is a fix for a preexisting wdl display bug, w and l were swapped while pondering.
Also some cleanups.

### DIFF
--- a/src/benchmark/benchmark.cc
+++ b/src/benchmark/benchmark.cc
@@ -106,8 +106,8 @@ void Benchmark::Run() {
           std::make_unique<CallbackUciResponder>(
               std::bind(&Benchmark::OnBestMove, this, std::placeholders::_1),
               std::bind(&Benchmark::OnInfo, this, std::placeholders::_1)),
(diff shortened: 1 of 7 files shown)


Naphthalin:
I believe I tested everything which is supposed to be tested, which includes
- behaviour of `go nodes`, `go ponder`, `ponderhit`, `go infinite`, `go movetime` with `--contempt-mode=play`
- behaviour of `go nodes`, `go ponder`, `ponderhit`, `go infinite` with contempt mode `white_side_analysis` and `black_side_analysis`
- WDL sharpening active with `disable`
- that contempt is automatically set to 0 with `disable`
- warning is only printed with `contempt != 0` and `play`
- WDL and eval display is now consistent with `go ponder` and `ponderhit`
- `WDL_mu` is functional with `--contempt-mode=disable`

Naphthalin (approved):

PR #1899, opened 2023-07-08
change default values for v0.30 release

Naphthalin:
Changed default values for v0.30 release:
- mlh-threshold to 0.8 as we now have a continuous behaviour. Training runs explicitly set threshold to 0, so no issues there.
- centipawn fallback for WDL_mu to only kick in if W or L < 0.2%, which should happen virtually never. Will still behave jumpy if around that threshold and needs a proper fix at some point, but eval at this point doesn't matter anyway anymore, 
- default scoretype to WDL_mu. Thanks to #1898, `WDL_mu` now is always calculated properly (even if contempt is disabled), and we use it everywhere anyway.

### DIFF
--- a/src/mcts/params.cc
+++ b/src/mcts/params.cc
@@ -510,11 +510,11 @@ void SearchParams::Populate(OptionsParser* options) {
                                          "Q",
                                          "W-L",
                                          "WDL_mu"};
-  options->Add<ChoiceOption>(kScoreTypeId, score_type) = "centipawn";
+  options->Add<ChoiceOption>(kScoreTypeId, score_type) = "WDL_mu";
   std::vector<std::string> history_fill_opt{"no", "fen_only", "always"};
   options->Add<ChoiceOption>(kHistoryFillId, history_fill_opt) = "fen_only";
   options->Add<FloatOption>(kMovesLeftMaxEffectId, 0.0f, 1.0f) = 0.0345f;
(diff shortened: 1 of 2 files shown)


yuzisee:
Out of curiosity, are https://lczero.org/play/flags/ and https://lczero.org/dev/wiki/lc0-options/ treated differently in terms of when each can be updated after a new release? Neither mentions `WDL_mu` (yet) but as far as I can tell: one of them requires a pull request, whereas anyone (even me) can edit the other.

mooskagh:
> Out of curiosity, are [lczero.org/play/flags](https://lczero.org/play/flags/) and [lczero.org/dev/wiki/lc0-options](https://lczero.org/dev/wiki/lc0-options/) treated differently in terms of when each can (should) be updated after a new release? Neither mentions `WDL_mu` (yet) but as far as I can tell: one of them requires a pull request, whereas anyone (even me) can edit the other.

The /wiki/ pages are imported from https://github.com/LeelaChessZero/lc0/wiki, while other pages (btw [this](https://github.com/LeelaChessZero/lczero.org/edit/master/content/play/configuration/flags/index.md) is the correct link to create the PR) require git repository change.
Both pages are intended to be easily changeable by everyone, but we had an incident of a spammer secretly adding links to malware to the page, so now changes have to go through the second pair of eyes (and it's generally useful, e.g. to spot spelling mistakes).

PR #1905, opened 2023-07-28
rework selection of cuda build architecture

borg323:
This pull request consists of 3 related parts:
1. Adds support for using `-arch=native` for cuda code (if nvcc supports it). This is used by default unless `-Dnative_cuda=false` is passed to the build or code generation for a specific CC is requested with e.g. `-Dcc_cuda=50`.
2. Disables fp16 cuda code that breaks the build when compiling for a CC less than 5.3 - this is needed to make `-arch=native` work even on older gpus.
3. Passes `-Dnative_cuda=false` to appveyor builds in order to generate code for multiple gpu architectures.


mooskagh (approved):

whelanh:
Compiling this using Windows, I received the following error message:
meson.build:507:26: ERROR: The `==` operator of str does not accept objects of type list ([])
[meson-log.txt](https://github.com/LeelaChessZero/lc0/files/12208311/meson-log.txt)

borg323:
@whelanh can you try the following fix?
```diff
--- a/meson.build
+++ b/meson.build
@@ -484,7 +484,7 @@ if get_option('build_backends')
     if cuda_cc != ''
       nvcc_extra_args = ['-arch=compute_' + cuda_cc, '-code=sm_' + cuda_cc]
     elif get_option('native_cuda') and nvcc_help.contains('-arch=native')
-      nvcc_extra_args = '-arch=native'
+      nvcc_extra_args = ['-arch=native']
     endif
     foreach x : get_option('cudnn_include')
       cuda_arguments += ['-I', x]
```

whelanh:
I did try that correction of line 487 changing to nvcc_extra_args =
['-arch=native'] and it runs fine without any errors.
Kind regards,
Hugh Whelan

PR #1917, opened 2023-09-13
Multiple head architecture for cuda backend

almaudoh:


src/neural/cuda/network_cudnn.cc
```diff
@@ -950,7 +954,7 @@ class CudnnNetwork : public Network {
   // Apparently nvcc doesn't see constructor invocations through make_unique.
   // This function invokes constructor just to please complier and silence
   // warning. Is never called (but compiler thinks that it could).
-  void UglyFunctionToSilenceNvccWarning() { InputsOutputs io(0, false, false); }
+  void UglyFunctionToSilenceNvccWarning() { InputsOutputs io(0, false, false, false); }
 
  private:
   const NetworkCapabilities capabilities_;
```
borg323, line 960: Later in the file, after line 1098: `case pblczero::NetworkFormat::NETWORK_ATTENTIONBODY_WITH_HEADFORMAT:` you should add the `case pblczero::NetworkFormat::NETWORK_ATTENTIONBODY_WITH_MULTIHEADFORMAT:` to switch to cuda.

src/neural/cuda/network_cuda.cc
```diff
@@ -359,6 +356,22 @@ class CudaNetwork : public Network {
 
     ActivationFunction act = mish_net ? ACTIVATION_MISH : ACTIVATION_RELU;
 
+    std::string policy_head =
+        options.GetOrDefault<std::string>("policy_head", "vanilla");
+    // Check that selected policy head exists.
+    if (policy_head == "optimistic") policy_head = "optimistic_st";
```
borg323, line 362: Better rename the map key in `network_legacy.cc` to `optimistic`.

src/neural/cuda/layers.cc
```diff
@@ -2073,12 +2104,13 @@ AttentionBody<DataType>::AttentionBody(const LegacyWeights& weights,
...
   int num_encoders = weights.encoder.size();
-  float alpha = (float)pow(2.0 * num_encoders, 0.25);
+  float alpha = (float)pow(2.0 * num_encoders, -0.25);
   for (const auto& enc : weights.encoder) {
     EncoderBlock<DataType>* pW = new EncoderBlock<DataType>(
         enc, scratch, encoder_head_count_, embedding_op_size_, alpha,
         smolgen_global_, smolgen_global_size_, max_batch_size,
-        activations_.smolgen_activation, activations_.ffn_activation);
+        activations_.smolgen_activation, activations_.ffn_activation,
+        new_encoding_ ? 1e-3 : 1e-6);
```
borg323, line 2113: Should the 1e-6 be scaled by depth for true equivalence with the old code?

src/neural/cuda/network_cudnn.cc
almaudoh: Done

src/neural/cuda/layers.cc
borg323: After doing some experiments, the scaling results in even lower epsilon values, so not worth the effort.

src/neural/cuda/network_cuda.cc
```diff
@@ -49,24 +49,25 @@ using namespace cudnn_backend;
 template <typename DataType>
 class CudaNetwork;
 
-static size_t getMaxAttentionHeadSize(const LegacyWeights& weights, int N) {
-  const size_t embedding_op_size = weights.ip_pol_b.size();
-  const size_t policy_d_model = weights.ip2_pol_b.size();
-  assert(policy_d_model == weights.ip3_pol_b.size());
+static size_t getMaxAttentionHeadSize(const MultiHeadWeights& weights, int N) {
```
borg323, line 52: I think a better approach is to pass a `MultiHeadWeights::PolicyHead&` instead of `LegacyWeights&`.

almaudoh: Done

src/neural/cuda/inputs_outputs.h
```diff
@@ -134,6 +136,8 @@ struct InputsOutputs {
 
   // cublas handle used to run the network
   cublasHandle_t cublas_;
+
+  bool has_moves_left_ = false;
```
mooskagh, line 140: Wouldn't it be possible to just check op_moves_left_mem_ to be nullptr? (and initialize it like that)

almaudoh: I did not find a very good way to test if the `op_moves_left_mem_` has been allocated. `op_moves_left_mem_ != nullptr` doesn't work as even the unallocated one still gets a pointer value. The only thing I have seen is that `op_moves_left_mem_ == op_moves_left_mem_gpu_` for cases where it has been allocated (and would be different if not allocated). Though I'm not sure if that is a safe or good way to go about it.

almaudoh: I didn't understand the comment initially. Fixed now. Initialized to `nullptr` and then used as the check for deallocation.

borg323 (approved):

Issue #1736, opened 2022-04-21
pytorch doesn't like onnx models that leela2onnx produces


mooskagh:
Code to reproduce.
```python
from onnx2pytorch import ConvertModel
onnx_model = onnx.load("generated_network.onnx")
onnx.checker.check_model(onnx_model)
pytorch_model = ConvertModel(onnx_model)
```
Error:
```
KeyError 'kernel_size'
```

Issue #1735, opened 2022-04-21
ONNX node names should not have leading slash


mooskagh:
Apparently, Keras doesn't like such onnx files, giving the error:

```
ValueError: '/inputconv_pad/' is not a valid root scope name. A root scope name has to match the following pattern: ^[A-Za-z0-9.][A-Za-z0-9_.\\/>-]*$
```

Fix is this:
Remove trailing slashes in this file https://github.com/LeelaChessZero/lc0/blob/master/src/neural/onnx/converter.cc
(basically, replace all `"/` which are not part of `+ "/` with `"`).

mooskagh:
Code to repro keras error:

```
from onnx2pytorch import ConvertModel
onnx_model = onnx.load("generated_network.onnx")
k_model = onnx_to_keras(onnx_model, ['/input/planes'])
```

mooskagh:
After removing leading slashes, next error is
```
ValueError: Layer block0/conv2/se/matmul1 weight shape (1, 32) is not compatible with provided weight shape (384, 32).
```

It seems that onnx2keras just doesn't understand onnx well, I'm inclined not to push the slash removal (https://github.com/mooskagh/lc0/tree/onnx_slash) either.

mooskagh:
I considered filing bugs in the onnx2keras repo, but the projects doesn't look alive, last commit was a year ago, probably not worth it.

https://github.com/gmalivenko/onnx2keras

PR #1924, opened 2023-10-10
Make leela2onnx Conv-nodes compatible with onnx2pytorch

patrik-ha:
Adds `kernel_shape` as a property to all generated Conv-nodes for generated ONNX-networks, to partially address #1736 . This was specifically done for the onnx2pytorch-package requiring. [The ONNX-standard](https://onnx.ai/onnx/operators/onnx__Conv.html) states that this field is optional, and if not provided it should be inferred just looking at the dimensionality of the weights. These changes make _some_ networks generated by `leela2onnx` readable by `onnx2pytorch`. 

I verified that it works with the network that comes bundled with the latest lc0-release (791556) at least. However, many of the networks fail since `onnx2pytorch` hasn't implemented layer-normalisation, (as was the case with many of the networks I tried (mostly from https://lczero.org/play/networks/bestnets/). I also considered making a PR over there implementing that, [but it seems pretty dead](https://github.com/Talmaj/onnx2pytorch).

I'm not sure if it is correct that this project should add stuff like this to accomodate such requirements in other projects. It is practical that it works with other software that people are likely to use, especially in this case since it's quite a small change (and follows an optional part of the standard). I'm a big advocate for making the barrier as low as possible for other people to tinker with lc0-networks (and for that it's nice that conversion between pytorch and ONNX works as well as it could), but probably not at the cost of more bloat. I haven't done any large contributions, so that would be up to someone else to judge, but I wouldn't lose any sleep if this wasn't merged.

### DIFF
--- a/src/neural/onnx/builder.cc
+++ b/src/neural/onnx/builder.cc
@@ -440,4 +440,40 @@ std::string OnnxBuilder::Mish(const std::string& name,
   return PopulateStdNodeFields(node, name, input, "Mish");
 }
 
+std::string OnnxBuilder::Sqrt(const std::string& name,
+                              const std::string& input) {
+  auto* node = model_.mutable_graph()->add_node();
+  return PopulateStdNodeFields(node, name, input, "Sqrt");
+}
+
+std::string OnnxBuilder::Reciprocal(const std::string& name,
+                                    const std::string& input) {
+  auto* node = model_.mutable_graph()->add_node();
+  return PopulateStdNodeFields(node, name, input, "Reciprocal");
+}
+
+std::string OnnxBuilder::Cast(const std::string& name, const std::string& input,
+                              pblczero::TensorProto::DataType type) {
+  auto* node = model_.mutable_graph()->add_node();
+  auto out = PopulateStdNodeFields(node, name, input, "Cast");
+  AddIntAttribute(node, "to", type);
+  return out;
+}
+
+std::string OnnxBuilder::ReduceMean(const std::string& name,
+                                    const std::string& input,
+                                    std::initializer_list<int> axes) {
+  auto* node = model_.mutable_graph()->add_node();
+  auto out = PopulateStdNodeFields(node, name, input, "ReduceMean");
+  if (opset_ < 18) {
+    AddIntsAttribute(node, "axes", axes);
+  } else {
+    node->add_input(AddInitializer(
+        name + "/axes",
+        Int64OnnxConst(std::vector<int64_t>(begin(axes), end(axes)),
+                       {static_cast<int>(axes.size())})));
+  }
+  return out;
+}
+
 }  // namespace lczero
--- a/src/neural/onnx/builder.cc
+++ b/src/neural/onnx/builder.cc
@@ -362,6 +362,21 @@ std::string OnnxBuilder::Sigmoid(const std::string& name,
   return PopulateStdNodeFields(node, name, input, "Sigmoid");
 }
 
+// This is only defined in opset 17 but onnxruntime supports it from 1.
+std::string OnnxBuilder::LayerNormalization(const std::string& name,
+                                            const std::string& input,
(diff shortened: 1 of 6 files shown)


src/neural/onnx/builder.cc
```diff
@@ -135,12 +135,15 @@ std::string PopulateStdNodeFields(pblczero::NodeProto* node,
 std::string OnnxBuilder::Conv(const std::string& name,
                               const std::string& input_name,
                               const OnnxConst& kernel_weights,
-                              const OnnxConst& bias_weights, int pads) {
+                              const OnnxConst& bias_weights, 
+                              int filters,
```
borg323, line 139: Why not just do a 'kernel_weights.GetDimensions().back()` instead of passing the number of filters?

borg323:
You can find an optional LayerNormalization replacement in https://github.com/borg323/lc0/commit/bc18becbc32241502ec62a94b43301c26713f76f

src/neural/onnx/builder.cc
patrik-ha: Agreed, that's better

patrik-ha:
Thanks for that. I confirmed that all the nets I tested can now be loaded correctly by onnx2pytorch using your alternative LayerNorm-definition. 

Is it worth considering making this the default implementation of LayerNorm for the ONNX converter? It introduces more complexity and makes the ONNX graph more complex to look at I guess, but since it's just the same operation defined more explicitly it's nice that it just works without any extra hassle. But that might again come down to how much one wants to accomodate projects like onnx2pytorch. (a somewhat dead project, but still the first thing that most people run into when they want to convert leela-onnx nets to pytorch)

I pushed some changes making that implementation the default for LayerNorm to this branch, but that can be ignored if it shouldn't be used.

borg323:
The last commit went too far, removing the `LayerNormalization()` operator unconditionally. I think the best way is to add a `--onnx2pytorch` switch to leele2onnx to control this. And while you are there, are there any other converter options that could be useful to make available to the user?

patrik-ha:
Yeah I understand, I've reverted that and put it behind a switch instead. I've made the switch for the user explicitly for onnx2pytorch, but internally I think it's nicer that that option toggles whatever needs to be different for the converter to make networks that are compatible. (I.e. if the user sets onnx2pytorch, it internally enables alt_ln or whatever relevant behaviour that's needed. The main thing is that there shouldn't be an internal option just for onnx2pytorch/compatibility I mean)

Regarding other relevant things, I'm not completely sure, but I don't think so. I was thinking if it was possible to put behaviour solving #1735 behind a similar switch, but in that case it doesn't look viable anyway. It might be nice to add explicit switches to enable the alternate implementations of mish and layer-norm since they already have options ready to go internally, but I don't know if it's it's that relevant for most people. They could maybe be added as hidden to avoid cluttering --help.

borg323:
Do you think options to change the onnx opset or the ability to generate a fp16 modes (with either fp16 or fp32 input/outputs) would be useful?
Regarding the alternative implementations we already have, LayerNormalization is a special case since onnxruntime accepts it for every opset even though it was only added in opset 17, so a separate switch makes sense.
This is all unrelated to this PR, just asking here since you are using the onnx models generated by lc0.

src/neural/onnx/converter.h
```diff
@@ -44,6 +44,7 @@ struct WeightsToOnnxConverterOptions {
   int batch_size = -1;
   int opset = 17;
   bool alt_mish = false;
+  bool alt_ln = false;
```
mooskagh, line 47: Maybe rename it so something slightly less cryptic? (`alternative_layer_normalization` or `alt_layer_norm` or at the very least add a comment). And I'd say the same with the command line option.

patrik-ha: I agree, fixed

mooskagh (approved):

PR #1928, opened 2023-10-25
onnx2leela options to fix tf exported onnx models

borg323:


scripts/compile_proto.py
```diff
@@ -378,8 +378,10 @@ def GenerateFunctionDeclarations(self, w):
                 w.Write("void add_%s(%s val);" % (name, cpp_type))
             w.Write("const std::vector<%s>& %s() const;" %
                     (var_cpp_type, name))
+            w.Write("std::vector<%s>* mutable_%s();" % (var_cpp_type, name))
```
mooskagh, line 381: Note that this breaks the API compatibility with the standard protobuf library (as they have own RepeatedPtrField container instead of a vector, and actually store pointers there instead of values).
I believe it's fine to leave like this though.

borg323: Is there an easy way to keep compatibility?

mooskagh: What probably would work is

1. Create `class RepeatedPtrField` and `class RepeatedField` in `protomessage.h`, which would take std::vector<> as a constructor parameter and wrap it with this https://protobuf.dev/reference/cpp/api-docs/google.protobuf.repeated_field/ interface.
2. Return `RepeatedPtrField` and `RepeatedField` instead of a raw vector.

(not saying it's necessary to do in this PR)

mooskagh: (note that the standard protobuf library likely returns the reference to these classes, while we would have to return then by value as they are created on the fly).
Also, in the case of "mutable_" functions, likely it's expected that pointer is returned, which makes it even harder.. Maybe we can approximate it with std::unique_ptr<>..

src/lc0ctl/onnx2leela.cc
```diff
@@ -240,6 +317,11 @@ void ConvertOnnxToLeela() {
 
   const OptionsDict& dict = options_parser.GetOptionsDict();
 
+  auto onnx_model = ReadFileToString(dict.Get<std::string>(kInputFilenameId));
```
mooskagh, line 320: Maybe isolate the code that modifies the ONNX from the code that embeds it into the Leela proto?

I can see two ways of doing this.
1. Either have a function that modifies ONNX first:
```cpp
std::string MaybeUpdateONNX(std::string input) {
   if (!fix50rule && !fixwdl) return input;
   pblczero::ModelProto model;
   model.ParseFromString(input);
   // Do the fixes.
   return model.OutputToString();
}

onnx->set_model(MaybeUpdateONNX(ReadFileToString(dict.Get<std::string>(kInputFilenameId))));
```

2. Or embed first and then postprocess:
```cpp
void MaybeUpdateWeights(pblczero::Net* out_weights) {
  if (!fix50rule && !fixwdl) return;
  // Otherwise parse from out_weights.mutable_onnx_model()->onnx(), fix and overwrite.
}

// And call only in the very end:
MaybeUpdateWeights(&out_weights);
WriteStringToGzFile(dict.Get<std::string>(kOutputFilenameId),
                      out_weights.OutputAsString());
```

Strictly speaking, fixing ONNX is not really part of the onnx to leela conversion (it may be a separate command or even tool), but I agree it's more convenient here. Do you think it would make sense also to have another command just for fixing the network? (maybe potentially not only ONNX).

src/lc0ctl/onnx2leela.cc
```diff
@@ -230,6 +236,77 @@ bool ValidateNetwork(const pblczero::Net& weights) {
   return true;
 }
 
+const std::vector<float> rule50weights = {
```
mooskagh, line 239: Constants should have names with `k-` prefix, like `kRule50weights`.
But maybe it's better to create dynamically in function?

```cpp
std::vector<float> MakeRule50Weights() {
  std::vector res(122, 1.0f);
  res[120] = 1.0r / 00;
  return res;
}
```

src/lc0ctl/onnx2leela.cc
```diff
@@ -230,6 +236,77 @@ bool ValidateNetwork(const pblczero::Net& weights) {
...
+    for (size_t j = 0; j < node.input_size(); j++) {
+      if (node.input(j) == in) {
+        CERR << "Inerting scaling between " << in << " and " << node.name();
+        model.mutable_graph()->mutable_node(i)->mutable_input()->at(j) =
+            std::string(name);
+        flag = true;
+      }
+    }
+  }
+
```
mooskagh, line 262: Isn't it an error if we reach this place and the flag is false?
(also the variable name `flag` here and in other places, if still needed, should probably be replaced with better names)

src/lc0ctl/onnx2leela.cc
borg323: Done

src/lc0ctl/onnx2leela.cc
borg323: Done, this was a remnant from an older iteration.

src/lc0ctl/onnx2leela.cc
borg323: I originally structured the code to avoid multiple `ParseFromString()` calls, so I went with a variation of option 1 that allowed me to keep it like that.
 
Regarding a separate command to modify the onnx model, I don't think it makes sense as those are not general fixes but rather adjustments we do for differences between the tensorflow generated files and what our backend expects. So I don't see a use case a separate command. Even working around our limitations in handling models with external data requires two passes (either the current onnx2leela+leela2onnx or a hypothetical fix+onnx2leela flow), and even that should be fixed to work with just onnx2leela.

scripts/compile_proto.py
borg323: In the end I just added a comment documenting out deviation from the standard protobuf API.

mooskagh (approved):

PR #1929, opened 2023-11-09
 Improving the MUtility function

ContradNamiseb:
In this PR I'm introducing 2 new tune able parameters, only the MovesLeftThreshold parameter is left from the old MUtility function. What this PR will do is prioritize shorter wins and longer losses.
The first parameter is MovesLeftMidpointMove this controls where the m-value crosses the y-axis as in moves with greater value than this parameter will be negative. This is done so moves that have shorter movesleft-value than MovesLeftMidpointMove can be prioritized higher than moves with longer movesleft-value than MovesLeftMidpointMove I would recomend (100) for this parameter.
The second parameter is the MovesLeftSteepnessFactor this should be the more tune-able parameter to match hardware as this parameter will differ depending on the nps. This sets the steepness of the curve, which is how quickly the m-value changes as the moves-left decrease or how high or low the m-value goes in simple terms. The higher this value is the greater the m-effect on node Q will be which might cause aggressive play and missing some deeper tactic's that might cause losses.

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -2,24 +2,28 @@
 setlocal
 
 rem 1. Set the following for the options you want to build.
-set CUDNN=true
-set CUDA=true
+set CUDNN=false
+set CUDA=false
 set DX12=false
-set OPENCL=false
+set OPENCL=true
 set MKL=false
-set DNNL=false
-set OPENBLAS=false
-set EIGEN=false
+set DNNL=true
+set OPENBLAS=true
+set EIGEN=true
 set TEST=false
+set ONNX=true
+
 
 rem 2. Edit the paths for the build dependencies.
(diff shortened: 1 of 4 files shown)


build.cmd
```diff
@@ -2,24 +2,28 @@
 setlocal
```
borg323, line 2: The changes to build.cmd are unrelated to this PR.

ContradNamiseb: Removed.

mooskagh:
To proceed, this change at least needs to show that it doesn't cause Elo or nps regression.

And preferably, also some better math-grounded justification for the changes, maybe discuss with @Naphthalin as he is the author of the original MUtility.

PR #1930, opened 2023-11-16
auto setting of threads and task workers

borg323:
This is the second part of the auto-configuration changes. It doesn't change the behavior for any explicitly set value, but the new defaults lead to the following:
1. Threads are set according to the backend specified value, plus 1 for non-cpu backends.
2. The backends request 1 thread except for cuda with multi-stream, multiplexing and round-robin.
3. The task workers are set to 0 for cpu backends, otherwise to cpu_cores/(threads-1) up to a max of 4.
4. The multiplexing and roundrobin backends suggest for minibatch the min of the values of the used backends. For the demux backend this is multiplied by the number of (backend) threads.

This is changing the default for cpu backends to 1 thread and may reduce task-workers (apparently this is generally expected of cpu engines). Multiplexing backends should now get a decent starting configuration.

### DIFF
--- a/src/neural/network.h
+++ b/src/neural/network.h
@@ -106,7 +106,9 @@ class Network {
  public:
   virtual const NetworkCapabilities& GetCapabilities() const = 0;
   virtual std::unique_ptr<NetworkComputation> NewComputation() = 0;
+  virtual int GetThreads() const { return 1; }
   virtual void InitThread(int /*id*/) {}
+  virtual bool IsCpu() const { return false; }
   virtual int GetMiniBatchSize() const { return 256; }
   virtual ~Network() = default;
 };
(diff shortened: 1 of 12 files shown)


src/neural/network.h
```diff
@@ -106,7 +106,9 @@ class Network {
  public:
   virtual const NetworkCapabilities& GetCapabilities() const = 0;
   virtual std::unique_ptr<NetworkComputation> NewComputation() = 0;
+  virtual int GetThreads() const { return 1; }
```
mooskagh, line 109: (optional)
I'd suggest to move GetThreads(), IsCpu() and GetMinibatchSize() into the Capabilities struct.
(or have a separate struct for them if you feel that they don't really belong to Capabilities -- although then better rename "Capabilities" to "Properties").
Otherwise it's kind of spoiling the interface -- it's clearly the information leakage, but at least it would be isolated.

mooskagh (approved):

PR #1938, opened 2024-01-01
simple multi-gpu configuration

borg323:
This adds a `--gpus=x` (0<=x<=8) to specify the number of gpus to use, which is ignored if `--backend-opts` is used. Calling meson build with `-Dgpus_opt=false` disables this option, and this is done for the blas and android appveyor builds. 

Using `--gpus=1` (default) is equivalent to `--backend-opts=gpu=0`, which is mainly affecting the onednn backend, making the gpu first choice, so I added a fallback to cpu there, if the gpu is not available. 

The old behavior is obtained by using `--gpus=0`, In practice this only makes sense with the onednn backend.

For a higher value of gpus, a backend options line of the form `(backend=X,gpu=0),(backend=X,gpu=1)` and so on is constructed, with X being either the backend specified with `--backend` (unless it is `multiplexing`, `demux` or `roundrobin`) or the default backend. The multiplexing backend is used, unless `demux` or `roundrobin` is specified with `--backend`.


borg323:
I'm having second thoughts about the (ab)use of backend options, made it draft for now.

PR #1940, opened 2024-01-07
Rename "win_percentage" UCI option to "expected_score"

HeyItsLeux:
By definition, "win_percentage" is the percentage chance of that side winning the game.
This worked when Leela didn't have that D value, but now it does.
Simply changing this name to a definition that makes more sense, to any user.


mooskagh:
(being discussed in Discord here: https://discord.com/channels/425419482568196106/438088534755377204/1212403332623241278)

PR #1942, opened 2024-01-13
fix rook positions after clearing board

borg323:


src/chess/board.cc
```diff
@@ -51,6 +51,7 @@ const BitBoard ChessBoard::kPawnMask = 0x00FFFFFFFFFFFF00ULL;
 
 void ChessBoard::Clear() {
   std::memset(reinterpret_cast<void*>(this), 0, sizeof(ChessBoard));
+  castlings_.SetRookPositions(FILE_A, FILE_H, FILE_A, FILE_H);
```
mooskagh, line 54: Maybe this would be more error-safe:
add `ChessBoard& operator=(const ChessBoard&) = default;`, and then

```cpp
void ChessBoard::Clear() {
  *this = ChessBoard();
}
```

?

borg323: I like it.

mooskagh (approved):

Issue #1843, opened 2023-02-09
Various issues w/ setoption


uwuplant:
BUG REPORT

**Describe the bug**
The setoption command does not properly handle multiple name or value arguments.

**Steps to Reproduce**
1. Send the command `setoption value livelogstats value true`
Expected behavior: Throw an error that no name is given
Observed behavior: lc0 returns `error Flag '--preload' must be either 'true' or 'false'`
Additionally: `setoption name a name b` returns `error Unknown option: ab`.

borg323:
Note, the error for `setoption name a name b` should be `error Unknown option: a name b` i.e. everything after the first name until the first value should be treated as the option.
Also doing `setoption name MaxPrefetch value 1 value 2` is accepted and sets `MaxPrefetch` to 12, which is obviously bad. There are two reasonable alternatives here, either complain about a bad value `1 value 2` or about a wrong option `MaxPrefetch value 1` depending on whether we use the first or last `value` token as the delimiter. I would prefer using the first `value` as delimiter to allow using the word `value` in string options, e.g. as part of a path name with spaces like `setoption name LogFile value c:\best value for money\log.txt`

KarlKfoury:
Has this issue been fixed?

borg323:
No, the issue is still open.

KarlKfoury:
i believe #2062 fixes the issue

Probatio-Diabolica:
Hi is this issue open?

borg323:
Yes, we are doing some major refactoring of the code so some stuff is still pending.

PR #1945, opened 2024-01-19
fixe setoption behaviour.

ContradNamiseb:
This pr partially fixes issue #1843

### DIFF
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -377,7 +377,7 @@ void EngineLoop::CmdSetOption(const std::string& name, const std::string& value,
     if (name.empty() || value.empty()) {
         throw Exception("Both name and value must be provided");
     } else {
-        // Split the names and values by space
+        // Split the names and values by semicolon.
         std::vector<std::string> names = SplitString(name, ';');
         std::vector<std::string> values = SplitString(value, ';');
         // Check that the number of names and values match
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -30,8 +30,7 @@
 #include <algorithm>
 #include <cmath>
 #include <functional>
-#include <string>
-#include <vector>
+
 
 
 #include "mcts/search.h"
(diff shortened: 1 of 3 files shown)


ContradNamiseb:
Here is an e.g of how it now works:

 `      _
|   _ | |
|_ |_ |_| v0.31.0-dev+git.dirty built Jan 21 2024
Found configuration file: C:\Users\cenam\Documents\lc0_onnxtest/lc0.config
setoption name cpuct threads value 2.14 2
setoption name cpuct hash value 2.14 2
error Unknown option: hash
setoption name cpuct threads value 2.14 2
setoption name cpuct threads value 2.14
error Mismatched number of names and values
setoption name cpuct value
error Provide a value for the option
setoption name  value 3
error An option name must be provided
value livelogstats value true
error Unknown command: value livelogstats value true
setoption value livelogstats value true
error An option name must be provided`

mooskagh:
It doesn't look like the code does what you described.  
And I don't think that the behaviour you described is what's needed.

```
setoption name a name b value 1 value 2
```

should set an option "`a name b`" to value "`1 value 2`"

ContradNamiseb:
> It doesn't look like the code does what you described. And I don't think that the behaviour you described is what's needed.
> 
> ```
> setoption name a name b value 1 value 2
> ```
> 
> should set an option "`a name b`" to value "`1 value 2`"

ContradNamiseb:
Sorry meant to edit the comments not close the pr.  I changed the described behavior of the code after talking with borg he recommended to keep this current changes in the pr.

ContradNamiseb:
@mooskagh this is what I originally had.

```void OptionsParser::SetUciOption(const std::string& name,
                                 const std::string& value,
                                 const std::string& context) {
   if (name.empty()) {
      throw Exception("An option name must be provided");
    } else {
        // Split the names and values by space.
        std::vector<std::string> names = SplitString(name, ' ');
        std::vector<std::string> values = SplitString(value, ' ');
        // Check that the number of names and values match
        if (names.size() != values.size()) {
            throw Exception("Mismatched number of names and values");
        }
        // Set the UCI options for each name-value pair
        for (size_t i = 0; i < names.size(); ++i) {
            const std::string& name = names[i];
            const std::string& value = values[i];
            auto option = FindOptionByUciName(name);
            if (option) {
            option->SetValue(value, GetMutableOptions(context));
            } else {
            // If name is not valid throw an error.
            throw Exception("Unknown option: " + name);
            }
        }
        return;
    }
}```

mooskagh:
It doesn't look to me that the code snippet from your latest comment does what is needed either.

PR #1915, opened 2023-09-07
Common changes for new multiple heads architecture.

almaudoh:
Changes shared by all backends for the new BT3 multiple policy and value heads architecture.

PR #1948, opened 2024-02-10
onnx support for multihead arch

borg323:


src/neural/loader.cc
```diff
@@ -140,6 +209,37 @@ void FixOlderWeightsFile(WeightsFile* file) {
       net->set_smolgen_activation(pblczero::NetworkFormat::ACTIVATION_SWISH);
     }
   }
+
+  // Get updated network format.
+  network_format = file->format().network_format().network();
+  auto embedding_type = file->format().network_format().input_embedding();
+  bool multihead_format = (network_format & 128) == 128;
```
mooskagh, line 216: Make a helper function instead of `& 128`

src/neural/loader.cc
```diff
@@ -140,6 +209,37 @@ void FixOlderWeightsFile(WeightsFile* file) {
...
+  if (!multihead_format && network_format != nf::NETWORK_ONNX) {
```
mooskagh, line 217: Maybe, to reduce nesting...

```cpp
if (network_format == nf::NETWORK_ONNX) return;

if (multihead_format) {
   // ...
   return;
}

if (weights.has_policy_heads() && weights.has_value_heads()) {
   // ...
   return;
}

// the rest.
```

src/neural/loader.cc
```diff
@@ -140,6 +209,37 @@ void FixOlderWeightsFile(WeightsFile* file) {
...
+    auto weights = file->weights();
+    if (weights.has_policy_heads() && weights.has_value_heads()) {
+      CERR << "Weights file has multihead format, updating format flag";
+      net->set_network(static_cast<pblczero::NetworkFormat::NetworkStructure>(network_format | 128));
```
mooskagh, line 221: Also it's better to have a helper function instead of `|128`

src/neural/loader.cc
```diff
@@ -104,6 +104,75 @@ std::string DecompressGzip(const std::string& filename) {
...
+    mutable_weights->mutable_ip_pol_b()->Clear();
+
+    // Some older attention policy nets have policy encoders.
+    for (auto enc : file->weights().pol_encoder()) {
+      *vanilla->add_pol_encoder() = enc;
+    }
+    vanilla->set_pol_headcount(file->weights().pol_headcount());
+  }
+
+  // Macro to move remaining shared weights around.
```
mooskagh, line 129: Would this maybe work? (Not sure whether it's less hacky, but at least it doesn't have macros)

```cpp
  vanilla->MergeFromString(file-weights().OutputAsString());
```

UPD.. Just reallized it's copying from different proto.
Then let's keep as it is, I cannot think of anything nicer.

src/neural/onnx/converter.cc
```diff
@@ -584,18 +596,62 @@ std::string Converter::MakeAttentionBody(OnnxBuilder* builder,
         "/attn_body/reshape2", flow,
         builder->AddInitializer("/const/att_body_shape2",
                                 Int64OnnxConst({-1, 176}, {2})));
+    fist_stage_out_C = 176;
+  } else if (input_embedding == network_format::INPUT_EMBEDDING_PE_DENSE) {
+    int embedding_dense_size = weights.ip_emb_preproc_b.size() / 64;
```
mooskagh, line 601: This function is huge, maybe move large blocks into separate functions?

mooskagh (changes requested):

src/neural/blas/network_blas.cc
```diff
@@ -1012,11 +1018,11 @@ std::unique_ptr<Network> MakeBlasNetwork(const std::optional<WeightsFile>& w,
   }
   const WeightsFile& weights = *w;
   if (weights.format().network_format().network() !=
-          pblczero::NetworkFormat::NETWORK_CLASSICAL_WITH_HEADFORMAT &&
+          pblczero::NetworkFormat::NETWORK_CLASSICAL_WITH_MULTIHEADFORMAT &&
```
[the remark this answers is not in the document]
borg323: Yes I noticed in the onnx converter. Will do it when I add multihead support here.

src/neural/loader.cc
borg323: Hopefully this can be removed completely after we finalize net.proto.

src/neural/loader.cc
borg323: This will likely be replaced by the code in #1915 (but still under revision)

src/neural/loader.cc
borg323: Also dealt with in #1915

src/neural/onnx/converter.cc
borg323: Done

src/lc0ctl/leela2onnx.cc
```diff
@@ -83,6 +87,10 @@ bool ProcessParameters(OptionsParser* options) {
   options->Add<StringOption>(kOutputValue) = "/output/value";
   options->Add<StringOption>(kOutputMlh) = "/output/mlh";
   options->Add<BoolOption>(kOnnxToPytorch) = false;
+  std::vector<std::string> value_head = {"winner", "q", "st"};
+  options->Add<ChoiceOption>(kValueHead, value_head) = "winner";
```
mooskagh, line 91: As it's passed to the converter as string anyway, and in the latest version of the proto we may have arbitrary head names, maybe just make it a `StringOption`, and list the "popular" options in the help line?

src/neural/onnx/converter.cc
```diff
@@ -148,6 +162,7 @@ class Converter {
   const pblczero::Net& src_;
   const WeightsToOnnxConverterOptions& options_;
   ActivationFunction default_activation_;
+  float default_eps_;
```
mooskagh, line 165: I'd make `default_eps_` and `default_activation_` const (so that when they are forgotten to be initialized, it's a compile-time error), and would initialize them in the initializer list of the constructor.

mooskagh (approved):

PR #1952, opened 2024-02-21
Fix unsigned int conversion from cpp to Python.

revalo:
The custom `pybind` implementation for the Python binding has a bug when returning the masks for the input planes.

`Py_BuildValue()` takes a [format string](https://docs.python.org/3/c-api/arg.html) for converting C++ values to Python integers. Lowercase `k` is used for `unsigned long`, and uppercase `K` is used for `unsigned long long`. For `uint64_t`, we want to use the latter.

Previously, the bits were being stripped to 32 bits.

### DIFF
--- a/scripts/pybind/retval.py
+++ b/scripts/pybind/retval.py
@@ -185,7 +185,7 @@ class IntegralTupleRetVal(RetVal):
     def parse_tuple_format(self):
         return {
             'i': 'i',
-            'u64': 'k',
+            'u64': 'K',
             'f32': 'f',
         }[self.type]
 
--- a/scripts/pybind/retval.py
+++ b/scripts/pybind/retval.py
@@ -103,7 +103,7 @@ class NumericRetVal(RetVal):
     def parse_tuple_format(self):
         return {
             'i': 'i',
-            'u64': 'k',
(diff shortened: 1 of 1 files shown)


scripts/pybind/retval.py
```diff
@@ -103,7 +103,7 @@ def cpp_type(self):
     def parse_tuple_format(self):
         return {
             'i': 'i',
-            'u64': 'k',
+            'u64': 'K',
```
Naphthalin, line 106: Is there a particular reason why line 106 would need to be changed but not line 186?

borg323:
Was this found on windows?

revalo:
> Was this found on windows?

Yes, is there a reason for the suspicion?

scripts/pybind/retval.py
borg323: From a quick look it seems the other one is not used, but better change it for consistency.

borg323:
> is there a reason for the suspicion?

If it wasn't observed on windows it would require deeper investigation, since the code works on linux and the size of `long` is a known difference between them.

borg323 (approved):

PR #1953, opened 2024-02-21
WDLRescale() cleanup

borg323:


src/mcts/search.cc
```diff
@@ -196,6 +196,12 @@ Search::Search(const NodeTree& tree, Network* network,
                            : ContemptMode::WHITE;
     }
   }
+  wdl_rescale_ratio_ = params_.GetWDLRescaleRatio();
+  wdl_rescale_diff_ =
+      contempt_mode_ == ContemptMode::NONE
+          ? 0
+          : params_.GetWDLRescaleDiff() * params_.GetWDLEvalObjectivity();
```
Naphthalin, line 203: That's unfortunately not how it works -- `WDLEvalObjectivity` is only supposed to be applied when `invert` is set.

Naphthalin: Best solution probably is to define a "forwards" and an "invert" `wdl_rescale_diff_` here if the goal is to avoid any `params_.Get...()` calls.

borg323: Ah, I missed that. I'll think about it.

borg323: I decided to store the value in `wdl_eval_objectivity_`, seems cleaner this way to keep all the logic inside `WDLRescale()`:
```c++
  float wdl_rescale_diff =
      invert ? -wdl_rescale_diff_ * wdl_eval_objectivity_ : wdl_rescale_diff_;
```

PR #1955, opened 2024-02-23
warning fixes

borg323:


src/chess/board.h
```diff
@@ -62,8 +62,6 @@ class ChessBoard {
   ChessBoard() = default;
   ChessBoard(const std::string& fen) { SetFromFen(fen); }
 
-  ChessBoard& operator=(const ChessBoard&) = default;
```
mooskagh, line 65: Wasn't this required for https://github.com/LeelaChessZero/lc0/pull/1942?

src/neural/cuda/winograd_helper.inc
```diff
@@ -68,6 +68,8 @@ __device__ __forceinline__ float activate(float cVal,
     case ACTIVATION_SWISH:
       cVal /= (1.0f + __expf(-cVal));
       break;
+    default:
```
mooskagh, line 71: I'd suggest to explicitly list all the enums, so that when a new enum is added, it's flagged with compile-time warning (or better error).
Also, shouldn't it be some kind of exception rather that ignoring this silently?

src/chess/board.h
borg323: Apple's clang complained about it:
```
../src/chess/board.h:65:15: warning: definition of implicit copy constructor for 'ChessBoard' is deprecated because it has a user-declared copy assignment operator [-Wdeprecated-copy]
  ChessBoard& operator=(const ChessBoard&) = default;
  ```

File: src/neural/cuda/winograd_helper.inc (hunk 2 above)
Comment (borg323) [reply src/neural/cuda/winograd_helper.inc:71]: Apparently cuda has a `__trap()` function to trigger an illegal instruction abort. Does that sound OK?

Comment (mooskagh) [reply src/neural/cuda/winograd_helper.inc]: Ah it's a CUDA code.. Yeah, sounds good!

File: src/chess/board.h (hunk 1 above)
Comment (mooskagh) [reply src/chess/board.h:65]: But how does it compile without this function? :thinking: 

Anyway, I think what they kind of suggest is to also add a default copy constructor:
```cpp
ChessBoard(const ChessBoard&) = default;
```

borg323: This works as well.

mooskagh (approved):

PR #1956, opened 2024-02-24
Support outputting HLO in leela2onnx

mooskagh:
In addition to (now optional) `--output=` parameter of `leela2onnx`,

* `--hlo-text-output=filename` to output as text HLO (or stdout if filename is `-`)
* `--hlo-proto-output=filename` to output to binary `HloModuleProto` file
* `--hlo-batch-size` -- for which batch size to generate HLO (333 is the default, to be noticeable in the output file)
* `--hlo-allow-partial-result` -- if conversion fails (i.e. due to ONNX op not being implemented), still output what was converted so far. Useful to look at the progress while implementing new ops.

### DIFF
--- a/src/lc0ctl/leela2onnx.cc
+++ b/src/lc0ctl/leela2onnx.cc
@@ -39,35 +39,33 @@
 namespace lczero {
 namespace {
 
-const OptionId kInputFilenameId{"input", "InputFile",
+const OptionId kInputFilenameId{"input", "",
                                 "Path of the input Lc0 weights file."};
-const OptionId kOutputFilenameId{"output", "OutputFile",
-                                 "Path of the output ONNX file."};
-const OptionId kHloTextOutputFilenameId = {
-    "hlo-text-output", "HloTextOutputFile", "Path of the output HLO file."};
+const OptionId kOutputFilenameId{"output", "", "Path of the output ONNX file."};
+const OptionId kHloTextOutputFilenameId = {"hlo-text-output", "",
+                                           "Path of the output HLO file."};
 const OptionId kHloProtoOutputFilenameId = {
-    "hlo-proto-output", "HloProtoOutputFile",
-    "Path of the output HLO proto file."};
-const OptionId kHloBatchSizeId{"hlo-batch-size", "HloBatchSize",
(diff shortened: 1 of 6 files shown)


src/lc0ctl/leela2onnx.cc
```diff
@@ -40,6 +43,16 @@ const OptionId kInputFilenameId{"input", "InputFile",
...
 const OptionId kOutputFilenameId{"output", "OutputFile",
                                  "Path of the output ONNX file."};
+const OptionId kHloTextOutputFilenameId = {
+    "hlo-text-output", "HloTextOutputFile", "Path of the output HLO file."};
+const OptionId kHloProtoOutputFilenameId = {
+    "hlo-proto-output", "HloProtoOutputFile",
+    "Path of the output HLO proto file."};
+const OptionId kHloBatchSizeId{"hlo-batch-size", "HloBatchSize",
+                               "Batch size to use for HLO conversion."};
+const OptionId kHloAllowPartialResultId = {
```
borg323, line 53: I think this should be a hidden option.

src/lc0ctl/leela2onnx.cc
```diff
@@ -40,6 +43,16 @@ const OptionId kInputFilenameId{"input", "InputFile",
                                 "Path of the input Lc0 weights file."};
 const OptionId kOutputFilenameId{"output", "OutputFile",
                                  "Path of the output ONNX file."};
+const OptionId kHloTextOutputFilenameId = {
+    "hlo-text-output", "HloTextOutputFile", "Path of the output HLO file."};
```
borg323, line 47: Why do we add uci names for the options? Just curious.

mooskagh: No reason. I thought it wouldn't harm, but actually they show up in `--help`, so better remove them. (Done)

src/lc0ctl/leela2onnx.cc
mooskagh: Indeed, done!

borg323:
xla_hlo_builder.cc needs a `#include <unordered_map>`
print_hlo.cc  needs a `#include <cctype>`

borg323 (approved):

PR #1957, opened 2024-02-25
upstream the rescorer

borg323:
The rescorer is being built as a separate executable using `rescorer_main.cc`, which has only the `rescore` mode and as such it can now be omitted from the comand line. Two new build options `lc0` and `rescorer` control which executable to build, and can both be set to `true` at the same time. Other changes relative to the `rescore_tb` branch are that the rescorer loop is now in a separate file with less meaningless options, the windows build script is `build_rescorer.cmd` and there is no appveyor support (yet).


Tilps:
Thanks for importing this for me!

Tilps (approved):
Even better :P

PR #1958, opened 2024-02-26
fix rescorer warnings

borg323:
I didn't want to add this to #1957


src/rescorer/rescoreloop.cc
```diff
@@ -542,7 +541,7 @@ void ProcessFile(const std::string& file, SyzygyTablebase* tablebase,
               fileContents[i].probabilities[j] = rootNode->policy[j];
             }
           }
-          if (i < fileContents.size() - 1) {
+          if (i <= fileContents.size()) {
```
Tilps, line 544: this is not equivalent.

Tilps (changes requested):

Tilps (approved):

PR #1959, opened 2024-02-26
Update ONNX-to-HLO converter to support T75 nets

mooskagh:
* Dependent computation support
* Helper class for HloShapes instead of suing protos directly
* Ops:
    * [x] GlobalAveragePool
    * [x] Squeeze
    * [x] Split
    * [x] Sigmoid
    * [x] Mul
    * [x] Gather

### DIFF
--- a/src/neural/xla/onnx2hlo.cc
+++ b/src/neural/xla/onnx2hlo.cc
@@ -166,7 +166,7 @@ pblczero::XlaLiteralProto OnnxTensorToXlaLiteral(
       break;
     case pblczero::TensorProto::INT32:
       convert(tensor.raw_data(), literal.mutable_s32s());
-      break;    
+      break;
     default:
       throw Exception("Cannot convert ONNX tensor to XLA literal for type " +
                       pblczero::XlaShapeProto::Type_Name(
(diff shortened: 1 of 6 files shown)


src/neural/xla/onnx2hlo.cc
```diff
@@ -313,45 +334,90 @@ class Onnx2HloConverter {
...
+  std::vector<HloFlow> OpMul(const pblczero::NodeProto& node) {
+    CheckKnownAttributes(node, 2, {});
+    auto* lhs = GetInput(node, 0);
+    auto* rhs = GetInput(node, 1);
+    std::tie(lhs, rhs) = EqualizeShape(lhs, rhs);
+    return {builder_.Multiply(lhs, rhs)};
+  }
+
+  std::vector<HloFlow> OpSplit(const pblczero::NodeProto& node) {
+    CheckKnownAttributes(node, 1, {"axis", "num_outputs"});
```
borg323, line 381: `Split` is different before opset 13, there was a "split" attribute instead of input: https://github.com/onnx/onnx/blob/main/docs/Changelog.md#attributes-187
Also "num_outputs" was only added in opset 18.

src/neural/xla/onnx2hlo.cc
```diff
@@ -381,8 +446,79 @@ class Onnx2HloConverter {
...
+    flow = builder_.Divide(flow, denominator);
+    HloTensorType output_shape(flow->shape());
+    while (output_shape.Rank() < lhs->shape().dimensions_size()) {
+      output_shape.AddDimension(1);
+    }
+    return {builder_.Reshape(flow, output_shape)};
+  }
+
+  std::vector<HloFlow> OpSqueeze(const pblczero::NodeProto& node) {
+    CheckKnownAttributes(node, 2, {});
```
borg323, line 476: Like `Split`, before opset 13 it used an "axes" attribute instead of input: https://github.com/onnx/onnx/blob/main/docs/Changelog.md#attributes-189

borg323 (commented):
Is this going to support multiple onnx opsets? The converter supports from 7 to 18 although I believe there are no functional changes affecting the generated models up to 21 (but needs testing).

src/neural/xla/print_hlo.cc
```diff
@@ -373,12 +412,13 @@ class HloPrettyPrinter {
...
     }
     s_ << "}\n";
-    current_computation_ = nullptr;
+    instructions_.clear();
   }
 
   PrettyPrintHloOptions options_;
   const pblczero::HloModuleProto* current_module_ = nullptr;
-  const pblczero::HloComputationProto* current_computation_ = nullptr;
+  std::unordered_map<size_t, const pblczero::HloInstructionProto*>
```
borg323, line 420: Missing `#include <unordered_map>`

src/neural/xla/onnx2hlo.cc
mooskagh: I've just added a check for the opcode version, to show that it's possible in principle to have different versions (what I'd do is to make `OpSplit()` call `OpSplit1()`, `OpSplit13()` etc depending on the version).

But for now the code just fails.

src/neural/xla/onnx2hlo.cc
mooskagh: Same as above.

src/neural/xla/print_hlo.cc
mooskagh: Done.

borg323 (approved):

PR #1961, opened 2024-03-01
Add support for min_batch to reduce non-determinism in cuda backends.

Tilps:
This is pre-work for the policy_tournament upstreaming.


Tilps:
Its been a while I tested a random old T80 net with my 4090 and I actually needed 16 to have determinism...
I wonder if 4 is no longer a sensible default.

Naphthalin:
While I'm not exactly sure whether reducing non-determinism justifies this, there are more reasons why adding a `min_batch_size_` is a good idea:
- makes things symmetric with max batch size
- should reduce high pitched and unstable coil whine (coming from small batches)
- very small patches might be slower due to optimizations.

Naphthalin (approved):

PR #1962, opened 2024-03-01
Policy tournament

Tilps:
Note that in preparing this for submission I renamed a number of things from PolicyMultiGames to MultiGames.

The intention is that after this is merged, MultiGames will get a replaceable multigame evaluator which is used to switch between policy and value modes - allowing for a cleaner merge of value_tournament, without massively redundant code.

This update preparing for merge has also changed the behaviour of policy tournament a bit.  Since the default value of the flag kTotalGames is -1, which doesn't make sense for policy tournament, you are now required to always provide a total games parameter if using multi game mode where before the code treated -1 like -2 and didn't handle -2 correctly at all.

Some additional checks have been added, no trying to play an odd number of games in multi-game mode.  Use of an opening book is now enforced, not just encouraged.

Also note the cuda/cudnn changes are obsoleted by PR1961 which I'll aim to submit first.


Tilps:
PR1961 merged and thus this is now ready to be reviewed and maybe submitted...

src/selfplay/tournament.cc
```diff
@@ -469,4 +636,26 @@ SelfPlayTournament::~SelfPlayTournament() {
...
+  auto p2name =
+      player_options_[1][0].Get<std::string>(NetworkFactory::kWeightsId);
+
+  output << std::endl;
+  output << "[White \"" << p1name << "\"]" << std::endl;
+  output << "[Black \"" << p2name << "\"]" << std::endl;
+  output << "[Results \"" << tournament_info_.results[0][0] << " "
+         << tournament_info_.results[2][0] << " "
+         << tournament_info_.results[1][0] << "\"]" << std::endl;
+  output << std::endl;
```
borg323, line 653: Isn't the second part only valid for policy tournament?

Tilps: I believe this is valid for any tournament where p1name and p2name are valid for both the white vs black and black vs white case. (I didn't feel like it was worth trying to handle the case where that is false...)

It is fake pgn regardless - no per-game moves, no tracking of openings, just raw results in a format which is consumable by my custom version of ordo...

borg323 (approved):

PR #1965, opened 2024-03-03
XLA support for T79 net

mooskagh:


src/neural/xla/onnx2hlo.cc
```diff
@@ -502,36 +575,81 @@ class Onnx2HloConverter {
     return flows;
   }
 
+  std::vector<HloFlow> OpSlice(const pblczero::NodeProto& node) {
```
borg323, line 578: Need a check for opset >= 10 here.

mooskagh: Done.

borg323 (approved):

PR #1967, opened 2024-03-05
Support ONNX batch size in leela2onnx

mooskagh:


borg323 (approved):
Does it makes sense to expose more converter options, like opset and fp16?

mooskagh:
> Does it makes sense to expose more converter options, like opset and fp16?

I believe it does, but I'm too lazy to do that in this PR.

borg323:
> I believe it does, but I'm too lazy to do that in this PR.

I'll do it together with a few other minor things I have planned.

PR #1969, opened 2024-03-06
bf16 support for onnx

borg323:


src/utils/bf16_utils.h
```diff
@@ -0,0 +1,49 @@
+/*
+  This file is part of Leela Chess Zero.
+  Copyright (C) 2022 The LCZero Authors
```
mooskagh, line 3: 2024

src/neural/onnx/network_onnx.cc
```diff
@@ -413,6 +427,12 @@ std::unique_ptr<Network> MakeOnnxNetwork(const std::optional<WeightsFile>& w,
         fp16 ? WeightsToOnnxConverterOptions::DataType::kFloat16
              : WeightsToOnnxConverterOptions::DataType::kFloat32;
 
+    if (opts.GetOrDefault<bool>("bf16", false)) {
```
mooskagh, line 430: I'd replace it with `datatype` = `f32`/`f16`/`bf16` parameter.
(see also this function https://github.com/LeelaChessZero/lc0/pull/1974/files#diff-4ede40c0bf9d771ac89c7f80c42de24aa6cb7bac22c215a95896b4ae28cf5709R1031 in my PR)

src/neural/onnx/converter.cc
```diff
@@ -170,19 +175,48 @@ std::unique_ptr<OnnxConst> Converter::GetWeghtsConverter(
...
   if (!options_.alt_mish || options_.opset < 9 ||
       options_.data_type_ !=
           WeightsToOnnxConverterOptions::DataType::kFloat32) {
     if (options_.opset >= 18) return builder->Mish(name, input);
-    auto flow = builder->Softplus(name + "/softplus", input);
+    std::string flow;
+    if (options_.fix_bf16 &&
+        options_.data_type_ ==
+            WeightsToOnnxConverterOptions::DataType::kBFloat16) {
+      flow = builder->Cast(name + "/to_float", input,
```
mooskagh, line 212: Maybe have two helper functions:
```
std::string MaybeFromBf16Fix(std::string name, std::string flow) {
   if (!options_.fix_bf16 ||
        options_.data_type_ !=
            WeightsToOnnxConverterOptions::DataType::kBFloat16)) return flow;
   return builder->Cast(name + "/to_float", input, pblczero::TensorProto::FLOAT);
}
std::string MaybeToBf16Fix(std::string name, std::string flow) { // similar
```

and then write without the need for conditional:
```
flow = MaybeFromBf16Fix(name, flow);
flow = builder->Softplus(name + "/softplus", input);
flow = MaybeToBf16Fix(name, flow);
```

src/neural/onnx/converter.cc
```diff
@@ -159,6 +160,14 @@ class Converter {
       const std::vector<float>&, std::initializer_list<int> dims,
       std::initializer_list<int> order = {});
 
+  std::unique_ptr<OnnxConst> GetScalarConverter(float in);
+
+  std::string OptionalBf16FixPrologue(OnnxBuilder* builder, std::string flow,
+                                      std::string name);
+
+  std::string OptionalBf16FixEpilogue(OnnxBuilder* builder, std::string flow,
```
mooskagh, line 168: Two functions have long but too similar names (only three letters in the middle are different). Was a bit confusing to read for me, but fine to keep as it is.

mooskagh (approved):

src/neural/onnx/converter.cc
borg323: Renamed to `StartOptionalBf16Fix` and `EndOptionalBf16Fix`.

mooskagh (approved):

PR #1971, opened 2024-03-08
Value only mode for uci.

Tilps:
This is quick and dirty...

I had intended to share code with multigame.cc - but that would need more refactoring than I originally planned - so for now I have not done so...

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -210,6 +210,7 @@ files += [
   'src/neural/onnx/converter.cc',
   'src/selfplay/game.cc',
   'src/selfplay/loop.cc',
+  'src/selfplay/multigame.cc',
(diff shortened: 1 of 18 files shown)


Tilps:
Note that only engine.cc is changed in this PR - the rest is from PR1966.

Tilps:
now just one file to review (since the other PR has been merged)

borg323:
What about adding something like https://github.com/borg323/lc0/commit/709571c85774f97ab504effe65f8544be6f2c9cb for minibatch-size support?

Tilps:
Seems reasonable, do you want to add it to my branch directly rather than have me copy it over?

PR #1976, opened 2024-03-11
Multihead arch with cutlass fused multihead attention

almaudoh:
cutlass implementation of fused multihead attention layer giving about 10% speedup on A100. 

@todo
- Fusing biases
- Fusing layernorms
- Fusing FFNs

### DIFF
--- a/src/neural/backends/cuda/cutlass_kernels.cu
+++ b/src/neural/backends/cuda/cutlass_kernels.cu
@@ -25,17 +25,17 @@
   Program grant you additional permission to convey the resulting work.
 */
 
-#include "cuda_common.h"
-
-#ifdef USE_CUTLASS
+#include "neural/backends/cuda/cuda_common.h"
 
 // Fused MHA implementation from cutlass example #41
 #include "fused_multi_head_attention/kernel_forward.h"
+#include "utils/exception.h"
 
-
+namespace lczero {
+namespace cudnn_backend {
 
 template <bool bias>
-bool fusedMHACutlass(void* output, void* q, void* k, void* v, void* skip,
+void fusedMHACutlass(void* output, void* q, void* k, void* v, void* skip,
                      int batch_size, int num_heads, int depth,
                      cudaStream_t stream) {
   cutlass::half_t* mha_q = (cutlass::half_t*)q;
@@ -54,6 +54,9 @@ bool fusedMHACutlass(void* output, void* q, void* k, void* v, void* skip,
                       false,  // Supports dropout
                       bias    // Supports bias
                       >;
+  static_assert(
+      !Attention::kNeedsOutputAccumulatorBuffer,
+      "Unhandled case in cutlass MHA: needs output accumulator buffer");
 
   typename Attention::Params p;
   {  // set parameters
@@ -62,10 +65,6 @@ bool fusedMHACutlass(void* output, void* q, void* k, void* v, void* skip,
     p.value_ptr = mha_v;
     p.logsumexp_ptr = nullptr;  // Only needed for bw
     p.output_accum_ptr = nullptr;
-    if (Attention::kNeedsOutputAccumulatorBuffer) {
-      // throw Exception("Unhandled case in cutlass MHA");
-      return false;
-    }
     p.output_ptr = (cutlass::half_t*)output;
     p.attn_bias_ptr = (cutlass::half_t*)skip;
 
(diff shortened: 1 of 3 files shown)


src/neural/backends/cuda/common_kernels.cu
```diff
@@ -1110,6 +1110,78 @@ __global__ void layer_norm_kernel(int N, int C, T* output, const T* input,
   }
 }
 
+__global__ void layer_norm_kernel_8_el_per_thread(
```
borg323: This seems to be unrelated.

src/neural/backends/cuda/inputs_outputs.h
```diff
@@ -35,7 +35,8 @@ namespace cudnn_backend {
 struct InputsOutputs {
   InputsOutputs(int maxBatchSize, bool wdl, bool moves_left,
                 size_t tensor_mem_size = 0, size_t scratch_size = 0,
-                bool cublasDisableTensorCores = false) {
+                bool cublasDisableTensorCores = false)
+      : has_moves_left_(moves_left) {
```
borg323: This seems unrelated as well.

build.cmd
```diff
@@ -11,6 +11,15 @@ set DNNL=false
 set OPENBLAS=false
 set EIGEN=false
 set TEST=false
+set CUTLASS=true
+
+if "%CUDA%"=="true" (
```
borg323, line 16: This block is repeated right after, probably a bad merge.

third_party/fused_multi_head_attention/CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
...
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice, this
+# list of conditions and the following disclaimer.
+#
+# 2. Redistributions in binary form must reproduce the above copyright notice,
+# this list of conditions and the following disclaimer in the documentation
+# and/or other materials provided with the distribution.
```
Menkib64, line 13: Do we include the copyright notice in our binary distributions?

src/neural/backends/cuda/cutlass_kernels.cu
```diff
@@ -0,0 +1,124 @@
...
+                         smem_bytes);
+  }
+  if (!Attention::check_supported(p)) {
+    // throw Exception("Unhandled case in cutlass MHA");
+    return false;
+  }
+
+  kernel_fn<<<p.getBlocksGrid(), p.getThreadsGrid(), smem_bytes, stream>>>(p);
+
+  // ReportCUDAErrors(cudaGetLastError());
```
Menkib64: Why is the error check commented out?

The caller does an error check but it would make logical sense to do the error check here.

src/neural/backends/cuda/cutlass_kernels.cu
```diff
@@ -0,0 +1,124 @@
...
+    p.bias_strideH = 64 * 64;
+    p.bias_strideM = 64;
+    p.bias_strideB = num_heads * p.bias_strideH;
+  }
+
+  constexpr auto kernel_fn = attention_kernel_batched_impl<Attention>;
+  int smem_bytes = sizeof(typename Attention::SharedStorage);
+  if (smem_bytes > 0xc000) {
+    cudaFuncSetAttribute(kernel_fn, cudaFuncAttributeMaxDynamicSharedMemorySize,
+                         smem_bytes);
```
Menkib64: This call needs an error check.

src/neural/backends/cuda/cutlass_kernels.cu
```diff
@@ -0,0 +1,124 @@
...
+
+  typename Attention::Params p;
+  {  // set parameters
+    p.query_ptr = mha_q;
+    p.key_ptr = mha_k;
+    p.value_ptr = mha_v;
+    p.logsumexp_ptr = nullptr;  // Only needed for bw
+    p.output_accum_ptr = nullptr;
+    if (Attention::kNeedsOutputAccumulatorBuffer) {
+      // throw Exception("Unhandled case in cutlass MHA");
```
Menkib64: It would be helpful if the error message would tell the problem is about an accumulator buffer.

src/neural/backends/cuda/cutlass_kernels.cu
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  constexpr auto kernel_fn = attention_kernel_batched_impl<Attention>;
+  int smem_bytes = sizeof(typename Attention::SharedStorage);
+  if (smem_bytes > 0xc000) {
+    cudaFuncSetAttribute(kernel_fn, cudaFuncAttributeMaxDynamicSharedMemorySize,
+                         smem_bytes);
+  }
+  if (!Attention::check_supported(p)) {
+    // throw Exception("Unhandled case in cutlass MHA");
```
Menkib64: Error message should tell `check_supported` failed.

src/neural/backends/cuda/layers.cc
```diff
@@ -1770,6 +1773,18 @@ void EncoderBlock<DataType>::Eval(int N, DataType* in_out_tensor,
...
 
+#ifdef USE_CUTLASS
+  if (use_fused_mha_) {
+    // TODO: check if we need skip in a different tensor than same tensor as
+    // output!
+    bool success =
+        fusedMHA(buffer2, mha_q, mha_k, mha_v, has_smolgen_ ? buffer2 : nullptr,
+                 N, encoder_heads_, depth, stream);
+
+    ReportCUDAErrors(cudaGetLastError());
```
Menkib64: It would be logical to have the error check in the same function as the kernel launch.

third_party/fused_multi_head_attention/CMakeLists.txt
borg323: Not yet, I need to fix that. The idea is to create `third_party/third_party.txt`with all the copyright notices and copy that.

almaudoh: This can be done in a new PR then?

src/neural/backends/cuda/cutlass_kernels.cu
almaudoh: I have restored this. Even though I don't know why @ankan-ban commented it out.

src/neural/backends/cuda/cutlass_kernels.cu
almaudoh: It seems here the preference is to not throw the exception inside the kernel, but rather outside it. We can pass the error message through a character array to the outside. Not sure if that's a great idea though.

PR #1978, opened 2024-03-13
Implement new multihead arch for blas / eigen / mkl backends

almaudoh:


src/neural/blas/network_blas.cc
```diff
@@ -141,12 +142,14 @@ class BlasComputation : public NetworkComputation {
   bool wdl_;
   bool moves_left_;
   bool conv_policy_;
+  bool attn_policy_;
+  bool attn_body_;
+  bool new_encoding_;
```
borg323, line 147: `new_encoding_` is not very descriptive.

borg323 (approved):

src/neural/blas/network_blas.cc
almaudoh: Renamed.

PR #1986, opened 2024-03-18
Debug printing of internal NN nodes in XLA backend

mooskagh:
* Also, don't compute heads that are not used.


mooskagh:
Just realized that I don't split individial samples of a batch, so converting it to draft for now.

mooskagh:
I'll redo the debug printing a bit differently, but overriding the outputs is useful in any case, so I extracted it into the separate PR https://github.com/LeelaChessZero/lc0/pull/1996

PR #1987, opened 2024-03-19
additional onnx options

borg323:


src/neural/onnx/converter.h
```diff
@@ -45,9 +45,10 @@ struct WeightsToOnnxConverterOptions {
   std::string output_mlh = "/output/mlh";
   int batch_size = -1;
   int opset = 17;
-  bool alt_mish = false;
-  bool alternative_layer_normalization = false;
-  bool fix_bf16 = false; // Work around missing bf16 support in onnx operators.
+  bool alt_mish = false;       // Use "Mish" approximation (fp32 only).
+  bool alt_layernorm = false;  // Discrete "LayerNormalization" implementation.
+  bool fix_bf16 = false;  // Work around missing bf16 support in onnx operators.
```
mooskagh, line 50: I'm not sure about "fix" word, but I cannot find better one.. "bf16_fallback"? "bf16_ok"? "override_bf16_ops"? "allow_nonstandard_bf16"? "deviant_bf16", "force_allow_bf16" "bf16_workaround"?

mooskagh (approved):

src/neural/onnx/converter.h
borg323: Maybe if I invert the meaning and make it `force_bf16` with comment "Ignore missing bf16 support in onnx operators"?

mooskagh: force_bf16 is still the same meaning, not inverted, no?
Inverted would be.. only_std_bf16? compatible_bf16? strict_bf16?

Hmm also I think soon we'll be adding fp8 etc and will have same problem there, so maybe don't mention bf16 in the option name at all? Like... strict_op_types? allow_nonstd_op_types?

borg323: No, it is inverted, the `fix_bf16`->cast to float, `force_bf16`->don't cast.
fp8 is a lot worse, almost no "interesting" operators support it, so I don't see a use case for not forcing it.

mooskagh (approved):

PR #2010, opened 2024-04-07
fix build outside x86_64/aarch64

nekopsykose:
the current logic includes the x86 simd header any time the platform is !arm, which is backwards


nekopsykose:
technically it could add `|| defined(__i386__)` since it's valid there too, not sure if you'd prefer that

borg323 (approved):

PR #2025, opened 2024-04-30
Minor changes to proto generation

mooskagh:
* `reserved` keyword is now supported.
    * It's possible to reserve fields by number, or by name (quotes are optional).
* string fields now have `mutable_X()` function.

### DIFF
--- a/scripts/compile_proto.py
+++ b/scripts/compile_proto.py
@@ -392,8 +392,9 @@ class ProtoFieldParser:
             w.Write("bool has_%s() const;" % (name))
             if self.type.IsMessage():
                 w.Write("const %s& %s() const;" % (cpp_type, name))
-                w.Write("%s* mutable_%s();" % (cpp_type, name))
-            else:
+            if self.type.IsMessage() or self.type.IsBytesType():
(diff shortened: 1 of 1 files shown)


borg323:
Is it hard to make it accept the `reserved 10 to 18;` syntax?

mooskagh:
Done (didn't test, should work)

scripts/compile_proto.py
```diff
@@ -630,6 +637,10 @@ def CheckReserved(self):
             if isinstance(r, int):
                 if any(x.number == r for x in self.fields):
                     raise ValueError(f'Field number [{r}] is reserved.')
+            elif isinstance(r, range):
+                if any(x.number in r for x in self.fields):
+                    raise ValueError(f'Field range [{r.start} to {r.stop}] '
```
borg323, line 642: I think this should be `r.stop - 1`.

mooskagh: Indeed, done!

borg323 (commented):
I did some testing of `reserved`, looks good.

borg323 (approved):
Approving again after rebase.

PR #2026, opened 2024-05-05
Save undecided games as draws

hans-ekbrand:
For context see https://discord.com/channels/425419482568196106/427066771627966466/1236740483091533825 but in short this patch makes sure that training games that are undecided after 450 are still saved (with game outcome DRAW). If that is a bad outcome, then the rescorer will likely be able to improve on it.

Without this patch, such long training games are not saved at all.


hans-ekbrand:
Sorry, this was not against master. I'll close this one and make a new one.

PR #2028, opened 2024-05-16
Make VerboseMoveStats consistent with UCI output if WDLEvalObjectivity is used

Naphthalin:
This fixes an oversight where the verbosemovestats output wasn't affected by WDLEvalObjectivity, which most notably causes inconsistencies when switcthing betweeen winrate and centipawn in the dialog box.


src/mcts/search.cc
```diff
@@ -493,13 +493,26 @@ std::vector<std::string> Search::GetVerboseStats(Node* node) const {
...
     if (n) {
-      print(oss, "(WL: ", sign * n->GetWL(), ") ", 8, 5);
-      print(oss, "(D: ", n->GetD(), ") ", 5, 3);
+      auto wl = sign * n->GetWL();
+      auto d = n->GetD();
+      auto is_perspective = ((contempt_mode_ == ContemptMode::BLACK) ==
+                             played_history_.IsBlackToMove())
+                                ? 1.0f
+                                : -1.0f;
+      auto mu_uci = WDLRescale(
```
borg323, line 502: `mu_uci` is not used anywhere.

Naphthalin: So, just call `WDLRescale()` without assigning the return value to a variable?

borg323: Exactly.

borg323 (approved):

PR #2039, opened 2024-06-24
Fix build for GCC 14.1

mooskagh:
Missing #include <algorithm> caused a cryptic error message.

### DIFF
--- a/src/neural/xla/xla_tensor.cc
+++ b/src/neural/xla/xla_tensor.cc
@@ -129,7 +129,7 @@ void XlaMutableTensor::Cast(pblczero::XlaShapeProto::Type new_type) {
         convert(FP32toBF16);
         break;
       case pblczero::XlaShapeProto::F8E5M2:
-        convert(FP32toFP8E5M2);
+        convert(FP32toFP8E5M2_Saturate);
         break;
       default:
         throw Exception("Unsupported cast F32 -> " +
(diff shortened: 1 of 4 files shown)


borg323 (approved):

borg323 (approved):

src/neural/xla/xla_tensor.cc
```diff
@@ -129,7 +129,7 @@ void XlaMutableTensor::Cast(pblczero::XlaShapeProto::Type new_type) {
         convert(FP32toBF16);
         break;
       case pblczero::XlaShapeProto::F8E5M2:
-        convert(FP32toFP8E5M2);
+        convert(FP32toFP8E5M2_Saturate);
         break;
```
borg323, line 133: Maybe we can remove this case completely now that f8e5m2 support has been removed from the onnx converter and rethink the interface when we need it again.

mooskagh: Sounds good, but I'll only be able to do that in the evening.  
So either merge now and follow up later, or wait for evening, or merge now and forget, or you push changes into this PR.

mooskagh: I tried several different variants (lambda, static_cast), but nothing really works.
I think it's better to keep this function anyway, so I'd merge it as it is for now.

PR #2040, opened 2024-06-28
circleci: build macos binaries on arm

borg323:
Circleci is phasing out their x86 mac build systems, so we have to migrate to the arm ones.


cn4750 (approved):
Rubber stamped as requested.

PR #2051, opened 2024-08-09
Fixing has_wdl typo affecting no WDL_mu centipawn output

Naphthalin:
v0.31 doesn't show the `WDL_mu` centipawn output because of a typo.


mooskagh (approved):

Naphthalin:
Quick confirmation that `"ScoreType": "WDL_mu"` now indeed behaves as intended, instead of simply producing the `"ScoreType: centipawn"` eval.

Issue #825, opened 2019-03-30
Incorrect comment on bit order in class BitBoard?


rhalbersma:
In bitboard.h, for `class BoardSquare` the comment on line 43/44 states:

```
// As a single number, 0 to 63, bottom to top, left to right.
// 0 is a1, 8 is a2, 63 is h8.
```

In contrast, for the `class BitBoard` the comment on line 88/89 states:

```
// Bit enumeration goes from bottom to top, from left to right:
// Square a1 is bit 0, square a8 is bit 7, square b1 is bit 8.
```

However, the subsequent `BitBoard` member functions `set`, `reset` etc. take a BoardSquare parameter. So in fact BitBoard has the same ordering as BoardSquare. 

Should this comment in `class BitBoard` be updated?

mooskagh:
Yeah initially it was wrong in both places, but then fixed in one but apparently not another.

rhalbersma:
There are some other improvements possible in bitboard.h. E.g. `BitBoard::Mirror()` could use the compiler intrinsics `_byteswap_uint64` (MSVC) or `__builtin_bswap64` (gcc/clang). This should reduce the current 15 instructions to 1 on modern architectures.

And `ReverseBitsInBytes` from mcts/node.cc could be moved to be a member function of BitBoard, and renamed `MirrorVertical` e.g. (and `Mirror` be renamed to `MirrorHorizontal`). 

I also don’t quite understand the rationale to vertically mirror the inputs for training, could you explain?

I could send a PR later. What do you think?

mooskagh:
Those are good suggestsions, PR will be appreciated!

TesseractA:
Still Relevant?

mooskagh:
Yes, I've just check, still wrong (the comments issue in the very first comment). :)

PR #2054, opened 2024-08-24
adressing issue #825

KarlKfoury:
The comment on the bitboard function previously stated that "a8 is bit 7, b1 is bit 8". 
All bitboard methods (set, get, reset...) use the following to set the position of the bit on the bitboard: std::uint64_t(1) << pos, with pos being the BoardSquare index that follows the horizontal indexing notation (a1=0, b1=1, c1=2...)
Meaning the bit's position on the bitboard is going to be 0 + pos which means the position of the bit should also follow the horizontal indexing notation.

### DIFF
--- a/src/chess/bitboard.h
+++ b/src/chess/bitboard.h
@@ -87,7 +87,7 @@ class BoardSquare {
 
 // Represents a board as an array of 64 bits.
 // Bit enumeration goes from bottom to top, from left to right:
-// Square a1 is bit 0, square a8 is bit 7, square b1 is bit 8.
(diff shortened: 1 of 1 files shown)


mooskagh:
There was also https://github.com/LeelaChessZero/lc0/pull/445, we should finally check what is right. :)

KarlKfoury:
BoardSquare.as_int() row*8 + col to pack square name into an int which is the horizontal indexing notation (a1 = 0, b1 = 1...). And all subsequent methods and functions (Move::as_packed_int for example) seem to follow this standard

mooskagh (approved):

PR #2055, opened 2024-08-27
Issue 825

KarlKfoury:
All bitboard methods (set, reset, get...) use std::uint64_t(1) << pos, moving the bit initiated at position 0 by 'pos' spaces. Making the bit position 0 + pos = pos. Pos is passed as argument to these methods as BoardSquare.as_int, a packed int representation of a boardsquare that follows horizontal indexing (a1=0, b1=1, ..., a2=8) which means bitboard bit location should follow a similar notation.


borg323:
Duplicate of #2054

Issue #2003, opened 2024-03-28
`go nodes 0` does infinite search


mooskagh:
It should stop immediately instead (or do one iteration when the tree is empty).

sovietpanzzer:
`class VisitsStopper : public SearchStopper {
 public:
  VisitsStopper(int64_t limit, bool populate_remaining_playouts)
    : nodes_limit_(limit ? limit : 4000000000ll), // here is the problem
        populate_remaining_playouts_(populate_remaining_playouts) {}
  int64_t GetVisitsLimit() const { return nodes_limit_; }
  bool ShouldStop(const IterationStats&, StoppersHints*) override;`

For some reason when limit = 0, nodes_limit is changed to 4000000000ll which makes "go nodes 0" similar to infinite search.h.
A simple fix is to remove this condition. Does it affect other parts of the code ?

sovietpanzzer:
Here seems to be the problem : 

In the file src/mcts/stoppers/stoppers.h
```
class PlayoutsStopper : public SearchStopper {
 public:
  PlayoutsStopper(int64_t limit, bool populate_remaining_playouts)
      : nodes_limit_(limit ? limit : 4000000000ll), \\Here seems to be the problem
        populate_remaining_playouts_(populate_remaining_playouts) {}
  int64_t GetVisitsLimit() const { return nodes_limit_; }
  bool ShouldStop(const IterationStats&, StoppersHints*) override;

 private:
  const int64_t nodes_limit_;
  const bool populate_remaining_playouts_;
};
```
For some reason when limit = 0, nodes_limit is changed to 4000000000ll which makes `go nodes 0` similar to infinite search.
A simple fix is to remove this condition. Does it affect other parts of the code ?

mooskagh:
This 4'000'000'000 was introduced to prevent number of nodes overflowing uint32 (e.g. in `go infinite` case), and we'd like to keep that check. But I believe it should be not in the stoppers constructors (so I agree with removing it from there), but in the place where the stoppers are created:


// common.cc
```cpp
  // "go nodes" stopper.
  int64_t node_limit = 0;
  if (params.nodes) {
    if (options.Get<bool>(kNodesAsPlayoutsId)) {
      stopper->AddStopper(std::make_unique<PlayoutsStopper>(
          *params.nodes, options.Get<float>(kSmartPruningFactorId) > 0.0f));
    } else {
      node_limit = *params.nodes;
    }
  }
[...]

PR #2056, opened 2024-08-29
Issue 2003

KarlKfoury:
since GoParams.nodes is an std::optional, it's initialization status can be checked through params.nodes.has_value() instead of params.nodes, which returns false if params.nodes is set to 0 (source of the problem)

### DIFF
--- a/src/mcts/stoppers/common.cc
+++ b/src/mcts/stoppers/common.cc
@@ -126,12 +126,12 @@ void PopulateCommonUciStoppers(ChainedSearchStopper* stopper,
 
   // "go nodes" stopper.
   int64_t node_limit = 0;
-  if (params.nodes) {
+  if (params.nodes.has_value()) {
     if (options.Get<bool>(kNodesAsPlayoutsId)) {
       stopper->AddStopper(std::make_unique<PlayoutsStopper>(
           *params.nodes, options.Get<float>(kSmartPruningFactorId) > 0.0f));
     } else {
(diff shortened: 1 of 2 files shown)


mooskagh:
just a note, the PR names would be better if they actually summarize the change rather than just named the issue number.
(and "Fixes #2003" in the actual text would link the issue and the PR)

KarlKfoury:
> just a note, the PR names would be better if they actually summarize the change rather than just named the issue number. (and "Fixes #2003" in the actual text would link the issue and the PR)
you prefer a summary this detailed and descriptive? or less?

src/mcts/stoppers/common.cc
```diff
@@ -126,12 +126,12 @@ void PopulateCommonUciStoppers(ChainedSearchStopper* stopper,
...
+  if (params.nodes.has_value()) {
     if (options.Get<bool>(kNodesAsPlayoutsId)) {
       stopper->AddStopper(std::make_unique<PlayoutsStopper>(
           *params.nodes, options.Get<float>(kSmartPruningFactorId) > 0.0f));
     } else {
-      node_limit = *params.nodes;
+      node_limit = 4000000000;
     }
   }
   // always limit nodes to avoid exceeding the limit 4000000000. That number is
```
VaeterchenFrost, line 137: The comment L138 should also be updated, no?

mooskagh (approved):

Issue #1597, opened 2021-07-01
Integer overflow when memory limit is less than NN cache memory usage


mooskagh:
Example with memory limit 1MB:

```
0630 18:21:32.324037 36208 c:\projects\lc0\src\mcts\stoppers\stoppers.cc:110] RAM limit 1MB. Cache takes 59MB. Remaining memory is enough for 92233720368256758 nodes.
```

ErdoganSeref:
Does the integer overflow occur regarding the number of available nodes? How can we reproduce the issue?

mooskagh:
I believe if you run `./lc0` with `--ramlimit-mb=1 "--logfile=<stderr>"`, the issue will show up.
(instead of `<stderr>`, you can write log to a logfile, `--logfile=filename.txt`)

ErdoganSeref:
When I run `./lc0 --ramlimit-mb=1 "--logfile=log.txt"` I get the following log:

`0516 17:20:23.509182 140325955516224 ../../src/main.cc:45] Lc0 started.
0516 17:20:23.509351 140325955516224 ../../src/main.cc:46]        _
0516 17:20:23.509421 140325955516224 ../../src/main.cc:47] |   _ | |
0516 17:20:23.509476 140325955516224 ../../src/main.cc:48] |_ |_ |_| v0.30.0-dev+git.dirty built May 10 2022
0516 17:20:23.525305 140325955516224 ../../src/utils/commandline.cc:56] Command line: ./lc0 --ramlimit-mb=1 --logfile=log.txt`

I can't see the issue in the log file.

mooskagh:
Run some actual search, with `go wtime 1000` or something like that.

ErdoganSeref:
I reproduced the issue by running `./lc0 --ramlimit-mb=1 "--logfile=log.txt"` followed by `go wtime 1000` as you've said. But I still don't understand what the exact issue is. What should the expected output be?

mooskagh:
When "negative" amount of RAM is available (cache use is larger than total avaialable), it should allocate `0` nodes and not `92233720368256758`.

Naphthalin:
Was this addressed lately when RAM limit was discussed?

KarlKfoury:
i believe #2058 fixes issue

PR #2058, opened 2024-08-30
node limit passed as 0 instead of a negative number to visits_stopper if nn cache > memory limit (prevents unsigned int variable overflow)

KarlKfoury:
if cache size > ram limit, passing (ram limit) - (cache size) to visitsstopper as argument for node limit, which is an unsigned int, results in an error. Which is why it is necessary to pass 0 instead of (ram limit) - (cache size) if cache size > ram limit. Adressing issue #1597


src/search/classic/stoppers/stoppers.cc
```diff
@@ -97,8 +97,11 @@ MemoryWatchingStopper::MemoryWatchingStopper(int ram_limit_mb,
                                              uint32_t nodes,
                                              bool populate_remaining_playouts)
     : VisitsStopper(
-          (ram_limit_mb * 1000000LL - total_memory + avg_node_size * nodes) /
-              avg_node_size,
+          (ram_limit_mb * 1000000LL + avg_node_size * nodes > total_memory)
```
mooskagh: Maybe something like:
```cpp
[&]() {
  const auto ram_limit = ram_limit_mb * 1000000LL;
  const auto nodes_memory = avg_node_size * nodes;
  if (ram_limit + nodes_memory > total_memory) return 0;
  return (ram_limit + nodes_memory - total_memory) / avg_node_size;
}()
```

mooskagh (approved):

PR #2072, opened 2024-10-06
add simple JS backend

zamfofex-dummy:
I was told this might be desirable. A couple things to note:

- I wish to be able to publish this as a package to npm eventually. (I hope this can be allowed!)
- I have only tried this with Emscripten v3.1.57, see: emscripten-core/emscripten#22394
- If there is anything I can improve, I’d love to try to!
- I set this as a draft pull request because I think the documentation could be improved, and also because I have set up some ORT‐web compatibility changes very directly, whereas they should probably be conditional.
- This only works well in browsers with WebGPU enabled (and in computers with a decent GPU), so this is currently fairly niche.

If anyone has any thoughts on whether this is warranted, I’d love to know! (Do let me know what you think!)

Some more technical details: This works by converting an Lc0 network to ONNX format (with `leela2onnx`) and then using the ORT‐web APIs to allow Lc0 to interface with the network through JavaScript.

### DIFF
--- a/src/chess/uciloop.cc
+++ b/src/chess/uciloop.cc
@@ -37,6 +37,10 @@
 #include <unordered_set>
 #include <utility>
 
+#ifdef __EMSCRIPTEN__
+#include <emscripten.h>
+#endif
+
 #include "utils/exception.h"
 #include "utils/logging.h"
 #include "utils/string.h"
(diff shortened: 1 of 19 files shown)


src/neural/onnx/converter.h
```diff
@@ -45,7 +45,7 @@ struct WeightsToOnnxConverterOptions {
   std::string output_mlh = "/output/mlh";
   int batch_size = -1;
   int opset = 17;
-  bool alt_mish = false;       // Use "Mish" approximation (fp32 only).
+  bool alt_mish = true;        // Use "Mish" approximation (fp32 only).
```
borg323, line 48: This is not the way to do it, add a flag in `leela2onnx.cc`.

src/neural/onnx/builder.cc
```diff
@@ -289,7 +290,9 @@ std::string OnnxBuilder::Identity(const std::string& name,
 std::string OnnxBuilder::Selu(const std::string& name,
                               const std::string& input) {
   auto* node = model_.mutable_graph()->add_node();
-  return PopulateStdNodeFields(node, name, input, "Selu");
+  auto out = PopulateStdNodeFields(node, name + "/elu", input, "Elu");
+  AddFloatAttribute(node, "alpha", 1.0507);
+  return Mul(name, out, FloatOnnxConst({1.67326}, {1}));
 }
```
borg323, line 296: I would suggest using a flag, but probably doesn't make a difference. Better keep the `Selu` operator unchanged and add an `Elu` one, then make the change in the converter (with a comment).

src/chess/uciloop.cc
```diff
@@ -127,12 +131,37 @@ bool ContainsKey(const std::unordered_map<std::string, std::string>& params,
...
+  char *cline = lc0web_get_line();
+  line = cline;
+  free(cline);
+  return true;
+}
+
+#else
+
+bool GetLine(std::string& line) {
+  return std::getline(std::cin, line);
```
borg323, line 155: This breaks the build, I think one solution is to change it to `return !!std::getline(std::cin, line);`.

PR #2091, opened 2024-12-16
go perft command

borg323:


src/chess/uciloop.h
```diff
@@ -47,6 +47,7 @@ struct GoParams {
   std::optional<int> depth;
   std::optional<int> mate;
   std::optional<int> nodes;
+  std::optional<int> perft;
```
mooskagh, line 50: Would it make more sense to have just `perft` command rather than `go perft`, and then route to a separate function rather than to EngineClassic?

Especially with many search algorithms, bringing perft closer to search complicates more than help.

Even if we'd keep `go perft` instead of `perft`, I'd prefer Uci layer to call completely different function rather than something in the Engine class.

borg323: It is `go perft` because it is modeled after the stockfish version. Regarding the rest, I wasn't planning on getting this merged. If you think it is worth having then I'll do my best.

mooskagh: I believe it makes sense to have it, but probably not through uci but from the command line `./lc0 perft` similarly to `bench` and `backendbench`.

PR #2092, opened 2024-12-16
new backend interface prototype

borg323:
This is a prototype for the new backend interface, that doesn't have any dependence on the position encoding. For now only the search-cache interface is changed, the cache-backend interface is still the old one. The main changes to the interface are:
```c++
 // Check if entry is in the cache.
  bool CacheLookup(const PositionHistory& history, const MoveList& moves = {},
                   CachedNNRequest* entry = nullptr);
  // Adds a sample to the batch. Also calls EncodePositionForNN() if needed.
  void AddInput(const PositionHistory& history, const MoveList& moves);
  ```
  The other change is that `GetPVal()` returns directly the 16 bit compressed float we used, now formalized to pfloat16:
  ```c++
  // Returns compressed P value @move_id of @sample.
  pfloat16 GetPVal(int sample, int move_ct) const;
  ```


borg323:
No longer needed.

PR #1150, opened 2020-03-25
Add visit-based policy temperature decay

farmersrice:
Currently policy temperature is static. These changes will make policy temperature dynamic and unique for each node. The policy temperature of edges originating from a node is `i - c * log2(1 + v)`, where `i` is the initial policy temperature (same as PolicyTemperature in the current master implementation), `c` is a constant specified by PolicyTemperatureDecay in UCI and `--policy-temp-decay` as a flag, and `v` is the number of visits of that node.

Tuning and testing are currently in progress and look favorable.

PR #2093, opened 2024-12-17
Positive policy decay

Naphthalin:
This is a reimplementation of #1173 and #1288 without the child visit boost part, aimed at supplementing `CPuct` scaling. Setting either `PolicyDecayFactor` or `PolicyDecayExponent` to 0 effectively turns the policy decay off, so it should be tuning friendly.

### DIFF
--- a/src/mcts/params.cc
+++ b/src/mcts/params.cc
@@ -299,6 +299,14 @@ const OptionId SearchParams::kPolicySoftmaxTempId{
     "policy-softmax-temp", "PolicyTemperature",
     "Policy softmax temperature. Higher values make priors of move candidates "
     "closer to each other, widening the search."};
+const OptionId SearchParams::kPolicyDecayExponentId{
+    "policy-decay-exponent", "PolicyDecayExponent",
+    "Policy decay exponent. Sets the exponent of the visit based policy decay "
+    "term."};
+const OptionId SearchParams::kPolicyDecayFactorId{
+    "policy-decay-factor", "PolicyDecayFactor",
+    "Policy decay factor. Scales the visit count for the visit based policy "
+    "decay term."};
 const OptionId SearchParams::kMaxCollisionVisitsId{
     "max-collision-visits", "MaxCollisionVisits",
     "Total allowed node collision visits, per batch."};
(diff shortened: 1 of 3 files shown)


Naphthalin:
Some words of explanation:
- PUCT in the way Alphazero uses it treats policy as a multiplicative factor to the U term, persistent even at millions of nodes. The original reference meanwhile treats policy as an additional "delaying" term, decaying to 0 with more nodes.
- this approach (first version was a semi April fools joke in 2020) picks up the general idea of decaying the policy effect, but instead by affecting the multiplicative `P%` term, keeping the nature of the PUCT formula intact.
- it's crucial that policy decay (and generally any modification to PUCT) is done in a way that the U term still gets reduced after a visit, and slightly increased after a sibling is visited. Hence why "positive policy decay" is needed, otherwise search inconsistencies like #1150 happen.
- what this does is effectively removing the "policy sums up to 100%" condition, resp. effectively increases cpuct especially in positions with multiple good moves in addition to cpuct scaling.
- mathematically, the formula converts policy into a logit, and adds a logarithmically growing term. With N->inf, all P% will grow towards 100%, though the initial P% value decides how much this growth is delayed.

I expect this to work best together with a reduced `CPuctFactor`, with `PolicyTemperature`, `CPuct` and `FpuValue` optimized for STC. A tuning could therefore happen in two steps; first the STC tune (~1k npm) with the 3 parameters, and then a LTC tune (>100k npm) with `CPuctFactor`, `PolicyDecayExponent` and `PolicyDecayFactor`.

zz4032:
Tuning result with T74 (≈100Knodes/move) vs. SF
320 iterations
landscape-20251115-160436-320_resized
Looks like it wants to turn it policy decay effect off with `PolicyDecayFactor: 1e-06`. Policy values are almost identical to initial ones at 100Knodes.
image

Naphthalin:
@zz4032 Thanks a lot for doing the tune! I unfortunately found out that the parametrization I'm using here basically acts like a (soft-capped) flat multiplier, while the intention was to phase out policy altogether once a (policy dependent) amount of visits was reached.

PR #2095, opened 2024-12-19
Spinlock warning fix

mooskagh:
++ for volatile variables is deprecated in C++20

### DIFF
--- a/src/utils/mutex.h
+++ b/src/utils/mutex.h
@@ -127,6 +127,13 @@ class CAPABILITY("mutex") SharedMutex {
 static inline void SpinloopPause() {
 #if defined(__x86_64__)
   _mm_pause();
+#else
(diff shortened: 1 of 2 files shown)


src/utils/mutex.h
```diff
@@ -127,6 +127,13 @@ class CAPABILITY("mutex") SharedMutex {
 static inline void SpinloopPause() {
 #if defined(__x86_64__)
```
borg323, line 128: This needs to be `defined(__x86_64__) || defined(_M_X64)` as msvc doesn't support `__asm` for x86_64 but does support `_mm_pause()`. This will also break arm cross builds with msvc (again no `__asm`), but I'll deal with it when I update #1801.

borg323 (approved):

PR #2098, opened 2024-12-22
New backend API and wrapper for old networks. 

mooskagh:
The API is probably not yet final -- will see.
The code is not plugged to anywhere, but it compiles and links.

Based on #2097 (to be rebased after it's merged)


src/neural/wrapper.cc
```diff
@@ -0,0 +1,125 @@
...
+class NetworkAsBackend : public Backend {
+ public:
+  NetworkAsBackend(std::unique_ptr<Network> network)
+      : network_(std::move(network)) {
+    const NetworkCapabilities& caps = network_->GetCapabilities();
+    attrs_.has_mlh = caps.has_mlh();
+    attrs_.has_wdl = caps.has_wdl();
+    attrs_.runs_on_cpu = network_->IsCpu();
+    attrs_.suggested_num_search_threads = network_->GetThreads();
+    attrs_.maximum_batch_size = network_->GetMiniBatchSize();
```
borg323, line 44: `GetMiniBatchSize()` returns a recommendation for good performance, not the max.

src/neural/backend.h
```diff
@@ -0,0 +1,119 @@
...
+
+namespace lczero {
+
+// Information about the backend or network that search may need.
+struct BackendAttributes {
+  bool has_mlh;
+  bool has_wdl;
+  bool runs_on_cpu;
+  size_t suggested_num_search_threads;
+  size_t maximum_batch_size;
```
borg323, line 48: Also need a `recommended_batch_size` or similar.

borg323 (commented):
We don't have a mechanism to get the max batch size from the backends.
I would also like a mechanism to pass the max batch size from the search to the backend.

src/neural/wrapper.cc
mooskagh: Done.

src/neural/backend.h
mooskagh: Done.

mooskagh:
Added softmax policy temperature too, also removed cached batched API.

src/neural/wrapper.cc
```diff
@@ -0,0 +1,148 @@
...
+                   float softmax_policy_temperature)
+      : network_(std::move(network)),
+        softmax_policy_temperature_(1.0f / softmax_policy_temperature) {
+    const NetworkCapabilities& caps = network_->GetCapabilities();
+    attrs_.has_mlh = caps.has_mlh();
+    attrs_.has_wdl = caps.has_wdl();
+    attrs_.runs_on_cpu = network_->IsCpu();
+    attrs_.suggested_num_search_threads = network_->GetThreads();
+    attrs_.recommended_batch_size = network_->GetMiniBatchSize();
+    attrs_.maximum_batch_size = std::numeric_limits<int>::max();
```
borg323, line 51: A safe value to use for `maximum_batch_size` is 1024, as this is the maximum batch size lc0 will make.

src/neural/wrapper.cc
```diff
@@ -0,0 +1,148 @@
...
+    input_format_ = caps.input_format;
```
borg323, line 52: I think the input format is only needed for encoding, so no need to expose it.

mooskagh: It's not exposed (not in the `attrs_`), it's copied to the local field of the wrapper so that wrapper knows how to encode the input.

src/neural/wrapper.cc
mooskagh: Updated. In general it's not nice to hardcode it like that, but given wrapper is only for older backends, ok.

borg323: This should definitely be provided by the backend even for old ones, but 1024 will do until we implement a mechanism for it.

src/neural/wrapper.cc
borg323: I misread the code and was commenting as if it was copied to `attrs_`

src/neural/wrapper.cc
```diff
@@ -0,0 +1,148 @@
...
+    transforms_.reserve(backend_->attrs_.maximum_batch_size);
+  }
+
+  size_t UsedBatchSize() const override { return computation_->GetBatchSize(); }
+
+  AddInputResult AddInput(const EvalPosition& pos,
+                          EvalResultPtr result) override {
+    int transform;
+    computation_->AddInput(EncodePositionForNN(backend_->input_format_, pos.pos,
+                                               8, FillEmptyHistory::FEN_ONLY,
```
borg323, line 81: Shouldn't FillEmptyHistory also be a backend option?

src/neural/wrapper.cc
```diff
@@ -0,0 +1,148 @@
...
+        moves.begin(), moves.end(), -std::numeric_limits<float>::infinity(),
+        [&, counter = 0](float max_p, const Move& move) mutable {
+          return std::max(max_p, dst[counter++] = computation->GetPVal(
+                                     idx, move.as_nn_index(transform)));
+        });
+    // Compute the softmax and compute the total.
+    const float temperature = backend_->softmax_policy_temperature_;
+    float total = std::accumulate(
+        dst.begin(), dst.end(), 0.0f, [&](float total, float& val) {
+          return total + (val = FastExp((val - max_p) * temperature));
```
borg323, line 115: Not a review comment but a note to myself to check whether `FastExp()` helps here.

mooskagh: I decided to keep it for now and then change it in a separate step when everything is more stable, so that if the performance degrades after the switch (which is likely), it would be easier to find a cause.

borg323 (approved):
I don't see any other issues, we can always change stuff on the way.

PR #2104, opened 2024-12-31
Move backend-specific options out of search params.cc

mooskagh:
* Have a separate file that defines backend options (that are likely used by all backends)
* Make Backend-to-Weights wrapper use these options.
* Move those options off params.cc (still accessible through that class because it will be changed anyway later)

### DIFF
--- a/src/neural/factory.cc
+++ b/src/neural/factory.cc
@@ -30,26 +30,11 @@
 #include <algorithm>
 
 #include "neural/loader.h"
+#include "neural/shared_params.h"
 #include "utils/commandline.h"
 #include "utils/logging.h"
 
 namespace lczero {
-
-const OptionId NetworkFactory::kWeightsId{
-    "weights", "WeightsFile",
-    "Path from which to load network weights.\nSetting it to <autodiscover> "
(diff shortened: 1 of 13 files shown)


src/neural/factory.cc
```diff
@@ -111,10 +85,15 @@ bool NetworkFactory::BackendConfiguration::operator==(
...
+  std::string net_path =
+      options.Get<std::string>(SharedBackendParams::kWeightsId);
+  const std::string backend =
+      options.Get<std::string>(SharedBackendParams::kBackendId);
   const std::string backend_options =
-      options.Get<std::string>(kBackendOptionsId);
+      options.Get<std::string>(SharedBackendParams::kBackendOptionsId);
+
+  constexpr char* kAutoDiscover = "<autodiscover>";
+  constexpr char* kEmbed = "<built in>";
```
borg323, line 96: The visual stdio compiler on appveyor doesn't like the `constexpr` here.

mooskagh: Done, let's see if it compiles. I actually fixed it similarly in one place hoping that another one will fail too in clang, but it didn't.

borg323 (approved):
Looks good

PR #2107, opened 2025-01-02
Plug new Search and Backend APIs to the engine

mooskagh:
This PR plugs everything together.
And the new `policyhead` search algorithm works, through the new Backend interface!

### DIFF
--- a/src/search/register.h
+++ b/src/search/register.h
@@ -42,7 +42,7 @@ class SearchManager {
   static SearchManager* Get();
   void AddSearchFactory(std::unique_ptr<SearchFactory>);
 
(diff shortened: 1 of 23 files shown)


src/search/register.h
```diff
@@ -42,6 +42,12 @@ class SearchManager {
   static SearchManager* Get();
   void AddSearchFactory(std::unique_ptr<SearchFactory>);
 
+  std::vector<std::string_view> GetSearchesList() const;
```
borg323, line 45: I don't like `GetSearchesList()`, what about a simple `GetList()`?

borg323 (approved):

PR #2109, opened 2025-01-05
Switch classic search to Backend interface.

mooskagh:


mooskagh:
Rebased, this is ready for review now.

src/selfplay/game.cc
```diff
@@ -86,7 +86,7 @@ SelfPlayGame::SelfPlayGame(PlayerOptions white, PlayerOptions black,
                 black.uci_options->Get<bool>(kUciChess960)},
       training_data_(classic::SearchParams(*white.uci_options).GetHistoryFill(),
                      classic::SearchParams(*black.uci_options).GetHistoryFill(),
-                     white.network->GetCapabilities().input_format) {
+                     pblczero::NetworkFormat::INPUT_CLASSICAL_112_PLANE) {
```
borg323, line 89: Why do we hardcode this? In the past we have used `INPUT_112_WITH_CANONICALIZATION_HECTOPLIES` as well.

src/search/classic/search.cc
```diff
@@ -2038,36 +2025,18 @@ void SearchWorker::ExtendNode(Node* node, int depth,
...
-  auto planes =
-      EncodePositionForNN(search_->network_->GetCapabilities().input_format,
-                          history_, 8, params_.GetHistoryFill(), &transform);
-
-  std::vector<uint16_t> moves;
-
+  std::vector<Move> moves;
   if (node && node->HasChildren()) {
     // Legal moves are known, use them.
+    assert(false);  // Why would we ever be here?
```
borg323, line 2031: Maybe make it an exception to get wider testing (in release builds)?

src/search/classic/search.h
```diff
@@ -334,17 +327,10 @@ class SearchWorker {
     bool is_collision = false;
     int probability_transform = 0;
```
borg323, line 328: Is this still used anywhere?

src/selfplay/game.cc
mooskagh: With new backend API, backends don't expose internal NN topology, so in order to pass it to the training data encoder, we'd have to leak it through the backend interface, that I don't like.

However,
* This is only about the training data generation, for normal play older networks still work
* All networks currently in use for selfplay (or planned to) have `INPUT_CLASSICAL_112_PLANE`
* I hope/intend to make training data format topology-agnostic.
* If we ever want other (112-)plane formats on the current data format, it's possible to do the conversion on the training site (information is not lost).

src/search/classic/search.cc
mooskagh: Done.

src/search/classic/search.h
mooskagh: Indeed it wasn't, removed.

mooskagh:
Testing at https://bench.lczero.org/test/571/ after the update (mostly to confirm that replacing assert with exception is fine).

src/selfplay/game.cc
borg323: In that case, what about removing the parameter entirely?

borg323 (approved):

PR #2113, opened 2025-01-21
Add a makefile for OpenBench

mooskagh:


borg323 (approved):
I have no idea what is expected here, but I can't see this causing any issues.

mooskagh:
This is to make the repository compliant to items 2,3,4 of https://github.com/AndyGrant/OpenBench/wiki/Requirements-For-Public-Engines#basic-requirements  

The last item (`EVALFILE`) is something that would also be nice to support, but I'm not sure what's the best way to support it.

PR #2116, opened 2025-01-27
Add AUTHORs file

mooskagh:
Generated using `git log --pretty=format:"%an" | sort -u`, and then added `Google LLC` manually as they hold the copyright of some code by some of us.

There are a few obvious duplicates with different spelling, but I believe it's more correct like this (plus less work to generate and update).

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -14,7 +14,6 @@ Anson Hu
 Asger Alstrup Palm
 baum
 benini
-Bonan
 borg323
 Boštjan Mejak
 Brandon Lin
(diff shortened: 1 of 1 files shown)


AUTHORS
```diff
@@ -0,0 +1,113 @@
...
+Aloril
+Andrew Grant
+Andy Olsen
+Ankan
+Ankan Banerjee
+Anson Hu
+Asger Alstrup Palm
+baum
+benini
+Bonan
```
ContradNamiseb, line 17: Can you add my name 'Contrad Namiseb(Bonan)' there instead. I know this was generated from the commits but I think I prefer my real name there.

borg323 (approved):

PR #2119, opened 2025-02-04
Introduce "openbench" build parameter

mooskagh:
To use: `./build.sh -Dopenbench=true`

Builds Lc0 in the mode which is usable with Openbench:
* `bench` command runs 10 positions, half a second each (instead of running for 6 minutes)
* Options are unhidden, like in "pro" mode.
* Openbench Makefile uses that parameter now


mooskagh:
Closing this for now, will redo differently (`bench` vs `benchmark`)

PR #2121, opened 2025-02-05
Introduce a "valuehead" search algorithm.

mooskagh:
Instead of it being an option.

### DIFF
--- a/src/search/instamove/instamove.cc
+++ b/src/search/instamove/instamove.cc
@@ -47,7 +47,7 @@ class PolicyHeadSearch : public InstamoveSearch {
   Move GetBestMove() final {
     const std::vector<Position> positions = game_state_.GetPositions();
     MoveList legal_moves = positions.back().GetBoard().GenerateLegalMoves();
-    std::vector<EvalResult> res = backend()->EvaluateBatch(
+    std::vector<EvalResult> res = context_.backend->EvaluateBatch(
         std::vector<EvalPosition>{EvalPosition{positions, legal_moves}});
     const size_t best_move_idx =
         std::max_element(res[0].p.begin(), res[0].p.end()) - res[0].p.begin();
@@ -76,7 +76,7 @@ class ValueHeadSearch : public InstamoveSearch {
 
   Move GetBestMove() final {
     std::unique_ptr<BackendComputation> computation =
-        backend()->CreateComputation();
+        context_.backend->CreateComputation();
 
     PositionHistory history(game_state_.GetPositions());
     const ChessBoard& board = history.Last().GetBoard();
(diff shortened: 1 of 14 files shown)


mooskagh:
I'll need to fix the registration, currently it's

```
../src/search/instamove/instamove.cc:148:1: error: redefinition of 'reg3b50Yalgorithm'
  148 | REGISTER_SEARCH(ValueHeadFactory);
      | ^
../src/search/register.h:65:34: note: expanded from macro 'REGISTER_SEARCH'
   65 |   static SearchManager::Register reg3b50Y##algorithm(std::make_unique<alg>()); \
      |                                  ^
<scratch space>:244:1: note: expanded from here
  244 | reg3b50Yalgorithm
      | ^
../src/search/instamove/instamove.cc:147:1: note: previous definition is here
  147 | REGISTER_SEARCH(PolicyHeadFactory);
      | ^
../src/search/register.h:65:34: note: expanded from macro 'REGISTER_SEARCH'
   65 |   static SearchManager::Register reg3b50Y##algorithm(std::make_unique<alg>()); \
      |                                  ^
<scratch space>:243:1: note: expanded from here
  243 | reg3b50Yalgorithm
      | ^
```

mooskagh:
Fixed the registration, ready for review.

src/search/instamove/instamove.cc
```diff
@@ -68,7 +69,85 @@ class PolicyHeadFactory : public SearchFactory {
...
+  Move GetBestMove() final {
+    std::unique_ptr<BackendComputation> computation =
+        backend()->CreateComputation();
+
+    PositionHistory history(game_state_.GetPositions());
+    const ChessBoard& board = history.Last().GetBoard();
+    const std::vector<Move> legal_moves = board.GenerateLegalMoves();
+    std::vector<EvalResult> results(legal_moves.size());
+
+    for (size_t i = 0; i < legal_moves.size(); i++) {
```
borg323, line 86: The loop should avoid exceeding the `maximum_batch_size`, the earlier version did.

mooskagh: For that we'll need to have a Backend layer that would handle this (and add `SearchCapabilities` to tell whether it needs that layer).  
I'll send a separate PR for that (to be merged before this).

borg323 (approved):

src/search/instamove/instamove.cc
borg323: You can merge it as is and do the other PR later.

mooskagh: Actually now I have an idea how to do it without SearchCapabilities -- will update this PR.

mooskagh: I've added a backend wrapper that watches for the computation not to overflow the maximum batch size.
However, it revealed a design flaw in the Search class hierarchy that I'd need to think about:
- It turned our to be "borderline elegant" to do the wrapping in `SearchEnvironment::SetBackend()`, but if we would want to do wrapping per-search rather than per-environment, it wouldn't be as nice (if we wanted `SearchBase::backend()` to return correct backend)
- We'd need to make wrapped backend ownership consistent:
    - MemCache wrapper takes unique_ptr of the nested backend as a parameter, as it's convenient in the EngineController not to own two backends.
    - However, for BatchSplittingBackend, we cannot do that as in Search classes we only have non-owning Backend pointer.
    - I'm inclined to make MemCache non-owning too.

PR #2124, opened 2025-02-20
Update logging.h

CallOn84:
Update logging.h to include `#include <chrono>`, making it compatible with recent VS 2022 changes.

Doing this will allow for lc0 to be compiled.


CallOn84:
Unfortunately, the issue of `error C1083: Cannot open include file: 'zlib.h': No such file or directory` still exists

PR #2125, opened 2025-02-20
Rescorer build fix

ContradNamiseb:
~Though there is one final change that needs to be made in ```subproject/gaviotatb/compression/wrap.c``` change #include "zlib.h" to #include "zlib/zlib.h".~ I'll see if I can add this but it seems like it will need a small touch from @borg323.


CallOn84:
Unfortunately, the issue of `error C1083: Cannot open include file: 'zlib.h': No such file or directory` still exists

CallOn84:
With the `#include "zlib/zlib.h"` replacement, I can confirm this works

ContradNamiseb:
Added a script to automate zlib include change.

borg323:
> Added a script to automate zlib.h include change.

You can probably just change https://github.com/LeelaChessZero/lc0/blob/master/subprojects/packagefiles/gaviotatb/meson.build instead.

ContradNamiseb:
@borg323  done.

src/trainingdata/rescoreloop.cc
```diff
@@ -254,7 +254,7 @@ void gaviota_tb_probe_hard(const Position& pos, unsigned int& info,
   unsigned char bpc[17];
 
   auto stm = pos.IsBlackToMove() ? tb_BLACK_TO_MOVE : tb_WHITE_TO_MOVE;
-  auto& board = pos.IsBlackToMove() ? pos.GetThemBoard() : pos.GetBoard();
+  auto& board = pos.GetBoard();
```
borg323, line 257: I'm quite sure you need to `Mirror()` the board if it is black to move.

ContradNamiseb: Done, but this comment also needs to be changed:
` // Gets board from the point of view of player to move.
  const ChessBoard& GetBoard() const { return us_board_; }`

borg323 (approved):

CallOn84:
There is an error when building it where the `gaviotatb/compression/wrap.c` can't access `zlib.h` from the zlib folder. This was fixed by changing line 45 in `wrap.c` to the following:

```
#if defined(ZLIB)
#include "compression/zlib/zlib.h"
#endif
```

borg323:
Can you merge with master once more to see if it resolves the circleci build issue?

Issue #2103, opened 2024-12-27
Wrong device specification for host to device in xla runners


PikaCat-OuO:
#2081 is not fixing the issue.

This line:
https://github.com/LeelaChessZero/lc0/blob/c3a160c9438c6c8aae48c42705dff64161100b1d/src/neural/backends/xla/xla_runner.cc#L201

Should be changed to:
```c
devices_.at(device_).get()
```

borg323:
Fixed with #2131

PR #2131, opened 2025-03-25
fix xla device selection

borg323:


mooskagh:
Fixes #2103

borg323:
> Fixes #2103

I knew I forgot something...

PR #2132, opened 2025-03-26
Use fp16 data type for all tensors.

almaudoh:


borg323:
Can you add a backend option to make the data type configurable?

Also is bf16 supported by any actual device? I see there is `MPSDataType.bFloat16` but is only available of macos 14, so it would have to be optional.

borg323:
I would also like to see some error histograms from the check backend with both fp16 and fp32.

PR #2137, opened 2025-03-30
Add caching layer in the "new" Engine controller

mooskagh:
Only classic EngineController created caching backend layer, now the new one does too.  
Also the caching layer now follows the options change to update the cache size.

It's however quite inconsistent now, with constructor taking size as a parameter but also update function taking options, so it looks like the caching backend both "owns" the parameter and expects it to be external.

The reason we need that is that in some cases (selfplay), we don't want to pass the options, and in that case we know we won't call "UpdateParameters" on backend.

### DIFF
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -32,6 +32,7 @@
 #include "chess/gamestate.h"
 #include "chess/position.h"
 #include "neural/backend.h"
+#include "neural/memcache.h"
 #include "neural/register.h"
 #include "neural/shared_params.h"
 
@@ -70,7 +71,8 @@ void Engine::UpdateBackendConfig() {
   if (!backend_ || backend_name != backend_name_ ||
       backend_->UpdateConfiguration(options_) == Backend::NEED_RESTART) {
     backend_name_ = backend_name;
-    backend_ = BackendManager::Get()->CreateFromParams(options_);
+    backend_ = CreateMemCache(BackendManager::Get()->CreateFromParams(options_),
+                              options_);
     search_->SetBackend(backend_.get());
   }
 }
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -68,12 +68,16 @@ void Engine::EnsureSearchStopped() {
 void Engine::UpdateBackendConfig() {
   const std::string backend_name =
       options_.Get<std::string>(SharedBackendParams::kBackendId);
+  const size_t cache_size =
+      options_.Get<int>(SharedBackendParams::kNNCacheSizeId);
   if (!backend_ || backend_name != backend_name_ ||
       backend_->UpdateConfiguration(options_) == Backend::NEED_RESTART) {
     backend_name_ = backend_name;
     backend_ = CreateMemCache(BackendManager::Get()->CreateFromParams(options_),
-                              options_);
+                              cache_size);
     search_->SetBackend(backend_.get());
+  } else {
+    backend_->SetCacheSize(cache_size);
   }
 }
 
--- a/src/engine.h
+++ b/src/engine.h
@@ -30,6 +30,7 @@
 #include <vector>
 
 #include "engine_loop.h"
+#include "neural/memcache.h"
 #include "search/search.h"
 
 namespace lczero {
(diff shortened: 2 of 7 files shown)


borg323:
Does this look like an improvement? (Not exactly what copilot suggested).
```diff
diff --git a/src/engine.cc b/src/engine.cc
index 7f8543f0..4a873562 100644
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -71,8 +71,7 @@ void Engine::UpdateBackendConfig() {
   if (!backend_ || backend_name != backend_name_ ||
       backend_->UpdateConfiguration(options_) == Backend::NEED_RESTART) {
     backend_name_ = backend_name;
-    backend_ = CreateMemCache(BackendManager::Get()->CreateFromParams(options_),
-                              options_);
+    backend_ = CreateMemCache(options_);
     search_->SetBackend(backend_.get());
   }
 }
diff --git a/src/engine_classic.cc b/src/engine_classic.cc
index a3135fb8..31f5eb09 100644
--- a/src/engine_classic.cc
+++ b/src/engine_classic.cc
@@ -145,8 +145,7 @@ void EngineClassic::UpdateFromUciOptions() {
   const auto network_configuration =
       NetworkFactory::BackendConfiguration(options_);
   if (network_configuration_ != network_configuration) {
-    backend_ = CreateMemCache(BackendManager::Get()->CreateFromParams(options_),
-                              options_);
+    backend_ = CreateMemCache(options_);
     network_configuration_ = network_configuration;
   }
 
diff --git a/src/neural/memcache.cc b/src/neural/memcache.cc
index 7ffa1a68..1c7ea482 100644
--- a/src/neural/memcache.cc
+++ b/src/neural/memcache.cc
@@ -28,6 +28,7 @@
 #include "neural/memcache.h"
 
 #include "neural/cache.h"
+#include "neural/factory.h"
 #include "neural/shared_params.h"
 #include "utils/atomic_vector.h"
 #include "utils/smallarray.h"
@@ -171,4 +172,8 @@ std::unique_ptr<CachingBackend> CreateMemCache(std::unique_ptr<Backend> wrapped,
   return std::make_unique<MemCache>(std::move(wrapped), options);
 }
 
+std::unique_ptr<CachingBackend> CreateMemCache(const OptionsDict& opts) {
+  return CreateMemCache(BackendManager::Get()->CreateFromParams(opts), opts);
+}
+
 }  // namespace lczero
diff --git a/src/neural/memcache.h b/src/neural/memcache.h
index d79a1905..f424a4de 100644
--- a/src/neural/memcache.h
+++ b/src/neural/memcache.h
@@ -43,4 +43,5 @@ class CachingBackend : public Backend {
 std::unique_ptr<CachingBackend> CreateMemCache(std::unique_ptr<Backend> parent,
                                                const OptionsDict& options);
 
+std::unique_ptr<CachingBackend> CreateMemCache(const OptionsDict& opts);
 }  // namespace lczero
diff --git a/src/selfplay/tournament.cc b/src/selfplay/tournament.cc
index 168fa275..1d99b116 100644
--- a/src/selfplay/tournament.cc
+++ b/src/selfplay/tournament.cc
@@ -221,8 +221,7 @@ SelfPlayTournament::SelfPlayTournament(
       if (!backends_.contains(config)) {
         backends_.emplace(
             config,
-            CreateMemCache(BackendManager::Get()->CreateFromParams(opts),
-                           options.GetSubdict(name)));
+            CreateMemCache(options.GetSubdict(name)));
       }
     }
   }
diff --git a/src/tools/benchmark.cc b/src/tools/benchmark.cc
index d62b009c..916f330c 100644
--- a/src/tools/benchmark.cc
+++ b/src/tools/benchmark.cc
@@ -72,8 +72,7 @@ void Benchmark::Run(bool run_shorter_benchmark) {
   try {
     auto option_dict = options.GetOptionsDict();
 
-    auto backend = CreateMemCache(
-        BackendManager::Get()->CreateFromParams(option_dict), option_dict);
+    auto backend = CreateMemCache(option_dict);
 
     const int visits = option_dict.Get<int>(kNodesId);
     const int movetime = option_dict.Get<int>(kMovetimeId);
```

borg323 (approved):

borg323:
I got carried away and changed the selfplay cal as well in the above diff. The rest should be good.

mooskagh:
I've actually changed my mind and think we should not let caching the layer handle options -- we should both pass cache size to the constructor and have "Resize()" function that takes a number. Instead of intercepting UpdateParameters(), Engine itself will check the parameter and resize.

I'll update this PR with that change.

borg323 (approved):

PR #2139, opened 2025-04-03
onnx-trt improvements

borg323:
The main improvements here are:
1. The `min_batch` option to specify minimum batch size (normally 1, 4 for onnx-trt). Only used with variable batch size.
2. Prefix and directory for engine and timing files.
3. A different engine file per batch size (for fixed batch size).
4. A small cleanup.

### DIFF
--- a/src/neural/backends/network_onnx.cc
+++ b/src/neural/backends/network_onnx.cc
@@ -407,12 +407,9 @@ OnnxNetwork::OnnxNetwork(const WeightsFile& file, const OptionsDict& opts,
       bf16_(file.onnx_model().data_type() == pblczero::OnnxModel::BFLOAT16),
       provider_(provider) {
   batch_size_ =
-      opts.GetOrDefault<int>("batch", provider == OnnxProvider::DML   ? 16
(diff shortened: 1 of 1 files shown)


src/neural/backends/network_onnx.cc
```diff
@@ -319,24 +323,39 @@ Ort::SessionOptions OnnxNetwork::GetOptions(int gpu, int threads,
     case OnnxProvider::TRT: {
       options.SetExecutionMode(ExecutionMode::ORT_SEQUENTIAL);
 
+      std::string cache_dir = CommandLine::BinaryDirectory() + "/trt_cache";
```
mooskagh, line 326: That's not linux-friendly (where we could have e.g. `/usr/bin/lc0`). `GetUserCacheDirectory()` may be more appropriate.

borg323: I did it explicitly like that because I'm not yet sure how portable are the engine files created. Once we know that the files are safe to keep around, then I agree `GetUserCacheDirectory()` is the right thing to use.

mooskagh (approved):

PR #2149, opened 2025-04-06
Call clock related search functions from new Engine API

mooskagh:


src/engine.cc
```diff
@@ -121,8 +128,9 @@ void Engine::EnsureSyzygyTablebasesLoaded() {
 
 void Engine::SetPosition(const std::string& fen,
                          const std::vector<std::string>& moves) {
-  UpdateBackendConfig();
   EnsureSearchStopped();
+  if (!options_.Get<bool>(kStrictUciTiming)) search_->StartClock();
```
borg323, line 132: Shouldn't this line be first to be safe?

mooskagh: The "contract" for the `search_` is that timer start on `StartClock()`, and stops either at `StopSearch()` (then time manager can do whatever calculations it needs) or `AbortSearch()` (then time manager ignores this move).

Because of that, we need to stop previous search before starting the clock for the next (otherwise it will be immediately cancelled).

I don't like this too much but that sounded reasonable.

Other options would be to have explicit StartClock()/StopClock() pair (not nice e.g. because of search-induced stops), or changing rules a bit (may be better):
* StartClock() starts the clock.
* StartClock() while clock is ticking resets the clock.
* AbortSearch() does nothing to the clock (i.e. clock is "ticking" while search is stopped)
* StopSearch() stops and commits the clock.

borg323 (approved):

PR #2150, opened 2025-04-07
Support for ponder in Engine

mooskagh:
Includes changes from https://github.com/LeelaChessZero/lc0/pull/2149

### DIFF
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -29,7 +29,6 @@
 
 #include <algorithm>
 
-#include "chess/gamestate.h"
 #include "chess/position.h"
 #include "neural/backend.h"
 #include "neural/memcache.h"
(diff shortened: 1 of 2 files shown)


borg323 (approved):
1. Do we have to be so strict with the uci `Ponder`  option?
2.  See my comment in #2149

mooskagh:
Regarding being string with Ponder, I assume you mean this check?:

```
  if (!ponder_enabled_ && params.ponder) {
    throw Exception(
        "Ponder is not enabled, but the ponder search is requested.");
  }
```

What happens if you remove the check, is you lose part of a tree.

What we do in non-ponder mode:
* `position` UCI command sets the tree to a given position
* `go` starts search

What we do in ponder mode:
* `position` remembers the position but doesn't reset the tree
* `go` resets the tree to a given position starts the search
* `go ponder` resets the tree to one ply earlier and starts the search.

What happens if you ponder in non-ponder mode:
* `position` UCI command sets the tree to a given position  (discarding other moves of the parent position)
* `go ponder` has to reset the tree to one ply earlier, so the tree is trimmed
It reduces the performance but may not be easy to notice. Therefore I believe it's better to return an error.

PR #1925, opened 2023-10-11
sycl backend support for AMD, NVIDIA, and Intel

KateBlueSky:
Hi, I'm not sure if this SYCL backend I developed is something the Lc0 team would be interested in or not?  The SYCL backend is based off the  CUDA backend and can run on Nvidia, AMD, or Intel GPUs. I've tested it using the benchmark functionality and the 752187.pb.gz network, but there still needs to be a lot more testing done before merging this code. I was wondering if there is any automated test that I could run or how do you suggest I test the code?  Also merging the mason build for the sycl code to the meson build on main needs to be done. 

To compile the SYCL code for AMD or Nvidia you would need to do the following: 
1.) Download and set up the  [DPC++ LLVM compiler](https://github.com/intel/llvm).
2.) Run the buildSycl batch script included with this code. For AMD use the following command: CC=clang CXX=clang++ ./buildSycl.sh -DUSE_AMD_BACKEND=true -DUSE_SM=gfx90a (Set is to whatever architecture your using -i.e. for MI100 use gfx908 for MI250 use gfx90a) For Nvidia do the following CC=clang CXX=clang++ ./buildSycl.sh -DUSE_NVIDIA_BACKEND=true -DUSE_SM=61 (Set is to whatever architecture your using -i.e. for A100 use 80 for H100 use 90) 

To compile the code for Intel GPUs you would need to do the following:
1.) Download OneAPI  from the Intel webiste.
2.)  Run the buildSycl batch script included with this code. Use the following command CC=icpx CXX=icpx ./buildSycl.sh -DUSE_L0_BACKEND=true

To run the SYCL on the benchmark you would do the following command:
./lc0_sycl benchmark -b sycl

Thanks

PR #2152, opened 2025-04-11
sycl backend

borg323:
This is an updated version of #1925 from @KateBlueSky, made as far as I can tell by using the dpct tool to translate the cuda backend and then fixing issues by hand.
The files in the sycl directory were not formatted to allow easy merging with the upstream project (most likely https://github.com/oneapi-src/Velocity-Bench).
To build a recent version of meson is needed and even that with a hack to rewrite the ninja build file to replace the hardcoded linker and fix the dependency tracking. The only compiler tested is oneAPI's icx on Windows, but icpx was working in earlier versions on Linux.
The original PR had support for running on amd and nvidia gpus. The code is still in (see above) but the build system support has been removed as it hasn't been tested at all.


mooskagh (approved):

rafbiels (changes requested):
Thank you @borg323 for picking up this excellent work by @KateBlueSky, it's great to see SYCL coming to upstream lc0! I've been contributing recently to the [Velocity-Bench](https://github.com/oneapi-src/Velocity-Bench/tree/main/lc0) version of the SYCL backend and I'd be interested in upstreaming those updates in the future.

I can help with some testing here. I managed to build and run this PR on Linux with icpx 2025.1, but I needed to apply three small changes - please see the suggestions.

I also have access to NVIDIA and AMD hardware and could offer some testing for the corresponding SYCL backends if you'd be interested in including them in this PR. Otherwise, we could also look into enabling them in a subsequent PR.

borg323:
Thanks for the suggestions @rafbiels. This PR will be merged very soon, and any additional changes are welcome.

Regarding the nvidia and amd code paths in the sycl backend, I would be very interested in enabling them, but this isn't urgent. 

I also noticed Velocity-Bench has an [amd backend](https://github.com/oneapi-src/Velocity-Bench/tree/main/lc0/src/neural/amd) that I tried to upstream as ["hip"](https://github.com/borg323/lc0/tree/amd) but apparently it doesn't work, so any suggestions about that are also welcome.

You can ping me at our discord [chat](http://lc0.org/chat) if you have any questions.

PR #2155, opened 2025-04-12
add dag-preview search algorithm

borg323:
This is mostly the work of @Etcaqab merged into master as a new search algorithm.
The full history of the `search/dag_classic` directory can be found in https://github.com/borg323/lc0/tree/dag-refresh-lite.


src/search/dag_classic/wrapper.cc
```diff
@@ -0,0 +1,144 @@
...
+  }
+  void StopSearch() override {
+    if (search_) search_->Stop();
+  }
+  void AbortSearch() override {
+    if (search_) search_->Abort();
+  }
+
+  const OptionsDict* options_;
+  std::unique_ptr<dag_classic::TimeManager> time_manager_;
```
mooskagh, line 68: Hmm as the `wrapper.cc` is already in the namespace `dag_classic`, there's no need to explicitly mention the namespace everywhere. I'll do the same change for `classic` `wrapper.cc`

mooskagh (changes requested):

src/search/dag_classic/search.cc
```diff
@@ -0,0 +1,2418 @@
...
+}  // namespace
+
+Search::Search(const NodeTree& tree, Backend* backend,
+               std::unique_ptr<UciResponder> uci_responder,
+               const MoveList& searchmoves,
+               std::chrono::steady_clock::time_point start_time,
+               std::unique_ptr<SearchStopper> stopper, bool infinite,
+               bool ponder, const OptionsDict& options, TranspositionTable* tt,
+               SyzygyTablebase* syzygy_tb)
+    : ok_to_respond_bestmove_(!infinite),
```
mooskagh, line 158: Non-dag version has `&& !ponder` here:

```
ok_to_respond_bestmove_(!infinite && !ponder),
```

I guess it's also needed here?

src/search/dag_classic/search.cc
```diff
@@ -0,0 +1,2418 @@
...
+}
+
+Search::~Search() {
+  Abort();
+  Wait();
+  {
+    SharedMutex::Lock lock(nodes_mutex_);
+    CancelSharedCollisions();
+
+#ifndef NDEBUG
```
mooskagh, line 1101: assert() is already a noop when `NDEBUG` is set, it's not need to guard it here.

src/search/dag_classic/search.h
```diff
@@ -0,0 +1,510 @@
...
+#include "search/dag_classic/params.h"
+#include "search/dag_classic/stoppers/timemgr.h"
+#include "syzygy/syzygy.h"
+#include "utils/logging.h"
+#include "utils/mutex.h"
+
+namespace lczero {
+namespace dag_classic {
+
+typedef std::vector<std::tuple<Node*, int, int>> BackupPath;
```
mooskagh, line 52: Do you know what these ints are? If yes, could you add a comment?

src/search/dag_classic/search.cc
```diff
@@ -0,0 +1,2418 @@
...
+  WaitForTasks();
+  for (int i = 0; i < static_cast<int>(picking_tasks_.size()); i++) {
+    for (int j = 0; j < static_cast<int>(picking_tasks_[i].results.size());
+         j++) {
+      minibatch_.emplace_back(std::move(picking_tasks_[i].results[j]));
+    }
+  }
+}
+
+// Depth starts with 0 at root, so number of plies in PV equals depth.
```
mooskagh, line 1524: Do you know what are the two numbers that the function is supposed to return? The first one seems to be the repetitions of this position, but the second one is some magic...

mooskagh (changes requested):

src/search/dag_classic/wrapper.cc
```diff
@@ -0,0 +1,144 @@
...
+      *move_start_time_, std::move(stopper), params.infinite, params.ponder,
+      *options_, &tt_, syzygy_tb_);
+
+  LOGFILE << "Timer started at "
+          << FormatTime(SteadyClockToSystemClock(*move_start_time_));
+  search_->StartThreads(options_->Get<int>(kThreadsOptionId));
+}
+
+class DagClassicSearchFactory : public SearchFactory {
+  std::string_view GetName() const override { return "dag-classic"; }
```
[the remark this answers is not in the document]
borg323: What about `dag-preview`?

src/search/dag_classic/stoppers/alphazero.cc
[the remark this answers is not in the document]
borg323: Unfortunately the tree size calculation in stoppers.cc seems to be making it hard.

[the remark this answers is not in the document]
borg323: Took a deeper look and the refactoring needed seems rather large, so I'll pass for now.

src/search/dag_classic/wrapper.cc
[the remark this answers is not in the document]
mooskagh: Sounds good.

src/search/dag_classic/search.cc
borg323: It is the number of moves (or plies?) left to the next repetition. As far as I can tell this (and the last int in BackupPath above) are only used to avoid repeat calls to 'GetPliesSincePrevRepetition()'. See https://github.com/Etcaqab/lc0/commit/fab32ead400a37217a5ef9a73e3a662040faacdf

mooskagh (approved):

PR #2159, opened 2025-04-15
Fix broken bench.

mooskagh:


cn4750 (approved):

borg323:
Should we add a flag to ignore the moves to keep results comparable with older ones? Apparently the old `ResetToPosition()` ignored additional input. Luckily, this seems to be the only place where it is noticeable (but only did a very quick look through the code).

mooskagh:
Likely not worth it. Alternatively, we can remove the `moves` part instead, but then the positions will be different from stockfish's.

borg323 (approved):
We can always add a compatibility flag later if it causes trouble.

PR #2162, opened 2025-05-03
dag GetVerboseStats() fix and assorted cleanup.

borg323:
This is an old bug in dag's `GetVerboseStats()` that was missing the second to last move when making a cache lookup. It was not getting caught by the `ApplyMove()` assert so a cache lookup was done with the wrong moves and the copy in `GetCachedEvaluation()` was accessing out of bounds. Cache lookups before the new backend interface were more forgiving so this was never noticed.

### DIFF
--- a/src/search/classic/search.cc
+++ b/src/search/classic/search.cc
@@ -516,7 +516,9 @@ std::vector<std::string> Search::GetVerboseStats(Node* node) const {
     if (n && n->IsTerminal()) {
       v = n->GetQ(sign * draw_score);
     } else {
-      std::optional<EvalResult> nneval = GetCachedNNEval(n);
+      auto history = GetPositionHistoryAtNode(node);
+      std::optional<EvalResult> nneval = backend_->GetCachedEvaluation(
+          EvalPosition{history.GetPositions(), {}});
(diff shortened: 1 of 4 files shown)


src/search/classic/search.cc
```diff
@@ -609,28 +609,11 @@ PositionHistory Search::GetPositionHistoryAtNode(const Node* node) const {
...
-    std::transform(node->Edges().begin(), node->Edges().end(),
-                   std::back_inserter(moves),
-                   [](const auto& edge) { return edge.GetMove(); });
-    return moves;
-  }
-  return board.GenerateLegalMoves();
-}
-}  // namespace
-
 std::optional<EvalResult> Search::GetCachedNNEval(const Node* node) const {
```
mooskagh, line 612: nit, optional: maybe rename the function so something like `GetCacheNNValueHead` to make it clear that it doesn't return policy now.

mooskagh (approved):

src/search/classic/search.cc
borg323: What about removing the function completely?

mooskagh: Yeah even better.

PR #2163, opened 2025-05-03
add some resiliency against cache hash collisions

borg323:


src/neural/memcache.cc
```diff
@@ -46,6 +46,7 @@ struct CachedValue {
   float q;
   float d;
   float m;
+  uint32_t move_ct;
```
mooskagh, line 49: Maybe `uint8_t` as more than 230 legal moves are not possible even in theory?
Also, maybe `num_moves`? I believe we used `num_` in a few places already, while `_ct` is something unusual.

mooskagh (approved):

PR #2165, opened 2025-05-04
Use FEN (as URL) for chess board DebugString()

mooskagh:
It's more convenient to have it as one string, and clickable link is quite useful too.

### DIFF
--- a/src/chess/board.cc
+++ b/src/chess/board.cc
@@ -1113,51 +1113,9 @@ bool ChessBoard::HasMatingMaterial() const {
 }
 
 std::string ChessBoard::DebugString() const {
-  std::string result;
-  for (int i = 7; i >= 0; --i) {
-    for (int j = 0; j < 8; ++j) {
-      File file = File::FromIdx(j);
-      Square square(file, Rank::FromIdx(i));
-      if (!our_pieces_.get(square) && !their_pieces_.get(square)) {
-        if (i == 2 && pawns_.get(Square(file, kRank1)))
-          result += '*';
-        else if (i == 5 && pawns_.get(Square(file, kRank8)))
-          result += '*';
-        else
-          result += '.';
-        continue;
-      }
-      if (our_king_ == square) {
-        result += 'K';
-        continue;
-      }
-      if (their_king_ == square) {
-        result += 'k';
-        continue;
-      }
-      char c = '?';
-      if ((pawns_ & kPawnMask).get(square)) {
-        c = 'p';
-      } else if (bishops_.get(square)) {
-        if (rooks_.get(square))
-          c = 'q';
-        else
-          c = 'b';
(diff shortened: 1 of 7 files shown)


src/chess/board.cc
```diff
@@ -1212,4 +1170,63 @@ Move ChessBoard::ParseMove(std::string_view move_str) const {
   return Move::White(from, to);
 }
 
+std::string BoardToFen(const ChessBoard& board) {
```
borg323, line 1173: We already have `GetFen()` in position.h/cc, we don't need both. From a quick look `GetFen()` seems cleaner, but I like the `BoardToFen()` interface better.

borg323: Obviously `BoardToFen()` can't output move count and rule50 plies, so we probably need to make `GetFen()` call `BoardToFen()`.

mooskagh: I remembered we have this, but couldn't find!

Did a few changes:
- Made GetPieceAt take square instead of ints.
- Moved the code (with some further cleanup) to BoardToFen()
- Renamed GetFen() to PositionToFen(), and made it call BoardToFen() plus append two more fields.
- Made DebugString() of both Position and Board to call corresponding *ToFen() func, also removed the hash output.

src/chess/board.cc
```diff
@@ -1113,52 +1113,7 @@ bool ChessBoard::HasMatingMaterial() const {
...
-    }
-    if (i == 0) {
-      result += " " + castlings_.DebugString();
-      result += flipped_ ? " (from black's eyes)" : " (from white's eyes)";
-      result += " Hash: " + std::to_string(Hash());
-    }
-    result += '\n';
-  }
-  return result;
+  return "https://lc0.org/fen/" + BoardToFen(*this);
```
borg323, line 1116: What about a lichess link instead? Would need replacing spaces with underscores or plus signs. Something like this:
```diff
--- a/src/chess/board.cc
+++ b/src/chess/board.cc
@@ -1113,7 +1113,9 @@ bool ChessBoard::HasMatingMaterial() const {
 }
 
 std::string ChessBoard::DebugString() const {
-  return "https://lc0.org/fen/" + BoardToFen(*this);
+  auto fen = BoardToFen(*this);
+  std::replace(fen.begin(), fen.end(), ' ', '_');
+  return "https://lichess.org/analysis/" + fen;
 }
 
 Move ChessBoard::ParseMove(std::string_view move_str) const {
```

mooskagh: I'd (weakly) prefer to point it to a website that we own. We can have our URL that redirect to lichess though.

borg323: Sure, that was just a proof of concept. Also replacing spaces with underscores seems to work fine with the current redirect.

PR #2168, opened 2025-05-16
OptionsDict refactoring part 1

mooskagh:
* The OptionsDict class now uses std::variant rather than multiple inheritance as it's much cleaner.
* No functional changes, except, that it's not allowed now to store values of different types under the same name. Attempt to do that results an exception, as would the attempt to read the existing key under wrong type.
* `OptionsDict` was used in two different contexts: for UCI/CommandLine params, and as single-line config for backends, time managers etc. Now these use cases are split:
    * The generic key-value store of variant type is a `CascadingDict` template.
    * The instantiation for this template for single-line configs is `InlineConfig`. It has `std::string` key and doesn't have `Button` value.
    * The instantiation for the UCI/CommandLine is `OptionsDict` for now, but will be renamed to `ProgramOptions` in the next PR. It has `OptionId*` as a key (actually now temporarily std::reference_wrapper in order not to touch many places of the code multiple times), and `Button` as possible value.


mooskagh:
Oh that touched quite a lots of files, not sure if it's worth it..

Some of files may not be formatted properly, I'll have another PR to format all files.

borg323 (commented):
This is what I have so far.

PR #2171, opened 2025-05-23
Fix: Improve setoption command parsing

AlexisOlson:
Addresses issue #1843

### DIFF
--- a/src/chess/uciloop.cc
+++ b/src/chess/uciloop.cc
@@ -95,9 +95,15 @@ ParseCommand(const std::string& line) {
       *value += whitespace + token;
       whitespace = " ";
     } else {
-      value = &params[token];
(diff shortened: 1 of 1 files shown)


AlexisOlson:
This handles a case that #2062 does not.

For example, `setoption name MaxPrefetch value 1 value 2` returns `error Flag '--max-prefetch' has trailing characters.` 
from  `src/utils/optionsparser.cc` instead of silently setting `MaxPrefetch` to 12.

However, I noticed that it still gives the odd behavior `error Flag '--preload' must be either 'true' or 'false'.` given the input `setoption value livelogstats value true`.

Maybe I'll merge #2062 into this one.

borg323:
superseded by #2207

PR #2172, opened 2025-05-25
Support simple/normal/pro mode with new search algorithms.

mooskagh:
This is orthogonal to https://github.com/LeelaChessZero/lc0/pull/2168 but as it moved around lots of stuff, it doesn't merge closely. I'll rebase https://github.com/LeelaChessZero/lc0/pull/2168 onto this, given that we want this in the upcoming release and https://github.com/LeelaChessZero/lc0/pull/2168 can wait.

### DIFF
--- a/src/search/classic/stoppers/common.cc
+++ b/src/search/classic/stoppers/common.cc
@@ -41,30 +41,44 @@ const OptionId kRamLimitMbId{
     "positions have 30 possible moves. When set to 0, no RAM limit is "
     "enforced."};
 const OptionId kMinimumKLDGainPerNodeId{
-    "minimum-kldgain-per-node", "MinimumKLDGainPerNode",
-    "If greater than 0 search will abort unless the last "
-    "KLDGainAverageInterval nodes have an average gain per node of at least "
-    "this much."};
+    {.long_flag = "minimum-kldgain-per-node",
+     .uci_option = "MinimumKLDGainPerNode",
+     .help_text = "If greater than 0 search will abort unless the last "
+                  "KLDGainAverageInterval nodes have an average gain per node "
+                  "of at least this much.",
+     .visibility_mask = OptionId::kProModeMask}};
 const OptionId kKLDGainAverageIntervalId{
-    "kldgain-average-interval", "KLDGainAverageInterval",
-    "Used to decide how frequently to evaluate the average KLDGainPerNode to "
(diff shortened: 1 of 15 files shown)


mooskagh:
I plan to move all option parameters (like min/max/default value) to the OptionId constructor, but that's later.

src/search/classic/stoppers/common.cc
```diff
@@ -75,17 +89,6 @@ void PopulateCommonStopperOptions(RunType for_what, OptionsParser* options) {
       (for_what == RunType::kUci ? 1.33f : 0.00f);
   options->Add<IntOption>(kMinimumSmartPruningBatchesId, 0, 10000) = 0;
   options->Add<BoolOption>(kNodesAsPlayoutsId) = false;
-
-  if (for_what == RunType::kUci || for_what == RunType::kSimpleUci) {
-    options->Add<IntOption>(kRamLimitMbId, 0, 100000000) = 0;
```
borg323, line 80: `kRamLimitMbId` is still needed.

src/search/dag_classic/params.cc
```diff
@@ -561,28 +588,6 @@ void SearchParams::Populate(OptionsParser* options) {
...
   options->Add<BoolOption>(kSearchSpinBackoffId) = false;
-
-  options->HideOption(kNoiseEpsilonId);
-  options->HideOption(kNoiseAlphaId);
-  options->HideOption(kLogLiveStatsId);
-  options->HideOption(kRootHasOwnCpuctParamsId);
-  options->HideOption(kCpuctAtRootId);
-  options->HideOption(kCpuctBaseAtRootId);
-  options->HideOption(kCpuctFactorAtRootId);
-  options->HideOption(kFpuStrategyAtRootId);
```
borg323, line 572: Those 4 are not marked as `kProOnly`. I also marked `kFpuStrategyAtRootId` by mistake.

mooskagh:
I believe I addressed all comments (also found a few more inconsistencies between classic/params.cc an dag_classic/params.cc)

approximately-normal:
Hi [mooskagh](https://github.com/mooskagh), is there any documentation around usage of the simple/normal/pro mode options? I think it might be the cause of some issues I've been having with the Temperature setting, but I'm struggling to find much information at all about usage.

Issue #2177, opened 2025-05-30
Add `wait` UCI command


mooskagh:
Add a non-standard `wait` UCI command.

The command should block the command input until the current search finishes. Particularly, it would prevent stopping the search when stdin closes.

It would allow to do this, which is useful e.g. for batch evals without need for UCI hosts like python-chess:

```sh
$ echo <<END | ./lc0
position ...
go nodes 100000
wait
position ...
go nodes 200000
wait
END
```

Without `wait` command, it would start `go nodes 100000`, then without waiting for the search to finish, `position` and `go nodes 200000` would be issued, aborting the ongoing search, and then as stdin is closed, the engine would immediately exit, without completing the `go nodes 200000`.

With `wait` commands, it would complete both searches.

mooskagh:
Sounds good!
Note that an hour ago the old search API was removed, so there are no `engine_classic.*` files anymore.

mooskagh:
@JulianHelmsen it seems that the commit in your fork works fine.

However, to make it merged, we'd need you to open a Pull Request (just a note for the case you are not familiar with the GitHub flow).

PR #2184, opened 2025-06-06
Add wait UCI command (LeelaChessZero#2177)

JulianHelmsen:
Introduce "wait" uci command which waits until the previous ongoing search is completed.
This addresses #2177.


JulianHelmsen:
Come to think about it, there is a problem when the search is unbounded since no further commands are being processed after "wait". 

One solution would be to not allow this command to be used when an unbounded search is going on.

borg323:
Something went wrong when updating the tree, needs merging with master to work. Can be done with the "Update branch" button below.

Issue #2179, opened 2025-06-02
Port "technical" backends to the new API


mooskagh:
Recently we've switched from the [old API](https://github.com/LeelaChessZero/lc0/blob/master/src/neural/network.h) to [new API](https://github.com/LeelaChessZero/lc0/blob/master/src/neural/backend.h) for backends.

The old backends are currently working through old-to-new [wrapper](https://github.com/LeelaChessZero/lc0/blob/master/src/neural/wrapper.h).

Eventually, it would be nice to port all backends to the new API, but most importantly, it would be nice to port technical pass-through backends first.

Here are such backends, roughly sorted by the importance of converting to the new API:
1. [`multiplexing`](https://github.com/LeelaChessZero/lc0/blob/master/src/neural/backends/network_mux.cc)
2. [`demux`](https://github.com/LeelaChessZero/lc0/blob/master/src/neural/backends/network_demux.cc)
3. [`check`](https://github.com/LeelaChessZero/lc0/blob/master/src/neural/backends/network_check.cc)
4. [`roundrobin`](https://github.com/LeelaChessZero/lc0/blob/master/src/neural/backends/network_rr.cc)
5. [`recordreplay`](https://github.com/LeelaChessZero/lc0/blob/master/src/neural/backends/network_record.cc)

PR #2188, opened 2025-06-13
make it possible to specify the default backend

borg323:


mooskagh:
oops, nice bug.

Maybe something like:

```
std::vector<std::string> BackendManager::GetBackendNames() const {
  std::vector<size_t> indices(algorithms_.size());
  std::iota(indices.begin(), indices.end(), 0);

  std::sort(indices.begin(), indices.end(), [this](size_t a, size_t b) {
    const auto& fa = algorithms_[a];
    const auto& fb = algorithms_[b];
#ifdef DEFAULT_SEARCH    
    return std::tie(fa->GetName() == DEFAULT_SEARCH, fa->GetPriority()) >
           std::tie(fb->GetName() == DEFAULT_SEARCH, fb->GetPriority());
#else
    return fa->GetPriority() > fb->GetPriority();
#endif
  });

  std::vector<std::string> names;
  names.reserve(indices.size());
  for (size_t i : indices) names.emplace_back(algorithms_[i]->GetName());
  return names;
}
```

borg323:
My current version isthis:
```diff
--- a/src/neural/register.cc
+++ b/src/neural/register.cc
@@ -52,6 +52,16 @@ std::vector<std::string> BackendManager::GetBackendNames() const {
   std::transform(priority_and_names.begin(), priority_and_names.end(),
                  std::back_inserter(result),
                  [](const std::pair<int, std::string>& p) { return p.second; });
+#ifdef DEFAULT_BACKEND
+#define STRINGIFY_INTERNAL(x) #x
+#define STRINGIFY(x) STRINGIFY_INTERNAL(x)
+  std::string name = STRINGIFY(DEFAULT_BACKEND);
+#undef STRINGIFY
+#undef STRINGIFY_INTERNAL
+  auto pos = std::find(result.begin(), result.end(), name);
+  if (pos == result.end()) throw Exception("Unknown backend: " + name);
+  std::rotate(result.begin(), pos, pos + 1);
+#endif
   return result;
 }
 
```
However fixing the above bug seems a bit involved, looks like I'll have to do #2179 as well. So I'll get back to this one later.

src/neural/register.cc
```diff
@@ -52,6 +53,12 @@ std::vector<std::string> BackendManager::GetBackendNames() const {
   std::transform(priority_and_names.begin(), priority_and_names.end(),
                  std::back_inserter(result),
                  [](const std::pair<int, std::string>& p) { return p.second; });
+#ifdef DEFAULT_BACKEND
```
mooskagh, line 56: (optional) Maybe this to keep relative order of other backends?
We don't use that now, but maybe one day backends would e.g. refuse to be created.

```
#ifdef DEFAULT_BACKEND
  auto it = std::find(result.begin(), result.end(), DEFAULT_BACKEND);
  if (it == result.end()) throw Exception("Unknown backend: " + std::string(DEFAULT_BACKEND));
  result.erase(it);
  result.insert(result.begin(), DEFAULT_BACKEND);
#endif
```

PR #2193, opened 2025-06-24
update selfplay and backendbench backend interface use

borg323:
Also use 'BackendManager' for the backend values in the options and invalidate the cache if more backend options change.

### DIFF
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -164,16 +164,15 @@ void Engine::EnsureSearchStopped() {
 void Engine::UpdateBackendConfig() {
   const std::string backend_name =
       options_.Get<std::string>(SharedBackendParams::kBackendId);
-  const size_t cache_size =
-      options_.Get<int>(SharedBackendParams::kNNCacheSizeId);
   if (!backend_ || backend_name != backend_name_ ||
       backend_->UpdateConfiguration(options_) == Backend::NEED_RESTART) {
     backend_name_ = backend_name;
     backend_ = CreateMemCache(BackendManager::Get()->CreateFromParams(options_),
-                              cache_size);
+                              options_);
     search_->SetBackend(backend_.get());
   } else {
-    backend_->SetCacheSize(cache_size);
+    backend_->SetCacheSize(
+        options_.Get<int>(SharedBackendParams::kNNCacheSizeId));
   }
 }
 
--- a/src/neural/backend.cc
+++ b/src/neural/backend.cc
@@ -27,6 +27,11 @@
 
 #include "neural/backend.h"
 
+#include <string>
+
+#include "neural/shared_params.h"
+#include "utils/hashcat.h"
+
 namespace lczero {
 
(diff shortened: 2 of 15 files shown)


src/engine.cc
```diff
@@ -164,16 +164,15 @@ void Engine::EnsureSearchStopped() {
...
   if (!backend_ || backend_name != backend_name_ ||
       backend_->UpdateConfiguration(options_) == Backend::NEED_RESTART) {
     backend_name_ = backend_name;
     backend_ = CreateMemCache(BackendManager::Get()->CreateFromParams(options_),
-                              cache_size);
+                              options_);
     search_->SetBackend(backend_.get());
   } else {
-    backend_->SetCacheSize(cache_size);
+    backend_->SetCacheSize(
```
mooskagh, line 174: This (whether `MemCacheBackend` takes the cache size from the `kNNCacheSizeId` or from function arguments) changed back and for several times already. :-P

I think if it IS aware of the parameter, it should update it in `UpdateConfiguration`, and we don't need `SetCacheSize` at all.

src/neural/backend.cc
```diff
@@ -46,4 +51,18 @@ std::vector<EvalResult> Backend::EvaluateBatch(
   return results;
 }
 
+uint64_t Backend::ConfigurationHash(const OptionsDict& options) const {
```
mooskagh, line 54: It would be cleaner to have this function in `BackendFactory` rather than `Backend` as there's nothing specific to the instance of backend, and there may be situations when we want to check whether configuration changed before we created any backend.

But then there's no way to go from `Backend` (that we pass around) to `BackendFactory`. Maybe add a function like `BackendFactory* Backend::GetMyFactory()`?

Hmm but that doesn't really work for multi-layered backends. With the current approach, it kind of works but still not fully.

Suppose we have backend A which uses --a-param, and backend B which uses --b-param. 

We have created our `--backend=multiplexing --backend-opts=backend=A --a-param=5`.  
Now we run `backend.ConfigurationHash(conf)`, where `conf` is `--backend=multiplexing --backend-opts=backend=B --a-param=7`. `multiplexing` will compute hash of `backend-opts` (and hopefully will NOT recursively call backend `A` because it's not in the new config), but in any case, it will leave `--b-param` not hashed.

Ok, so I believe the way out of it is still to create `BackendFactory::ConfigurationHash`, and for wrapper backends make this function look up factories of nested backends by name.

Then it probably means that MemCacheBackend also has to have own factory (just not registered)..

src/engine.cc
mooskagh: ...and then `Engine` class may go back to storing `std::unique_ptr<Backend>` rather than `std::unique_ptr<MemCacheBackend>)`.

(and also see below that MemCacheBackend should probably have own BackendFactory)

borg323: I was leaving this for a follow-up.

PR #2195, opened 2025-06-29
Update of CONTRIBUTING.md

mooskagh:
Old one is a bit outdated and contained some weird stuff.

I've started to write the dev documentation in lczero.org, and realized that many of that would make more sense in CONTRIBUTING.md

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,18 +10,23 @@ guidelines for contributing to the codebase.
 * All contributors are encouraged to join our Discord server at
   <https://lc0.org/chat>.
 * Refer to [README.md](README.md) for building and running instructions.
+  * The protobufs that are shared with the training code are located in a
+    separate repository. Don't forget to run `git submodule update --init` to
+    fetch them.
 * Familiarize yourself with the developer documentation at
   <https://lczero.org/dev/>.
-* We use the [Meson](https://mesonbuild.com/) build system.
-  * In Linux, using `builddir/` is recommended for development
-    (`meson setup builddir/`), as VSCode recognizes it and all the development
-    and debugging tools work there (ask in Discord if you have issues).
-    * In the `builddir/`, run `ninja lc0` (which is faster than just `ninja`).
-      Run `ninja test` to run the tests.
-  * In Windows, `meson setup build/debug` generates a Visual Studio solution
-    that can be used for development.
-  * Check `meson_options.txt` for the available build options (to use them,
-    pass `-D<name>=<value>` to `meson setup`).
+
+We use the [Meson](https://mesonbuild.com/) build system.
+
+* In Linux, using `builddir/` is recommended for development
+  (`meson setup builddir/`), as VSCode recognizes it and all the development and
+  debugging tools work there (ask in Discord if you have issues).
+  * In the `builddir/`, run `ninja lc0` (which is faster than just `ninja`). Run
+    `ninja test` to run the tests.
+* In Windows, `meson setup build/debug` generates a Visual Studio solution that
+  can be used for development.
+* Check `meson_options.txt` for the available build options (to use them, pass
+  `-D<name>=<value>` to `meson setup`).
 
 ## Sending Pull Requests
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -1,81 +1,132 @@
...
+  knowledge.
+* **Core vs. auxiliary code**: While AI assistance works well for auxiliary code
+  (website, documentation, simple tools), it has failed so far on core lc0
+  engine code. We discourage using agentic coding for core engine components but
+  welcome it for simpler projects within the Leela Chess Zero organization.
+
+Remember: You are responsible for the quality and correctness of all code you
+submit, regardless of how it was generated.
+
+Thank you for helping make LCZero better!
```
borg323: "LCZero" or "Leela Chess Zero"?

CONTRIBUTING.md
```diff
@@ -1,81 +1,132 @@
...
+* **References vs Pointers**: Non-const reference parameters are neither
+  encouraged nor discouraged over pointers (in the Google style guide, they used
+  to be discouraged, and now they are encouraged).
+* **Formatting**: Run `clang-format` on all code before committing.
+* **RValue references**: Rvalue reference function parameters are allowed, not
+  only for constructors and assignment operators. However, use them only if they
+  benefit readability or have significant performance gain.
+* Every new file should contain a GPLv3 banner with:
+  * Additional exception under GNU GPL version 3 section 7 regarding NVIDIA
+    libraries (see examples in other files).
```
borg323: We also allow backends to be GPL only without the exception..

CONTRIBUTING.md
```diff
@@ -1,81 +1,132 @@
...
+  name but ending with `_test.cc`.
+
+## Style Guidelines
+
+We follow the
+[Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html) with
+these modifications:
+
+* **Header guards**: Use `#pragma once` instead of traditional header guards.
+* **Exceptions**: are allowed, but only one: `lczero::Exception`.
```
borg323, line 98: I'm quite sure there are some exceptions here, but I can't check at the moment.

CONTRIBUTING.md
```diff
@@ -1,81 +1,132 @@
...
+  <https://lczero.org/dev/>.
+
+We use the [Meson](https://mesonbuild.com/) build system.
+
+* In Linux, using `builddir/` is recommended for development
+  (`meson setup builddir/`), as VSCode recognizes it and all the development and
+  debugging tools work there (ask in Discord if you have issues).
+  * In the `builddir/`, run `ninja lc0` (which is faster than just `ninja`). Run
+    `ninja test` to run the tests.
+* In Windows, `meson setup build/debug` generates a Visual Studio solution that
```
borg323, line 26: Are you sure about that? I think ninja is the default under windows as well. BTW I have started using ninja under windows as well, I find it easier to work with.

borg323 (approved):
You don't have to do anything about my 4 minor comments, I just added them as reminders.

CONTRIBUTING.md
mooskagh: That's how it was in 2018 when I still used windows. Maybe it's changed or there are several options.

At least `build.cmd` seems to invoke `msbuild` and not `ninja`.

CONTRIBUTING.md
mooskagh: It was in the previous version of CONTRIBUTING.md, and quick grepping didn't find anything other being thrown.

CONTRIBUTING.md
mooskagh: Updated.

CONTRIBUTING.md
borg323: The other uses are in the opencl backend and in pjrt.cc, so I think the text is fine as is.

CONTRIBUTING.md
borg323: Yes, `ninja` is not always installed with visual studio, IIRC you get it if you ask for ~~`clang`~~`cmake` during installation, so it is safer to use `msbuild` in `build.cmd`. However I think to get a VS solution you need to pass the correct `--backend` option to `meson`. No worries, I'll revise this section when I get back my windows development laptop to run a few tests

borg323 (approved):

PR #2196, opened 2025-07-01
Improve windows build script

xfearofdarkness:
Update build.cmd to use CUDA_PATH from the system variables automatically set by the installer instead of hard coding it. If it is not defined, clear the paths related to it to avoid setting up invalid paths. If cuda is enabled and CUDA_PATH is not defined, exit as it will not compile. As a bonus, the user can also set the CUDA_PATH variable before executing manually to point to another cuda installation. Any style/logic improvements are very welcome.

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -13,13 +13,37 @@ set EIGEN=false
 set TEST=false
 
 rem 2. Edit the paths for the build dependencies.
-set CUDA_PATH=C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v10.0
-set CUDNN_PATH=%CUDA_PATH%
+if "%CUDA%"=="false" (
+    echo CUDA build option is OFF. Skipping CUDA/cuDNN path setup.
+    rem Clear CUDA-related paths when CUDA backend is off
+    set "CUDA_PATH="
+    set "CUDNN_PATH="
+    set "OPENCL_LIB_PATH="
+    set "OPENCL_INCLUDE_PATH="
+) else (
+    echo CUDA build option is ON. Validating CUDA_PATH...
+    if defined CUDA_PATH (
+        echo CUDA_PATH found in system environment: "%CUDA_PATH%"
+        rem Proceed with derived paths as CUDA_PATH is defined
+    ) else (
+        echo.
+        echo ERROR: CUDA build option is ON, but CUDA_PATH environment variable is NOT set.
+        echo Please ensure CUDA is installed and CUDA_PATH is correctly defined
+        echo as a system or user environment variable, then restart your console/reboot.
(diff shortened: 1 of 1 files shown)


borg323:
The intention of `build.cmd` is to make it easy for people to build lc0 on windows with minimal effort. In this respect, using a global `CUDA_PATH` to avoid needing to edit the file is a definite improvement. However adding more complicated checks in the section titled `2. Edit the paths for the build dependencies.` is not making it any easier to see what needs changing.
Moreover, erroring out if `CUDA_PATH` is not set if `CUDA` is set makes it harder for people who use the script for other backends, as they now have to set it to false explicitly. The idea is that the build should go on, ignoring paths that are not available.

xfearofdarkness:
I deleted the error and exit logic, as it might require more work for the user than absolutely necessary. I kept the warning about the `CUDA_PATH` variable, but maybe there should be more clear warnings. Users who just run the script and wanted to build with other backends enabled might not know that some variables are derived from the `CUDA_PATH`. It would need some more checking for conditions, but maybe if `CUDA_PATH` is not defined, and either `CUDA`, `CUDNN` or `OPENCL` is selected, then throw this warning with a little bit more explanation. This would make missing dependencies much easier to fix. If the derived paths need to be edited this would also make it harder now. Maybe I could also revise the logic in that case? Maybe it would be better to add commented out lines for the user for editing these variables.

xfearofdarkness:
My proposal would be to keep the warning but implement setting the `CUDA_PATH` variable like you originally intended, with some minor adjustment like this:
```cmd
rem 2. Edit the paths for the build dependencies.

if "%CUDA%"=="true" (
  if not defined CUDA_PATH (
    echo WARNING: CUDA_PATH environment variable not found. Using default value.
  ) else (
    echo CUDA_PATH found in system environment: "%CUDA_PATH%"
  )
)

if not defined CUDA_PATH set CUDA_PATH=C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v12.9
set CUDNN_PATH=%CUDA_PATH%
set OPENCL_LIB_PATH=%CUDA_PATH%\lib\x64
set OPENCL_INCLUDE_PATH=%CUDA_PATH%\include
set OPENBLAS_PATH=C:\OpenBLAS
set MKL_PATH=C:\Program Files (x86)\IntelSWTools\compilers_and_libraries\windows\mkl
set DNNL_PATH=C:\dnnl_win_1.1.1_cpu_vcomp
```

borg323:
This one is perfectly fine and clear. Maybe the `if "%CUDA%"=="true"` block would be better before the `rem` statement but I'm not certain, I'll leave that decision to you.

xfearofdarkness:
Sure I moved it before the comment. Thanks for the feedback

PR #2202, opened 2025-07-10
update roundrobin backend

borg323:


src/neural/backends/network_rr.cc
```diff
@@ -25,83 +25,80 @@
...
+      AddBackend(backends[0], options);
     }
 
     for (const auto& name : parents) {
-      AddBackend(name, weights, options.GetSubdict(name));
+      std::string s;
+      // Make the subdict into a string for backend-opts, excluding the options
+      // used here.
+      for (const auto& e : options.GetSubdict(name).AsStringVec()) {
+        if (e.find("backend") == 0) continue;
```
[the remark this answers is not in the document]
mooskagh: Note that both key and value can be in quotes, e.g. `"backend"="cuda"` also work, so it would be better to parse it into OptionDict rather than searching for a string.
And if we do match string, we could use regex, I believe all supported compilers have it in standard library (I believe it appeared in C++11, but got real support much later)

src/utils/optionsdict.h
```diff
@@ -69,7 +69,21 @@ class TypeDict {
       }
     }
   }
-
+  std::vector<std::string> AsStringVec() const {
+    std::vector<std::string> r;
+    for (auto const& option : dict_) {
+      std::string t = option.first + '=';
```
mooskagh: I'd send both key and string value through some "GetStringLiteral" function, which would put quotes around, and replaced existing `"` with `\"`.

src/neural/register.h
```diff
@@ -71,9 +71,9 @@ class BackendManager {
   std::vector<std::unique_ptr<BackendFactory>> algorithms_;
 };
 
-#define REGISTER_BACKEND(factory)                   \
-  namespace {                                       \
-  static SearchFactory::Register reg29c93##factory( \
-      std::make_unique<factory>());                 \
+#define REGISTER_BACKEND(factory)                    \
+  namespace {                                        \
+  static BackendManager::Register reg29c93##factory( \
```
mooskagh, line 77: Oh we didn't properly register backends?..

src/neural/backend.h
```diff
@@ -47,6 +47,16 @@ struct BackendAttributes {
   int suggested_num_search_threads;
   int recommended_batch_size;
   int maximum_batch_size;
+
+  void Merge(BackendAttributes& other) {
```
mooskagh: **const** BackendAttributes&

src/neural/backends/network_rr.cc
```diff
@@ -25,83 +25,80 @@
...
+      options.GetMutableSubdict(name)->Set<std::string>(
+          SharedBackendParams::kBackendOptionsId, s);
+      AddBackend(name, options.GetSubdict(name));
     }
   }
 
-  void AddBackend(const std::string& name,
-                  const std::optional<WeightsFile>& weights,
-                  const OptionsDict& opts) {
+  void AddBackend(const std::string& name, const OptionsDict& opts) {
```
mooskagh: I assume the RoundRobin part is not yet finalized, right? (I expected that `AddBackend` wouldn't query the backend itself but rather take it from the parameter, as we discussed)

src/neural/backends/network_rr.cc
```diff
@@ -25,83 +25,80 @@
...
+      auto backends = BackendManager::Get()->GetBackendNames();
+      AddBackend(backends[0], options);
     }
 
     for (const auto& name : parents) {
-      AddBackend(name, weights, options.GetSubdict(name));
+      std::string s;
+      // Make the subdict into a string for backend-opts, excluding the options
+      // used here.
+      for (const auto& e : options.GetSubdict(name).AsStringVec()) {
```
mooskagh: So we return the vector of strings rather than a string, only to filter `backend` out?

Seems ugly, but I cannot come up with better suggestion...

Maybe just leave it as it is?

It will probably trigger "unread key" error, but actually even with this part it's problematic because technical backends have own options.
E.g. `--backend=multiplexing --backend-opts=max_batch=128,backend=cuda,(gpu=0),(gpu=1)`  (`max_batch` will not be read by cuda backend)
or `--backend=check --backend-opts=mode=histo,...`

src/utils/optionsdict.cc
```diff
@@ -340,4 +340,25 @@ void OptionsDict::CheckAllOptionsRead(
   }
 }
 
+std::vector<std::string> OptionsDict::AsStringVec() const {
```
mooskagh:
```
std::vector<std::string> OptionsDict::AsStringVec() const {
  std::vector<std::string> result;
  
  for (auto append : {&TypeDict<bool>::AsStringVec(),
                      &TypeDict<int>::AsStringVec, 
                      &TypeDict<float>::AsStringVec,
                      &TypeDict<std::string>::AsStringVec}) {
    auto vec = (this->*append)();
    result.insert(result.end(), vec.begin(), vec.end());
  }
  
  for (const auto& [name, dict] : subdicts_) {
    auto entries = dict.AsStringVec();
    result.push_back(name + "(" + absl::StrJoin(entries, ",") + ")");
  }
  
  return result;
}
```

src/neural/backends/network_rr.cc
borg323: Thinking about it some more, the issue is with the wrapper: The `backend` option and any other option read by the technical backend should have the `was_read_since_last_set_` flag set so the backend can safely ignore it, but unfortunately the wrapper makes a new `OptionsDict` so the flag is once more clear.

PR #2204, opened 2025-07-14
build support for onnx-trt

borg323:


dist/README-onnx-trt.txt
```diff
@@ -0,0 +1,53 @@
+Lc0
+
+Lc0 is a UCI-compliant chess engine designed to play chess via
+neural network, specifically those of the LeelaChessZero project
+(https://lczero.org).
+
+To run this version you will also need several dll files from NVIDA's
+CUDA, cuDNN and TensorRT. Those dlls can either be on the path from
+an existing (or new) installation, or can be in the Lc0 folder.
```
mooskagh: This sentence is a bit confusing, maybe:

Those dlls can either be on the system path from a separate installation of these libraries, or can be placed directly in the Lc0 folder.

PR #2208, opened 2025-07-17
Simplify task reading code

Menkib64:
Atomic variables can be used without locks. We want an id from the shared taksks_taken_ counter. We want to increment the counter while staying less than or equal to the task_count_. This can be done using compare and exchange operation. We call compare exchange in a loop because it will either success soon or other thread increment counters to an equal value.

I have been learning the code base for a few days. I'm thinking about updating opencl backend to handle latest network files. I noticed potential for a minor simplification while reading the code. I decided to make the small changes to learn how pull request are handled.


mooskagh:
We are currently in the middle of release, so we are holding off non-urgent PR review for now. Will take a look soon.

Menkib64:
I think I will drop this pull request. I will instead make dag code shared for both. This should reduce code duplication for different implementations.

PR #2210, opened 2025-07-18
update authors, changelog and version before branch

borg323:


mooskagh (approved):
Looks good (except what copilot already noticed)

PR #2213, opened 2025-07-20
Update opencl.hpp

Menkib64:
There has been incompatible extension changes in opencl.h. My system opencl.h doesn't compile with old opencl.hpp because type names has changed. The new opencl.hpp adds version checks for incompatible extensions.
There was a typo in the last release which was fixed by KhronosGroup/OpenCL-CLHPP#321. Therefore choice is the latest version from KhronosGroup/OpenCL-CLHPP@2a60842

### DIFF
--- a/src/neural/backends/opencl/OpenCL.h
+++ b/src/neural/backends/opencl/OpenCL.h
@@ -36,7 +36,13 @@ using net_t = float;
 #include <string>
 #include <vector>
 
+#if __has_include("CL/opencl.hpp")
+#include "CL/opencl.hpp"
+#elif __has_include("OpenCL/opencl.hpp")
+#include "OpenCL/opencl.hpp"
+#else
 #include "opencl.hpp"
+#endif
 
 #include "neural/backends/opencl/OpenCLBuffers.h"
 #include "neural/backends/opencl/OpenCLParams.h"


Menkib64:
opencl.h is 2024.10.24 and a patch which adds cl_amd_device_memory_flags extension.

Alternative would be using the last release but manually apply the bug fix.

The problematic changes are D3D integration, cl_khr_command_buffer_mutable_dispatch, and cl_khr_command_buffer

borg323:
> Alternative would be using the last release but manually apply the bug fix.

Nah, don't worry. I'll approve and merge soon (after a local build test).

borg323:
An alternative approach is to remove the file completely from `third_party` and assume the system provides it:
```diff
--- a/src/neural/backends/opencl/OpenCL.h
+++ b/src/neural/backends/opencl/OpenCL.h
@@ -36,7 +36,7 @@ using net_t = float;
 #include <string>
 #include <vector>
 
-#include "opencl.hpp"
+#include "CL/opencl.hpp"
 
 #include "neural/backends/opencl/OpenCLBuffers.h"
 #include "neural/backends/opencl/OpenCLParams.h"
```
This works for me just fine.

borg323:
Thanks. Something like the following should do it.
```c++
#if __has_include("CL/opencl.hpp")
#include "CL/opencl.hpp"
#else
#include "opencl.hpp"
#endif
```

borg323:
To test you can just edit `.circleci/config.yml`, wait for the test and then revert the commit. But I don't expect any trouble, I assume if there is a local version it will be compatible with the corresponding `opencl.h` and if not the `third_party` version is working AFAIK.

borg323:
Thanks for doing this. Also, are you on discord by any chance?

PR #2214, opened 2025-07-22
add onnx model hash to trt cache prefix

borg323:
It seems tensorrt's engine cache doesn't differentiate between models with the same architecture. Adding the hash of the full onnx model to the engine prefix should take care of it.

### DIFF
--- a/src/neural/backends/network_onnx.cc
+++ b/src/neural/backends/network_onnx.cc
@@ -103,7 +103,8 @@ class OnnxNetwork : public Network {
   }
   bool IsCpu() const override { return provider_ == OnnxProvider::CPU; }
 
-  Ort::SessionOptions GetOptions(int gpu, int threads, int batch_size);
+  Ort::SessionOptions GetOptions(int gpu, int threads, int batch_size,
+                                 uint64_t hash);
 
   Ort::Env onnx_env_;
   // Prepare sessions for this many multiples of the batch size;
@@ -296,7 +297,7 @@ void OnnxComputation<DataType>::ComputeBlocking() {
 }
 
 Ort::SessionOptions OnnxNetwork::GetOptions(int gpu, int threads,
-                                            int batch_size) {
+                                            int batch_size, uint64_t hash) {
   Ort::SessionOptions options;
   options.SetIntraOpNumThreads(threads);
   options.SetGraphOptimizationLevel(GraphOptimizationLevel::ORT_ENABLE_ALL);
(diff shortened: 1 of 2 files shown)


src/neural/backends/network_onnx.cc
```diff
@@ -331,8 +332,10 @@ Ort::SessionOptions OnnxNetwork::GetOptions(int gpu, int threads,
       trt_options["trt_max_partition_iterations"] = "1000";
       trt_options["trt_min_subgraph_size"] = "1";
       trt_options["trt_engine_cache_enable"] = "1";
+      // We need the batch size as well as the hash, as it is set after loading.
       trt_options["trt_engine_cache_prefix"] =
-          "Lc0_ONNX_TRT_batch_" + std::to_string(batch_size) + "_";
+          "Lc0_ONNX_TRT_ORT_" + Ort::GetVersionString() + "_batch_" +
```
mooskagh, line 337: There is now `absl::StrCat`, which is usually nicer. It does std::to_string() for non-string parameters automatically, and also faster (not that we care about it here).

There's also `absl::StrFormat`,probably less convenient in this case.

src/neural/backends/network_onnx.cc
```diff
@@ -457,11 +460,25 @@ OnnxNetwork::OnnxNetwork(const WeightsFile& file, const OptionsDict& opts,
   std::transform(outputs_.begin(), outputs_.end(),
                  std::back_inserter(outputs_cstr_),
                  [](const auto& x) { return x.c_str(); });
+  uint64_t hash = 0;
+  if (provider == OnnxProvider::TRT) {
+    auto sv = md.model();
+    // Filter out 16 random chars in the name of our conversions.
```
mooskagh: Maybe let's just remove that random name from the converter? Or even not fill that model at all.

Alternatively, we can populate it with some stable hash while converting. Like maybe hash of the source weights file.

Also, it seems that domain+name is something that could be use as a key rather than ONNX file hash:
https://github.com/onnx/onnx/blob/c25eebcf51b781dbfcc75a9c8bdf5dd1781367fe/onnx/onnx.proto#L471-L473

(and also it seems that `org.lczero` part should go to domain?)

borg323: I went for the hash so that we can distinguish conversions of the same weights file with different options. Moreover all tf2onnx converted models have `tf2onnx` as name and no domain...

borg323: I went with removing the random name. When we move to the new backend interface I'll change it to the network name.

mooskagh (approved):

PR #2215, opened 2025-07-22
Sycl for AMD build improvements

Menkib64:
I managed to compile Sycl for AMD. I used AMD provided ROCm packages for HIP support. I used Intel provided oneAPI packages for Sycl runtime and DPC++. Both install everything under /opt. Here are default paths to help others repeat the compile without manual paths. Path uses symlinks which point to the latest versioned path.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -677,6 +677,26 @@ if get_option('build_backends')
       add_project_arguments('-fsycl', language : 'cpp')
       add_project_link_arguments('-fsycl', language : 'cpp')
 
+      dpct_args = []
+      foreach onemkl_include : get_option('onemkl_include')
+        if run_command('scripts/checkdir.py', onemkl_include, check : false).returncode() == 0
+          includes += include_directories(onemkl_include, is_system: true)
+          dpct_args += '-I' + onemkl_include
+        endif
+      endforeach
+      # Check for oneMKL header which is used by DPC++ CT. It could also depend
+      # on oneMATH headers but L0 Sycl uses oneMKL kernels too.
+      cc.has_header('oneapi/mkl.hpp', required: true, args: dpct_args)
+
+      foreach dpct_include : get_option('dpct_include')
+        if run_command('scripts/checkdir.py', dpct_include, check : false).returncode() == 0
+          includes += include_directories(dpct_include, is_system: true)
+          dpct_args += '-I' + dpct_include
+        endif
+      endforeach
+      # Check for DPC++ Compatibility Tool header.
+      cc.has_header('dpct/dpct.hpp', required: true, args: dpct_args)
+
       files += 'src/neural/backends/sycl/layers.cc.dp.cpp'
       files += 'src/neural/backends/sycl/network_sycl.cc.dp.cpp'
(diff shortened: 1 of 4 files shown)


meson.build
```diff
@@ -670,6 +670,14 @@ if get_option('build_backends')
       add_project_arguments('-fsycl', language : 'cpp')
       add_project_link_arguments('-fsycl', language : 'cpp')
 
+      dpct_args = ['-fsycl']
+      foreach d : get_option('dpct_include')
+        includes += include_directories(d)
```
borg323: Is this really needed? It helped with a build issue I had, but that was eventually due to an older mkl library I had forgotten I had installed,

meson.build
```diff
@@ -670,6 +670,14 @@ if get_option('build_backends')
...
+        dpct_args += '-I' + d
+      endforeach
+      # dpct.hpp includes headers from other packages. Check that it compiles.
+      cc.check_header('dpct/dpct.hpp', required: true, args: dpct_args)
```
borg323: What is this supposed to do? It didn't catch the build issue I had with the old mkl.

Menkib64: It makes sure that dpct.hpp can be included. My builds first failed because I didn't have mkl library installed. oneapi dpccpp-ct package didn't have mkl dependency.

borg323: Fine, I always installed the full oneapi base package, so I never saw that.

borg323: BTW, why `check_header()` and not just `has_header()`?

meson.build
Menkib64: dpct has internal dependency to mkl. Therefore using dpct.hpp requires paths for both libraries. Maybe it could be better to check for mkl library explicitly.

meson.build
Menkib64: Because of internal dependency in the library. But I failed to post my reply to your other comment at the same time with the first reply.

I will change the detection code to explicitly check for mkl library later today.

meson.build
borg323: At least on my install the paths are already there. Anyway, this isn't a big deal but make sure you check the directory exists to keep windows builds working and add `is_system: true` to avoid some warnings.

meson.build
borg323: Please do, I'm leaning towards picking this for rc2.

meson.build
```diff
@@ -691,8 +711,20 @@ if get_option('build_backends')
...
+          if run_command('scripts/checkdir.py', dpct_include, check : false).returncode() == 0
+            includes += include_directories(hip_include, is_system: true)
+            hip_args += '-I' + hip_include
+          endif
+        endforeach
+        deps += cc.find_library('hipblas', dirs: hip_libdirs, required: true)
+        cc.has_header('hipblas/hipblas.h', required: true, args: hip_args)
+        deps += cc.find_library('amdhip64', dirs: hip_libdirs, required: true)
+        cc.has_header('hip/hip_runtime.h', required: true, args: hip_args)
+        includes += include_directories(hip_include)
```
Menkib64: I was resolving a conflict here. I noticed that I accidentally left the last line in when it should be deleted. I will update the branch asap.

PR #2222, opened 2025-07-28
Support new hipBLAS 3 API

Menkib64:
ROCm version 7 is under development. It will include hipBLAS 3. The new API uses different enums for datatype and computetype. We can update our code to use the new version while macros map the new compute type to old values for version 2 support.

This change is low priority because AMD only recently released the 6.4.2. I'm the odd one compiling it from the sources.

### DIFF
--- a/src/neural/backends/sycl/layers.cc.dp.cpp
+++ b/src/neural/backends/sycl/layers.cc.dp.cpp
@@ -52,6 +52,10 @@
 
 
 #ifdef USE_HIPBLAS
+#if hipblasVersionMajor < 3
+#define HIPBLAS_COMPUTE_16F HIPBLAS_R_16F
+#define HIPBLAS_COMPUTE_32F HIPBLAS_R_32F
+#endif
 #define transpose_type hipblasOperation_t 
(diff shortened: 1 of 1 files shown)


src/neural/backends/sycl/layers.cc.dp.cpp
```diff
@@ -52,6 +52,10 @@
 
 
 #ifdef USE_HIPBLAS
+#if hipblasVersionMajor < 3
+#define HIPBLAS_COMPUTE_16F HIPBLAS_R_16F
+#define HIPBLAS_COMPUTE_32F HIPBLAS_R_32F
+#endif
```
borg323, line 58: What about doing it like this:
```c++
#if !defined(HIPBLAS_COMPUTE_16F)
#define HIPBLAS_COMPUTE_16F HIPBLAS_R_16F
#endif
#if !defined(HIPBLAS_COMPUTE_32F)
#define HIPBLAS_COMPUTE_32F HIPBLAS_R_32F
#endif
```

Menkib64: It is an enum value in the header. I don't know if there is a better way than a version check.

PR #2228, opened 2025-07-31
Update `sycl` backend

ContradNamiseb:
- add auto threading and batching.
- add a case insensitive way to select device platform.
- add show platforms to see the supported platforms by device.
- add enhancements to show deviceinfo.

### DIFF
--- a/src/neural/backends/sycl/network_sycl.cc.dp.cpp
+++ b/src/neural/backends/sycl/network_sycl.cc.dp.cpp
@@ -204,22 +204,29 @@ class SyclNetwork : public Network {
 
     // Get all available platforms
     auto platforms = sycl::platform::get_platforms();
+    
+    if (platforms.empty()) {
+      throw Exception("No SYCL platform found.");
+    }
     showPlatformInfo(platforms);
+    
+    // A vector to store all sycl devices.
+    std::vector<sycl::device> devices;
 
     for (const auto& platform : platforms) {
        auto platform_devices = platform.get_devices();
-       devices_.insert(devices_.end(), platform_devices.begin(), platform_devices.end());
+       devices.insert(devices.end(), platform_devices.begin(), platform_devices.end());
     }
 
-    if (gpu_id_ >= (int)devices_.size() || gpu_id_ < 0)
+    if (gpu_id_ >= (int)devices.size() || gpu_id_ < 0)
       throw Exception("Invalid GPU Id: " + std::to_string(gpu_id_));
     
-    // Get the sycl device.
-    device_ = devices_[gpu_id_];
+    // Is it a cpu device?
+    is_cpu_ = devices[gpu_id_].is_cpu();
     // Get the number of compute units(execution units).
-    compute_units_ = device_.get_info<sycl::info::device::max_compute_units>();
+    compute_units_ = devices[gpu_id_].get_info<sycl::info::device::max_compute_units>();
     // Get context.
-    sycl::context context{device_};
+    sycl::context context{devices[gpu_id_]};
     auto exceptions_handler = [&] (sycl::exception_list exceptions) {
         for (std::exception_ptr const& e : exceptions) {
            try {
@@ -234,7 +241,7 @@ class SyclNetwork : public Network {
         }
     };
     
-    sycl_queue_ = new sycl::queue{context, device_, 
+    sycl_queue_ = new sycl::queue{context, devices[gpu_id_], 
               exceptions_handler, sycl::property_list{sycl::property::queue::in_order{}} };
 
     showDeviceInfo(*sycl_queue_);
@@ -929,12 +936,12 @@ class SyclNetwork : public Network {
   }
 
   // Check if device is the cpu for thread handling.
-  bool IsCpu() const override { return device_.is_cpu(); }
+  bool IsCpu() const override { return is_cpu_; }
 
   int GetThreads() const override { return 1 + multi_stream_; }
 
   int GetMiniBatchSize() const override {
-     if (device_.is_cpu()) { return 47; }
+     if (is_cpu_) return 47;
        // Simple heuristic that seems to work for a wide range of GPUs.
        return 2 * compute_units_;
     }
@@ -984,13 +991,12 @@ class SyclNetwork : public Network {
   bool multi_stream_;                     // run multiple parallel network evals
   bool allow_cache_opt_;  // try to fit residual block activations in L2 cache
 
+
   // Currently only one NN Eval can happen a time (we can fix this if needed
   // by allocating more memory).
   mutable std::mutex lock_;
   sycl::queue* sycl_queue_;
-  sycl::device device_;
-  // A vector to store all sycl devices.
-  std::vector<sycl::device> devices_;
+  bool is_cpu_;
 
 
   int numBlocks_;
@@ -1025,8 +1031,7 @@ class SyclNetwork : public Network {
   std::list<std::unique_ptr<InputsOutputs>> free_inputs_outputs_;
 
   void showDeviceInfo(const sycl::queue &mqueue) const {
-    CERR <<
-          "Device-Info...";
+    CERR << "Device-Info...";
     CERR << "Platform: " 
          << mqueue.get_device().get_platform().get_info<sycl::info::platform::name>() 
          << " selected";
@@ -1045,21 +1050,11 @@ class SyclNetwork : public Network {
     CERR << "Global memory size: " 
          << mqueue.get_device().get_info<sycl::info::device::global_mem_size>() / (1024 * 1024) 
          << " MB";         
-    CERR <<
-          "...Device-Info-End"
-            << "\n";
+    CERR << "...Device-Info-End";
     }
     
     void showPlatformInfo(const std::vector<sycl::platform>& platforms) {
-       if (platforms.empty()) {
-           CERR << "No SYCL platform found." << "\n";
-           return;
-        }
-       
-       CERR << "\n";
-       CERR <<
-          "Platform-List...";
-        
+       CERR << "Platform-List...";
        for (size_t i = 0; i < platforms.size(); ++i) {
            std::string version = platforms[i].get_info<sycl::info::platform::version>();
            
@@ -1079,9 +1074,7 @@ class SyclNetwork : public Network {
             }
         }
         
-        CERR <<
-           "...Platform-List-End"
-             << "\n";
+        CERR << "...Platform-List-End";
     }
 };
 


src/neural/backends/sycl/network_sycl.cc.dp.cpp
```diff
@@ -997,15 +1050,61 @@ class SyclNetwork : public Network {
...
+              << std::endl;
+    std::cerr << device_type << " clock frequency: " 
+              << mqueue.get_device().get_info<sycl::info::device::max_clock_frequency>() 
+              << " MHz" 
+              << std::endl;
+    std::cerr << "L2 cache capacity: " 
+              << mqueue.get_device().get_info<sycl::info::device::local_mem_size>() / (1024) 
+              << " KB" 
+              << std::endl;
+    std::cerr << "Global memory size: "
```
borg323: Why are all these `std:cerr` instead of `CERR`?

src/neural/backends/sycl/network_sycl.cc.dp.cpp
```diff
@@ -202,28 +202,55 @@ class SyclNetwork : public Network {
...
+          auto platform_devices = platform.get_devices();
+          devices.insert(devices.end(), platform_devices.begin(), platform_devices.end());
+        }
+    }
 
-    int total_gpus = dpct::dev_mgr::instance().device_count();
+    // Count the GPU's.
+    for (const auto& device : devices) {
+        if (device.is_gpu()) { total_gpus_ += 1; }
+    }
```
borg323: This will change the numbering from what is shown in `sycl-ls`. Is this something we want?

src/neural/backends/sycl/network_sycl.cc.dp.cpp
```diff
@@ -915,7 +942,30 @@ class SyclNetwork : public Network {
...
+  std::vector<sycl::device> devices;
+
+  // Check if device is the cpu for thread handling.
+  bool IsCpu() const override { return device_.is_cpu(); }
 
+  // 2 threads for cpu and 1 + total_gpu's for multi gpu configs. 
+  int GetThreads() const override { return device_.is_cpu() ? 2 : 1 + total_gpus_; }
+
+  int GetMiniBatchSize() const override {
+     if (device_.is_cpu()) { return 47; }
```
borg323: Will this ever be true now that only gpus are ennumerated? Also 47 seems on the high side for a cpu.

src/neural/backends/sycl/network_sycl.cc.dp.cpp
```diff
@@ -915,7 +942,30 @@ class SyclNetwork : public Network {
...
+       // Simple heuristic that seems to work for a wide range of GPUs.
+       return 2 * compute_units_;
```
borg323, line 946: Has this been tested?

src/neural/backends/sycl/network_sycl.cc.dp.cpp
ContradNamiseb: Oh, sorry this was from the old part of the code  before the new contribution_md, will update in a moment.

src/neural/backends/sycl/network_sycl.cc.dp.cpp
ContradNamiseb: This doesn't change the devices vector so the numbering will still just be the same. I know total_gpus_ was used differently in cuda, in here I repurposed it to count the total gpu's on the config so that the thread counting will be auto counted like `threads` = 1 + total_gpu(which can be 1, 2, 3, or 4) to reflect the gpu's on the config.

src/neural/backends/sycl/network_sycl.cc.dp.cpp
ContradNamiseb: Yes it will be true the device here is an `sycl::device device_`, as I mentioned in the other comment this does not change  the selected device with `gpu_id` so when a cpu device is selected it will be true.

src/neural/backends/sycl/network_sycl.cc.dp.cpp
ContradNamiseb: Yes though only on my igpu:

but the testing we did was more to see which approach we would take  `(1 * compute_units)`, `(2 * compute_units)` or `(3 * compute_units)`

here are results:
`(1 * compute_units)`,
===========================
Total time (ms) : 343605
Nodes searched  : 671416
Nodes/second    : 1954

`(2 * compute_units)` 
===========================
Total time (ms) : 345992
Nodes searched  : 750640
Nodes/second    : 2170

`(3 * compute_units)`
===========================
Total time (ms) : 348520
Nodes searched  : 726827
Nodes/second    : 2085

Though not that much of a difference  between the three the `2 * compute_units` was a bit better.

src/neural/backends/sycl/network_sycl.cc.dp.cpp
ContradNamiseb: Example:

```
 lc0-dag-preview benchmark --backend=sycl-auto -o gpu=2         in cmd at 09:14:42
       _
|   _ | |
|_ |_ |_| v0.33.0-dev+git.dirty built Jul 31 2025
Loading weights file from: C:\Users\PC\Documents\lc0-binaries\Sycl-test/791556.pb.gz
Trying to switch to [sycl-fp16]...
Switched to [sycl-fp16]...
Platform 0 (version: 1.6): GPU
 Device: Intel(R) oneAPI Unified Runtime over Level-Zero

Platform 1 (version: OpenCL 3.0 ): GPU
 Device: Intel(R) OpenCL Graphics

Platform 2 (version: OpenCL 3.0 WINDOWS): CPU
 Device: Intel(R) OpenCL

Platform 3 (version: OpenCL 3.0 WINDOWS): CPU
 Device: Intel(R) OpenCL

Platform 4 (version: OpenCL 3.0 WINDOWS): CPU
 Device: Intel(R) OpenCL

PLATFORM: Intel(R) OpenCL selected
CPU: 11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz
CPU: 5965 MB
CPU clock frequency: 2400 MHz
L2 cache capacity: 256 KB
Global memory size: 11930 MB
Using Fp16
```

ContradNamiseb: sycl-ls e.g:

```
sycl-ls                                                        in cmd at 09:16:24
[level_zero:gpu][level_zero:0] Intel(R) oneAPI Unified Runtime over Level-Zero, Intel(R) Iris(R) Xe Graphics 12.0.0 [1.6.33890]
[opencl:gpu][opencl:0] Intel(R) OpenCL Graphics, Intel(R) Iris(R) Xe Graphics OpenCL 3.0 NEO  [32.0.101.6913]
[opencl:cpu][opencl:1] Intel(R) OpenCL, 11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz OpenCL 3.0 (Build 0) [2024.18.10.0.08_160000]
[opencl:cpu][opencl:2] Intel(R) OpenCL, 11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz OpenCL 3.0 (Build 0) [2024.18.10.0.08_160000]
[opencl:cpu][opencl:3] Intel(R) OpenCL, 11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz OpenCL 3.0 (Build 0) [2024.18.11.0.13_160000]
```

Sorry don't mean to spam.

Menkib64:
Overall the changes look something that I have thinking would be good changes.

It is important to note that benchmarking current backend won't give reliable results. The backend breaks GPU pipeline which can affect results unpredictable ways. Still 2 * compute_units would match Cuda backend if I remember correctly. I'm working on fixing pipelining issues.

I will take a more detailed look later today.

src/neural/backends/sycl/network_sycl.cc.dp.cpp
```diff
@@ -997,15 +1050,61 @@ class SyclNetwork : public Network {
...
-    CERR << "Global memory Size: " << mqueue.get_device().get_info<sycl::info::device::global_mem_size>();
-
-  } 
+  void showDeviceInfo(const sycl::queue &mqueue) const {
+    CERR << "PLATFORM: " 
+         << mqueue.get_device().get_platform().get_info<sycl::info::platform::name>() 
+         << " selected"
+         << "\n";
+    std::string device_type = mqueue.get_device().is_gpu() ? "GPU" : "CPU";
+    CERR << mqueue.get_device().is_gpu() ? "GPU" : "CPU" << ": "
```
ContradNamiseb: Will change this `mqueue.get_device().is_gpu() ? "GPU" : "CPU"` to `device_type` later.

src/neural/backends/sycl/network_sycl.cc.dp.cpp
```diff
@@ -915,7 +942,30 @@ class SyclNetwork : public Network {
   const NetworkCapabilities& GetCapabilities() const override {
     return capabilities_;
   }
+  
+  // A vector to store all sycl devices.
+  std::vector<sycl::device> devices;
```
Menkib64: devices would need a trailing underscore to mark it as a member variable.

src/neural/backends/sycl/network_sycl.cc.dp.cpp
```diff
@@ -915,7 +942,30 @@ class SyclNetwork : public Network {
...
+
+  // Check if device is the cpu for thread handling.
+  bool IsCpu() const override { return device_.is_cpu(); }
 
+  // 2 threads for cpu and 1 + total_gpu's for multi gpu configs. 
+  int GetThreads() const override { return device_.is_cpu() ? 2 : 1 + total_gpus_; }
```
Menkib64: I think total_gpus_ should be active_gpus_ instead because queue creation code uses only one GPU currently. This can be changed later if we add multi GPU support for it.

src/neural/backends/sycl/network_sycl.cc.dp.cpp
Menkib64: How is hardcoded 47 selected for CPU device?
I suspect it should also scale based on number of CPU cores.  I have no idea what is a good scaling for CPU backend. I haven't yet even tested how it works.
It is good to remember that it be useful to use CPU also to do some evaluation. It of course depends on the system power budget and cooling. GPU is usually more power efficient for compute which could slow down total compute in a power or cooling limited system.

src/neural/backends/sycl/network_sycl.cc.dp.cpp
```diff
@@ -202,28 +202,55 @@ class SyclNetwork : public Network {
...
+    std::string search_platform = to_lower(requested_platform);
+
+    // Get all available platforms
+    auto platforms = sycl::platform::get_platforms();
+    showPlatformInfo(platforms);
+
+    for (const auto& platform : platforms) {
+       // Check if this platform matches our search criteria
+       std::string platform_name_lower = to_lower(platform.get_info<sycl::info::platform::name>());
+       if (platform_name_lower.find(search_platform) != std::string::npos) {
```
Menkib64: How about  using case-sensitive full matches only?

src/neural/backends/sycl/network_sycl.cc.dp.cpp
```diff
@@ -202,28 +202,55 @@ class SyclNetwork : public Network {
 
     max_batch_size_ = options.GetOrDefault<int>("max_batch", 1024);
 
+    std::string requested_platform = options.GetOrDefault<std::string>("platform", "OpenCL");
```
Menkib64: I think default selection should be sycl default selection. If we create sycl::queue without device, sycl will select the system default device. Default device could be from any platform.

After looking more into the code. I'm starting to think that requested_patform isn't useful for users. Only time when platform selection would matter is a shared context between multiple devices. But I don't see how we could use that for anything. We would still need to schedule work using separate queues.

src/neural/backends/sycl/network_sycl.cc.dp.cpp
```diff
@@ -202,28 +202,55 @@ class SyclNetwork : public Network {
...
+    // Get all available platforms
+    auto platforms = sycl::platform::get_platforms();
+    showPlatformInfo(platforms);
+
+    for (const auto& platform : platforms) {
+       // Check if this platform matches our search criteria
+       std::string platform_name_lower = to_lower(platform.get_info<sycl::info::platform::name>());
+       if (platform_name_lower.find(search_platform) != std::string::npos) {
+          auto platform_devices = platform.get_devices();
+          devices.insert(devices.end(), platform_devices.begin(), platform_devices.end());
```
Menkib64: This enumerates devices from one platform only. That means multi-platform systems won't find all devices. This would affect ordering for gpu id selection.

src/neural/backends/sycl/network_sycl.cc.dp.cpp
ContradNamiseb: Done.

src/neural/backends/sycl/network_sycl.cc.dp.cpp
ContradNamiseb: This can be done, but I know when I'm in a hurry to test something I dont want to be checking if my wording matches the case just want to have the correct spelling.

src/neural/backends/sycl/network_sycl.cc.dp.cpp
ContradNamiseb: When I was doing this, what I had in mind was how we run multiple gpu's in Lc0. 
IIRC we run `backend=onnx-trt,(gpu=0),(gpu=1)` with Demux or Multiplexing. So it can be `backend=sycl,(gpu=0),(gpu=1),(gpu=2)`.

Menkib64: Then there shouldn't be any filters on platform. We need to check all platforms for devices. A Frankenstein system could include Intel iGPU, NVidia and AMD GPUS. In that case Sycl has devices in opencl (GPU+CPU), cuda and hip platforms. We would need to list all of them.

src/neural/backends/sycl/network_sycl.cc.dp.cpp
borg323: Just leave it without override, one thread is all that is needed.

src/neural/backends/sycl/network_sycl.cc.dp.cpp
borg323: We just did a couple of benchmarks several months ago and 47 seems like a good choice. Certainly better than the default 256.

src/neural/backends/sycl/network_sycl.cc.dp.cpp
ContradNamiseb: Like in the above reply we can run 2 different platforms with `(opencl)` since most gpu's have opencl support.  But first we would have to adjust the build structure so we can build for multiple platforms at the same time . For Now we have  `l0, amd or nvidia` which are separate  builds meaning we do not have a unified built to support multiple platforms at once. Thus if we run multiple gpus it has to be from the same platform even if they are from different generations e.g: RTX4090 and RTX5090.

Menkib64: That is the current case. It should be relatively small change to change the backend support runtime selection using template parameters.

But even in current situation filter platforms is wrong approach. For example my system lists hip as the first platform which include a radeon device. Then it list opencl as the second platform with CPU only. If we filter for opencl platform, then CPU device will be id 0 which won't match sycl-ls. 

Sycl exposes only Intel GPUs using opencl platform.

ContradNamiseb: Ah, makes sense then did some research I didn't know amd opencl support was this poor.

ContradNamiseb:
Okay so moving forward we can remove the filter check, and with it the case insensitive code as well.  Though I'm not sure about  changing `total_gpus_` to `active_gpus_` meaning there are inactive gpus somewhere.

ContradNamiseb:
Though it my seem like a good idea to run cpu and igpu/gpu together it would cause a big slow down since. The cpu code is not optimised to even go above 100nps with the cpu for the default t79 net. But yeah I'll make the necessary changes for removing the filter & case insensitive code, and ask for a review

borg323:
> It sounds more like backend is supposed to handle a device at the time. That means it shouldn't care about total_gpus_ or active_gpus_. that would mean GetThreads should probably have the same logic as Cuda backend.

This is how we do it at the moment, each gpu gets a different backend. However, there is no reason forbidding a backend handling more devices if there is a technical advantage in doing so (but sharing weight memory seems iffy).

For now we just need a consistent way to enumerate devices, so that the numbers are the same across different instances of the same backend.

src/neural/backends/sycl/network_sycl.cc.dp.cpp
borg323: Or return `1 + multi_stream_` to match cuda. Whether multi_stream_ is working is something to test separately.

src/neural/backends/sycl/network_sycl.cc.dp.cpp
```diff
@@ -997,15 +1033,65 @@ class SyclNetwork : public Network {
...
-    CERR << "GPU: " << mqueue.get_device().get_info<sycl::info::device::name>();
-    CERR << "GPU memory: " << mqueue.get_device().get_info<sycl::info::device::max_mem_alloc_size>();
-    CERR << "GPU clock frequency: " << mqueue.get_device().get_info<sycl::info::device::max_clock_frequency>(); 
-    CERR << "L2 cache capacity: " << mqueue.get_device().get_info<sycl::info::device::local_mem_size>();
-    CERR << "Global memory Size: " << mqueue.get_device().get_info<sycl::info::device::global_mem_size>();
-
-  } 
+  void showDeviceInfo(const sycl::queue &mqueue) const {
+    CERR <<
+          "=================================Device-Info========================================";
```
borg323: Those long lines are annoying. Use something like `* Device-Info` if you want it to stand out.

src/neural/backends/sycl/network_sycl.cc.dp.cpp
```diff
@@ -916,6 +928,18 @@ class SyclNetwork : public Network {
     return capabilities_;
   }
 
+  // Check if device is the cpu for thread handling.
+  bool IsCpu() const override { return device_.is_cpu(); }
+
+  // 2 threads for cpu and 1 + total_gpu's for multi gpu configs. 
+  int GetThreads() const override { return device_.is_cpu() ? 2 : 1 + multi_stream_; }
```
borg323: This should just be 
```c++
int GetThreads() const override { return 1 + multi_stream_; }
```

src/neural/backends/sycl/network_sycl.cc.dp.cpp
```diff
@@ -963,7 +987,10 @@ class SyclNetwork : public Network {
   // Currently only one NN Eval can happen a time (we can fix this if needed
   // by allocating more memory).
   mutable std::mutex lock_;
-  sycl::queue * sycl_queue_;
+  sycl::queue* sycl_queue_;
+  sycl::device device_;
+  // A vector to store all sycl devices.
+  std::vector<sycl::device> devices_;
```
borg323: `devices_` doesn't need to be stored anymore, and neither does `device_` which was used just to call `is_cpu()`:
```diff
--- a/src/neural/backends/sycl/network_sycl.cc.dp.cpp
+++ b/src/neural/backends/sycl/network_sycl.cc.dp.cpp
@@ -206,20 +206,23 @@ class SyclNetwork : public Network {
     auto platforms = sycl::platform::get_platforms();
     showPlatformInfo(platforms);
 
+    // A vector to store all sycl devices.
+    std::vector<sycl::device> devices;
+
     for (const auto& platform : platforms) {
        auto platform_devices = platform.get_devices();
-       devices_.insert(devices_.end(), platform_devices.begin(), platform_devices.end());
+       devices.insert(devices.end(), platform_devices.begin(), platform_devices.end());
     }
 
-    if (gpu_id_ >= (int)devices_.size() || gpu_id_ < 0)
+    if (gpu_id_ >= (int)devices.size() || gpu_id_ < 0)
       throw Exception("Invalid GPU Id: " + std::to_string(gpu_id_));
-    
-    // Get the sycl device.
-    device_ = devices_[gpu_id_];
+
+    // Is it a cpu device?
+    is_cpu_ = devices[gpu_id_].is_cpu();
     // Get the number of compute units(execution units).
-    compute_units_ = device_.get_info<sycl::info::device::max_compute_units>();
+    compute_units_ = devices[gpu_id_].get_info<sycl::info::device::max_compute_units>();
     // Get context.
-    sycl::context context{device_};
+    sycl::context context{devices[gpu_id_]};
     auto exceptions_handler = [&] (sycl::exception_list exceptions) {
         for (std::exception_ptr const& e : exceptions) {
            try {
@@ -234,7 +237,7 @@ class SyclNetwork : public Network {
         }
     };
     
-    sycl_queue_ = new sycl::queue{context, device_, 
+    sycl_queue_ = new sycl::queue{context, devices[gpu_id_],
               exceptions_handler, sycl::property_list{sycl::property::queue::in_order{}} };
 
     showDeviceInfo(*sycl_queue_);
@@ -929,16 +932,16 @@ class SyclNetwork : public Network {
   }
 
   // Check if device is the cpu for thread handling.
-  bool IsCpu() const override { return device_.is_cpu(); }
+  bool IsCpu() const override { return is_cpu_; }
 
   int GetThreads() const override { return 1 + multi_stream_; }
 
   int GetMiniBatchSize() const override {
-     if (device_.is_cpu()) { return 47; }
-       // Simple heuristic that seems to work for a wide range of GPUs.
-       return 2 * compute_units_;
-    }
-  
+    if (is_cpu_) return 47;
+    // Simple heuristic that seems to work for a wide range of GPUs.
+    return 2 * compute_units_;
+  }
+
   std::unique_ptr<NetworkComputation> NewComputation() override {
     // Set correct gpu id for this computation (as it might have been called
     // from a different thread).
@@ -988,10 +991,7 @@ class SyclNetwork : public Network {
   // by allocating more memory).
   mutable std::mutex lock_;
   sycl::queue* sycl_queue_;
-  sycl::device device_;
-  // A vector to store all sycl devices.
-  std::vector<sycl::device> devices_;
-
+  bool is_cpu_;
 
   int numBlocks_;
   int numFilters_;
```

src/neural/backends/sycl/network_sycl.cc.dp.cpp
```diff
@@ -997,15 +1024,65 @@ class SyclNetwork : public Network {
...
-    CERR << "PLATFORM: " << mqueue.get_device().get_platform().get_info<sycl::info::platform::name>();
-    CERR << "GPU: " << mqueue.get_device().get_info<sycl::info::device::name>();
-    CERR << "GPU memory: " << mqueue.get_device().get_info<sycl::info::device::max_mem_alloc_size>();
-    CERR << "GPU clock frequency: " << mqueue.get_device().get_info<sycl::info::device::max_clock_frequency>(); 
-    CERR << "L2 cache capacity: " << mqueue.get_device().get_info<sycl::info::device::local_mem_size>();
-    CERR << "Global memory Size: " << mqueue.get_device().get_info<sycl::info::device::global_mem_size>();
-
-  } 
+  void showDeviceInfo(const sycl::queue &mqueue) const {
+    CERR <<
```
borg323: You can compress the following a bit:
```diff
@@ -1025,8 +1025,7 @@ class SyclNetwork : public Network {
   std::list<std::unique_ptr<InputsOutputs>> free_inputs_outputs_;
 
   void showDeviceInfo(const sycl::queue &mqueue) const {
-    CERR <<
-          "Device-Info...";
+    CERR << "Device-Info...";
     CERR << "Platform: " 
          << mqueue.get_device().get_platform().get_info<sycl::info::platform::name>() 
          << " selected";
@@ -1045,20 +1044,15 @@ class SyclNetwork : public Network {
     CERR << "Global memory size: " 
          << mqueue.get_device().get_info<sycl::info::device::global_mem_size>() / (1024 * 1024) 
          << " MB";         
-    CERR <<
-          "...Device-Info-End"
-            << "\n";
+    CERR << "...Device-Info-End\n";
     }
     
     void showPlatformInfo(const std::vector<sycl::platform>& platforms) {
        if (platforms.empty()) {
-           CERR << "No SYCL platform found." << "\n";
-           return;
-        }
-       
-       CERR << "\n";
-       CERR <<
-          "Platform-List...";
+          CERR << "No SYCL platform found.\n";
+          return;
+       }
+       CERR << "\nPlatform-List...";
         
        for (size_t i = 0; i < platforms.size(); ++i) {
            std::string version = platforms[i].get_info<sycl::info::platform::version>();
@@ -1078,10 +1072,7 @@ class SyclNetwork : public Network {
                      << device.get_platform().get_info<sycl::info::platform::name>() << ")";
             }
         }
-        
-        CERR <<
-           "...Platform-List-End"
-             << "\n";
+        CERR << "...Platform-List-End\n";
     }
 };
 
```

src/neural/backends/sycl/network_sycl.cc.dp.cpp
```diff
@@ -997,15 +1024,65 @@ class SyclNetwork : public Network {
...
+            << "\n";
+    }
+    
+    void showPlatformInfo(const std::vector<sycl::platform>& platforms) {
+       if (platforms.empty()) {
+           CERR << "No SYCL platform found." << "\n";
+           return;
+        }
+       
+       CERR << "\n";
```
borg323: This prints two empty lines, no need for that, see the diff above for alternative.

src/neural/backends/sycl/network_sycl.cc.dp.cpp
```diff
@@ -997,15 +1024,65 @@ class SyclNetwork : public Network {
...
+         << mqueue.get_device().get_info<sycl::info::device::global_mem_size>() / (1024 * 1024) 
+         << " MB";         
+    CERR <<
+          "...Device-Info-End"
+            << "\n";
+    }
+    
+    void showPlatformInfo(const std::vector<sycl::platform>& platforms) {
+       if (platforms.empty()) {
+           CERR << "No SYCL platform found." << "\n";
```
borg323: This should be an exception, but move it outside the function.

src/neural/backends/sycl/network_sycl.cc.dp.cpp
ContradNamiseb: Done.

src/neural/backends/sycl/network_sycl.cc.dp.cpp
ContradNamiseb: Cleaned up as recomended.

src/neural/backends/sycl/network_sycl.cc.dp.cpp
ContradNamiseb: done.

src/neural/backends/sycl/network_sycl.cc.dp.cpp
ContradNamiseb: Removed from the function.

PR #2242, opened 2025-08-09
search-contempt implementation for classic search

amjshl:
-search-contempt implementation on the current master branch
-added support for negative search-contempt, first suggested by Lucario6607 from discord
-added a new parameter ScTemperature, which does a softmax on the visit distribution of frozen nodes. This might mitigate some of the "blunders" observed in the odds bot, though it is recommended to leave it at the default value of 1
-no hit in nps observed with the default value of ScLimit which is 1000000000
-with lower values of ScLimit, nps actually improves

### DIFF
--- a/src/neural/encoder.cc
+++ b/src/neural/encoder.cc
@@ -134,7 +134,7 @@ int TransformForPosition(pblczero::NetworkFormat::InputFormat input_format,
 InputPlanes EncodePositionForNN(
     pblczero::NetworkFormat::InputFormat input_format,
     std::span<const Position> history, int history_planes,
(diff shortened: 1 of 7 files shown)


amjshl:
The changes in encoder.* are intentional. This is to support the SwapColors options implemented by Naphthalin. This is required to run the odds bot from the black side. Currently I have hardcoded it as false, but this can be easily fixed, which I wanted to discuss here as well. Any easy way of accessing the options_.Get<float>(kSwapColorsId) in the encoder.cc file? Something similar to "options.Get<float>(SharedBackendParams::kPolicySoftmaxTemp" in the neural/wrapper.cc file?

PR #2243, opened 2025-08-10
Add improved SYCL device FP16 capability detection and device selection logic

ContradNamiseb:


src/neural/backends/sycl/network_sycl.cc.dp.cpp
```diff
@@ -261,8 +261,7 @@ class SyclNetwork : public Network {
     //sycl_queue_->get_device().get_device_info(deviceProp);
 
 
-    if (fp16) {
-      dpct::has_capability_or_fail(sycl_queue_->get_device(), {sycl::aspect::fp16});
+    if (fp16 && sycl_queue_->get_device().has(sycl::aspect::fp16)) {
```
Menkib64: If fp16 is requested on device without support, we should throw an Exception. The network layer code won't work correctly without support.

src/neural/backends/sycl/network_sycl.cc.dp.cpp
```diff
@@ -1145,20 +1137,23 @@ std::unique_ptr<Network> MakeSyclNetworkAuto(
...
   try {
     CERR << "Trying to switch to [sycl-fp16]...";
-    dpct::has_capability_or_fail(dpct::dev_mgr::instance().get_device(gpu_id),
-                                 {sycl::aspect::fp16});
-    CERR << "Switched to [sycl-fp16]...";
-    return MakeSyclNetwork<sycl::half>(weights, options);
+    if (devices[gpu_id].has(sycl::aspect::fp16)) {
+        CERR << "Switched to [sycl-fp16]..."; 
+        return MakeSyclNetwork<sycl::half>(weights, options);     
+    }
```
Menkib64: It would be useful have else branch reporting that the device doesn't support fp16. The same message as catch for an exception prints.

ContradNamiseb: That is handled by the catch block.

Menkib64: catch block executes only when there is an exception. has_capability_or_fail throws an exception when device doesn't support requested feature. The new code doesn't throw anything which means catch block won't execute.

src/neural/backends/sycl/network_sycl.cc.dp.cpp
ContradNamiseb: If the device doesn't support fp16 an exception will be thrown when making the network , before we get here.

Menkib64: I can make has_capability_or_fail throw if I try to use an opencl backend without fp16 support:
```
$ gdb --args ./lc0 bench -b sycl-fp16 -o gpu=2
(gdb) b main
Breakpoint 1 at 0x40a0199: file ../src/main.cc, line 81.
(gdb) r
Thread 1 "lc0" hit Breakpoint 1, main (argc=6, argv=0x7fffffffdcf8) at ../src/main.cc:81
81	  LOGFILE << "Lc0 started.";
(gdb) catch throw 
Catchpoint 2 (throw)
(gdb) c
Continuing.
Device-Info...
Platform: rusticl selected
GPU: AMD Radeon Graphics (radeonsi, phoenix, LLVM 19.1.7, DRM 3.61, 6.14.0-27-generic)
GPU: 2047 MB (max allocation)
GPU clock frequency: 2799 MHz
L2 cache capacity: 64 KB
Global memory size: 512 MB
...Device-Info-End

Thread 1 "lc0" hit Catchpoint 2 (exception thrown), 0x00007fffcd6c132a in __cxa_throw () from /lib/x86_64-linux-gnu/libstdc++.so.6
(gdb) bt
#0  0x00007fffcd6c132a in __cxa_throw () from /lib/x86_64-linux-gnu/libstdc++.so.6
#1  0x0000000004253b07 in dpct::has_capability_or_fail (dev=..., props=...) at /opt/intel/oneapi/dpcpp-ct/latest/include/dpct/device.hpp:436
#2  0x00000000042567d3 in lczero::SyclNetwork<sycl::_V1::detail::half_impl::half>::SyclNetwork (this=0x1748b370, file=..., options=...)
    at ../src/neural/backends/sycl/network_sycl.cc.dp.cpp:259
```

ContradNamiseb: Can you try that with this branch I want to see what it will displays? Use: `sycl-auto`.

Menkib64:
```
$  ./lc0 bench -b sycl-auto -o gpu=2
Trying to switch to [sycl-fp16]...
Switched to [sycl]...
Platform-List...
...Device-Info-End
Using Fp32 
Unhandled exception: Device does not support USM device allocations!
```
The device is using opencl driver which implements only embedded profile. There is no USM support.

ContradNamiseb: Thanks I understand now what you mean.

ContradNamiseb: @Menkib64 this should be okay now.

Menkib64 (approved):
Looks good now.

borg323 (approved):

PR #2246, opened 2025-08-11
rescorer interface

borg323:
...and some warning and dependencies fixes.

### DIFF
--- a/src/trainingdata/rescorer.h
+++ b/src/trainingdata/rescorer.h
@@ -35,6 +35,11 @@
 namespace lczero {
 
 void RunRescorer();
+
+// Interface for external use.
+bool RescorerDeblunderSetup(float threshold, float width);
+bool RescorerGaviotaSetup(std::string dtmPaths);
+bool RescorerPolicySubstitutionSetup(std::string policySubsDir);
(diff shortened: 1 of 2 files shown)


src/trainingdata/rescorer.h
```diff
@@ -27,13 +27,17 @@
...
-#include "chess/uciloop.h"
-#include "utils/optionsparser.h"
+#include "syzygy/syzygy.h"
+#include "trainingdata/trainingdata_v6.h"
 
 namespace lczero {
 
 void RunRescorer();
+std::vector<V6TrainingData> RescoreTrainingData(
+    const std::vector<V6TrainingData>& fileContents, SyzygyTablebase* tablebase,
```
mooskagh: I believe all updates of the training data are in-place.

If so, I believe it makes sense either to pass the vector by reference and make it update in-place, or

(probably easier) pass it by value so that it's moved into in the call and then moved back on return.

That would avoid an extra copy.

borg323: Argh, I realized this right after I pushed, and then only fixed `ProcessFileInternal()`...

mooskagh (approved):

PR #2247, opened 2025-08-11
Support of default value in our protobufs

mooskagh:


mooskagh:
Once it's reviewed, I also plan to reformat the file and then push it with formatting change.
(doing it right away makes it hard to see what actually changed)

borg323 (approved):

PR #2251, opened 2025-08-15
fix for eigen dependency issue

borg323:
Also simplify eigen includes and update eigen.wrap to have a `source_fallback_url` as the default Eigen server seems to have connectivity problems once in a while.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -338,7 +338,13 @@ if get_option('build_backends')
 
     endif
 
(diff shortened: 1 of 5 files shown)


borg323:
This is to fix the issue reported in https://github.com/LeelaChessZero/lc0/discussions/2249#discussioncomment-14098172 where the dependency is found but the include path is not correct, even though the .pc file seems OK.

PR #2253, opened 2025-08-19
readme update

borg323:


README.md
```diff
@@ -42,174 +40,135 @@ Having successfully acquired Lc0 via either of these methods, proceed to the bui
...
-7. Run `./build.sh -Dgtest=false` (needs step 5)
+The Intel tools can be found in either the "oneAPI Base Toolkit" or "C++ Essentials" packages that can be downloaded from
+<https://www.intel.com/content/www/us/en/developer/tools/oneapi/base-toolkit-download.html>, while hipBLAS can be downloaded from
+<https://rocm.docs.amd.com/projects/hipBLAS/en/latest/>
 
-### Raspberry Pi
+The compiler for C code is icx and for C++ code is icx on Windows but icpx on Linux.
 
-You'll need to be running the latest Raspberry Pi OS "buster".
+To build Lc0 with SYCL you need to set the `sycl` build option using `-Dsycl=l0` (that is el zero) for an inter GPU or `-Dsycl=amd` for (you guessed it) an AMD GPU.
```
Menkib64: s/an inter GPU/an Intel GPU/

README.md
```diff
@@ -42,174 +40,135 @@ Having successfully acquired Lc0 via either of these methods, proceed to the bui
...
 ```shell
-git clone https://github.com/xianyi/OpenBLAS.git
-cd OpenBLAS/
-make
-sudo make PREFIX=/usr install
-cd ..
+. /opt/intel/oneapi/setvars.sh --include-intel-llvm
+CC=icx CXX=icpx AR=llvm-ar ./build.sh release -Dgtest=false -Dsycl=l0
 ```
+The first line is to initialize the build environment and is only needed once per session, while the build line may need modification as described above.
```
Menkib64, line 157: Should we mention that Meson reads AR environment variable every time configure is run?
This is unexpected behavior because people expect meson to remember the AR setting from the initial setup. Meson using wrong AR can produce a confusing build error. It could help people if they try to recompile same build directory after changing configuration options.
I suspect this is a bug in Meson. I haven't reported it yet because I'm not sure if it is a bug.

README.md
```diff
@@ -42,174 +40,135 @@ Having successfully acquired Lc0 via either of these methods, proceed to the bui
...
 
 #### Generic
 
-1. Install backend:
-    - If you want to use NVidia graphics cards Install [CUDA](https://developer.nvidia.com/cuda-zone) and [cuDNN](https://developer.nvidia.com/cudnn).
-    - If you want to use AMD graphics cards install OpenCL.
-    - if you want OpenBLAS version Install OpenBLAS (`libopenblas-dev`).
+1. Install backend (also read the detailed instructions in later sections):
+    - If you want to use NVidia graphics cards Install [CUDA](https://developer.nvidia.com/cuda-zone) (and optionally [cuDNN](https://developer.nvidia.com/cudnn)).
+    - If you want to use AMD or intel graphics cards you can try SYCL.
```
Menkib64: Intel would need a capital letter.

README.md
```diff
@@ -42,174 +40,135 @@ Having successfully acquired Lc0 via either of these methods, proceed to the bui
...
+<https://rocm.docs.amd.com/projects/hipBLAS/en/latest/>
 
-### Raspberry Pi
+The compiler for C code is icx and for C++ code is icx on Windows but icpx on Linux.
 
-You'll need to be running the latest Raspberry Pi OS "buster".
+To build Lc0 with SYCL you need to set the `sycl` build option using `-Dsycl=l0` (that is el zero) for an inter GPU or `-Dsycl=amd` for (you guessed it) an AMD GPU.
 
-1. Install OpenBLAS
+You may also have to set the `dpct_include` option to point to the DPC++ Compatibility Tool includes, the `onemkl_include` similarly for the oneMKL includes, or `hip_libdirs` and `hip_include` to the AMD HIP libraries and includes respectively.
```
Menkib64, line 151: I just remember an important detail. AMD support requires targeting a specific GPU. That requires using `-Damd_gfx=<gpu core version number>`  Example for NAVI 31 (product names include RX 7900) is `-Damd_gfx=1100`. ROCm package  rocminfo includes `rocm_agent_enumerator` which prints the version number in format `gfx1100`. Meson option expects the value without gfx prefix.

This makes me think that I should add automatic detection for the system GPU core version using the agent tool.

borg323: If you can make this patch today, I'm fine with adding it to the release even at this late stage. The option being without the `gfx` prefix is probably due to what we do with cuda CC, where we use the value for both `arch` and `sm`. Feel free to change it, as it isn't documented yet.

PR #2254, opened 2025-08-19
Use rocm_agent_enumerator to detect AMD GPU core version

Menkib64:


meson.build
```diff
@@ -715,11 +715,37 @@ if get_option('build_backends')
...
+                      required: false)
+          if not agent_enum.found()
+            warning( '\'rocm_agent_enumerator\' not found. AMD GPU detection doesn\'t work. You can install rocminfo or set -Damd_gfx.')
+          elif meson.version().version_compare('<1.2.0')
+            warning( 'Automatic AMD GPU detection requires Meson 1.2.0')
+          else
+            agents = run_command(agent_enum, check : false).stdout()
+            agent_list = agents.splitlines()
+            foreach agent : agent_list
+              if agent.startswith('gfx')
```
borg323, line 731: Isn't it better to get an error if a wrong prefix is used than ignoring it?

meson.build
```diff
@@ -715,11 +715,37 @@ if get_option('build_backends')
...
+        else
+          amd_gfx = ['amd_gpu_' + amd_gfx]
+        endif
+        if amd_gfx.length() == 0
+          error('-Dsycl=amd requires specifying -Damd_gfx architecture identifier (e.g. gfx90a, gfx1100 or similar)')
         endif
-        add_project_arguments('-fsycl-targets=amd_gpu_gfx'+get_option('amd_gfx'), language : 'cpp')
-        add_project_link_arguments('-fsycl-targets=amd_gpu_gfx'+get_option('amd_gfx'), language : 'cpp')
+        add_project_arguments('-fsycl-targets='+','.join(amd_gfx), language : 'cpp')
+        add_project_link_arguments('-fsycl-targets='+','.join(amd_gfx), language : 'cpp')
```
borg323: Needs spaces around the `+`.

meson.build
Menkib64: I added it to make sure that empty string won't be added if agent enumeration doesn't find and GPUs. It will produce error message after the loop. But that was for the case when using split instead of splitlines. The tool should produce only lines with gfx prefix. I guess it could also print some python error message if there is an unexpected error.

PR #2264, opened 2025-08-24
Add onnx-openvino

ContradNamiseb:
This pr adds openvino for intel igpu's, may be useful for igpu's that don't support sycl(e,g intel 10th gen and below).
This supports:
- OnnxRuntime 1.22.0 Release and 
- OpenVINO™ 2025.1.0 Release

### DIFF
--- a/src/neural/backends/network_onnx.cc
+++ b/src/neural/backends/network_onnx.cc
@@ -38,6 +38,14 @@
 #define USE_DML
 #endif
 
+#if __has_include("CL/opencl.hpp")
+#include "CL/opencl.hpp"
+#elif __has_include("OpenCL/opencl.hpp")
+#include "OpenCL/opencl.hpp"
+#else
+#include "opencl.hpp"
+#endif
+
 #include "cpu_provider_factory.h"
 #include "neural/factory.h"
 #include "neural/loader.h"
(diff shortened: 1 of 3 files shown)


src/neural/backends/network_onnx.cc
```diff
@@ -38,6 +38,14 @@
 #define USE_DML
 #endif
 
+#if __has_include("CL/opencl.hpp")
+#include "CL/opencl.hpp"
+#elif __has_include("OpenCL/opencl.hpp")
+#include "OpenCL/opencl.hpp"
+#else
+#include "opencl.hpp"
+#endif
```
borg323, line 48: This should be guarded by a `#if __has_include("CL/opencl.h")` and should define a `USE_OPENVINO` or `USE_OPECL`  symbol to exclude the code that will fail to build otherwise.

Also `OpenCL/opencl.hpp` is only for apple systems AFAIK, does openvino support them?

src/neural/backends/network_onnx.cc
```diff
@@ -376,6 +423,25 @@ Ort::SessionOptions OnnxNetwork::GetOptions(int gpu, int threads,
       api.ReleaseTensorRTProviderOptions(trt_options_v2);
       break;
     }
+    case OnnxProvider::OPENVINO: {
+      auto ocl_instance = std::make_shared<OpenCL>();
+      cl_context opencl_context = ocl_instance->context_.get();
```
borg323: Is this necessary or optional?

src/neural/backends/network_onnx.cc
```diff
@@ -376,6 +423,25 @@ Ort::SessionOptions OnnxNetwork::GetOptions(int gpu, int threads,
...
+      options.SetGraphOptimizationLevel(GraphOptimizationLevel::ORT_DISABLE_ALL);
+      std::unordered_map<std::string, std::string> openvino_options;
+      openvino_options["device_type"] = "GPU"; 
+      openvino_options["precision"] = "FP16";
+      openvino_options["num_of_threads"] = "8";
+      openvino_options["num_streams"] = "8";
+      openvino_options["cache_dir"] = "openvino_cache";
+      openvino_options["context"] = 
+           std::to_string((unsigned long long)(void*)opencl_context);
+      openvino_options["enable_qdq_optimizer"] = "False";
```
borg323: Setting this to true shouldn't matter for our graphs, but may make a difference for an imported onnx graph with quantization (unless of course more stuff is needed to make it work).

src/neural/backends/network_onnx.cc
ContradNamiseb: Yes, it should be [supported](https://docs.openvino.ai/2025/about-openvino/release-notes-openvino/system-requirements.html) by apple but that was not my intent here.

src/neural/backends/network_onnx.cc
ContradNamiseb: No. its optional. When you don't want to use cl_context that is.

borg323: Then make it conditional on `USE_OPENCL` defined earlier in the file if `__has_include("CL/opencl.h")` is true.

ContradNamiseb:
@copilot review this pull request

PR #2267, opened 2025-08-26
Optimize dag node debugging code

Menkib64:
TreeWalk is about 6 times faster than the old non-recursive version. The old version was mostly wasting time in memory management which recursion handles efficiently using stack and a member variable.

It also changes dot graph printing to write directly to a file instead of a string object. This avoids huge string memory allocations when dumping a large tree.

### DIFF
--- a/src/search/dag_classic/node.cc
+++ b/src/search/dag_classic/node.cc
@@ -32,6 +32,7 @@
 #include <cmath>
 #include <cstring>
 #include <iostream>
+#include <fstream>
 #include <list>
 #include <sstream>
 #include <thread>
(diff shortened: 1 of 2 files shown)


src/search/dag_classic/node.cc
```diff
@@ -514,83 +565,37 @@ std::string Node::DotGraphString(bool as_opponent) const {
...
-      auto child_low_node = child->GetLowNode().get();
-      if (child_low_node != nullptr &&
-          (seen.find(child_low_node) == seen.end())) {
-        seen.insert(child_low_node);
-        unvisited_fifo.push_back(std::pair(child, !parent_as_opponent));
-      }
-    }
-  }
+  TreeWalk(this, !as_opponent,
+    [&](const LowNode* low_node, bool) {
```
borg323, line 569: Is the unused second argument there for planned expansion?

Menkib64: I was first implementing `TreeWalk` when I noticed dot printing requires as_opponent. Late the old dot printing function when I noticed as_opponent is only used for edges. I didn't think about simplify the internal API.

PR #2276, opened 2025-09-01
guard against inf fp16 input in softmax kernel

borg323:
This seems to have fixed the NaN values reported with BT4. 

I'm not so sure about the clamp value, maybe something closer to the fp16 range is better.

### DIFF
--- a/src/neural/backends/cuda/common_kernels.cu
+++ b/src/neural/backends/cuda/common_kernels.cu
@@ -813,6 +813,10 @@ __device__ __forceinline__ float clamp(float val, float low, float high) {
   return fminf(fmaxf(val, low), high);
 }
 
+namespace {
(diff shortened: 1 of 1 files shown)


SunnyWar:
and to deal with both changes and be completely branchless:
#include <type_traits>

constexpr float kHalfMax    = 65504.0f;                    // largest finite half float
constexpr float kSumHalfMax = kHalfMax * 2.0f;  // max of two halfs = 131008.0
constexpr float kDefaultClamp = 1e6f;                      // your existing guard for fp32

template<typename T>
__device__ __forceinline__ float clamp_inf(float v) {
    if constexpr (std::is_same<T, half>::value) {
        return fminf(fmaxf(v, -kSumHalfMax), kSumHalfMax);
    } else {
        return fminf(fmaxf(v, -kDefaultClamp), kDefaultClamp);
    }
}

borg323:
Won't merge without more tests.

PR #2282, opened 2025-09-07
Fix param order

SunnyWar:
fix header param order to correspond with definition and usage


john-sp (approved):
> header is wrong, code is correct, all use cases use it like the code specifies. -Tilps

I am trusting Tilps, LGTM

PR #2283, opened 2025-09-09
add support for fp16 encoded weights

borg323:
Uses the `net.proto` from https://github.com/LeelaChessZero/lczero-common/pull/22

### DIFF
--- a/src/utils/weights_adapter.cc
+++ b/src/utils/weights_adapter.cc
@@ -25,19 +25,47 @@
   Program grant you additional permission to convey the resulting work.
 */
 
-#include "src/utils/weights_adapter.h"
+#include "utils/weights_adapter.h"
+
+#include "utils/bf16_utils.h"
+#include "utils/exception.h"
+#include "utils/fp16_utils.h"
 
 namespace lczero {
+
 float LayerAdapter::Iterator::ExtractValue(const uint16_t* ptr,
                                            const LayerAdapter* adapter) {
-  return *ptr / static_cast<float>(0xffff) * adapter->range_ + adapter->min_;
+  switch (adapter->encoding_) {
+    case pblczero::Weights::Layer::LINEAR16:
+      return *ptr / static_cast<float>(0xffff) * adapter->range_ +
+             adapter->min_;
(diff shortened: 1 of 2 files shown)


src/utils/weights_adapter.cc
```diff
@@ -25,19 +25,47 @@
...
+
 float LayerAdapter::Iterator::ExtractValue(const uint16_t* ptr,
                                            const LayerAdapter* adapter) {
-  return *ptr / static_cast<float>(0xffff) * adapter->range_ + adapter->min_;
+  switch (adapter->encoding_) {
+    case pblczero::Weights::Layer::LINEAR16:
+      return *ptr / static_cast<float>(0xffff) * adapter->range_ +
+             adapter->min_;
+    case pblczero::Weights::Layer::FLOAT16:
+      return FP16toFP32(*ptr) * adapter->max_;
```
mooskagh: Why do we multiply by `max_`?
If we'd like to scale (which I'm not sure we do), I believe a separate field with more correct name (factor? scale?) would be better.

borg323: Seemed like a good idea to increase the range, but thinking about it, it won't work when running with fp16 so indeed there is no point.

src/utils/weights_adapter.cc
```diff
@@ -25,19 +25,56 @@
...
+      return adapter->min_ * (1 - theta) + adapter->max_ * theta;
+    }
+    case pblczero::Weights::Layer::FLOAT16:
+      return FP16toFP32(*ptr);
+    case pblczero::Weights::Layer::BFLOAT16:
+      return BF16toFP32(*ptr);
+    [[unlikely]] default:  // To silence a couple of warnings.
+#if defined(ABSL_UNREACHABLE)
+      ABSL_UNREACHABLE();
+#elif defined(__GNUC__)
```
mooskagh, line 52: Do we need all that given that we #include <absl/base/optimization.h>?

borg323: Unfortunately my Ubuntu 24.04 has an abseil version from 2022 while `ABSL_UNREACHABLE` was introduced in 2023.

PR #2285, opened 2025-09-11
warning fixes

borg323:


src/neural/backends/cuda/network_cuda.cc
```diff
@@ -1015,7 +1015,8 @@ class CudaNetwork : public Network {
     }
   }
 
-  void showDeviceInfo(const cudaDeviceProp& deviceProp, int deviceId) const {
+  void showDeviceInfo(const cudaDeviceProp& deviceProp,
+                      [[maybe_unused]] int deviceId) const {
```
mooskagh, line 1019: I find the "old way" of commenting out the parameter name (i.e. `void showDeviceInfo(const cudaDeviceProp& deviceProp, int /*deviceId*/)`) a bit cleaner (less characters, not possible to forget to undo the change when the argument becomes used).

But no strong preference, maybe I'm not accustomed to the C++ attributes yet.

borg323: I agree, but in this case the parameter is needed for cuda 13 (there is an `#if` a bit further down).

PR #2288, opened 2025-09-14
Refactor dag time keeping

Menkib64:
We overestimate NPS when clock starts after the first batch. We are counting nodes from the first batch which were calculated before the timing starts.

Changes move time keeping start before we evaluate the first NN batch. It still overestimates NPS a little. Any earlier logical point seem to underestimate more. GatherMinibatch typically operates at the same time as previous batch is evaluate. It is extra latency but shouldn't affect throughput.

Locking change from counter_mutex_ to nodes_mutex_ avoids taking counter_mutex_ in performance critical path. There is also plan to reduce nodes_mutex_ use. Cached time difference is used instead of recalculating the same value multiple times.

### DIFF
--- a/src/search/dag_classic/search.cc
+++ b/src/search/dag_classic/search.cc
@@ -27,6 +27,8 @@
 
 #include "search/dag_classic/search.h"
 
+#include <absl/cleanup/cleanup.h>
+
 #include <algorithm>
 #include <array>
 #include <chrono>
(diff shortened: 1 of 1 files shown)


src/search/dag_classic/search.cc
```diff
@@ -1376,6 +1368,16 @@ void SearchWorker::GatherMinibatch() {
 
   int thread_count = search_->thread_count_.load(std::memory_order_acquire);
 
+  struct RecordBatchStartTime {
```
borg323: Would `absl::Cleanup()` be better for this?

PR #2291, opened 2025-09-18
fixed size transposition table

borg323:
Can be enabled by the `FIX_TT` build option (on by default).
Introduces a new `Hash` uci option to control the size of the transposition table in MB (default 50).

### DIFF
--- a/src/search/dag_classic/search.cc
+++ b/src/search/dag_classic/search.cc
@@ -2245,32 +2245,30 @@ void SearchWorker::DoBackupUpdateSingleNode(
         node_to_process.hash, node_to_process.tt_low_node);
 #endif
     if (is_tt_miss) {
-#ifndef FIX_TT
-      assert(!tt_iter->second.expired());
-#endif
-      node_to_process.node->SetLowNode(node_to_process.tt_low_node);
 #ifdef FIX_TT
       search_->tt_->Insert(node_to_process.hash,
                            std::make_unique<std::weak_ptr<LowNode>>(
                                node_to_process.tt_low_node));
-#endif
-    } else {
-#ifndef FIX_TT
-      auto tt_low_node = tt_iter->second.lock();
 #else
+      assert(!tt_iter->second.expired());
+#endif
+      node_to_process.node->SetLowNode(node_to_process.tt_low_node);
+    } else {
+#ifdef FIX_TT
       auto tt_low_node = entry->lock();
       search_->tt_->Unpin(node_to_process.hash, entry);
+#else
+      auto tt_low_node = tt_iter->second.lock();
 #endif
       if (!tt_low_node) {
-#ifndef FIX_TT
-        tt_iter->second = node_to_process.tt_low_node;
-#endif
-        node_to_process.node->SetLowNode(node_to_process.tt_low_node);
 #ifdef FIX_TT
         search_->tt_->Insert(node_to_process.hash,
                              std::make_unique<std::weak_ptr<LowNode>>(
                                  node_to_process.tt_low_node));
+#else
+        tt_iter->second = node_to_process.tt_low_node;
 #endif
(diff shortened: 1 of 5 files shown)


src/search/dag_classic/wrapper.cc
```diff
@@ -126,7 +142,10 @@ void DagClassicSearch::StartSearch(const GoParams& params) {
       options_->Get<int>(SharedBackendParams::kNNCacheSizeId);
   // FIXME: This is too conservative.
   const size_t kAvgNodeSize =
-      sizeof(Node) + sizeof(LowNode) + sizeof(TranspositionTable::slot_type) +
+      sizeof(Node) + sizeof(LowNode) +
+#ifndef FIX_TT
+      sizeof(TranspositionTable::slot_type) +
+#endif
```
Menkib64: Fixed size hash should be included in `total_memory`.

src/search/dag_classic/search.cc
```diff
@@ -2072,13 +2074,23 @@ void SearchWorker::ExtendNode(NodeToProcess& picked_node) {
...
   // NN evaluation.
   picked_node.hash = history.HashLast(params_.GetCacheHistoryLength() + 1);
+#ifndef FIX_TT
   auto tt_iter = search_->tt_->find(picked_node.hash);
   // Transposition table entry might be expired.
   if (tt_iter != search_->tt_->end()) {
     picked_node.tt_low_node = tt_iter->second.lock();
   }
+#else
+  auto entry = search_->tt_->LookupAndPin(picked_node.hash);
```
Menkib64, line 2105: We could insert nodes here if insertions avoid random slowdowns of the dynamic version. For the testing purposes it is better to delay insert. That makes it easier to compare how the fixed implementation compares to the dynamic implementation.

src/search/dag_classic/search.cc
```diff
@@ -2225,18 +2237,42 @@ void SearchWorker::DoBackupUpdateSingleNode(
...
+    bool is_tt_miss = !entry;
+#else
     auto [tt_iter, is_tt_miss] = search_->tt_->try_emplace(
         node_to_process.hash, node_to_process.tt_low_node);
+#endif
     if (is_tt_miss) {
+#ifdef FIX_TT
+      search_->tt_->Insert(node_to_process.hash,
+                           std::make_unique<std::weak_ptr<LowNode>>(
+                               node_to_process.tt_low_node));
```
Menkib64: I think this needs a comment that insertions aren't thread safe if a hash is the same. Current implementation silently ignores insertion if hash is already present in the table.

I was thinking about last inserted eviction policy. It isn't perfect match how search access nodes. We can end up evicting nodes which are old insertion which are still close to the PV line. A better eviction policy would be least recently accessed but that would be expensive to implement without major refactoring.
We could reduce risk of eviction in insertion if garbage collector evicts expired weak pointers. I don't think initial version needs this. It can be an useful change when playing matches against strong engines where both agree on a long line.

The table implementation requires unique_ptr storage. Could we modify storage to support weak_ptr too? I think this is also a change that shouldn't be part of initial version. It could be a future improvement. It could be also redundant work if we choose to give ownership to the hash table to allow cycles in the tree.

borg323: I don't see the thread safety issue, can you elaborate? The insertion failing shouldn't cause any correctness issues.

Menkib64: It would mean transposition might have different low node. Search would work but transpositions wouldn't share information which would affect evaluation. It should be simple change to make Insert return the existing value or an inserted value. The caller can then use the returned value which would be known to be shared between all transpositions.
GatherMinibatch found the same transposition from different paths fairly frequently when I was experiment putting TT Insertion into GatherMinibatch.

borg323:
I'm having crashes with the last commit when the TT is small, (e.g. Hash=1). Converted to draft while I investigate.

PR #2292, opened 2025-09-21
Fix piece moves marked en passant

Menkib64:
Fix a bug where a piece move starting from opponent's backrank was marked as en passant. This could eliminate one of opponents pieces from the board state.

```
position fen r1bqk1nr/pp1nbpp1/2pp3p/3Pp3/2P1P3/2N1B3/PP2NPPP/R2QKB1R w KQkq - 0 1 moves e2g3 g7g6 f1e2 c6c5 d1d2 e7g5 g3f1 d8e7 f2f3 g5e3 f1e3 d7b6 h2h4 g8f6 g2g4 c8d7 b2b3 h6h5 g4g5 f6h7 a2a4 f7f6 g5f6 e7f6 e1c1 b6c8 d1f1 c8e7 e3g2 e8c8 f3f4 h8f8 c1c2 a7a6 a4a5 d8e8 g2e1 e5f4 e1d3 f6g7 f1f4 h7f6 b3b4 c5b4 d3b4 f6g4 h1f1 f8f4 d2f4 g7d4 e2g4 h5g4 c2b3 c8c7 f1d1 d4c5 e4e5 d6e5 f4e5 c7d8 e5b8 e7c8 d5d6 b7b5 b8c7
go nodes 2
bestmove a1a1
```

### DIFF
--- a/src/chess/board.cc
+++ b/src/chess/board.cc
@@ -598,6 +598,7 @@ bool ChessBoard::ApplyMove(Move move) {
   struct BoardValidate {
     const ChessBoard& board_;
     const Move move_;
+    BoardValidate(const ChessBoard& b, Move m) : board_(b), move_(m) {}
     ~BoardValidate() {
       assert(board_.IsValid(move_) && "after ChessBoard::ApplyMove");
     }
--- a/src/chess/board.cc
+++ b/src/chess/board.cc
@@ -34,6 +34,7 @@
 #include <cstring>
 #include <sstream>
 #include <utility>
+#include <absl/cleanup/cleanup.h>
 
 #include "utils/exception.h"
 
@@ -587,7 +588,11 @@ bool ChessBoard::IsValid(Move move) const {
       (rooks() & queens()).as_int() ||
       (rooks() & kings()).as_int() ||
       (queens() & kings()).as_int()) {
-    CERR << DebugString() << " " << move.ToString(false);
+    std::string move_string = "";
+    if (move.raw_data()) {
+      move_string = " move " + move.ToString(false);
+    }
+    CERR << DebugString() << move_string;
     return false;
   }
   return true;
@@ -595,14 +600,9 @@ bool ChessBoard::IsValid(Move move) const {
 
 bool ChessBoard::ApplyMove(Move move) {
   assert(our_pieces_.intersects(BitBoard::FromSquare(move.from())));
-  struct BoardValidate {
-    const ChessBoard& board_;
-    const Move move_;
-    BoardValidate(const ChessBoard& b, Move m) : board_(b), move_(m) {}
-    ~BoardValidate() {
-      assert(board_.IsValid(move_) && "after ChessBoard::ApplyMove");
-    }
-  } validator(*this, move);
+  absl::Cleanup validate = [&] {assert(IsValid(move) &&
+                                       "after ChessBoard::ApplyMove");};
+
   const Square& from = move.from();
   const Square& to = move.to();
   const Rank from_rank = from.rank();
--- a/src/chess/board.cc
+++ b/src/chess/board.cc
@@ -573,8 +573,35 @@ MoveList ChessBoard::GeneratePseudolegalMoves() const {
   return result;
 }  // namespace lczero
 
+bool ChessBoard::IsValid(Move move) const {
+  const auto all = ours() | theirs();
+  auto check = all | pawns() | bishops() | rooks() | queens() | kings();
+  if (check != all ||
+      (pawns() & bishops()).as_int() ||
+      (pawns() & rooks()).as_int() ||
+      (pawns() & queens()).as_int() ||
+      (pawns() & kings()).as_int() ||
+      (bishops() & rooks()).as_int() ||
+      (bishops() & queens()).as_int() ||
+      (bishops() & kings()).as_int() ||
+      (rooks() & queens()).as_int() ||
+      (rooks() & kings()).as_int() ||
+      (queens() & kings()).as_int()) {
+    CERR << DebugString() << " " << move.ToString(false);
+    return false;
+  }
+  return true;
+}
+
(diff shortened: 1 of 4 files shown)


borg323 (approved):

src/chess/board.cc
```diff
@@ -573,8 +573,36 @@ MoveList ChessBoard::GeneratePseudolegalMoves() const {
...
 bool ChessBoard::ApplyMove(Move move) {
   assert(our_pieces_.intersects(BitBoard::FromSquare(move.from())));
+  struct BoardValidate {
+    const ChessBoard& board_;
+    const Move move_;
+    BoardValidate(const ChessBoard& b, Move m) : board_(b), move_(m) {}
+    ~BoardValidate() {
+      assert(board_.IsValid(move_) && "after ChessBoard::ApplyMove");
+    }
+  } validator(*this, move);
```
mooskagh: There is `absl::Cleanup` which is arguably less boilerplate, but I'm not sure about it being optimized out in release build:
```
absl::Cleanup validate = [&] { assert(IsValid(move)); };
```

But also I don't understand the meaning of the check?.. I read it as it checks that the move that we just applied is valid?..

Menkib64: It validates CheckBoard state is consistent. This bug removed the queen from their_pieces_ which can be detected because there was a rook+bishop without matching side bit. The parameter is just used to give content for the debug output.

I will check if abls::Cleanup optimises as expected. If it is eliminate, I will use it.

mooskagh: Ah it just uses move for printing. I assumed that it checks that the move is valid looking at the signature.
I did read the function, but a few lines below I forgot it.

It feels wrong having to provide a move to the `IsValid()` function as it may be useful in situations where we haven't just applied a move. I wonder whether there are other ways to print the move and still be a noop in release build (to me, both not printing move, and even using explicit #ifndef NDEBUG sound better).

Menkib64: How about making it default to Move() and skipping printing it if no move passed?

mooskagh: I'm risking to go into overengineering area, but maybe add `utils/debug_cleanup.h` with something like this:

```
#ifdef NDEBUG
class [[nodiscard]] DebugCleanup final {
 public:
  template<typename T>
  Cleanup(T&&) = default;
  void Cancel() && {}
  void Invoke() && {}
};
#else
using DebugCleanup = absl::Cleanup;
#endif
```

Then there's no need to hope for compiler optimization or hide everything in assert. We can do anything, like

```
DebugCleanup validate = [&] { if (!IsValid()) {
   CERR << "Bad board state " << DebugString() << " after move " << move.ToString(false);
   abort();  // Exception unfortunately won't work here as we are in destructor.
});
```

Let's see what Copilot suggests!

Menkib64: I tested gcc and clang. absl version optimised out as I would expect. I think we would need to go about 10 years back to find a compiler which doesn't optimise it correctly.

borg323: I think this is cleaner:
```diff
--- a/src/chess/board.cc
+++ b/src/chess/board.cc
@@ -574,7 +574,7 @@ MoveList ChessBoard::GeneratePseudolegalMoves() const {
   return result;
 }  // namespace lczero
 
-bool ChessBoard::IsValid(Move move) const {
+bool ChessBoard::IsValid() const {
   const auto all = ours() | theirs();
   auto check = all | pawns() | bishops() | rooks() | queens() | kings();
   if (check != all ||
@@ -588,11 +588,6 @@ bool ChessBoard::IsValid(Move move) const {
       (rooks() & queens()).as_int() ||
       (rooks() & kings()).as_int() ||
       (queens() & kings()).as_int()) {
-    std::string move_string = "";
-    if (move.raw_data()) {
-      move_string = " move " + move.ToString(false);
-    }
-    CERR << DebugString() << move_string;
     return false;
   }
   return true;
@@ -600,9 +595,15 @@ bool ChessBoard::IsValid(Move move) const {
 
 bool ChessBoard::ApplyMove(Move move) {
   assert(our_pieces_.intersects(BitBoard::FromSquare(move.from())));
-  absl::Cleanup validate = [&] {assert(IsValid(move) &&
-                                       "after ChessBoard::ApplyMove");};
-
+#ifndef NDEBUG
+  absl::Cleanup validate = [&] {
+    if (!IsValid()) {
+      CERR << "Move " + move.ToString(true) +
+                  " resulted in invalid board: " + DebugString();
+      assert(false);
+    }
+  };
+#endif
   const Square& from = move.from();
   const Square& to = move.to();
   const Rank from_rank = from.rank();
@@ -1141,7 +1142,9 @@ bool ChessBoard::HasMatingMaterial() const {
 }
 
 std::string ChessBoard::DebugString() const {
-  return "https://lc0.org/fen/" + BoardToFen(*this);
+  auto fen = BoardToFen(*this);
+  std::replace(fen.begin(), fen.end(), ' ', '_');
+  return "https://lc0.org/fen/" + fen;
 }
 
 Move ChessBoard::ParseMove(std::string_view move_str) const {
diff --git a/src/chess/board.h b/src/chess/board.h
index bef0787f..d455fcb6 100644
--- a/src/chess/board.h
+++ b/src/chess/board.h
@@ -231,9 +231,8 @@ class ChessBoard {
  private:
   // Sets the piece on the square.
   void PutPiece(Square square, PieceType piece, bool is_theirs);
-  // Check internal state is consistent after state transformations. Move can be
-  // passed to provide context when transformation is related to making a move.
-  bool IsValid(Move move = Move()) const;
+  // Check internal state is consistent after state transformations.
+  bool IsValid() const;
 
   // All white pieces.
   BitBoard our_pieces_;
```
Also changes `move.ToString(false)` to `true` to avoid issues with FRC castling and fixes an oversight in `DebugString()` (last time we fixed only the `Position` one and forgot this).

borg323 (approved):

PR #2296, opened 2025-09-25
Remove Sycl dpct dependency

Menkib64:
Nightly build of DPC++ renames some experimental functions. These functions are used by dpct tool. This forced me to finally remove the dpct dependency from sycl.

I'm not 100% sure if my sub group communication and atomic changes are correct. Sycl documentation makes me thing this should be correct. I may have missed some key details because I was reading it for the first time.

### DIFF
--- a/src/neural/backends/sycl/winograd_helper.h
+++ b/src/neural/backends/sycl/winograd_helper.h
@@ -20,12 +20,12 @@
 */
 
 #include <sycl/sycl.hpp>
-#include "dpct/dpct.hpp"
 
 namespace lczero {
 namespace sycldnn_backend {
 
-__dpct_inline__ float mishActivate(float el) {
+[[gnu::always_inline]]
+inline float mishActivate(float el) {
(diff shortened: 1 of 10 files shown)


src/neural/backends/sycl/winograd_helper.h
```diff
@@ -69,7 +70,8 @@ __dpct_inline__ float activate(float cVal, ActivationFunction activation) {
 }
 
 template <typename T, int M, int N, int K>
-__dpct_inline__ void matrixMul_gpu_serial(T* c, const T* a, const T* b) {
+[[gnu::always_inline]]
+inline void matrixMul_gpu_serial(T* c, const T* a, const T* b) {
 #ifndef SKIP_FP16_BITS
```
borg323: While you are at it, can you also remove those `SKIP_FP16_BITS` tests? They are leftovers from cuda code that needs to support compilation for CC < 6 which doesn't support fp16.

PR #2303, opened 2025-10-04
Add Dockerfile for lc0

john-sp:
Simple Dockerfile. Does not support downloading networks, nor is any network provided.

Example usage
```bash
docker build -t lc0 .

# Interactive mode:
docker run -gpus all -it -v /local/weights:/chess/weights lc0 --weights=/chess/weights/BT4-1740.pb.gz

## Training mode:
docker run --gpus all d lc0 -e TRAINING=true -e TRAINING_USER=foo -e TRAINING_PASSWORD=bar
```

Deepseek aided in the creation of this file, I have verified that everything is working.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,74 @@
+FROM nvidia/cuda:12.9.1-cudnn-devel-ubuntu24.04
+
+ENV TRAINING=false
+ENV TRAINING_USER=foo
+ENV TRAINING_PASSWORD=bar
+
+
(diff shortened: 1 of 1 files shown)


john-sp:
Docker vs native give different outputs for best move. Draft until I figure out why

john-sp:
I tested more, 744706 seems odd, but BT4, T1 and 768x15x24h-t82-swa-7464000 are identical docker and native. As does. I am confident that it is fine

PR #2307, opened 2025-10-05
Onnx memory management and onnx-trt optimisation

Menkib64:


src/neural/backends/network_onnx.cc
```diff
@@ -483,21 +823,28 @@ OnnxNetwork::OnnxNetwork(const WeightsFile& file, const OptionsDict& opts,
...
-                 [](const auto& x) { return x.c_str(); });
   uint64_t hash = 0;
   if (provider == OnnxProvider::TRT) {
     hash = std::hash<std::string_view>()(md.model());
   }
+  switch (provider) {
+    case OnnxProvider::TRT:
+    case OnnxProvider::CUDA:
+#if USE_ONNX_CUDART
+      ReportCUDAErrors(cudaSetDevice(gpu_));
```
borg323, line 856: Apparently this is per thread, the cuda backend also sets the gpu number in `NewComputation()`.

Menkib64: Yes. cuda uses some thread local state. It is a mix of old fashioned graphics apis where driver stored all state in thread local structures. Most of cuda uses a newer better performing idea where used passes all objects pointers to functions. It doesn't user to manage context object for all calls.

borg323: With this patch it runs on all gpus:
```diff
@@ -156,6 +156,11 @@ class OnnxNetwork final : public Network {
               OnnxProvider provider, bool cpu_wdl);
   ~OnnxNetwork();
   std::unique_ptr<NetworkComputation> NewComputation() override {
+#ifdef USE_ONNX_CUDART
+    if (provider_ == OnnxProvider::TRT || provider_ == OnnxProvider::CUDA) {
+      ReportCUDAErrors(cudaSetDevice(gpu_));
+    }
+#endif
     if (fp16_) {
       return std::make_unique<OnnxComputation<Ort::Float16_t>>(this);
     } else if (bf16_) {
```

Menkib64: I noticed the same issue when I read your comment.

src/neural/backends/network_onnx.cc
```diff
@@ -287,20 +507,115 @@ void OnnxComputation<DataType>::ComputeBlocking() {
...
+      ReportCUDAErrors(cudaEventRecord(inputs_outputs_->inputs_processed_event_,
+                                       network_->upload_stream_));
+      if (i == 0) {
+        ReportCUDAErrors(
+            cudaStreamWaitEvent(network_->compute_stream_,
+                                inputs_outputs_->outputs_download_event_));
+      }
+      options.AddConfigEntry("disable_synchronize_execution_providers", "1");
+    } else {
+      binding.SynchronizeInputs();
```
borg323, line 568: Shouldn't this be outside the `#if` block? Something like
```diff
@@ -560,10 +565,12 @@ void OnnxComputation<DataType>::ComputeBlocking() {
                                 inputs_outputs_->outputs_download_event_));
       }
       options.AddConfigEntry("disable_synchronize_execution_providers", "1");
-    } else {
+    } else
+#endif
+    {
       binding.SynchronizeInputs();
     }
-#endif
+
     network_->session_[step - 1].Run(options, binding);
 #ifdef USE_ONNX_CUDART
     if (network_->provider_ == OnnxProvider::TRT ||
```
Same for `SynchronizeOutputs()` a bit further down.

borg323:
While `network_onnx.cc` and the related cuda files should eventually move to `src/network/backends/onnx`, I prefer to leave that for a followup to avoid complicating ongoing development.

src/neural/backends/cuda/onnx_kernels.h
```diff
@@ -0,0 +1,40 @@
...
+*/
+
+#pragma once
+
+#include "cuda_common.h"
+
+namespace lczero {
+namespace cudnn_backend {
+
+template <typename DataType>
```
mooskagh, line 37: It makes sense to add a short comment what this kernel does.

src/neural/backends/network_onnx.cc
mooskagh, line 1: It would be nice to somehow have the CUDA part isolated from the pure ONNX, but I don't immediately see a good way to do it. I'll give it a thought.

Menkib64: #2328 tries to separate EP specific code using a provider template type. It could be probably improved.

src/neural/backends/network_onnx.cc
```diff
@@ -133,46 +216,178 @@ class OnnxNetwork : public Network {
...
+  }
+#endif
   raw_input_.emplace_back(input);
   if (raw_input_.size() > network_->max_batch_size_) {
     throw Exception("NN input exceeds max batch size of " +
                     std::to_string(network_->max_batch_size_) + ".");
   }
 }
+template <typename DataType>
+int OnnxComputation<DataType>::GetBatchSize() const {
```
borg323, line 381: `GetBatchSize()` is a remnant of the old API that isn't currently used (except in the python bindings that need updating anyway). I think it would be a lot cleaner to use `input_size_` everywhere, and it will make it easy to get rid of `GetBatchSize()` in the near future.

borg323 (approved):

PR #2308, opened 2025-10-05
Clamp negative orig_m

mooskagh:


borg323 (approved):

Tilps:
I think this is the wrong approach, we should just drop weird data.  If we want to fix something it should be on the generating side.

borg323:
The backend bug (mish instead of relu as final activation) is fixed, but is the `orig_m` value this critical?

PR #2319, opened 2025-10-16
onnx-coreml backend

borg323:
Currently the onnxruntime coreml provider doesn't support everything required, the following three patches are needed for both fp32 and fp16 with fixed batch size (default for now).
~~https://github.com/microsoft/onnxruntime/pull/26443~~ (merged)
~~https://github.com/microsoft/onnxruntime/pull/26442~~ (merged)
~~https://github.com/microsoft/onnxruntime/pull/26462~~ (merged)

For variable batch size, hopefully the fix for issue https://github.com/microsoft/onnxruntime/issues/26328 is simple.

If someone wants to try it out, the default onnxruntime branch should work. ~~The last outstanding patch is for `Gather` fp16 support, which is the last kernel before the policy output, so doing it on the cpu shouldn't cause a huge performance drop.~~

### DIFF
--- a/src/neural/backends/onnx/network_onnx.cc
+++ b/src/neural/backends/onnx/network_onnx.cc
@@ -928,8 +928,10 @@ std::unique_ptr<Network> MakeOnnxNetwork(const std::optional<WeightsFile>& w,
     } else {
       bool fp16 = kProvider == OnnxProvider::CPU ? false : true;
 #if ORT_API_VERSION < 24
-      CERR << "WARNING: CoreML with onnxruntime before v1.24 is very slow.";
-      if (kProvider == OnnxProvider::COREML) fp16 = false;
+      if (kProvider == OnnxProvider::COREML) {
+        CERR << "WARNING: CoreML with onnxruntime before v1.24 is very slow.";
+        fp16 = false;
+      }
 #endif
       fp16 = opts.GetOrDefault<bool>("fp16", fp16);
       datatype = fp16 ? "f16" : "f32";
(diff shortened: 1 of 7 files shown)


borg323:
Some preliminary tests using `lc0 bench` with 791556 on a Apple M3 Pro.

fp32:
```
Total time (ms) : 5217
Nodes searched  : 13762
Nodes/second    : 2637
```
fp16
```
Total time (ms) : 5203
Nodes searched  : 20807
Nodes/second    : 3998
```
fp16 with [PR26442](https://github.com/microsoft/onnxruntime/pull/26442) applied:
```
Total time (ms) : 5179
Nodes searched  : 26833
Nodes/second    : 5180
```

src/neural/backends/onnx/network_onnx.cc
```diff
@@ -648,6 +648,14 @@ Ort::SessionOptions OnnxNetwork::GetOptions(int threads, int batch_size,
       options.AppendExecutionProvider("DML", dml_options);
       break;
     }
+    case OnnxProvider::COREML: {
+      std::unordered_map<std::string, std::string> provider_options;
+      provider_options["ModelFormat"] = "MLProgram";
+      provider_options["ProfileComputePlan"] = "1";
```
ChinChangYang, line 654:
```suggestion
provider_options["MLComputeUnits"] = compute_units_;
```

Onnx runtime fails optimizing performance by this. It would be better to let lc0 to benchmark `MLComputeUnits`. Then, add `compute_units_` to the member field like this: https://github.com/borg323/lc0/pull/10/changes#diff-e18d2fce5100e036be2f7e918b15b27c2453f75cbc8d4143b9ad86bcd03317f7R212

mooskagh (approved):

PR #2327, opened 2025-10-27
Make StrictTiming default and allow check for isready

Menkib64:
This makes `StrictTiming` default. It sounds like it is an expected behavior for users. At the same time I decided to add a TCEC specific mode which uses `StrictTiming` if we see `isready` command.

### DIFF
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -67,7 +67,7 @@ const OptionId kPreload{"preload", "",
 void Engine::PopulateOptions(OptionsParser* options) {
   options->Add<BoolOption>(kPonderId) = false;
   options->Add<StringOption>(kSyzygyTablebaseId);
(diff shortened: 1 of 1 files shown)


src/engine.cc
```diff
@@ -67,7 +67,7 @@ const OptionId kPreload{"preload", "",
 void Engine::PopulateOptions(OptionsParser* options) {
   options->Add<BoolOption>(kPonderId) = false;
   options->Add<StringOption>(kSyzygyTablebaseId);
-  options->Add<BoolOption>(kStrictUciTiming) = false;
+  options->Add<BoolOption>(kStrictUciTiming) = true;
```
borg323: Keep the default as `false` for maximum compatibility.

PR #2329, opened 2025-10-28
Support network evaluations per second in uci info

Menkib64:
This is based on @dje-dev 's proposal that we should report network evaluation numbers.

Example using `classic` search
```setoption name UCI_ShowEPS value true 
go
Loading weights file from: ./t3-512x15x16h-distill-swa-2767500.pb.gz
Switching to [cuda-fp16]...
info depth 1 seldepth 2 time 1305 nodes 2 score cp 8 nps 1000 eps 1000 tbhits 0 pv d2d4 d7d5
info depth 2 seldepth 3 time 1307 nodes 3 score cp 7 nps 750 eps 750 tbhits 0 pv d2d4 d7d5 c2c4
info depth 2 seldepth 4 time 1313 nodes 16 score cp 9 nps 1600 eps 600 tbhits 0 pv d2d4 d7d5 c2c4 c7c6
info depth 3 seldepth 5 time 1316 nodes 26 score cp 9 nps 2000 eps 1230 tbhits 0 pv d2d4 d7d5 c2c4 c7c6 c4d5

info depth 11 seldepth 46 time 12224 nodes 436658 score cp 12 nps 39983 eps 15751 tbhits 0 pv d2d4 d7d5 c2c4 c7c6 c4d5 c6d5 c1f4 g8f6 e2e3 b8c6 f1b5 c8g4 g1f3 a8c8 e1g1 g4f3 d1f3 e7e6 f1c1```

### DIFF
--- a/src/chess/callbacks.h
+++ b/src/chess/callbacks.h
@@ -66,6 +66,8 @@ struct ThinkingInfo {
   int64_t nodes = -1;
   // Nodes per second.
   int nps = -1;
+  // Evaluations per second.
+  int eps = -1;
   // Hash fullness * 1000
   int hashfull = -1;
   // Moves to mate.
--- a/src/chess/uciloop.cc
+++ b/src/chess/uciloop.cc
@@ -54,6 +54,11 @@ const OptionId kShowWDL{{.long_flag = "show-wdl",
                          .uci_option = "UCI_ShowWDL",
                          .help_text = "Show win, draw and lose probability.",
                          .visibility = OptionId::kAlwaysVisible}};
+const OptionId kShowEPS{
+    {.long_flag = "show-eps",
+     .uci_option = "UCI_ShowEPS",
+     .help_text = "Show neural network evaluations per second.",
+     .visibility = OptionId::kAlwaysVisible}};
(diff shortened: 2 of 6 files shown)


mooskagh:
There's a long need to output various stats, e.g. whether we want to include cache hits into visits, whether we want to count leaves or internal nodes (Monty style), whether we count terminals first time or every time we hit them, we count multivisits as one or as multivisit value. Whether we want these numbers per second or total.

With this, it appears that adding new values to uci info line is not scalable.

It's not clear what would be the best way to output such data. We considered outputting them as jsonl (e.g. this branch https://github.com/mooskagh/lc0/tree/json started to implement it).  
For `lc3` we also plan to collect and output the detailed stats: https://github.com/mooskagh/lc0/blob/new-search-draft/src/search/lc3/metrics/nodes_metric.h

jsonl-like interface is what I'd also steer UCI protocol to (as it will allow more advanced features like realtime search tree introspection).

As with other features (like tracing), it's not clear whether we should do something simple now, or do it "properly" some time in future.

mooskagh:
"evaluations per second" term is a bit misleading to my ear, it may mean "batches per second" or (also useful) "evaluated positions per second".

mooskagh (approved):

PR #2331, opened 2025-10-29
Add basic NVTX tracing support

Menkib64:
This adds a few useful basic annotations to Nsight system profiles. It helps compare CPU execution speed to GPU speed. This adds only a few annotations to most likely suspects to cause issues.

### DIFF
--- a/src/utils/trace.h
+++ b/src/utils/trace.h
@@ -57,15 +57,15 @@ namespace lczero {
   TRACE_EVENT("lc0", name)
 
 #elif USE_NVTX_TRACE
-#define LCTRACE_DEFINE_CATEGORIES /* nop */
-#define LCTRACE_INITIALIZE        /* nop */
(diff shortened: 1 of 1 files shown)


mooskagh:
FYI I'm adding Perfetto-based traces into our search (especially `lc3`, where there's more interaction between threads to watch far), which is I believe more portable and controllable thing than do NVIDIA-based traces.

Until it's ready, it's probably fine to keep these calls in the search, but after that, I'd prefer to replace it and only keep NVIDIA traces inside NVIDIA backends.

Menkib64:
As an example an evaluation from `nsys profile -w true -t cuda,nvtx,osrt,cuda-hw --cuda-graph-trace=graph --gpu-metrics-devices=all --cuda-event-trace=false ./lc0 benchmark -b onnx-trt --weights=../../t1-256x10-distilled-swa-2432500.pb.gz --num-positions=1` 
image

PR #2333, opened 2025-10-31
Cuda backend graph support

Menkib64:


src/neural/backends/cuda/common_kernels.cu
```diff
@@ -437,54 +439,48 @@ __global__ void batchNorm_kernel(T* output, const T* input, const T* skipInput,
...
-    shVals[threadIdx.x] = values[planeIndex + threadIdx.x];
-  }
-  __syncthreads();
-
-  uint64_t mask = shMasks[threadIdx.x >> 6];
+  uint64_t mask = masks[planeIndex];
 
   int sqIndex = index & 0x3F;
-  float op = 0;
+  half op[2] = {0, 0};
```
borg323: Why is this `half`?

Menkib64: It is a mistake. I should remove the IOType template parameter because it isn't used anymore. I should also follow the typical way of using template parameter to select the type.

src/neural/backends/cuda/network_cudnn.cc
```diff
@@ -664,86 +678,132 @@ class CudnnNetwork : public Network {
...
+    ReportCUDAErrors(cudaEventRecord(io->upload_done_event_, upload_stream_));
+    ReportCUDAErrors(
+        cudaStreamWaitEvent(compute_stream_, io->upload_done_event_, 0));
+  }
+
+  void GraphLaunch(InputsOutputs<DataType>* io, int batchSize) {
+    UploadInputs(io, batchSize);
+
+    // Make sure graph has completed upload before launching it.
+    ReportCUDAErrors(cudaStreamSynchronize(io->exec_stream_));
```
borg323, line 729: is this only needed for `CUDART_VERSION >= 11010`?

src/neural/backends/cuda/network_cuda.cc
```diff
@@ -593,32 +620,101 @@ class CudaNetwork : public Network {
...
+          cudaEventRecord(io->download_done_event_, compute_stream_));
+    } else
+#endif
+    {
+      io->cuda_graphs_[batchSize - 1].Launch(io->exec_stream_);
+#if !CUDA_GRAPH_SUPPORTS_EXTERNAL_EVENTS
+      ReportCUDAErrors(
+          cudaEventRecord(io->download_done_event_, io->exec_stream_));
+#endif
+    }
```
borg323: I think this would be cleaner as
```c++
#if CUDA_GRAPH_SUPPORTS_EXTERNAL_EVENTS
    io->cuda_graphs_[batchSize - 1].Launch(io->exec_stream_);
#else
    if (!multi_stream_) {
      UploadInputs(io, batchSize);

      io->cuda_graphs_[batchSize - 1].Launch(compute_stream_);
      ReportCUDAErrors(
          cudaEventRecord(io->download_done_event_, compute_stream_));
    } else {
      io->cuda_graphs_[batchSize - 1].Launch(io->exec_stream_);
      ReportCUDAErrors(
          cudaEventRecord(io->download_done_event_, io->exec_stream_));
    }
    ```

PR #2335, opened 2025-11-01
Nvcc from cudnn path and debug symbols

Menkib64:


meson.build
```diff
@@ -450,10 +452,16 @@ if get_option('build_backends')
   ## cuDNN
   ## ~~~~~
   cudnn_libdirs = get_option('cudnn_libdirs')
+  nvcc_paths = []
+  foreach p : cudnn_libdirs
+    nvcc_paths += fs.parent(p) + '/bin/nvcc'
+  endforeach
+  nvcc_paths += ['nvcc', '/usr/local/cuda/bin/nvcc', '/opt/cuda/bin/nvcc']
```
borg323, line 459: Do we still need this?

Menkib64: I don't know. I have too little knowledge how different nvidia packages work. I suspect we have to keep `nvcc`. Others might be useless when using library path to look for the compiler. I kept the old defaults to avoid accidentally breaking builds.

borg323 (approved):

PR #2360, opened 2025-12-05
restore support for v2 weights files

borg323:


src/neural/loader.cc
```diff
@@ -190,6 +190,88 @@ WeightsFile ParseWeightsProto(const std::string& buffer) {
   return net;
 }
 
+using FloatVector = std::vector<float>;
+using FloatVectors = std::vector<FloatVector>;
+
+FloatVectors LoadFloatsFromFile(std::string* buffer) {
+  // Parse buffer.
+  FloatVectors result;
```
mooskagh: Maybe something like this? (but as far as I understnand, it also deletes the first line?)

```cpp
FloatVectors LoadFloatsFromFile(std::string_view buffer) {
  FloatVectors result;
  float val;
  
  for (std::string_view line : absl::StrSplit(buffer, absl::ByAnyChar("\n\r"), absl::SkipEmpty())) {
    FloatVector row;
    for (std::string_view token : absl::StrSplit(line, absl::ByAnyChar(" \t"), absl::SkipEmpty())) {
      if (!absl::SimpleAtof(token, &val)) throw Exception("Bad file.");
      row.push_back(val);
    }
    if (!row.empty()) result.push_back(std::move(row));
  }
  return result;
}
```

src/neural/loader.cc
```diff
@@ -110,6 +113,11 @@ void FixOlderWeightsFile(WeightsFile* file) {
   auto network_format = file->format().network_format().network();
   const auto has_network_format = file->format().has_network_format();
 
+  // The version should be more fine grained, for now use latest.
+  file->mutable_min_version()->set_major(LC0_VERSION_MAJOR);
```
mooskagh: I believe that will just override future (proto) nets to the current version so the version check won't work?

borg323: Right, I'll add a check for earlier version first.

mooskagh: Or maybe add an override not inside this function, but outside of it only in .txt pass?

borg323: > Or maybe add an override not inside this function, but outside of it only in .txt pass?

Eventually I took this approach.

PR #2375, opened 2026-01-20
feat: Add ROCm backend with attention network support

johnnytshi:
Implemented a complete ROCm backend for AMD GPUs, enabling support for modern attention-based chess networks on RDNA 3.5 and other AMD architectures.

Implementation Details:
- Added full ROCm backend in src/neural/backends/rocm/
- Implemented attention network architecture (multi-head self-attention, FFN, embeddings)
- Used rocBLAS for GEMM operations and MIOpen for convolutions
- NCHW layout optimized for FP16 performance on RDNA 3.5
- Three backend variants: rocm (FP32), rocm-fp16 (FP16), rocm-auto (auto-detect)
- MIOpen is required dependency (similar to cuDNN for CUDA)
- Automatic AMD GPU architecture detection via rocm_agent_enumerator
- Build option: -Drocm=true -Damd_gfx=gfx1151 (or auto-detect)

Key Files:
- src/neural/backends/rocm/network_rocm.cc - Main network implementation
- src/neural/backends/rocm/layers.{cc,h} - Layer implementations
- src/neural/backends/rocm/*.hip - GPU kernels (FP16 and FP32)
- meson.build, meson_options.txt - Build configuration

Performance Notes:
- FP16 performance: >2000 nps on Strix Halo (Radeon 8060S, gfx1151)
- Automatic batch size tuning (min_batch=64 for RDNA 3.5)
- Tested rocWMMA but rocBLAS provided better performance

OpenCL/SYCL Compatibility:
- Preserved existing OpenCL/SYCL AMD backend (uses hip_* naming)
- ROCm backend separate from SYCL backend (uses rocm_* naming)

Verification (Strix Halo - Radeon 8060S, gfx1151):
- Tested models: 768x15x24h-t82-swa-7464000.pb.gz and maia-1900.pb.gz
- Backend: rocm-fp16 functional and producing correct moves
- ROCm 7.2.53150, MIOpen 3.5.1
- Only tested on RDNA 3.5; other AMD architectures not verified


borg323:
There is a lot to discuss here, can you join our [Discord chat](https://discord.gg/pKujYxD)?

meson.build
```diff
@@ -402,6 +402,76 @@ if get_option('build_backends')
...
+  hipcc = find_program('hipcc', '/opt/rocm/bin/hipcc', required: false)
+
+  if get_option('rocm') and rocblas_lib.found() and amdhip_lib.found() and miopen_lib.found() and hipcc.found()
+    deps += [rocblas_lib, hipblas_lib, amdhip_lib, miopen_lib]
+
+    rocm_files = ['src/neural/backends/rocm/rocm_common.cc',
+                  'src/neural/backends/rocm/layers.cc',
+                  'src/neural/backends/rocm/network_rocm.cc']
+
+    add_project_arguments('-DUSE_MIOPEN', language : 'cpp')
```
Menkib64, line 432: `USE_MIOPEN` seem redundant. The code can be enabled always.

meson.build
```diff
@@ -402,6 +402,76 @@ if get_option('build_backends')
...
+
+    foreach d : get_option('hip_include')
+      if run_command('scripts/checkdir.py', d, check : false).returncode() == 0
+        includes += include_directories(d, is_system: true)
+      endif
+    endforeach
+    includes += include_directories('src/neural/backends/rocm/')
```
Menkib64, line 439: Backend should work without adding an include path.

meson.build
```diff
@@ -402,6 +402,76 @@ if get_option('build_backends')
...
+
+    add_project_arguments('-DUSE_HIP', '-D__HIP_PLATFORM_AMD__', language : 'cpp')
```
Menkib64, line 441: `USE_HIP` looks redundant.

`__HIP_PLATFORM_AMD__`  can be set using `dependency('hip')`

src/neural/backends/rocm/network_rocm.cc
```diff
@@ -0,0 +1,1280 @@
...
+    // Architecture-specific batch size optimization for RDNA 3.5 (gfx1151)
+    // Based on benchmarking: FP16 peaks at batch 49-73 (1,750-1,800 nps)
+    // Only apply if user hasn't explicitly set min_batch
+    std::string arch_name(deviceProp.gcnArchName);
+    if (arch_name.find("gfx1151") != std::string::npos &&
+        !options.Exists<int>("min_batch")) {
+      constexpr bool fp16 = std::is_same<half, DataType>::value;
+      if (fp16) {
+        // FP16: optimal batch 49-73, use 64 as default
+        min_batch_size_ = 64;
```
Menkib64, line 243: Minimum batch size of 64 doesn't sound optimal when search wants to evaluate only a position. Latency of evaluation matters too.

If kernels would be optimized well for AMD, then I would expect that optimal batch size would be multiple of `N * compute_units/2`. `N` would depend on the size of network in use. For BT4 it should be 1 and smaller networks would require higher N.

src/neural/backends/rocm/network_rocm.cc
```diff
@@ -0,0 +1,1280 @@
...
+    const bool mish_net = file.format().network_format().default_activation() ==
+                          pblczero::NetworkFormat::DEFAULT_ACTIVATION_MISH;
+
+    // 1. Allocate scratch space (used internally by cudnn to run convolutions,
+    //     and also for format/layout conversion for weights).
+    miopenTensorDescriptor_t wDesc;
+    miopenConvolutionDescriptor_t convDesc;
+    miopenTensorDescriptor_t xDesc;
+    // MIOpen uses miopenCreateTensorDescriptor for both filters and tensors
+    miopenCreateTensorDescriptor(&wDesc);
```
Menkib64, line 413: miopen initialization and allocations should happen only when network needs miopen. The important networks use transformers only currently.

src/neural/backends/rocm/network_rocm.cc
```diff
@@ -0,0 +1,1280 @@
...
+      network_[l++]->Eval(batchSize, spare2, spare1, nullptr, scratch_mem_,
+                          scratch_size_, cudnn_, cublas_, stream);  // moves FC1
+
+      // Moves left FC2
+      network_[l++]->Eval(batchSize, (DataType*)opMov, spare2, nullptr,
+                          scratch_mem_, scratch_size_, cudnn_, cublas_, stream);
+    }
+
+    // Defer policy copy until after all GPU work is queued (Optimization #3)
+    ReportHIPErrors(hipMemcpyAsync(io->op_policy_mem_, io->op_policy_mem_gpu_,
```
Menkib64, line 1002: There should be a separate stream for output downloads. A separate stream allows download happen at the same time while other computations are happening.

The latest version of cuda backend has an example how to do it for both uploads and downloads.

src/neural/backends/rocm/network_rocm.cc
```diff
@@ -0,0 +1,1280 @@
...
+                                   sizeof(float) * kNumOutputPolicy * batchSize,
+                                   hipMemcpyDeviceToHost, stream));
+    // Record event after policy transfer
+    ReportHIPErrors(hipEventRecord(policy_ready_event_, stream));
+
+    // Wait ONLY for value (needed for CPU softmax)
+    ReportHIPErrors(hipEventSynchronize(value_ready_event_));
```
Menkib64, line 1009: `hipEventSynchronize` should be outside locked section. It will be blocking until GPU has completed operations. We want to queue the next batch before the current batch completes the computation.

meson.build
```diff
@@ -402,6 +402,76 @@ if get_option('build_backends')
 
   endif
 
+  ## ~~~~~
+  ## ROCm
+  ## ~~~~~
+  rocm_libdirs = get_option('hip_libdirs')
+  rocblas_lib = cc.find_library('rocblas', dirs: rocm_libdirs, required: false)
+  hipblas_lib = cc.find_library('hipblas', dirs: rocm_libdirs, required: false)
+  miopen_lib = cc.find_library('MIOpen', dirs: rocm_libdirs, required: false)
```
Menkib64: Library discovery should use `dependency()` functions if possible. It makes it easier to discovery correct compiler and linker options.

src/neural/backends/rocm/network_rocm.cc
```diff
@@ -0,0 +1,1280 @@
...
+    CERR << "MIOpen created";
+    ReportROCBLASErrors(rocblas_create_handle(&cublas_));
+    CERR << "rocBLAS created";
+
+    // Default layout is nchw.
+    nhwc_ = false;
+    constexpr bool fp16 = std::is_same<half, DataType>::value;
+
+    if (fp16) {
+      // Check if the GPU support FP16.
```
Menkib64, line 291: [Documentation](https://rocm.docs.amd.com/projects/HIP/en/latest/reference/low_fp_types.html) claims fp16 support for all hardware. This makes me thing that support check can be removed.

meson.build
```diff
@@ -402,6 +402,76 @@ if get_option('build_backends')
...
+  amdhip_lib = cc.find_library('amdhip64', dirs: rocm_libdirs, required: false)
```
Menkib64, line 413: `dependecy()` function makes discovering libraries and flags easier. It should be used if possible. You can see my quick hacky build changes from #2365 for an example. My changes requires using hipcc for everything because it made flag handling simpler for quick testing. It should be possible to use hipcc only for kernels like your changes do.

Menkib64 (commented):
#2365 has an automated conversion from a newer version of cuda backend. I stopped attempting to fix bugs which I found because onnx-migraphx is much faster. My guess is that most kernels would have to be tuned specifically for AMD.

I discovered a few bugs which I documented in the pull request message.

> Automatic batch size tuning (min_batch=64 for RDNA 3.5)

This sounds like you might be optimizing for a wrong target. Optimization target is a combination of batch latency and throughput. Search cannot always produce optimally sized batches which requires backend to evaluate suboptimal batches at the lowest possible latency. Backend or user should tell the optimal batch for the search which should be the smallest batch size which reaches the best or close to the best throughput. Running `backendbench` with `--clippy`  can be used to estimate the best trade of between latency and throughput.

> Tested rocWMMA but rocBLAS provided better performance

I haven't used rocWMMA. My assumption is that either rocWMMA, composable kernels or MigraphX offers the best performance potential. User code has to implement the same optimizations which are included in rocBLAS. It can be hard to optimize GEMM correctly because small changes can prevent HW from doing memory operations at the same time as computation. These libraries should allow user to fuse other work with matrix multiplications. Fused kernels avoid overhead from small kernels which allow higher network evaluation performance. The first step would be reaching rocBLAS performance with custom matrix multiplication kernels.

I mostly skipped GPU code. There are too many changes to quickly check everything.

src/neural/backends/rocm/network_rocm.cc
Technetium1: @Menkib64 
> This precision format is supported across all GPU architectures. The HIP types and functions are available for use in both host and device code, **with implementation handled by the compiler and device libraries**.

I'm **not** an expert here, but I think there are some older GPUs that don't support it directly, so a check seems right.

Menkib64: [D3D optimization](https://gpuopen.com/learn/first-steps-implementing-fp16/) article claims that RX Vega was the first generation which supported fp16. ROCm HW support starts from the next generation (first RDNA and CDNA architectures).

meson.build
johnnytshi: Let me give onnx-migraphx a try on my system

I've just implemented fused kernel for attention on my local, which improved T82 backendbench from 2000 nps, to 2500 nps. Right now I am playing around with multi stream, supposely it would improve another 30%.

johnnytshi:
building onnx-migraph is criminal. I had to build rocMLIR, AMDMIGraphX, then onnxruntime. at lower batch size, migraph is faster, but max is under 2000 nps for T82 across different batch sizes

ROCm with handcrafted fused kernel hit 2500 nps for T82, with batch size of 64

justinkb:
I'll be trying this on NixOS with gfx1100. I've gotten the SYCL amd backend working just now and I'd like to compare the performance between this and that. I'll let you know once I get all this up and running

PR #2376, opened 2026-01-28
show correct castling move for frc

borg323:


Menkib64:
When was the castling wrong for frc?

I noticed that search claims that we are always playing frc. uci loop uses `kUciChess960` to select if output should be frc or standard.

borg323:
It is a display issue with FRC positions when `UCI_Chess960` is false:
```
position fen 2rkr3/2p1p3/8/8/8/8/8/RRK5 w B - 0 1
go nodes 2
info depth 1 seldepth 2 time 101 nodes 615 score mate 1 nps 12551 tbhits 0 pv c1c1
bestmove c1c1
```

Menkib64:
> It is a display issue with FRC positions when `UCI_Chess960` is false:
> 
> ```
> position fen 2rkr3/2p1p3/8/8/8/8/8/RRK5 w B - 0 1
> go nodes 2
> info depth 1 seldepth 2 time 101 nodes 615 score mate 1 nps 12551 tbhits 0 pv c1c1
> bestmove c1c1
> ```

This makes sense. FRC option isn't set. Would it be a better fix if using FRC fen would automatically set `UCI_Chess960`?

src/chess/types.h
```diff
@@ -229,9 +229,12 @@ inline PieceType PieceType::Parse(char c) {
 }
 
 inline std::string Move::ToString(bool is_chess960) const {
-  if (is_castling() && !is_chess960) {
-    return from().ToString() + (to().file() > from().file() ? "g" : "c") +
-           to().rank().ToString();
+  if (is_castling() && !is_chess960 && from().file() == kFileE) {
+    if (to().file() == kFileA) {
+      return from().ToString() + "c" +  to().rank().ToString();
+    } else if (to().file() == kFileH) {
```
mooskagh, line 235: super-optional: maybe drop this "else"?

mooskagh (approved):

src/chess/types.h
borg323: The issue is that `is_chess960` is false if the `UCI_Chess960` option is not set correctly, possibly giving invalid castling moves.

Menkib64: > The issue is that `is_chess960` is false if the `UCI_Chess960` option is not set correctly, possibly giving invalid castling moves.

I guess UCI way of thinking would be that it is a GUI bug if `UCI_Chess960` isn't set. If we want to auto detect the 960 position, I think more logical place is `position fen` handling checking if fen is 960 positions and setting a flag. Then we output 960 castling if either option is set or flag is set.

borg323: I would totally agree if we used the option anywhere else, but it is only used for showing the castling moves (after the refactoring), so I don't think it is worth the added complexity.

Menkib64: I think this detection misses some FRC positions because it only checks one of rooks. Initial fen might be stated as FRC position with castling rights specifying rook files. FEN parser would also accept FRC position with standard castling flags but one of rooks on a different file. These cases makes me think that the correct place to check for FRC is when handling uci `position` command.

borg323: After the refactoring, all moves are encoded as FRC internally - only the uci output changes depending on the option, and this code path just checks for the two classical castling positions to output the king move instead of the king takes rook convention (final `return` statement).

Menkib64: image
If we get this position as a fen, we should be outputting FRC castling. But only short castle will be detected as FRC castling move.

There is also possibility that fen has `AHah` castling rights with king on e file. The code would choose standard chess. Should this fen signal that user wants 960? I would think that it means 960 if we are doing automatic detection for the variant.

mooskagh: > super-optional: maybe drop this "else"?

I meant just dropping the token "else" literally.

```
if (blah) {
  return
} else if (...
```
would be the same as
```
if (blah) {
  return
}
if (...
```

PR #2382, opened 2026-02-07
Add first batch time to backendbench

Menkib64:


mooskagh:
I'm not sure what is the intention...

We run several threads in parallel, and then computes deltas between finishing first eval in each thread. I assume it's expected that ComputeBlocking() is single threaded and serializes computations, but still I'm not sure what it's supposed to measure?

If it's time to warmup, then why we do it in every thread rather than just once?

Menkib64:
> I'm not sure what is the intention...
> 
> We run several threads in parallel, and then computes deltas between finishing first eval in each thread. I assume it's expected that ComputeBlocking() is single threaded and serializes computations, but still I'm not sure what it's supposed to measure?
> 
> If it's time to warmup, then why we do it in every thread rather than just once?

The main target is to catch performance problems for the first batch. I don't know if the only first is enough to capture all weird performance problems. A potential example problem could be Windows power management that can sometimes behavior unexpectedly.

src/tools/backendbench.cc
```diff
@@ -215,8 +228,22 @@ void BackendBenchmark::Run() {
                      " max nps,"
                      "  median,"
                      " min nps,"
+                     " first max,"
+                     " first mean"
                   << std::endl;
       }
+      double first_max = 0.0;
```
mooskagh, line 236: It's still quite questionable to me that this is useful metric, but it doesn't harm to have it in any case.

But it can be rewritten a bit more idiomatically:
1. Instead of multiplying seconds by 1000, we can takes measurements in milliseconds
2. We cut interval into n-1 pieces and compute average, it's the same as taking the entire interval as divide by n-1.

i.e.

```
using DoubleMs = std::chrono::duration<double, std::milli>;

double first_max = 0.0;
for (size_t j = 1; j < num_threads; ++j) {
  first_max = std::max(first_max,
      DoubleMs(
        first_batch_times[j] - first_batch_times[j - 1]
      ).count());
}

double first_mean = DoubleMs(first_batch_times.back() - first_batch_times.front())
    .count() / threads;
```

mooskagh (approved):

PR #2388, opened 2026-02-19
Fix dag n_in_flight underflow

Menkib64:
ShouldStopPickingHere can change return value for a node if the garbage collection thread frees an alternative parent while node is considered for visits. Initially picking assumed that it would stop here and generate conflicts which would skip n_in_flight increments. Later decision would change to visit children which would assume n_in_flight was incremented.

These changes were originally part of #2318. They fix an underflow bug. I think this is the smallest set of changes which can be copied with minimal modifications from the big pull request.

### DIFF
--- a/src/search/dag_classic/search.cc
+++ b/src/search/dag_classic/search.cc
@@ -1717,17 +1717,18 @@ void SearchWorker::PickNodesToExtendTask(
   current_path.emplace_back(collision_limit, true, visit_root, stop_root, 0);
   // take base sqrt(2) logarithm of root node for large subbranch N limit. It is
   // used to split tasks.
-  const unsigned large_branch_limit = std::bit_width(search_->root_node_->GetN()) * 2;
+  const unsigned large_branch_limit =
+      std::bit_width(search_->root_node_->GetN()) * 2;
 
   while (current_path.size() > 0) {
     assert(full_path.size() >= path.size());
     // First prepare visits_to_perform.
-    int cur_limit = current_path.back().bits_.visits_;
+    int cur_limit = current_path.back().visits_;
     // First prepare visits_to_perform.
     // First check if node is terminal or not-expanded.  If either than create
     // a collision of appropriate size and pop current_path.
-    if (current_path.back().bits_.stop_picking_) {
-      if (current_path.back().bits_.visit_child_) {
+    if (current_path.back().stop_picking_) {
+      if (current_path.back().visit_child_) {
         cur_limit -= 1;
         receiver->push_back(NodeToProcess::Visit(full_path, history));
       }
@@ -1743,7 +1744,7 @@ void SearchWorker::PickNodesToExtendTask(
       }
       bool saved_is_last_child;
       do {
-        saved_is_last_child = current_path.back().bits_.last_child_;
+        saved_is_last_child = current_path.back().last_child_;
         current_path.pop_back();
         if (current_path.size() == 0) break;
         history.Pop();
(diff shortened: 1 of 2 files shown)


src/search/dag_classic/search.h
```diff
@@ -396,18 +397,55 @@ class SearchWorker {
           repetitions(std::get<1>(path.back())) {}
   };
 
+  // Combine visits to perform, index, and node state flags into a packed
+  // variable. Packed value stores required visit infromation which can be
+  // pushed into the current_path stack.
+  struct CurrentPath {
+    uint32_t visits_ : 20;       // <= collision limit
```
mooskagh, line 404: `kMaxCollisionVisits` currently has upper bound of `100000000`, which is 27 bits.  
But I guess collision_visits in our gather code usually means "target batch size"?

src/search/dag_classic/search.h
```diff
@@ -396,18 +397,55 @@ class SearchWorker {
...
+                "CurrentPath must be packed into 32 bits");
+
   // Holds per task worker scratch data
   struct TaskWorkspace {
     std::array<Node::Iterator, 256> cur_iters;
-    std::vector<std::unique_ptr<std::array<int, 256>>> vtp_buffer;
-    std::vector<std::unique_ptr<std::array<int, 256>>> visits_to_perform;
-    std::vector<int> vtp_last_filled;
-    std::vector<int> current_path;
+    std::vector<CurrentPath> current_path;
```
mooskagh, line 446: Not sure how much of optimization it would be, but there's also absl::InlinedVector<CurrentPath, 30>

src/search/dag_classic/search.cc
```diff
@@ -1896,63 +1879,97 @@ void SearchWorker::PickNodesToExtendTask(
...
         if (task_workers_ > 0 &&
-            child_limit > params_.GetMinimumWorkSizeForPicking() &&
-            child_limit <
-                ((collision_limit - passed_off - completed_visits) * 2 / 3) &&
-            child_limit + passed_off + completed_visits <
-                collision_limit -
-                    params_.GetMinimumRemainingWorkSizeForPicking()) {
-          Node* child_node = cur_iters[i].GetOrSpawnNode(/* parent */ node);
-          history.Append(cur_iters[i].GetMove());
+            ((is_large_main_branch && is_large_branch) ||
```
mooskagh, line 1960: Why doesn't `is_large_branch` imply `is_large_main_branch`? Is that due to transpositions?

mooskagh (approved):

src/search/dag_classic/search.h
Menkib64: Yes. The current upper limit is set by following `std::min`
```    PickNodesToExtend(
        std::min({collisions_left, target_minibatch_size_ - minibatch_size,
                  max_out_of_order_ - number_out_of_order_}));
```
I plan to increase it to `target_minibatch_size + upper_bound_of_out_of_order + uppper_bound_of_collisions`. Realistically this should be limited much lower than current `kMaxCollisionVisits` because it would likely require extremely bad tree state where it is likely better to evaluate smaller batches until the tree recovers to a healthier state.

src/search/dag_classic/search.h
Menkib64: We need `max_depth * selected_children_per_level`. This should be more than hundred in a typical local analysis. TCEC like scenarios could reach thousands because larger minibatch and deeper tree.

src/search/dag_classic/search.cc
Menkib64: `is_large_main_branch` is for the child which is selected to have the most visits. The most visited child is always handled by the current thread. The above rule is making decision if it is worth sending other child visits to a task or do them locally. The main branch might not be the most visited child if a less visited has higher Q.

src/search/dag_classic/search.h
mooskagh: InlinedVector would have 30 inlined (i.e. it would work like std::array up to 30), and when number crosses 30, it would behave like a standard vector with allocation on heap.

Menkib64: I don't know if inline is worth it because we would be anyway doing heap allocations even in bullet games. I guess it could make sense if we would avoid heap allocations in most cases. Would it be worth reserving enough inline space for most usecases?

PR #2395, opened 2026-03-08
fix: prevent segfault in Python bindings when calling `Weights.filters()` on transformer networks

sparsh2:
Calling `Weights.filters()` via the Python bindings on a T1/smolgen/transformer network causes a **segfault**, crashing the Python process with no recoverable error.

This happens because `filters()` unconditionally accesses `residual(0)` to compute the filter count, but transformer-based networks store their architecture differently and have `residual_size() == 0`. The out-of-bounds access on the empty repeated field causes undefined behavior (segfault on most platforms).

**Reproduction:**

```python
from lczero.backends import Weights

w = Weights("t1-512x15x8h-distilled-swa-3395000.pb.gz")
w.blocks()   # returns 0 (no crash, but misleading)
w.filters()  # segfault — process killed
```

### DIFF
--- a/src/python/weights.h
+++ b/src/python/weights.h
@@ -71,6 +71,11 @@ class Weights {
   }
   int blocks() const { return weights_.weights().residual_size(); }
   int filters() const {
+    if (weights_.weights().residual_size() == 0) {
+      throw Exception(
+          "filters() is not available for this network architecture "
+          "(no residual blocks found — likely a transformer/T1 network).");
+    }
     return weights_.weights().residual(0).conv1().weights().params().size() /
            2304;
   }
--- a/src/python/weights.h
+++ b/src/python/weights.h
@@ -72,9 +72,7 @@ class Weights {
   int blocks() const { return weights_.weights().residual_size(); }
   int filters() const {
     if (weights_.weights().residual_size() == 0) {
-      throw Exception(
-          "filters() is not available for this network architecture "
-          "(no residual blocks found - likely a transformer/T1 network).");
+      throw Exception("This network does not use residual blocks. There are no filters.");
     }
     return weights_.weights().residual(0).conv1().weights().params().size() /
            2304;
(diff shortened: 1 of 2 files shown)


src/python/weights.h
```diff
@@ -71,6 +71,11 @@ class Weights {
   }
   int blocks() const { return weights_.weights().residual_size(); }
   int filters() const {
+    if (weights_.weights().residual_size() == 0) {
+      throw Exception(
+          "filters() is not available for this network architecture "
+          "(no residual blocks found — likely a transformer/T1 network).");
```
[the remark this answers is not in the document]
sparsh2: done

src/python/weights.h
```diff
@@ -71,6 +71,11 @@ class Weights {
   }
   int blocks() const { return weights_.weights().residual_size(); }
   int filters() const {
+    if (weights_.weights().residual_size() == 0) {
+      throw Exception(
+          "filters() is not available for this network architecture "
+          "(no residual blocks found - likely a transformer/T1 network).");
```
Menkib64: I think it should say more like
`This network doesn't use residual blocks. There is no filters.`

Menkib64 (changes requested):

src/python/weights.h
mooskagh: I'd say it's better to rename it to "channels", and make it return number of filters for resnets and embedding size for transformers.
Also I believe similar change would make sense to do with blocks.

sparsh2: @mooskagh that would be a breaking change, no? I can add additional transformer specific APIs in the follow up PR like `is_transformer()`, `attention_heads()`, `embedding_size()`

Menkib64 (approved):

src/python/weights.h
mooskagh: Yeah but nobody will notice because there's hardly any using of the bindings.

PR #2398, opened 2026-03-13
Replace WDL_mu fallback with thresholds.

Menkib64:
This centipawn conversion formula has been fitted trought 3 fixed points. 75% expected score is 1.00 centipawns. 100% expected score is 128.0 centipawns. 50% expected score is 0.0 centipawns.

Default score type is changed to centipawn because WDL_mu doesn't match how search ranks moves.

### DIFF
--- a/src/search/dag_classic/search.cc
+++ b/src/search/dag_classic/search.cc
@@ -273,13 +273,25 @@ namespace {
 // WDL conversion formula based on random walk model.
 inline double WDLRescale(float& v, float& d, float wdl_rescale_ratio,
                          float wdl_rescale_diff, float sign, bool invert,
-                         float max_reasonable_s) {
+                         float max_reasonable_s, bool use_threshold = false) {
   if (invert) {
     wdl_rescale_diff = -wdl_rescale_diff;
     wdl_rescale_ratio = 1.0f / wdl_rescale_ratio;
   }
   auto w = (1 + v - d) / 2;
   auto l = (1 - v - d) / 2;
+  if (use_threshold) {
+    // Keep centipawn scores within resonable limits when l or d becomes close
+    // to zero.
+    const float wl_threshold = 0.006f;
+    const float d_threshold = 0.000492143f;
+    w = std::max(w, wl_threshold);
+    l = std::max(l, wl_threshold);
+    float d_copy = std::max(d, d_threshold);
+    const float scale = 1.0f / (w + l + d_copy);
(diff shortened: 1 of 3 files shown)


src/search/dag_classic/search.cc
```diff
@@ -367,7 +367,7 @@ void Search::SendUciInfo(const classic::IterationStats& stats)
     } else if (score_type == "centipawn_with_drawscore") {
       uci_info.score = 90 * tan(1.5637541897 * q);
     } else if (score_type == "centipawn") {
-      uci_info.score = 90 * tan(1.5637541897 * wl);
+      uci_info.score = 100.79055 * tan(1.56292199291664 * wl);
```
mooskagh, line 382: Usually when we update centipawn formula, we keep previous as `centipawn_<year>` (where year is when it was introduced)

Naphthalin:
I'm strongly against changing the default centipawn formula back to being Q based for several reasons:
- Q (as derived from WDL) is inherently dependent on the playing conditions (specified through `WDLCalibrationElo`), which means Q values for the same position aren't comparable 
- the primary objective of an eval is to communicate something about a position, and the absolute meaning is in 95% of use cases more important than multiPV moves being in the "right" order
- to add on that, ordering of moves isn't something universal either, and again depends on the playing conditions -- if you have e.g. a +2.5 position, there might be a riskier move which (when played through precisely) maintains the +2.5 eval, and a safe alternative which is only +2.2 but has no counterplay. I personally fail to see the downside of Leela recommending the +2.2 line over the +2.5 line, and trust the user to not get confused by this.
- and ultimately, the several mathematical properties of `WDL_mu` mean it can 1:1 be translated to SF evals (assuming they're perfectly normalized) in the range of [-2.5, +2.5] which wasn't the case for any Q based eval we tried so far, nor will it be the case for this formula. The fallback behavior (which is unfortunately necessary) isn't ideal, but only rarely relevant for TCEC conditions which shouldn't be our primary development target.

Furthermore, search doesn't order moves by Q anyway, but by nodes -- which is of course closely correlated with Q, but so is the `WDL_mu` eval. 

In total, reverting back to a Q based centipawn formula would put a significant burden in interpreting and translating Leela's eval, including suffering from old issues where different nets give vastly different evals at default settings because of having different inherent draw rates. I'm not generally against adding a new `centipawn_2026` score type option, but definitely don't want to see it as the default.

PR #2409, opened 2026-04-22
revert ponder behavior

borg323:
The ponder behavior changed with v0.32, and this led to some issues: https://github.com/orgs/LeelaChessZero/discussions/2354#discussioncomment-15799597
I have identified 3 differences with the previous version:
1. Default `Ponder` is now false.
2. `go ponder` is not accepted if `Ponder` if false
3. `go ponder` is not accepted at startpos

Reviewing the uci spec, it seems the `go ponder` behavior should not change depending on the `Ponder` value:
> 		<id> = Ponder, type check
> 			this means that the engine is able to ponder.
> 			The GUI will send this whenever pondering is possible or not.
> 			Note: The engine should not start pondering on its own if this is enabled, this option is only
> 			needed because the engine might change its time management algorithm when pondering is allowed.

This PR reverts the behavior of points 2 and 3 above since SF is also accepting `go ponder` at startpos and has false as the default `Ponder` value.

### DIFF
--- a/src/engine.cc
+++ b/src/engine.cc
@@ -56,8 +56,8 @@ const OptionId kPonderId{
     {.long_flag = "",
      .uci_option = "Ponder",
      .help_text =
-         "Indicates to the engine that it will be requested to ponder. This "
-         "postpones resetting the search tree until the search is started.",
+         "Indicates to the engine that it will be requested to ponder. Not "
+         "currently used.",
      .visibility = OptionId::kAlwaysVisible}};
 
 const OptionId kPreload{"preload", "",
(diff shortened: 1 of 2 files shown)


src/engine.cc
```diff
@@ -98,9 +98,9 @@ class Engine::UciPonderForwarder : public UciResponder {
   }
   void OutputThinkingInfo(std::vector<ThinkingInfo>* infos) override {
     if (!wrapped_) return;
-    if (engine_->last_go_params_ && engine_->last_go_params_->ponder) {
-      assert(engine_->last_position_ &&
-             !engine_->last_position_->moves.empty());
+    if (engine_->last_go_params_ && engine_->last_go_params_->ponder &&
+        !engine_->last_position_->moves.empty()) {
+      assert(engine_->last_position_);
```
mooskagh: It's a bit weird to have assert here after we already would crash if it was false (as copilot mentioned). Maybe move it outside of the if as `assert(!engine_->last_go_params_ || engine_->last_position_);`?

mooskagh (approved):

PR #2410, opened 2026-04-22
Limit demux minibatch to maximum batch

Menkib64:


borg323 (approved):
Long term we need to rethink this mechanism.

Menkib64:
> Long term we need to rethink this mechanism.

I think the backend API will offer a better solution. It allows demux to set higher maximum batch size.

PR #2412, opened 2026-05-05
Add uncertainty weighting to dag-preview

Menkib64:
Uncertainty weighting uses networks error estimation to reduce weight on a node if error estimate is high. The weighing aims to improve value backpropagation with reduced weight when value error is likely to be higher.

This branch later modified weighting to also affect node selection to account for the new weight. This transfer a few more nodes towards branches which have high error estimation which are typically tactical line.


Menkib64:
This is @daniel-monroe's uncertainty weighting with a few changes.

The most important change is replacing visit count (`N`) with weight (`W`). I thought about this change to keep data structure within 64 byte limit. It allocates a few extra nodes towards uncertain branches. Visit counts are stored as a double precision float. It is used as a 32 bit float when calculating visit distribution.

The minor search focus change tested as a positive change using many different time controls and search parameters:
https://bench.lczero.org/test/901/
https://bench.lczero.org/test/902/
https://bench.lczero.org/test/903/
https://bench.lczero.org/test/907/

This branch backports changes from TCEC branch. I was trying to merge this for FRD event but there was issues getting it work before deadline. Validation test for this branch is scheduled after 907 completes:
https://bench.lczero.org/test/922/

I'm still going to run a test if logistic function would improve the error output conversion formula. Tune is scheduled after 907 completes:
https://bench.lczero.org/tune/923/

Menkib64:
It looks like logistic formula is a minor gain. I switched this pull request to use it [logistic vs exponent 8+0.08](https://bench.lczero.org/test/983/), [logistic vs exponent 40+0.4](https://bench.lczero.org/test/984/), and [TCEC logistic vs exponent 40+0.4](https://bench.lczero.org/test/982/)

PR #2415, opened 2026-05-26
Improve policy training targets

Menkib64:
Post processing aims to improve policy training targets. The expected improvements are reduced policy for one move blunders, better policy focus towards search discovered good move, and potential ability fix some single position evaluation mistakes.

Changes are taken from KataGo training improvements paper. KataGo has added more post-processing improvements later. I decided to keep the initial testing version based on paper version to keep changes minimal. If this change improves training targets, we can look into implementing other Katago improvements too.

### DIFF
--- a/src/search/classic/params.cc
+++ b/src/search/classic/params.cc
@@ -551,6 +551,14 @@ const OptionId SearchParams::kForcedExplorationVisitsId{
          "can include noise. This aims to overcome a problem if the first "
          "visit to a child evaluates it incorrectly which would suppress "
          "further visits even when noised policy wants to explore it."}};
+const OptionId SearchParams::kForcedExplorationMaxPolicyId{
+    {.long_flag = "forced-exploration-max-policy",
+     .uci_option = "ForcedExplorationMaxPolicy",
+     .help_text =
+         "Force minimum visits at root based on policy which can include "
+         "noise. This parameter scales down higher policy children visits to "
+         "avoid over exploration when policy would force enough exploration in "
+         "a normal search."}};
 const OptionId SearchParams::kSingleChildForcedBoostId{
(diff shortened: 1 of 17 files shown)


src/selfplay/tournament.cc
```diff
@@ -142,6 +142,7 @@ void SelfPlayTournament::PopulateOptions(OptionsParser* options) {
   defaults->Set<int>(classic::SearchParams::kMiniBatchSizeId, 32);
   defaults->Set<float>(classic::SearchParams::kCpuctId, 1.2f);
   defaults->Set<float>(classic::SearchParams::kCpuctFactorId, 0.0f);
+  defaults->Set<int>(classic::SearchParams::kForcedExplorationVisitsId, 220);
```
john-sp, line 145: This should have to be explicitly enabled in the RL run training params, not defaulted to enabled. While tests are showing this is likely an improvement, we don't want to change T91 on accident as users are updating.

src/search/classic/params.cc
```diff
@@ -637,6 +688,15 @@ void SearchParams::Populate(OptionsParser* options) {
   BaseSearchParams::Populate(options);
   options->Add<IntOption>(kMaxPrefetchBatchId, 0, 1024) = DEFAULT_MAX_PREFETCH;
   options->Add<IntOption>(kSolidTreeThresholdId, 1, 2000000000) = 100;
+  options->Add<BoolOption>(kUsePolicyPostProcessingId) = true;
```
john-sp, line 691: Similar here, we don't want to necessarily enable this for every training run with out an explicit override to disabled. We couldn't disable it universally via Command Line Args because clients in v0.32 and earlier will exit on the unknown setting in selfplay

PR #2420, opened 2026-06-11
Add a native HIP (ROCm) backend for AMD GPUs

jeffdaily:
Add a first-class ROCm/HIP backend (`hip` / `hip-fp16` / `hip-auto`) that compiles lc0's own CUDA NN-backend kernels (the cuBLAS + custom-kernel `network_cuda` path) with hipcc, so AMD GPUs run lc0's optimized kernels directly on ROCm without the Intel DPC++/SYCL toolchain. This is distinct from the existing SYCL backend (a different toolchain) and from onnx-rocm (which runs the network inside onnxruntime rather than lc0's own kernels).

Two NVIDIA-specific optional paths are intentionally not part of this backend. The separate cuDNN backend is a distinct backend (cuDNN is opt-in), and the cuBLAS `network_cuda` path ported here does not use cuDNN, so MIOpen is not required. The optional CUTLASS fused multi-head attention is also not used: with CUTLASS off (the default for many nets even on NVIDIA, where `use_fused_mha` is only enabled on sm80+ fp16) attention runs as the unfused cuBLAS path -- batched Q.K^T, softmax, then .V -- which is exactly what the NVIDIA build runs without CUTLASS and produces identical results. An AMD-native fused attention kernel (Composable Kernel / ck_tile) is a possible future optimization; the benefit is bounded here because lc0's attention is over the 64 board squares, where the unfused path is already cheap.

This work was authored with the assistance of Claude (Anthropic), an AI coding assistant, and validated on real AMD GPU hardware (see Test Plan).

## Review order

1. `src/neural/backends/cuda/hip_compat.h` (new) -- the single file that knows HIP. Force-included into every HIP `.cu` and pulled in by `cuda_common.h` under `USE_HIP`, it aliases the `cuda*`/`cublas*` spellings to `hip*`/`hipblas*` so the rest of the backend keeps CUDA spelling. It also (a) defines `CUDART_VERSION` to a pre-11.0 value so the NVIDIA-only L2-persistence and CUDA-graph external-event blocks compile out while `showInfo()`'s arithmetic still works, (b) wraps `hipblasGemmStridedBatchedEx` to translate the v2 compute-type (lc0 passes `CUDA_R_16F/32F` where hipBLAS wants `HIPBLAS_COMPUTE_16F/32F`), (c) wraps `cublasHgemm`/`HgemmBatched` to reinterpret `__half*` as hipBLAS's `hipblasHalf`, and (d) defines `__CUDA_ARCH__` for the HIP device pass (see the fp16 paragraph below).
2. `meson.build` / `meson_options.txt` -- a parallel `-Dhip` build branch that reuses the existing `-Dsycl=amd` hipBLAS/amdhip64/amd_gfx discovery and compiles `common_kernels.cu` + `fp16_kernels.cu` through hipcc custom_targets. Arch is baked from `-Damd_gfx` (autodetected via `rocm_agent_enumerator`, default gfx90a), so other AMD GPU architectures need only `-Damd_gfx` changed.
3. The kernel/host edits, all behind `USE_HIP` except the arch-unified layer_norm barrier fix, so the NVIDIA build is byte-identical.

## fp16 conv-SE kernels (the SE-ResNet path)

`fp16_kernels.cu` and the fp16 bodies in `winograd_helper.inc` gate their device code on `__CUDA_ARCH__ >= 530` (NVIDIA's "this GPU has native fp16"), and skip the bodies when `__CUDA_ARCH__ < 530`. hipcc never defines `__CUDA_ARCH__` (it uses `__HIP_DEVICE_COMPILE__`), so on HIP those guards made the fp16 squeeze-excitation and fused conv-transform kernels compile to empty bodies -- they launched, wrote nothing, and left SE-scaled output uninitialized, so any fp16 conv-residual-SE net (the dominant historical Leela architecture, `kUseFusedSELayer` on by default) silently produced wrong evals on hip-fp16. Fixed in `hip_compat.h` by defining `__CUDA_ARCH__` for the device pass; gfx90a has native fp16 + MFMA so those bodies are correct to compile in. The define is gated on `__HIP_DEVICE_COMPILE__` and placed after `<hip/hip_runtime.h>` on purpose: HIP's `hip_common.h` derives `__HIP_DEVICE_COMPILE__` from `__CUDA_ARCH__`, so an unconditional or earlier define would wrongly mark the host pass as a device compile. The only `__CUDA_ARCH__` readers reachable from the HIP-compiled translation units are exactly these fp16 capability gates, so the define activates no NVIDIA-specific path.

## Three wave64 correctness fixes (gfx90a is a 64-lane wavefront; NVIDIA is 32)

- `__shfl_*_sync` mask. ROCm requires a 64-bit mask and asserts at runtime that the mask equals the active lane set (`mask == __ballot(true)`). A literal full mask faults whenever a block is not a whole multiple of 64 lanes -- which is what crashed odd batch sizes >= 5, because layer_norm launches (32,1,z) blocks whose last wavefront is half-populated. Fixed by using `__activemask()` as the mask on HIP (CUDA keeps its `0xFFFFFFFF` literal).
- `softmax_opt_64_kernel` broadcast. A 32-lane warp reduction followed by a broadcast of lane 0 of the whole wavefront fed lanes 32-63 (a different row) the wrong row's max/sum on wave64. Fixed with a width-32 subgroup broadcast (read lane 0 of the caller's 32-lane subgroup); a no-op on wave32.
- layer_norm divergent `__syncthreads()`. The `if (n >= N) return` early-return left padding rows out of the block-wide barrier in `shared_sum_for_layer_norm`; on wave64 a padding row shares a wavefront with a valid row, making it a partial-wavefront barrier. Folded `n>=N` into the existing oobThread flag so all threads reach the barriers (padding rows own their own shared slot, so the result is identical on NVIDIA).

The capability gating, which decided FP16 support and tensor-core math from CUDA SM numbers and threw on gfx90a, gets a `USE_HIP` branch that enables the FP16/MFMA path; `cublasSetMathMode` (no hipBLAS peer) is guarded out at both call sites; CUTLASS fused-MHA stays off.

`README.md` documents the backend in lc0's per-backend style (the `-Dhip` option, the `hip` / `hip-fp16` / `hip-auto` backends, ROCm prerequisites, arch selection, and the `-Db_lto=false` requirement).

## Test Plan (real GPU, ROCm 7.2.1)

Validated on Linux on gfx90a (MI250X, CDNA2, wave64) and gfx1100 (RDNA3, wave32). Two nets: a conv-residual-SE net (maia-1100: NETWORK_SE, 6 SE blocks, 64 filters, conv policy -- exercises the fp16 SE / conv-transform kernels above) and a 256x10 attention-body net (the cuBLAS attention fallback path).

```
meson setup build-hip -Dhip=true -Damd_gfx=gfx90a -Dplain_cuda=false \
  -Dcudnn=false -Dcutlass=false -Dnvcc=false -Dgtest=true -Dblas=true \
  -Dopencl=false -Donnx=false -Db_lto=false \
  -Dhip_libdirs=/opt/rocm/lib -Dhip_include=/opt/rocm/include
ninja -C build-hip -j 16

# conv-residual-SE net, fp32 hip vs the trusted CPU blas backend
./build-hip/lc0 backendbench --backend=check \
  --backend-opts="hip(),blas(),mode=check,atol=1e-3,rtol=1e-2,freq=1.0" \
  --weights=maia-1100.pb.gz --start-batch-size=1 --max-batch-size=55 --batches=4
# -> 222/222 "Check passed", 0 failures (SE/conv-transform kernels exact vs blas in fp32)

# same conv-SE net, fp16 hip-fp16 vs blas (fp16-appropriate tolerance)
./build-hip/lc0 backendbench --backend=check \
  --backend-opts="hip-fp16(),blas(),mode=check,atol=1.1e-1,rtol=2e-1,freq=1.0" \
  --weights=maia-1100.pb.gz --start-batch-size=1 --max-batch-size=55 --batches=4
# -> 222/222 pass; value within fp16 envelope, matching bestmoves

# attention-body net, fp32 and fp16 vs blas
./build-hip/lc0 backendbench --backend=check \
  --backend-opts="hip(),blas(),mode=check,atol=1e-3,rtol=1e-2,freq=1.0" \
  --weights=attn.pb.gz --start-batch-size=1 --max-batch-size=32 --batches=4
# -> 130/130 passed, 0 failures
./build-hip/lc0 backendbench --backend=check \
  --backend-opts="hip-fp16(),blas(),mode=check,atol=2.5e-2,rtol=1e-1,freq=1.0" \
  --weights=attn.pb.gz --start-batch-size=1 --max-batch-size=32 --batches=4
# -> 130/130 passed, 0 failures

# benchmark sweep batch 1-256, fp32 and fp16, fault-free
./build-hip/lc0 backendbench --backend=hip --weights=maia-1100.pb.gz --batches=3
./build-hip/lc0 backendbench --backend=hip-fp16 --weights=maia-1100.pb.gz --batches=3

# CPU gtest regression set stays green
meson test -C build-hip            # 8/8 OK
```

fp32 hip matches blas exactly on both nets on gfx90a and gfx1100; fp16 is within fp16 tolerance with matching policy argmax; eval/PV are identical run-to-run (no reduction race); named lc0 kernels confirmed dispatching on device via `AMD_LOG_LEVEL=3`; gtest 8/8 OK on both architectures. The NVIDIA build is unchanged (every edit is behind `USE_HIP` / the `-Dhip` meson option, except the layer_norm barrier fix which is identical on NVIDIA).


src/neural/backends/cuda/common_kernels.cu
```diff
@@ -935,9 +938,17 @@ __global__ void layer_norm_kernel(int N, int C, T* output, const T* input,
...
   int c = (threadIdx.y * 32 + threadIdx.x) * 16;
-  bool oobThread = c >= C;
+  // An out-of-range row (n >= N) must NOT early-return: shared_sum_for_layer_norm
+  // calls __syncthreads(), and on a 64-lane wavefront two threadIdx.z rows share
+  // one wavefront, so returning the padding row while its partner survives is a
+  // partial-wavefront barrier -> GPU fault on AMD (it is benign on NVIDIA only
+  // because whole 32-lane warps exit). Instead fold it into oobThread so the
+  // padding row skips every load/store (each guarded by !oobThread) yet still
+  // reaches both barriers. Padding rows own their own sum[threadIdx.z] slot, so
+  // valid rows are never corrupted. Arch-unified: identical results on NVIDIA.
```
Menkib64, line 957: There are other places too where extra threads do early return. Why is this only place that causes a fault?
Is there an alternative to __syncthreads() which would only wait for active thread mask?
I remember that hardware supports waiting based on active thread mask. I don't remember how to map to use generate those instructions.
RDNA and CDNA documentation have following text:
> If a wave executes an S_BARRIER before all of the waves of the work-group have been created, the wave waits
until the work-group is complete.
Any wave may terminate early using S_ENDPGM, and the barrier is considered satisfied when the remaining
live waves reach their barrier instruction

meson.build
```diff
@@ -609,6 +609,86 @@ if get_option('build_backends')
...
+        agents = run_command(agent_enum, check : false).stdout()
+        foreach agent : agents.splitlines()
+          if hip_gfx == '' and agent.startswith('gfx') and agent != 'gfx000'
+            hip_gfx = agent
+          endif
+        endforeach
+      endif
+      if hip_gfx == ''
+        hip_gfx = 'gfx90a'
+        warning('Could not autodetect AMD GPU arch; defaulting to gfx90a. Set -Damd_gfx to override.')
```
Menkib64: I think failure to detect an architecture should error or build for multiple generic architectures (eg. `gfx9-generic`, `gfx11-generic`).

meson.build
```diff
@@ -609,6 +609,86 @@ if get_option('build_backends')
...
+    hipcc_args = ['--offload-arch=' + hip_gfx, '-x', 'hip', '-std=c++17'] + hipcc_fpic + [
+                  '-O3', '-DUSE_HIP', '-D__HIP_PLATFORM_AMD__',
+                  '-I', meson.current_source_dir() + '/src',
+                  '-I', meson.current_source_dir() + '/third_party',
+                  '-include', hipcc_compat]
+    hipcc_args += hip_args
+    hipcc_io_args = ['-c', '@INPUT@', '-o', '@OUTPUT@']
+
+    cuda_files = ['src/neural/backends/cuda/layers.cc',
+                  'src/neural/backends/cuda/network_cuda.cc']
```
Menkib64, line 682: Is it possible to compile both hip and cuda backends at the same time?
It should be possible to compile both or hip backend will broken all the time.

README.md
```diff
@@ -51,7 +51,7 @@ Given those basics, the OS and backend specific instructions are below.
 
 1. Install backend (also read the detailed instructions in later sections):
     - If you want to use NVidia graphics cards Install [CUDA](https://developer.nvidia.com/cuda-zone) (and optionally [cuDNN](https://developer.nvidia.com/cudnn)).
-    - If you want to use AMD or Intel graphics cards you can try SYCL.
```
Menkib64, line 54: This line should be preserved for Intel only.

src/neural/backends/cuda/hip_compat.h
```diff
@@ -0,0 +1,262 @@
...
+// itself derives __HIP_DEVICE_COMPILE__ from __CUDA_ARCH__ in hip_common.h, so an
+// unconditional or pre-include define would wrongly mark the HOST pass as a device
+// compile. By the time we get here HIP has already set __HIP_DEVICE_COMPILE__, so
+// gating on it confines __CUDA_ARCH__ to the device pass. The only __CUDA_ARCH__
+// readers reachable from the two HIP-compiled .cu are exactly these fp16 gates (no
+// PTX, sm-specific intrinsic, or other arch branch), so this does not activate any
+// NVIDIA-only path.
+#if defined(__HIP_DEVICE_COMPILE__) && __HIP_DEVICE_COMPILE__
+#ifndef __CUDA_ARCH__
+#define __CUDA_ARCH__ 800
```
Menkib64: I think the better way would be to change `SKIP_FP16_BITS` define to `HAS_FP16_SUPPORT` and inverse the  `ifndef` to `ifdef` check. Then hip could set it enabled unconditionally because fp16 support was added in GCN 5.0. I don't think that we are gone support older cards using hip.

src/neural/backends/cuda/network_cuda.cc
```diff
@@ -343,11 +356,15 @@ class CudaNetwork : public Network {
     }
 
     bool use_fused_mha = false;
+#if !defined(USE_HIP)
+    // The fused-MHA kernel is CUTLASS-only (USE_CUTLASS), which does not build on
+    // ROCm; HIP always takes the cuBLAS attention fallback, so keep this false.
```
Menkib64, line 362: Would there be an easy way to map cutlass code to composable kernel code?
Transformers would need better fused kernels that can avoid kernel transition costs. Nvidia side would likely use cutlass to implement custom fused matrix multiplication kernels.

meson.build
```diff
@@ -609,6 +609,86 @@ if get_option('build_backends')
...
+        hip_gfx = 'gfx90a'
+        warning('Could not autodetect AMD GPU arch; defaulting to gfx90a. Set -Damd_gfx to override.')
+      endif
+      message('HIP target architecture: ' + hip_gfx)
+    endif
+
+    hipcc_compat = meson.current_source_dir() + '/src/neural/backends/cuda/hip_compat.h'
+    # -fPIC is invalid for x86_64-pc-windows-msvc (clang rejects it on Windows)
+    hipcc_fpic = host_machine.system() != 'windows' ? ['-fPIC'] : []
+    hipcc_args = ['--offload-arch=' + hip_gfx, '-x', 'hip', '-std=c++17'] + hipcc_fpic + [
```
Menkib64: It should be `-std=c++20` but nvcc had some issues enabling it consistently on all host compilers.

Menkib64 (commented):
What version of ROCm was used for testing?

I tested automatic conversion from cuda to hip using hipify a few months ago. The conversion crashed en external events because some internal event state pointer was null.
Also fp16 mode convoluting networks (old networks like T60) produced garbage output. That might have been an error in my conversion attempt.

Menkib64:
> What version of ROCm was used for testing?
> 

It might help to read carefully the commit message too. The testing was done using 7.2.1.

I decided try test it locally. I noticed that 7.2 hipcc ldd links to old libxml2 which isn't available in my recently upgraded system. I choose to install 7.13 preview packages to test it.

My first compile attempt hit a few conflicts between hip and cuda headers. I have to disable cuda to test hip. 

```$ ./lc0 backendbench -b hip-auto --max-batch-size=16 -w ../../BT4-tf13.pb.gz 
       _
|   _ | |
|_ |_ |_| v0.33.0-dev+git.d83b6d17 built Jun 13 2026
Loading weights file from: ../../BT4-tf13.pb.gz
Weights file has multihead format, updating format flag
Switching to [cuda-fp16]...
CUDA Runtime version: 71399.0.4
WARNING: CUDA Runtime version mismatch, was compiled with version 10.2.0
Latest version of CUDA supported by the driver: 71399.0.4
GPU: AMD Radeon Graphics
GPU memory: 15.0198 Gb
GPU clock frequency: 2799 MHz
GPU compute capability: 11.0
L2 cache capacity: 2097152
size, mean nps, mean ms,   sdev,     cv, max nps,  median, min nps, first max, first mean
   1,       35,   28.93, 0.8742, 0.0302,      35,      35,      27,   29.3517,    29.3517
   2,       71,   28.26, 0.7504, 0.0266,      74,      70,      68,   29.2424,    29.2424
   3,      108,   27.69, 0.2754, 0.0099,     111,     108,     106,   27.9299,    27.9299
   4,      144,   27.86, 0.2529, 0.0091,     147,     143,     141,   28.3749,    28.3749
   5,      157,   31.82, 0.1639, 0.0052,     158,     157,     155,   32.0573,    32.0573
   6,      178,   33.77, 0.2705, 0.0080,     181,     178,     174,   33.4783,    33.4783
   7,      171,   41.05, 0.3744, 0.0091,     173,     171,     166,   41.7865,    41.7865
   8,      186,   42.92, 0.2751, 0.0064,     189,     187,     183,   43.0242,    43.0242
   9,      178,   50.49, 0.4833, 0.0096,     181,     178,     175,   49.8067,    49.8067
  10,      190,   52.53, 0.6331, 0.0121,     193,     191,     172,   53.5222,    53.5222
  11,      185,   59.59, 0.2673, 0.0045,     187,     185,     183,   59.6822,    59.6822
  12,      174,   68.99, 0.4085, 0.0059,     176,     174,     169,   70.6075,    70.6075
  13,      166,   78.53, 0.4487, 0.0057,     167,     166,     162,   79.0430,    79.0430
HW Exception by GPU node-1 (Agent handle: 0x59c8e0f3eb40) reason :GPU Hang
```
There are a few string and version checks which would need to be improved.  The GPU hand at the end is something random going wrong. I don't know if it is related to random calculation errors which I noticed last autumn. Performance looks fine but it is a little slower than `onnx-migraphx`.  Too bad `onnx-migraphx` doesn't work with variable batch size.

jeffdaily:
Pushed a80a7be addressing the review: a hard error on arch-autodetect failure, -std=c++20 for hipcc, the README SYCL line scoped to Intel with AMD pointed at the HIP backend, and the SKIP_FP16_BITS -> HAS_FP16_SUPPORT inversion (which also removed the __CUDA_ARCH__ shim).

Testing was on ROCm 7.2.1. Re-validated on gfx90a (MI250X): gtest 8/8, and hip / hip-fp16 versus the CPU reference pass on a conv-residual-SE net and an attention net in both fp32 and fp16, with batch 1-256 running fault-free.

If you can share the exact header-conflict errors you saw with the ROCm 7.13 preview, that would help; I can try to reproduce them here.

src/neural/backends/cuda/common_kernels.cu
jeffdaily: The ISA text you quoted covers a whole wave terminating early via S_ENDPGM: once every lane of a wave has exited, that wave is dropped from the barrier's wait set. That case is handled and is fine here. The layer_norm fault is a different situation, intra-wave divergence rather than whole-wave termination.

The kernel is launched with blockDim = (32, 1, z), so one threadIdx.z row is 32 lanes. On a 64-lane wave two z-rows share a single wave. When a padding row (n >= N) hits `if (n >= N) return;` while its partner row (n < N) keeps going, only half the lanes of a still-live wave leave. The wave never executes S_ENDPGM, so the barrier still expects it, and it arrives at S_BARRIER with a partial exec mask. That is the divergent-barrier hazard. On NVIDIA each 32-lane z-row is a whole warp, so a row either all-returns or all-continues (warp granularity), which is why it is benign there and only shows up on wave64.

On why this is the only faulting site: I went through every __syncthreads in the HIP-compiled files (common_kernels.cu, fp16_kernels.cu, winograd_helper.inc). The other early-return kernels, e.g. `if (k >= C) return;` in OutputTransform_relu_InputTransform_kernel, have no barrier after the return, so an early-returning lane never abandons a barrier. Every barrier-containing kernel that has no early return (softmax_kernel, promotion_logits_kernel, the SE kernels, the winograd SE blocks) is reached by all launched threads, and the SE barriers sit inside `if (use_se)`, which is a template parameter and so block-uniform. layer_norm was the only barrier downstream of a per-row early return whose row granularity (32) is smaller than the wave (64).

On an active-mask barrier: there is no workgroup barrier that waits on a lane mask; S_BARRIER is wave-level, one arrival per wave, not per lane. The portable fix is to keep control flow to the barrier uniform, which is what folding the early return into the oobThread guard does (the padding row owns its own sum[threadIdx.z] slot, so it does not corrupt the valid row, and the result is identical on NVIDIA).

meson.build
jeffdaily: Done in a80a7be. Autodetect failure is now a hard `error()` that tells the user to set -Damd_gfx explicitly, instead of silently defaulting to gfx90a.

meson.build
jeffdaily: Not at the moment, and it is more than a build-file change. Both backends compile the same translation units (network_cuda.cc, layers.cc, common_kernels.cu, fp16_kernels.cu) and register the same backend factory symbols, so building both into one binary would collide. Separating them cleanly needs a namespacing refactor of the shared cuda backend, which seems better left to the maintainers than done as part of this port.

Worth noting for keeping it from breaking: the HIP path can be compile-checked by anyone with hipcc installed, since it only needs AMD hardware to run, not to build.

README.md
jeffdaily: Done in a80a7be. The SYCL bullet is now scoped to Intel, and AMD points at the HIP backend.

src/neural/backends/cuda/hip_compat.h
jeffdaily: Done in a80a7be. Renamed to HAS_FP16_SUPPORT with `#ifdef` guards. The NVIDIA build defines it for compute capability >= 5.3, which is exactly where SKIP_FP16_BITS was previously absent, so the nvcc preprocessed result is unchanged; HIP defines it unconditionally. This also let me drop the `#define __CUDA_ARCH__ 800` shim, since those fp16 gates were its only reader.

meson.build
jeffdaily: Done in a80a7be. hipcc now uses -std=c++20; only the nvcc path stays on c++17.

Menkib64:
Thanks.

S_BARRIER must be reached by all lanes in a wave. I guess compiler cannot enforce it because S_BARRIER could be inside a function call if a kernel uses functions for some reason.

I tried to quickly find the hip documentation which would give the  `__syncthread` must be called by all lanes requirement. I failed to find it. I found that cooperate groups have more flexible thread synchronization options. [It](https://rocm.docs.amd.com/projects/HIP/en/latest/reference/hip_runtime_api/modules/cooperative_groups_reference.html) didn't document all lanes requirement but I suspect it would have similar requirement.

RNDA should be able to execute both wave 64 and wave 32 code. It looks like experimental support might be there but [documentation](https://rocm.docs.amd.com/projects/HIP/en/latest/reference/hardware_features.html) claims HIP runtime won't support it. I wonder if there is a way for me to test wave 64 mode on RDNA.
> The RDNA architectures feature an experimental compiler option called mwavefrontsize64, which determines the wavefront size for kernel code generation. When this option is disabled, the native wavefront size of 32 is used, when enabled wavefront size 64 is used. This option is not supported by the HIP runtime.

If testing it is going to be hard, I'm thinking that `__syncthreads` will require a wrapper which validates the requirement in debug builds to avoid future surprise bugs caused by it.

My compiler error was about CUDA headers being included because compiler command included cuda include paths. There was many redefinition errors. A quick fix was to remove cuda include path which made compile work fine.

As for compiling cuda together with HIP. I think I know a simple way to make both compile at the same time. This would make it more likely that potential future breakage will be caught earlier.

It is late for me. I will check and test new changes tomorrow.

Menkib64:
I experiment building both hip and cuda together. It does work fairly small build changes. I pushed the experimental idea to https://github.com/Menkib64/lc0/tree/hip_shared_backend
I also tested running them at the same time using `check` backend. This run uses `hip-auto` as the main backend while `cuda-auto` is used as the reference. The errors are expected magnitude for two different fp16 implementations.
```
$ ninja && ./lc0 -b check -o mode=display,freq=1.0,hip-auto,cuda-auto -w ../../t1-256x10-distilled-swa-2432500.pb.gz -v
ninja: no work to do.
       _
|   _ | |
|_ |_ |_| v0.33.0-dev+git.dirty built Jul  3 2026
Search algorithm: classic
go nodes 1000
Loading weights file from: ../../t1-256x10-distilled-swa-2432500.pb.gz
Working backend set to cuda-auto.
Reference backend set to hip-auto.
Creating backend [cuda-auto]...
Switching to [cuda-fp16]...
CUDA Runtime version: 13.3.0
Latest version of CUDA supported by the driver: 13.3.0
Compiled with CUTLASS enabled
GPU: NVIDIA GeForce RTX 5080
GPU memory: 15.5084 Gb
GPU clock frequency: 2670 MHz
GPU compute capability: 12.0
L2 cache capacity: 67108864
Creating backend [hip-auto]...
Switching to [hip-fp16]...
HIP Runtime version: 71399.0.4
Latest version of HIP supported by the driver: 71399.0.4
GPU: AMD Radeon Graphics
GPU memory: 15.0198 Gb
GPU clock frequency: 2799 MHz
GPU compute capability: 11.0
L2 cache capacity: 2097152
Check mode: error display.
Check rate: 100%.
maximum error for a batch of 1:
  value: absolute: 7.2e-04, relative: 1.8e-02.
  policy: absolute: 9.1e-04, relative: 8.5e-03.
maximum error for a batch of 20:
  value: absolute: 2.8e-03, relative: 1.7e-01.
  policy: absolute: 5.8e-03, relative: 7.8e-02.
info depth 1 seldepth 2 time 1662 nodes 2 score cp 16 nps 62 tbhits 0 pv d2d4 d7d5
maximum error for a batch of 19:
  value: absolute: 2.4e-03, relative: 1.5e-01.
  policy: absolute: 8.2e-03, relative: 7.7e-02.
info depth 1 seldepth 2 time 1686 nodes 6 score cp 16 nps 107 tbhits 0 pv g1f3 d7d5
maximum error for a batch of 20:
  value: absolute: 4.8e-03, relative: 2.0e-02.
  policy: absolute: 7.0e-03, relative: 4.2e-02.
info depth 2 seldepth 3 time 1706 nodes 8 score cp 12 nps 106 tbhits 0 pv d2d4 d7d5 c2c4
maximum error for a batch of 28:
  value: absolute: 5.2e-03, relative: 1.9e-01.
  policy: absolute: 8.0e-03, relative: 3.6e-02.
maximum error for a batch of 16:
  value: absolute: 2.0e-03, relative: 2.2e-01.
  policy: absolute: 5.6e-03, relative: 3.2e-02.
info depth 2 seldepth 4 time 1741 nodes 15 score cp 14 nps 136 tbhits 0 pv d2d4 d7d5 c2c4 c7c6
maximum error for a batch of 29:
  value: absolute: 2.4e-03, relative: 1.2e-01.
  policy: absolute: 1.2e-02, relative: 8.2e-02.
maximum error for a batch of 19:
  value: absolute: 2.0e-03, relative: 1.1e-01.
  policy: absolute: 8.4e-03, relative: 6.6e-02.
info depth 2 seldepth 5 time 1774 nodes 19 score cp 15 nps 131 tbhits 0 pv d2d4 d7d5 c2c4 c7c6 c4d5
maximum error for a batch of 28:
  value: absolute: 2.3e-03, relative: 9.0e-02.
  policy: absolute: 4.1e-03, relative: 5.2e-02.
info depth 3 seldepth 5 time 1792 nodes 27 score cp 14 nps 167 tbhits 0 pv d2d4 d7d5 c2c4 d5c4 e2e3
maximum error for a batch of 8:
  value: absolute: 2.2e-03, relative: 6.4e-01.
  policy: absolute: 4.2e-03, relative: 3.5e-02.
info depth 3 seldepth 6 time 1800 nodes 32 score cp 15 nps 189 tbhits 0 pv d2d4 d7d5 c2c4 d5c4 e2e3
maximum error for a batch of 10:
  value: absolute: 1.6e-03, relative: 7.3e-01.
  policy: absolute: 3.4e-03, relative: 3.5e-02.
maximum error for a batch of 3:
  value: absolute: 2.8e-03, relative: 2.6e-02.
  policy: absolute: 3.6e-03, relative: 2.6e-02.
maximum error for a batch of 25:
  value: absolute: 2.7e-03, relative: 4.7e-01.
  policy: absolute: 5.3e-03, relative: 4.8e-02.
maximum error for a batch of 32:
  value: absolute: 2.0e-03, relative: 4.0e-01.
  policy: absolute: 1.0e-02, relative: 4.3e-02.
info depth 4 seldepth 7 time 1850 nodes 102 score cp 16 nps 465 tbhits 0 pv d2d4 d7d5 c2c4 d5c4 e2e3 a7a6
maximum error for a batch of 27:
  value: absolute: 2.8e-03, relative: 1.8e+00.
  policy: absolute: 5.4e-03, relative: 6.2e-02.
maximum error for a batch of 32:
  value: absolute: 3.1e-03, relative: 2.4e-01.
  policy: absolute: 6.1e-03, relative: 6.2e-02.
info depth 4 seldepth 8 time 1886 nodes 163 score cp 16 nps 636 tbhits 0 pv d2d4 d7d5 c2c4 c7c6 c4d5 c6d5 c1f4
maximum error for a batch of 35:
  value: absolute: 4.6e-03, relative: 2.0e-01.
  policy: absolute: 5.2e-03, relative: 4.8e-02.
maximum error for a batch of 39:
  value: absolute: 2.7e-03, relative: 1.9e-01.
  policy: absolute: 6.6e-03, relative: 5.2e-02.
info depth 4 seldepth 9 time 1932 nodes 276 score cp 17 nps 916 tbhits 0 pv d2d4 d7d5 c2c4 c7c6 c4d5 c6d5 c1f4
maximum error for a batch of 30:
  value: absolute: 2.7e-03, relative: 1.9e-01.
  policy: absolute: 8.0e-03, relative: 5.5e-02.
maximum error for a batch of 118:
  value: absolute: 3.4e-03, relative: 8.7e-01.
  policy: absolute: 9.0e-03, relative: 8.6e-02.
info depth 5 seldepth 10 time 2014 nodes 472 score cp 16 nps 1232 tbhits 0 pv d2d4 d7d5 c2c4 d5c4 g1f3 g8f6 e2e3 a7a6
maximum error for a batch of 64:
  value: absolute: 3.6e-03, relative: 7.1e-01.
  policy: absolute: 1.3e-02, relative: 8.8e-02.
maximum error for a batch of 139:
  value: absolute: 2.9e-03, relative: 1.4e+00.
  policy: absolute: 9.9e-03, relative: 6.6e-02.
maximum error for a batch of 45:
  value: absolute: 3.2e-03, relative: 4.3e-01.
  policy: absolute: 9.3e-03, relative: 5.5e-02.
maximum error for a batch of 73:
  value: absolute: 3.1e-03, relative: 4.2e-01.
  policy: absolute: 1.1e-02, relative: 6.4e-02.
maximum error for a batch of 29:
  value: absolute: 3.9e-03, relative: 2.8e-01.
  policy: absolute: 9.3e-03, relative: 5.5e-02.
maximum error for a batch of 55:
  value: absolute: 2.7e-03, relative: 1.3e-01.
  policy: absolute: 1.0e-02, relative: 7.6e-02.
info depth 5 seldepth 11 time 2241 nodes 985 score cp 16 nps 1612 tbhits 0 pv d2d4 d7d5 c2c4 c7c6 c4d5 c6d5 c1f4 g8f6 e2e3
info string g2g4  (378 ) N:       1 (+ 0) (P:  0.87%) (WL: -0.31490) (D: 0.419) (M: 156.8) (Q: -0.31490) (U: 0.25090) (S: -0.06155) (V: -0.3124) 
info string b1a3  (34  ) N:       1 (+ 0) (P:  0.90%) (WL: -0.19756) (D: 0.489) (M: 167.5) (Q: -0.19756) (U: 0.25837) (S:  0.05871) (V: -0.1997) 
info string f2f3  (346 ) N:       1 (+ 0) (P:  0.97%) (WL: -0.18381) (D: 0.493) (M: 169.0) (Q: -0.18381) (U: 0.27848) (S:  0.09274) (V: -0.1857) 
info string g1h3  (161 ) N:       1 (+ 0) (P:  1.08%) (WL: -0.14240) (D: 0.527) (M: 173.4) (Q: -0.14240) (U: 0.31068) (S:  0.16908) (V: -0.1416) 
info string h2h4  (403 ) N:       2 (+ 0) (P:  1.11%) (WL: -0.15427) (D: 0.522) (M: 173.2) (Q: -0.15427) (U: 0.21386) (S:  0.06031) (V: -0.1097) 
info string b2b4  (234 ) N:       2 (+ 0) (P:  1.27%) (WL: -0.10774) (D: 0.540) (M: 180.2) (Q: -0.10774) (U: 0.24331) (S:  0.13569) (V: -0.0516) 
info string f2f4  (351 ) N:       2 (+ 0) (P:  1.26%) (WL: -0.10121) (D: 0.511) (M: 172.1) (Q: -0.10121) (U: 0.24126) (S:  0.14102) (V: -0.0728) 
info string a2a4  (207 ) N:       6 (+ 0) (P:  2.13%) (WL: -0.04384) (D: 0.570) (M: 182.3) (Q: -0.04384) (U: 0.17508) (S:  0.12947) (V: -0.0143) 
info string h2h3  (400 ) N:      11 (+ 0) (P:  3.03%) (WL: -0.01555) (D: 0.590) (M: 184.3) (Q: -0.01555) (U: 0.14561) (S:  0.12861) (V:  0.0082) 
info string d2d3  (288 ) N:      11 (+ 0) (P:  3.21%) (WL: -0.01939) (D: 0.556) (M: 180.1) (Q: -0.01939) (U: 0.15393) (S:  0.13439) (V:  0.0167) 
info string a2a3  (204 ) N:      11 (+ 0) (P:  3.14%) (WL: -0.01512) (D: 0.581) (M: 183.3) (Q: -0.01512) (U: 0.15063) (S:  0.13446) (V:  0.0022) 
info string c2c3  (259 ) N:      19 (+ 0) (P:  3.79%) (WL:  0.01587) (D: 0.558) (M: 178.4) (Q:  0.01587) (U: 0.10911) (S:  0.12513) (V:  0.0251) 
info string b1c3  (36  ) N:      22 (+ 0) (P:  4.37%) (WL:  0.01187) (D: 0.580) (M: 179.3) (Q:  0.01187) (U: 0.10948) (S:  0.12212) (V:  0.0210) 
info string b2b3  (230 ) N:      22 (+ 0) (P:  4.32%) (WL:  0.01604) (D: 0.546) (M: 177.6) (Q:  0.01604) (U: 0.10818) (S:  0.12404) (V:  0.0436) 
info string e2e3  (317 ) N:      53 (+ 0) (P:  7.18%) (WL:  0.03792) (D: 0.585) (M: 185.2) (Q:  0.03792) (U: 0.07659) (S:  0.11681) (V:  0.0523) 
info string e2e4  (322 ) N:      82 (+ 0) (P:  9.54%) (WL:  0.04581) (D: 0.575) (M: 181.5) (Q:  0.04581) (U: 0.06625) (S:  0.11411) (V:  0.0493) 
info string g2g3  (374 ) N:      85 (+ 0) (P:  9.07%) (WL:  0.05319) (D: 0.585) (M: 184.8) (Q:  0.05319) (U: 0.06075) (S:  0.11625) (V:  0.0668) 
info string c2c4  (264 ) N:      96 (+ 0) (P:  9.06%) (WL:  0.05294) (D: 0.578) (M: 183.3) (Q:  0.05294) (U: 0.05384) (S:  0.10894) (V:  0.0596) 
info string g1f3  (159 ) N:     253 (+ 1) (P: 15.67%) (WL:  0.06958) (D: 0.576) (M: 184.0) (Q:  0.06958) (U: 0.03542) (S:  0.10681) (V:  0.0745) 
info string d2d4  (293 ) N:     303 (+ 0) (P: 18.05%) (WL:  0.07219) (D: 0.575) (M: 183.1) (Q:  0.07219) (U: 0.03421) (S:  0.10811) (V:  0.0731) 
info string node  (  20) N:     985 (+ 1) (P: 100.0%) (WL:  0.05384) (D: 0.575) (M: 184.0) (Q:  0.05384) (V:  0.0389) 
bestmove d2d4 ponder d7d5
maximum error for a batch of 9:
  value: absolute: 3.1e-03, relative: 6.7e-02.
  policy: absolute: 5.3e-03, relative: 4.6e-02.
quit
```

Menkib64 (approved):

meson.build
jeffdaily: Done in 223ee639. I pulled the four commits from your hip_shared_backend branch onto the PR branch (kept as your commits): the NS_BACKEND namespace split (cudnn_backend vs hip_backend) plus building each backend as its own static_library so -DUSE_HIP / -DUSE_CUDNN / -DUSE_CUTLASS stay per-backend, then linking the extracted objects. Both backends now build into one binary; with -Dhip=true -Dplain_cuda=true -Dnvcc=true the --backend list has hip, cuda, hip-fp16, cuda-fp16, hip-auto, cuda-auto, and the single-backend builds are unchanged. Verified the co-build on gfx90a plus nvcc 12.6. One note: extract_objects() on the generated object targets wants meson >= 0.61 but project() declares >= 0.60, so 0.60 emits a feature-version warning; may be worth bumping the minimum.

src/neural/backends/cuda/common_kernels.cu
jeffdaily: Added the debug wrapper you suggested, in 223ee639: lc0SyncThreads() replaces __syncthreads() at the barrier sites and, in a HIP debug build, asserts that every lane the block placed in the wave is still active before the barrier (computed from blockDim/warpSize, so a legitimately partial final wave is allowed). No-op on CUDA and in release.

src/neural/backends/cuda/network_cuda.cc
jeffdaily: I looked at mapping the HIP fused-MHA path onto Composable Kernel (ck_tile fmha) and benchmarked it before deciding; I'm keeping the cuBLAS fallback. First, lc0's smolgen additive attention bias isn't supported by the available CK flash interface -- flash_attn rejects a non-null mask, and the low-level ck_tile fmha_fwd that does have a bias_ptr ships declared-only, so using it means running CK's codegen and vendoring the generated instances, a much larger dependency than the CUDA CUTLASS subproject. Second, I benchmarked CK flash against the hipBLAS 3-step at lc0's shape (S=64, head_dim 32/64, fp16) even without the bias: it is about 1.9x slower at batch 1-8, roughly even at 32, and only wins (~2.3x) at batch 128-256. Search inference is dominated by small batches, so the fused path would cost latency in the common case and help only at the high end. So it is a real dependency cost for a bias-unsupported, batch-limited win; the cuBLAS attention stays and matches the blas reference within tolerance across batch 1-256. Worth revisiting if CK ships a bias-capable fmha as a prebuilt instance.

PR #2423, opened 2026-06-28
Add opening seed control for selfplay opening randomness

john-sp:
Adds a new selfplay flag, --opening-seed, to control randomness used by shuffled and random opening modes.

### Changes:
- Adds --opening-seed with default -1 to preserve existing nondeterministic behavior.
- Uses the seeded opening RNG for: --openings-mode=shuffled and --openings-mode=random
- Adds a non-breaking Random(int seed) constructor while leaving Random::Get() unchanged.
- Keeps unrelated tournament randomness on the existing global RNG.

### DIFF
--- a/src/selfplay/tournament.cc
+++ b/src/selfplay/tournament.cc
@@ -92,6 +92,9 @@ const OptionId kOpeningsMirroredId{
     "Not really compatible with openings mode random."};
 const OptionId kOpeningsModeId{"openings-mode", "OpeningsMode",
                                "A choice of sequential, shuffled, or random."};
+const OptionId kOpeningSeedId{
+    "opening-seed", "OpeningSeed",
+    "Seed for shuffled and random openings mode. Set to -1 for a random seed."};
 const OptionId kSyzygyTablebaseId{
     "syzygy-paths", "SyzygyPath",
     "List of Syzygy tablebase directories, list entries separated by system "
@@ -134,6 +137,7 @@ void SelfPlayTournament::PopulateOptions(OptionsParser* options) {
   std::vector<std::string> openings_modes = {"sequential", "shuffled",
                                              "random"};
   options->Add<ChoiceOption>(kOpeningsModeId, openings_modes) = "sequential";
+  options->Add<IntOption>(kOpeningSeedId, -1, 2147483647) = -1;
 
   options->Add<StringOption>(kSyzygyTablebaseId);
   SelfPlayGame::PopulateUciParams(options);
@@ -165,6 +169,7 @@ SelfPlayTournament::SelfPlayTournament(const OptionsDict& options,
                        options.GetSubdict("player1").GetSubdict("black")},
                       {options.GetSubdict("player2").GetSubdict("white"),
                        options.GetSubdict("player2").GetSubdict("black")}},
+      opening_random_(options.Get<int>(kOpeningSeedId)),
       uci_responder_(uci_responder),
       game_callback_(game_info),
       tournament_callback_(tournament_info),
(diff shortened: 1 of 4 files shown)


src/selfplay/tournament.cc
```diff
@@ -134,6 +137,7 @@ void SelfPlayTournament::PopulateOptions(OptionsParser* options) {
   std::vector<std::string> openings_modes = {"sequential", "shuffled",
                                              "random"};
   options->Add<ChoiceOption>(kOpeningsModeId, openings_modes) = "sequential";
+  options->Add<IntOption>(kOpeningSeedId, -1, 2147483647) = -1;
```
Menkib64: Seed should be unsigned integer. This would mean option should accept full range from `INT_MIN` to `INT_MAX`. Then the value can be cast to unsigned to use full 32bit unsigned range. Then option parsing needs to first check if option is set using `Exists(kOpeningSeedId)`.

src/selfplay/tournament.cc
```diff
@@ -185,7 +190,7 @@ SelfPlayTournament::SelfPlayTournament(const OptionsDict& options,
     book_reader.AddPgnFile(book);
     openings_ = book_reader.ReleaseGames();
     if (options.Get<std::string>(kOpeningsModeId) == "shuffled") {
-      Random::Get().Shuffle(openings_.begin(), openings_.end());
```
Menkib64, line 188: Random is supposed to be a singleton because mersenne twister uses too much memory. It is possible to do something like
```
if (options.Exists(kOpeningSeedId)) {
  Random::Get().Seed(static_cast<unsigned>(options.Get<int>(kOpeningSeedId)));
}
Random::Get().Shuffle(openings_.begin(), openings_.end());

// random.cc
void Random::Seed(unsigned seed) {
  eng_.seed(seed);
}
```

john-sp: I checked the other `Random::Get()` call sites, and I think seeding the singleton makes the behavior broader than this PR intends.

With the`Random::Get().Seed(...)` approach, `--opening-seed` does control shuffled/random openings, but it also makes later singleton draws deterministic in the same process, including:

- first-game color randomization in `SelfPlayTournament`
- discarded-start chance and discard-pile selection/pruning
- resign playthrough sampling
- `opening-stop-prob` early exit inside `SelfPlayGame`
- classic/dag search Dirichlet noise
- classic/dag root move sampling with temperature
- training data random directory suffixes if training output is written

So I agree with keeping `Random` as a singleton if we want to avoid another `mt19937`, but seeding the singleton directly seems to leak the opening seed into unrelated tournament/search randomness.

Menkib64: Good point.
A solution could be to allocate a temporary random engine in stack which takes the given seed.

john-sp: That sounds good. For shuffled that should be easy enough to implement.
One extra wrinkle is `--openings-mode=random`: right now it calls `Random::Get()` for each new game, so unlike shuffled mode there is not just one up-front operation to seed. I can think of 2 options off the top of my head: 1) Precompute a random opening order from the temporary seeded engine at tournament initialization, then consume that order during play. For endless selfplay or huge number of games (thinking 50k+), regenerate another batch/order when exhausted, likely using an extra uint32 as the seed for the next batch. 2) Use some stateless deterministic mapping from `(opening_seed, game_number)` to an opening index, avoiding a long-lived `mt19937`. I’d need to think a bit more about how to do that cleanly without introducing bias.

Menkib64: We could also use 64 bit linear congruential engine which has much smaller footprint. It should be comparable runtime performance when set correctly to avoid integer division. A potential good generation could be defined based on[ an old recommendation](https://nuclear.llnl.gov/CNP/rng/rngman/node4.html).

```
// https://nuclear.llnl.gov/CNP/rng/rngman/node4.html
using GeneratorType = std::linear_congruential_engine<uint64_t, 2862933555777941757ULL, 3037000493ULL, 0>;
```

I think it should be good enough quality for shuffling. If there is demand for higher quality random numbers with reduced memory footprint, there are other newer alternatives (bit manipulation of LCE or bit operations without Mersene memory requirement). Too bad they haven't made it into standard so they require more code. I think there are C++ libraries for some of them.

Menkib64 (approved):

Issue #2185, opened 2025-06-08
Save TensorRT engine in weights file for stable performance.


borg323:
When using the onnx-trt backend, the generated engine performance is not always the same. However there is an option to save a modified onnx model with the engine embedded, giving consistent performance. Details on how to do this are given [here](<https://onnxruntime.ai/docs/execution-providers/TensorRT-ExecutionProvider.html#tensorrt-ep-caches>). Such a generated onnx model would have to be packaged in a weights file using `onnx2leela`, but ideally this step should also be automated.

saithepaithewhyyy:
Hi!
I'd like to work on this. Can I take this up, considering I'm a year late?

As far as I understand, theres a couple of things that need to be done:
- Configure the onnx trt options to generate the embedded engine and package this in a new weights file
- Automate the whole process, so that during run with the onnx-trt backend (after all the other caches like timing caches, etc are created), the above step takes place. During subsequent runs, the new weights file is picked

Do correct me if I am wrong anywhere, I had a thorough read of the repo, and this was my understanding of the issue
Thanks!

Issue #2414, opened 2026-05-19
lc0 with ONNX backend has hardcoded cache directories in the binary directory


oati:
See:
https://github.com/LeelaChessZero/lc0/blob/d8ce48258c39d331c119f8c8729374ceb3df8409/src/neural/backends/onnx/network_onnx.cc#L678
https://github.com/LeelaChessZero/lc0/blob/d8ce48258c39d331c119f8c8729374ceb3df8409/src/neural/backends/onnx/network_onnx.cc#L758

I am trying to package lc0 for Nix. This issue prevents lc0 from running, since the binary directory is read only in Nix.

The cache directory should be configurable in some way, rather than being hardcoded.

PR #2428, opened 2026-07-13
Save TensorRT engine in weights file with onnx2leela and automate during runtime

saithepaithewhyyy:
With respect to the following issue: https://github.com/LeelaChessZero/lc0/issues/2185

This PR adds the following:-
- Added a protobuf field for ONNX EPContext model marking (is_ep_context). Basically, so that a net can be marked as either a regular ONNX model or an EPContext model (embedded engine)
- Onnx2leela can now determine if an ONNX model is an embedded engine model or not. If it is an embedded engine model, it packages the relevant _ctx.onnx file into an output weights file
- The above packaging is automated verbatim in network_onnx during runtime. Given a <weights>.pb.gz file, it builds the embedded engine (_ctx.onnx) and packages it in a new <weights>-embedded.pb.gz file and marks the proto to true.


src/neural/backends/onnx/network_onnx.cc
```diff
@@ -896,6 +905,10 @@ OnnxNetwork::OnnxNetwork(const WeightsFile& file, const OptionsDict& opts,
     mlh_head_ = outputs_.size();
     outputs_.emplace_back(md.output_mlh());
   }
+
+  cache_dir_ = (std::filesystem::path(CommandLine::BinaryDirectory()) / "trt_cache").string();
+  is_ep_context_ = md.is_ep_context();
```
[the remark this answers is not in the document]
saithepaithewhyyy: This is not required. The reason for this is that when lc0 with onnx-trt is run for the first time, it anyway scans for the trt_cache folder and the relevant timing caches. If it doesnt exist, the directory is created and the relevant caches are generated. This part of the code is only reached when all the timing, engine and embedded engine caches are generated and stored in trt_cache folder

mooskagh:
I'll take a look later, but just want to say that the suggestions for the Copilot are often too "defensive", and often extra checks that it suggests are not worth it (use your judgement).

src/neural/backends/onnx/network_onnx.cc
```diff
@@ -916,10 +929,68 @@ OnnxNetwork::OnnxNetwork(const WeightsFile& file, const OptionsDict& opts,
...
-    session_.emplace_back(onnx_env_, file.onnx_model().model().data(),
-                          file.onnx_model().model().size(),
+  for (int step = 1; step <= steps_; step++) {
+    if (provider_ == OnnxProvider::TRT && is_ep_context_) {
+      std::filesystem::path ctx_path = std::filesystem::path(cache_dir_) / "_ctx.onnx";
+      
+      {
+        std::ofstream f(ctx_path, std::ios::binary);
+        f.write(file.onnx_model().model().data(), file.onnx_model().model().size());
+      }
```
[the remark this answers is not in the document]
saithepaithewhyyy: Same reason as the above one, the cache directory will exist. As for the _ctx.onnx not existing, this is not possible, since the options that are required for it to be generated are added. 

On second thought, sure this can be added, but the exception breaks the flow.
Even in the (imopssible) case of it not being there, itll just silently fail and the embedded packaged weights wont be created, but the code doesnt break and will continue running normally

src/neural/backends/onnx/network_onnx.cc
```diff
@@ -916,10 +929,68 @@ OnnxNetwork::OnnxNetwork(const WeightsFile& file, const OptionsDict& opts,
...
+      for (const std::string& suffix : {".pb.gz", ".pb"}) {
+        if (net_path.size() > suffix.size() &&
+            net_path.compare(net_path.size() - suffix.size(), suffix.size(),
+                              suffix) == 0) {
+          net_path.resize(net_path.size() - suffix.size());
+          break;
+        }
+      }
+      net_path = net_path + "-embedded.pb.gz";
+    }
```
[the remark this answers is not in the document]
saithepaithewhyyy: again, net_path either comes from user input (if present) or from DiscoverWeightsFile().  So it can't be empty

src/tools/onnx2leela.cc
```diff
@@ -481,15 +494,35 @@ void ConvertOnnxToLeela() {
...
+        if (onnx->has_output_wdl()) {
+          onnx->set_output_wdl(name);
+        } else if (onnx->has_output_value()) {
+          onnx->set_output_value(name);
+        }
+      }
+    }
     onnx->set_model(onnx_model);
+    onnx->set_is_ep_context(true);
+  } else {
```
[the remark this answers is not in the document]
saithepaithewhyyy: Sure, this can be added as a sanity check. I'll add this

saithepaithewhyyy:
> I'll take a look later, but just want to say that the suggestions for the Copilot are often too "defensive", and often extra checks that it suggests are not worth it (use your judgement).

Sure, thanks a lot! Yup seems like copilot suggestions are a bit defensive. Either ways, I have replied to all

src/tools/onnx2leela.cc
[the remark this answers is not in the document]
saithepaithewhyyy: added

src/neural/backends/onnx/network_onnx.cc
```diff
@@ -916,10 +929,68 @@ OnnxNetwork::OnnxNetwork(const WeightsFile& file, const OptionsDict& opts,
...
+                        GetOptions(threads, batch_size_ * step, hash, optimize));
+      session_.push_back(std::move(sess));
+    } else {
+      session_.emplace_back(onnx_env_, file.onnx_model().model().data(),
+                        file.onnx_model().model().size(),
                           GetOptions(threads, batch_size_ * step, hash, optimize));
+    }
+  }
+
+  if (provider_ == OnnxProvider::TRT && !is_ep_context_) {
```
Menkib64: I don't think it is good idea to load a different file to what user specified. We should let user decide which network is used for a specific run.

src/tools/onnx2leela.cc
```diff
@@ -481,15 +494,36 @@ void ConvertOnnxToLeela() {
     onnx->set_output_mlh(dict.Get<std::string>(kOnnxOutputMlhId));
   }
 
-  if (MaybeFixOnnx(model, dict, data_type)) {
-    onnx->set_model(model.OutputAsString());
-  } else {
+  if (is_ctx) {
+    for (const auto& out : model.graph().output()) {
+      const auto& name = out.name();
+      if (onnx->has_output_policy() && name.find("policy") != std::string::npos) {
```
Menkib64: It is possible to get onnx models where output names don't follow naming convention. The names are provided as command line arguments to let user choose name maps to which output or input.

src/tools/onnx2leela.cc
```diff
@@ -448,6 +458,9 @@ void ConvertOnnxToLeela() {
     auto in = dict.Get<std::string>(kOnnxInputId);
     onnx->set_input_planes(in);
     data_type = GetDataType(model, in);
+    if(is_ctx){
+      data_type = GetDataType(model, "");
```
Menkib64: Why does empty string override the input data_type?

Menkib64 (commented):
Do we have to use file to store the context temporary everytime?
I haven't figured out how the loading code passes these to TensorRT but it would be useful if it could be loaded as string like we do for other models. If string cannot be passed, then I think it is better to use a temporary file/directory to pass the model and make sure it is deleted. The loading should also work if there is multiple threads running at the same time loading different engines.

We have a need to store many engines together. We use engines specifically optimized for a small range of batch sizes. For example 5090 can use optimized backend options with `-o batch=7,steps=12`.  This requires loading 12 slightly different engines created from the same network. This would be an important feature to include when embedding the engine to a network file. There would be reason to add a couple of more engines to optimize a few key batch sizes even more. It could make sense to add special engines which handle only batch size 1 and exact minibatch size. These should offer a little faster evaluation for these special case. There might be need to add more small batch size kernels because TRT is bad at optimizing kernels for multiple batch sizes.

saithepaithewhyyy:
> Do we have to use file to store the context temporary everytime? I haven't figured out how the loading code passes these to TensorRT but it would be useful if it could be loaded as string like we do for other models. If string cannot be passed, then I think it is better to use a temporary file/directory to pass the model and make sure it is deleted. The loading should also work if there is multiple threads running at the same time loading different engines.
> 
> We have a need to store many engines together. We use engines specifically optimized for a small range of batch sizes. For example 5090 can use optimized backend options with `-o batch=7,steps=12`. This requires loading 12 slightly different engines created from the same network. This would be an important feature to include when embedding the engine to a network file. There would be reason to add a couple of more engines to optimize a few key batch sizes even more. It could make sense to add special engines which handle only batch size 1 and exact minibatch size. These should offer a little faster evaluation for these special case. There might be need to add more small batch size kernels because TRT is bad at optimizing kernels for multiple batch sizes.

Firstly, thanks a lot for having a look at this!

Yes this is a valid point, the context file does take up quite some space (171 MB on my system currently). I'll have to take a look if we can bypass the step of having to save this in the trt_cache folder. It should be possible, but in the case that this isnt found, it can always be manually deleted once the weights file is generated

EDIT: I have now added code that removes the _ctx.onnx when the run finished. However it still remains in trt_cache folder until the embedded weights are packaged in the file given in the dump_embedded_weights option

src/neural/backends/onnx/network_onnx.cc
saithepaithewhyyy: This block of code only saves the embedded engine in a <network>-embedded.pb.gz weights file, where the network would come from the --weights input if provided by the user or from the DiscoverWeightsFile if no input is given. The net_path is never passed on to the Ort::Session

The engine is loaded with only the net file the user provides.
Please do correct me if I am wrong somewhere

src/tools/onnx2leela.cc
saithepaithewhyyy: ah good point, ill use the cli arguments instead of the hardcoded strings

src/neural/backends/onnx/network_onnx.cc
Menkib64: Right. I missed the point that it was about saving instead of load. There is options which are passed to the engine. We would need to define an option to give the desired embedded context file path. Backend should only generate the file when user requests it.

saithepaithewhyyy: Right, I can do this, I was not sure if I can create an option for this and instead hardcode it, but I'll create that which would save it only when requested

src/tools/onnx2leela.cc
saithepaithewhyyy: The reason for the empty-string fallback is that ctx models can have their input node renamed by TRT compilation

I didnt intend the overriding; it should have been an if else chunk

Menkib64: Shouldn't we save the renamed node names for inputs and outputs and pass the new name here?

src/neural/backends/onnx/network_onnx.cc
saithepaithewhyyy: Done i added the option dump_embedded_weights. When not used, the engine does not store the _ctx model in a new package weight file. If mentioned, it stores it at the path mentioned

also the _ctx model is now deleted at the end, so it does not get saved in trt_cache after run. I found no way to have this not save on runtime however (like model.onnx). Do you think it is possible?

src/tools/onnx2leela.cc
saithepaithewhyyy: I removed this part altogether. Why set it again, when it was set before. I missed this previously

src/tools/onnx2leela.cc
saithepaithewhyyy: Yes youre right, added that in

src/neural/backends/onnx/network_onnx.cc
```diff
@@ -687,6 +693,9 @@ Ort::SessionOptions OnnxNetwork::GetOptions(int threads, int batch_size,
       trt_options["trt_max_partition_iterations"] = "1000";
       trt_options["trt_min_subgraph_size"] = "1";
       trt_options["trt_engine_cache_enable"] = "1";
+      trt_options["trt_dump_ep_context_model"] = "1";
+      trt_options["trt_ep_context_file_path"] = cache_dir;
+      trt_options["trt_ep_context_embed_mode"] = "1";
```
Menkib64, line 698: `trt_ep_context_file_path` needs similar prefix as cache file uses. This makes sure that concurrent exports work if building different engines for different batch sizes on multiple GPUs.

src/neural/backends/onnx/network_onnx.cc
```diff
@@ -916,10 +929,56 @@ OnnxNetwork::OnnxNetwork(const WeightsFile& file, const OptionsDict& opts,
       break;
   }
 
-  for (int step = 1; step <= steps_; step++)
-    session_.emplace_back(onnx_env_, file.onnx_model().model().data(),
-                          file.onnx_model().model().size(),
+  std::filesystem::path ctx_path = std::filesystem::path(cache_dir_) / "_ctx.onnx";
+  for (int step = 1; step <= steps_; step++) {
+    if (provider_ == OnnxProvider::TRT && is_ep_context_) {
```
Menkib64: This branch should be removed. We should be able to provide a context model just like an ordinary model. There is no need for special handling here.

It is open question if we have to disable cache uses when loading a context model.

src/neural/backends/onnx/network_onnx.cc
```diff
@@ -916,10 +929,56 @@ OnnxNetwork::OnnxNetwork(const WeightsFile& file, const OptionsDict& opts,
...
+    if(!net_path.empty()){
+      const std::string ctx = ReadFileToString(ctx_path.string());
+      pblczero::ModelProto model;
+      model.ParseFromString(ctx);
+      pblczero::Net out = file;
+      auto* md_out = out.mutable_onnx_model();
+      for (const auto& output : model.graph().output()) {
+        const auto& name = output.name();
+        if (md_out->has_output_policy() &&
+            name.find("policy") != std::string::npos) {
```
Menkib64: Should this use the name from original model?

Original model could have outputs named in any possible way. There is no requirement to include any specific string in the name.

src/neural/backends/onnx/network_onnx.cc
```diff
@@ -916,10 +929,56 @@ OnnxNetwork::OnnxNetwork(const WeightsFile& file, const OptionsDict& opts,
...
+    } else {
+      session_.emplace_back(onnx_env_, file.onnx_model().model().data(),
+                        file.onnx_model().model().size(),
                           GetOptions(threads, batch_size_ * step, hash, optimize));
+    }
+  }
+
+  if (provider_ == OnnxProvider::TRT && !is_ep_context_ &&
+    opts.Exists<std::string>(SharedBackendParams::kDumpEmbeddedWeightsId)) {
+    std::string net_path = opts.Get<std::string>(SharedBackendParams::kDumpEmbeddedWeightsId);
```
Menkib64: Backend specific options don't have help messages currently. Backends just implement them using direct key string lookup. You can see how `"gpu"`, `"batch"`, `"steps"` are handled currently.

`SharedBackendParams` makes it look like option should be passed directly to Leela. But the option is parsed for `--backend-options dump-embedded-weights=file_path`.

src/neural/backends/onnx/network_onnx.cc
```diff
@@ -916,10 +929,56 @@ OnnxNetwork::OnnxNetwork(const WeightsFile& file, const OptionsDict& opts,
...
+          md_out->set_output_mlh(name);
+        } else if (md_out->has_output_wdl()) {
+          md_out->set_output_wdl(name);
+        } else if (md_out->has_output_value()) {
+          md_out->set_output_value(name);
+        }
+      }
+
+      md_out->set_model(ctx);
+      md_out->set_is_ep_context(true);
```
Menkib64: We should store `trt_profile_min_shapes`, `trt_profile_max_shapes` and `trt_profile_opt_shapes` with the model. These parameters restrict when the engine is valid. For example 5090 could be configured using `--backend-options batch=7,steps=12` which generates 12 different engines. Each engine is valid for only 7 batch sizes. Loading has to verify that the context file matches requested optimization state.

This also puts another requirement to context files. We want to embed many engines together to provide optimized engines for full batch range.

src/neural/backends/onnx/network_onnx.cc
```diff
@@ -916,10 +929,56 @@ OnnxNetwork::OnnxNetwork(const WeightsFile& file, const OptionsDict& opts,
       break;
   }
 
-  for (int step = 1; step <= steps_; step++)
-    session_.emplace_back(onnx_env_, file.onnx_model().model().data(),
-                          file.onnx_model().model().size(),
+  std::filesystem::path ctx_path = std::filesystem::path(cache_dir_) / "_ctx.onnx";
```
Menkib64: This path would have had race condition when initializing many paths. All temporary files must use unique filename like `tmpfile` created file.

I had a quick look into the TensorRT provider code. It looks to me that we can just use the existing loading code as is. Context file is a drop in replacement for onnx models. No need for temporary files.

src/neural/backends/onnx/network_onnx.cc
```diff
@@ -701,7 +710,7 @@ Ort::SessionOptions OnnxNetwork::GetOptions(int threads, int batch_size,
       trt_options["trt_timing_cache_path"] = cache_dir;
       trt_options["trt_layer_norm_fp32_fallback"] = "1";
       trt_options["trt_force_sequential_engine_build"] = "1";
-      trt_options["trt_context_memory_sharing_enable"] = "1";
+      trt_options["trt_context_memory_sharing_enable"] = is_ep_context_ ? "0" : "1";
```
Menkib64: Why is context sharing disabled?
Did you see a bug when loading them?
Did you test a multi-engine configuration like `batch=7,steps=4`?

src/neural/backends/onnx/network_onnx.cc
```diff
@@ -209,6 +212,9 @@ class OnnxNetwork final : public Network {
   // The lower limit for variable batch size.
   int min_batch_size_;
   int gpu_;
+  // trt cache directory and flag for trt EPcontext model
+  std::string cache_dir_;
```
Menkib64: We don't have to store the cache directory because it is only used in constructor. The code can use helper function to return the string when it is needed in different parts of code.

src/neural/backends/onnx/network_onnx.cc
```diff
@@ -209,6 +212,9 @@ class OnnxNetwork final : public Network {
...
+  bool is_ep_context_;
```
Menkib64: Runtime doesn't need to know about context file. We should keep it as a stack variable in constructor.

src/tools/onnx2leela.cc
```diff
@@ -446,8 +456,21 @@ void ConvertOnnxToLeela() {
       NetworkFormat::InputFormat_AllValues, NetworkFormat::InputFormat_Name));
   if (dict.OwnExists<std::string>(kOnnxInputId)) {
     auto in = dict.Get<std::string>(kOnnxInputId);
-    onnx->set_input_planes(in);
-    data_type = GetDataType(model, in);
+    if (is_ctx) {
+      bool found = false;
```
Menkib64: There shouldn't be any need for special case here. Caller should provide input and output names to converter.

saithepaithewhyyy:
Thanks a lot @Menkib64 for having a look, ill have a relook on the code. Apologies if it was a bit bothersome to review this.

src/neural/backends/onnx/network_onnx.cc
saithepaithewhyyy: Yes, i recall that the build failed to load the model when this was set, so i had it disabled for ep context models. But after reading your other reviews, i'll make some changes to the model loading to see if that solves it. If it does, I shall remove this diff

saithepaithewhyyy:
Made a couple of changes. Here's a rundown:-

- Added prefix naming convention for the ctx files. 
- Embedded package weights support multiple shapes configuration
- Removed loading logic for embedded engines. Uses the same logic (except for multi step embedded engines)

Here is some testing I did with the generated embedded package weights. Since each embedded weight has the engine baked into it, a particular shape can only be run for an embedded weight file that has the same shape baked into it. I have added an exception for this case.

  ### **Default:-**
```
size, mean nps, mean ms,   sdev,     cv, max nps,  median, min nps, first max, first mean
   1,       20,   50.63, 0.0379, 0.0007,      20,      20,      20,   51.6496,    51.1538
   2,       40,   50.63, 0.0209, 0.0004,      40,      40,      39,   51.2964,    50.9542
   3,       59,   50.63, 0.0240, 0.0005,      59,      59,      59,   51.3002,    50.9201
   4,       79,   50.63, 0.0153, 0.0003,      79,      79,      79,   51.3491,    50.9868
   5,       86,   58.11, 0.0178, 0.0003,      86,      86,      86,   58.9883,    58.5488
   6,       96,   62.59, 0.0142, 0.0002,      96,      96,      96,   63.5354,    63.0272
   7,       91,    76.5, 0.0467, 0.0006,      92,      92,      91,   77.5483,    77.0046
   8,       94,    85.4, 0.0150, 0.0002,      94,      94,      94,   86.6241,    85.9815
   9,       97,   92.41, 0.0376, 0.0004,      98,      97,      97,   93.4660,    92.9277
  10,      103,   96.71, 0.0156, 0.0002,     103,     103,     103,   97.6259,    97.1589
  11,      107,     103, 0.0179, 0.0002,     107,     107,     107,  103.9298,   103.4571
  12,      110,   109.5, 0.2333, 0.0021,     110,     110,     109,  111.0847,   110.6124
  13,      110,   117.9, 0.0802, 0.0007,     110,     110,     110,  118.8826,   118.3230
  14,      116,     121, 0.0180, 0.0001,     116,     116,     116,  121.9979,   121.5165
  15,      109,   137.9, 0.0214, 0.0002,     109,     109,     109,  138.7651,   138.4088
  16,      112,   142.3, 0.0288, 0.0002,     113,     112,     112,  143.5514,   142.9252
  17,      113,   149.8, 0.1647, 0.0011,     114,     113,     113,  150.9601,   150.2807
  18,      116,   155.5, 0.0158, 0.0001,     116,     116,     116,  156.4069,   155.9458
  19,      119,   159.8, 0.0192, 0.0001,     119,     119,     119,  160.7277,   160.2635
  20,      122,   164.2, 0.0216, 0.0001,     122,     122,     122,  165.3464,   164.7481
  21,      114,   183.9, 0.1302, 0.0007,     114,     114,     114,  185.0707,   184.6698
  22,      116,   190.3, 0.0913, 0.0005,     116,     116,     115,  191.1741,   190.7254
  23,      118,   195.7, 0.2512, 0.0013,     118,     118,     117,  197.0801,   196.2794
  24,      119,   201.7, 0.1748, 0.0009,     119,     119,     119,  202.5590,   202.1118
  25,      120,   207.9, 0.0312, 0.0002,     120,     120,     120,  208.8152,   208.3384
  26,      123,   212.2, 0.0365, 0.0002,     123,     123,     122,  213.3785,   212.7528
  27,      124,     217, 0.1300, 0.0006,     125,     124,     124,  218.0712,   217.4922
  28,      126,   222.5, 0.0943, 0.0004,     126,     126,     126,  223.5003,   222.8751
```

### **Batch=8, Steps=2**
```
size, mean nps, mean ms,   sdev,     cv, max nps,  median, min nps, first max, first mean
   1,       60,   16.59, 0.0519, 0.0031,      61,      60,      60,   26.5929,    21.5859
   2,       83,   24.04, 0.0635, 0.0026,      84,      83,      82,   25.0528,    24.5386
   3,      100,   30.06, 0.0296, 0.0010,     100,     100,     100,   31.0184,    30.5089
   4,      105,   38.22, 0.0340, 0.0009,     105,     105,     104,   39.4662,    38.8522
   5,      113,   44.32, 0.0826, 0.0019,     113,     113,     112,   45.3204,    44.8599
   6,      116,    51.7, 0.0728, 0.0014,     116,     116,     115,   52.5185,    52.1559
   7,      121,   57.72, 0.0451, 0.0008,     121,     121,     121,   58.6213,    58.1818
   8,      121,   66.27, 0.0772, 0.0012,     121,     121,     120,   67.2074,    66.7385
   9,      116,   77.39, 0.0362, 0.0005,     116,     116,     116,   78.3962,    77.8936
  10,      122,   82.01, 0.0319, 0.0004,     122,     122,     122,   83.0998,    82.5115
  11,      122,   90.28, 0.0365, 0.0004,     122,     122,     122,   91.4701,    90.8238
  12,      126,   94.99, 0.0623, 0.0007,     127,     126,     126,   95.9134,    95.4385
  13,      123,   105.6, 0.2348, 0.0022,     124,     123,     123,  106.2901,   105.8451
  14,      127,   110.1, 0.0731, 0.0007,     127,     127,     127,  111.0326,   110.5533
  15,      125,   119.6, 0.0721, 0.0006,     126,     125,     125,  120.6484,   120.1364
  16,      126,   126.5, 0.0848, 0.0007,     127,     127,     126,  127.3285,   126.9024
  17,      119,   143.1, 0.0889, 0.0006,     119,     119,     119,  144.1346,   143.6372
  18,      120,   150.5, 0.0797, 0.0005,     120,     120,     119,  152.4475,   151.4872
  19,      121,   156.6, 0.0970, 0.0006,     121,     121,     121,  157.7536,   157.1516
  20,      121,   164.7, 0.0634, 0.0004,     122,     121,     121,  165.8092,   165.2976
  21,      123,   170.8, 0.0856, 0.0005,     123,     123,     123,  171.9790,   171.3635
  22,      123,   178.2, 0.0574, 0.0003,     124,     123,     123,  179.2062,   178.7313
  23,      125,   184.3, 0.0507, 0.0003,     125,     125,     125,  185.5579,   184.9609
  24,      124,   193.1, 0.0661, 0.0003,     124,     124,     124,  194.1754,   193.7434
  25,      122,   205.3, 0.1315, 0.0006,     122,     122,     122,  205.4495,   205.3043
  26,      124,   209.9, 0.2285, 0.0011,     124,     124,     123,  210.1727,   209.9657
  27,      124,   218.3, 0.0892, 0.0004,     124,     124,     123,  218.7345,   218.5696
  28,      126,     223, 0.2170, 0.0010,     126,     126,     125,  223.5206,   223.2151
```

### **Batch=14, Steps=1**
```
size, mean nps, mean ms,   sdev,     cv, max nps,  median, min nps, first max, first mean
  1,       42,    23.6, 0.0136, 0.0006,      42,      42,      42,   24.7826,    24.1808
   2,       70,   28.44, 0.0548, 0.0019,      71,      70,      70,   29.3406,    28.8656
   3,       92,   32.63, 0.0331, 0.0010,      92,      92,      91,   33.4935,    33.1242
   4,       90,   44.62, 0.0718, 0.0016,      90,      90,      89,   45.5048,    45.0513
   5,       98,   50.99, 0.0889, 0.0017,      98,      98,      97,   51.8807,    51.4316
   6,      107,   56.05, 0.0125, 0.0002,     107,     107,     107,   57.3853,    56.7048
   7,      117,   59.61, 0.0254, 0.0004,     118,     117,     117,   60.4652,    60.0540
   8,      110,   72.52, 0.0196, 0.0003,     110,     110,     110,   73.5161,    73.0239
   9,      112,   80.13, 0.0198, 0.0002,     112,     112,     112,   81.2692,    80.6559
  10,      120,   83.18, 0.0212, 0.0003,     120,     120,     120,   84.3465,    83.7862
  11,      114,   96.32, 0.0407, 0.0004,     114,     114,     114,   97.4976,    96.8986
  12,      121,    99.4, 0.0620, 0.0006,     121,     121,     120,  100.3288,    99.8679
  13,      121,   107.6, 0.0376, 0.0003,     121,     121,     121,  108.4957,   108.0673
  14,      128,   109.8, 0.0435, 0.0004,     128,     128,     127,  110.6513,   110.2467
  15,      111,   134.6, 0.2697, 0.0020,     112,     111,     110,  135.0925,   134.7902
  16,      115,   139.4, 0.0977, 0.0007,     115,     115,     114,  139.7017,   139.4867
  17,      118,   143.6, 0.1053, 0.0007,     119,     118,     118,  144.1462,   143.9622
  18,      116,   155.6, 0.1669, 0.0011,     116,     116,     115,  156.0126,   155.7785
  19,      117,   162.1, 0.2642, 0.0016,     117,     117,     116,  165.2834,   164.0508
  20,      120,   166.9, 0.0737, 0.0004,     120,     120,     120,  167.2829,   167.1104
  21,      123,   170.5, 0.0786, 0.0005,     123,     123,     123,  170.9780,   170.7624
  22,      120,   183.4, 0.0793, 0.0004,     120,     120,     120,  183.6726,   183.5862
  23,      120,     191, 0.1272, 0.0007,     120,     120,     120,  191.5909,   191.3423
  24,      124,     194, 0.0574, 0.0003,     124,     124,     124,  194.3530,   194.2014
  25,      121,   207.1, 0.2031, 0.0010,     121,     121,     120,  208.0990,   207.6935
  26,      124,   209.7, 0.2301, 0.0011,     124,     124,     123,  210.0023,   209.8257
  27,      124,   217.7, 0.0859, 0.0004,     124,     124,     124,  218.0631,   217.8639
  28,      128,     219, 0.0653, 0.0003,     128,     128,     128,  219.8228,   219.3561
```

src/neural/backends/onnx/network_onnx.cc
saithepaithewhyyy: This is the exact error i got while running the embedde engine weights file, with this setting set to true:

```
2026-08-02 12:01:10.7885568 [E:onnxruntime:lc0, tensorrt_execution_provider.h:90 onnxruntime::TensorrtLogger::log] [2026-08-02 06:31:10   ERROR] IExecutionContext::setDeviceMemory: Error Code 3: API Usage Error (Parameter check failed, condition: memory != nullptr || mEngine.getDeviceMemorySizeInternal(mOptimizationProfile, false) == 0. setDeviceMemory: Cannot set memory to nullptr.)
2026-08-02 12:01:10.7978102 [E:onnxruntime:lc0, tensorrt_execution_provider.h:90 onnxruntime::TensorrtLogger::log] [2026-08-02 06:31:10   ERROR] IExecutionContext::enqueueV3: Error Code 3: API Usage Error (Parameter check failed, condition: noDeviceMemory. The engine requires 6727168 device memory. The IExecutionContext is created with ExecutionContextAllocationStrategy::kUSER_MANAGED or ICudaEngine::createExecutionContextWithoutDeviceMemory. IExecutionContext::setDeviceMemoryV2 should be called before enqueue/execute.)
2026-08-02 12:01:10.8144406 [E:onnxruntime:, sequential_executor.cc:615 onnxruntime::ExecuteKernel] Non-zero status code returned while running TRTKernel_graph_TRTKernel_graph_org.lczero/converted_12010072553396873023_0_10041065298831557996_0 node. Name:'TensorrtExecutionProvider_TRTKernel_graph_TRTKernel_graph_org.lczero/converted_12010072553396873023_0_10041065298831557996_0_0' Status Message: TensorRT EP execution context enqueue failed.
Unhandled exception in worker thread: Non-zero status code returned while running TRTKernel_graph_TRTKernel_graph_org.lczero/converted_12010072553396873023_0_10041065298831557996_0 node. Name:'TensorrtExecutionProvider_TRTKernel_graph_TRTKernel_graph_org.lczero/converted_12010072553396873023_0_10041065298831557996_0_0' Status Message: TensorRT EP execution context enqueue failed.
```
I have a feeling this is related to the setting itself. What do you think?

src/neural/backends/onnx/network_onnx.cc
saithepaithewhyyy: Added the prefix based naming

src/neural/backends/onnx/network_onnx.cc
saithepaithewhyyy: Removed

src/neural/backends/onnx/network_onnx.cc
saithepaithewhyyy: Yup, have added these. Also tested with various settings with backendbench

src/neural/backends/onnx/network_onnx.cc
saithepaithewhyyy: Removed

src/tools/onnx2leela.cc
saithepaithewhyyy: Removed this case

src/neural/backends/onnx/network_onnx.cc
Menkib64: The error message looks like it might be a bug in onnxruntime side. It fails to allocated context memory if loading embedded model.

src/neural/backends/onnx/network_onnx.cc
```diff
@@ -687,21 +694,27 @@ Ort::SessionOptions OnnxNetwork::GetOptions(int threads, int batch_size,
...
       trt_options["trt_engine_cache_enable"] = "1";
+      trt_options["trt_dump_ep_context_model"] = "1";
+      trt_options["trt_ep_context_embed_mode"] = "1";
       // We need the batch size as well as the hash, as it is set after loading.
       std::ostringstream oss;
       oss << std::hex << hash;
-      trt_options["trt_engine_cache_prefix"] =
-          "Lc0_ONNX_TRT_ORT_" + Ort::GetVersionString() + "_batch_" +
+      std::string cache_prefix =
+          "Lc0_ONNX_TRT_ORT_" + Ort::GetVersionString() + "_gpu_" +
```
Menkib64: Different gpu are expected to use the same cache file. Current initialization is done in the main thread only. I have a branch where backends can be initialized concurrently. This branch has extra initialization locking for onnx-trt so each GPU loads different batch size range. All GPUs should share the same cache file for the same batch size range.

I'm thinking that the same locking rule would apply to saving context files. It would have minimal help to use many GPUs to build context file concurrently because onnxruntime has some unnecessary locking in tensorrt execution provider.

src/neural/backends/onnx/network_onnx.cc
```diff
@@ -896,6 +909,35 @@ OnnxNetwork::OnnxNetwork(const WeightsFile& file, const OptionsDict& opts,
     mlh_head_ = outputs_.size();
     outputs_.emplace_back(md.output_mlh());
   }
+
+  std::string cache_dir =
+      (std::filesystem::path(CommandLine::BinaryDirectory()) / "trt_cache")
+          .string();
+  is_ep_context_ = md.is_ep_context();
```
Menkib64: `is_ep_context_` could be a stack variable.

src/neural/backends/onnx/network_onnx.cc
```diff
@@ -916,10 +958,64 @@ OnnxNetwork::OnnxNetwork(const WeightsFile& file, const OptionsDict& opts,
...
+
+  bool multi_step_embedded = is_ep_context_ && md.step_models_size() > 0;
+  for (int step = 1; step <= steps_; step++) {
+    std::string_view model = multi_step_embedded
+                                  ? md.step_models(step - 1)
+                                  : std::string_view(file.onnx_model().model());
+    session_.emplace_back(
+        onnx_env_, model.data(), model.size(),
+        GetOptions(threads, batch_size_ * step, hash, optimize,
+                  dump_weights && !multi_step_embedded ? &ctx_paths[step - 1] : nullptr));
```
Menkib64, line 993: Should dump_weights be always false when the input has an embedded context?
Now I think that an error check together with size checks would make sense. It would throw an error if trying to dump from an embedded context.

src/neural/backends/onnx/network_onnx.cc
```diff
@@ -687,21 +694,27 @@ Ort::SessionOptions OnnxNetwork::GetOptions(int threads, int batch_size,
       trt_options["trt_max_partition_iterations"] = "1000";
       trt_options["trt_min_subgraph_size"] = "1";
       trt_options["trt_engine_cache_enable"] = "1";
+      trt_options["trt_dump_ep_context_model"] = "1";
```
Menkib64: We should only enable dump context conditionally when constructor is asking for the context path.

src/neural/backends/onnx/network_onnx.cc
saithepaithewhyyy: Yeah, fair.
Yes dump weights should be false when the input itself has an embedded context, since the engine is basically just one node right, the EPContext stub. Theres nothing to dump.

I added some error checks around this for all possible cases. Theres an additional case if trying to dump but the provider itself isnt onnx-trt but any of the other ones. Another case is if the weight file is ep context but provider is not trt. 

I think this covers all cases?

src/neural/backends/onnx/network_onnx.cc
saithepaithewhyyy: thanks yes youre right

src/neural/backends/onnx/network_onnx.cc
saithepaithewhyyy: cool, ill use the existing prefix then

src/neural/backends/onnx/network_onnx.cc
```diff
@@ -687,21 +694,26 @@ Ort::SessionOptions OnnxNetwork::GetOptions(int threads, int batch_size,
       trt_options["trt_max_partition_iterations"] = "1000";
       trt_options["trt_min_subgraph_size"] = "1";
       trt_options["trt_engine_cache_enable"] = "1";
+      trt_options["trt_dump_ep_context_model"] = ep_context_path ? "1" : "0";
+      trt_options["trt_ep_context_embed_mode"] = "1";
```
Menkib64, line 698: Too bad embedded context is very slow to load. This makes me think that trt_pe_context_embed_mode must be disable to make loading faster.

Too bad onnxruntime prevents dumping cache referencing models if cached engine already exists. I wouldn't want to delete all cached files when dumping context models. This makes me think that we need to remove all cached engines which have the matching prefix for a given step. This way we can make sure there won't be a matching cached engine while keeping cache files for different networks.

src/neural/backends/onnx/network_onnx.cc
```diff
@@ -687,21 +694,26 @@ Ort::SessionOptions OnnxNetwork::GetOptions(int threads, int batch_size,
...
       oss << std::hex << hash;
-      trt_options["trt_engine_cache_prefix"] =
+      std::string cache_prefix =
           "Lc0_ONNX_TRT_ORT_" + Ort::GetVersionString() + "_batch_" +
           (batch_size < 0 ? std::to_string(batch_size)
                           : std::to_string(batch_size - batch_size_ + 1) + "-" +
                                 std::to_string(batch_size)) +
           "_" + std::to_string(optimize) + "_" + oss.str() + "_";
+      trt_options["trt_engine_cache_prefix"] = cache_prefix;
+      trt_options["trt_ep_context_file_path"] = cache_dir + "/" + cache_prefix + "ctx.onnx";
```
Menkib64, line 709: TRT works only if context file doesn't include cache_dir. This makes me think we must use `= cache_prefix + "ctx.onnx"`.

I think we should set `trt_ep_context_file_path` only when dumping context model.

src/neural/backends/onnx/network_onnx.cc
Menkib64, line 1044: The last boolean parameter needs to be set true for context models because context model uses CPU side wdl softmax. If it isn't set, value output is raw logits instead of win, draw, lose probabilities.

src/neural/wrapper.cc
```diff
@@ -183,7 +183,7 @@ std::unique_ptr<Backend> NetworkAsBackendFactory::Create(
     const OptionsDict& options) {
   const std::string backend_options =
       options.Get<std::string>(SharedBackendParams::kBackendOptionsId);
-  OptionsDict network_options;
+  OptionsDict network_options(&options);
```
Menkib64, line 186: Backend should see only `--backend-opts` options (alts: `-o` or UCI `BackendOptions`).

Menkib64 (commented):
I was meaning to test it. Too bad I kept on failing to find time to do local testing. The testing made me notice that my earlier idea was flawed. There is some performance problems if we pass the embedded context model in a string.

src/neural/backends/onnx/network_onnx.cc
Menkib64, line 652: TRT log message recommends that we should disable graph optimisation.
```
if (is_context_ep) {
  level = GraphOptimizationLevel::ORT_DISABLE_ALL;
}
```

src/neural/backends/onnx/network_onnx.cc
saithepaithewhyyy: Yeah, I agree, do you think we should remove the conditional and disable trt_ep_context_embed_mode altogether

src/neural/backends/onnx/network_onnx.cc
saithepaithewhyyy: Done!

src/neural/backends/onnx/network_onnx.cc
saithepaithewhyyy: Added a conditional for this

src/neural/wrapper.cc
saithepaithewhyyy: removed this diff

saithepaithewhyyy:
> I was meaning to test it. Too bad I kept on failing to find time to do local testing. The testing made me notice that my earlier idea was flawed. There is some performance problems if we pass the embedded context model in a string.

Thanks a lot for looking at this again @Menkib64 ! 
Also I did find another machine with a more modern 3050 GPU (better than my current 1650 atleast) and I plan to run some more benchmarks

Do you suggest anything particular that I can test apart from the backendbench and increasing the (batch,step) ?

src/neural/backends/onnx/network_onnx.cc
Menkib64: It would be possible to pass the return value of `is_ep_context` call like `return std::make_unique<OnnxNetwork>(*w, opts, kProvider, md.is_ep_context());`

src/neural/backends/onnx/network_onnx.cc
Menkib64: I think we have to set `trt_ep_context_embed_mode` to `"0"`. That mode requires that cache is configured when dumping a context model. Cached files must stay at the same location when trying to use a context model.

It will also require using std::filesystem to remove matching cache prefixes. std::filesystem::directory_iterator should be able to scan the cache directory and names can be filtered for a matching prefix. Matches would have to be removed to avoid execution provider skipping TRT build and context dump.

I don't remember if TRT recommended disabling cache when context model was used. Can you check if  disabling cache has any effect when loading a context mode?

I did use command `(echo go nodes 1; sleep 20) | ./lc0 dag-preview -b onnx-trt -o batch=7,steps=12 -w context_test.pb.gz`  when I was testing loading performance on 5080. The info line includes milliseconds since reading the `go nodes 1` command which includes backend loading in this case.

Menkib64 (commented):
Techpowerup claims that 3050 has 20 SMs. My guess for optimised backend configuration is `batch=10,steps=2` or `batch=5,steps=4` because `2 * 2 * 5 == 20`. I don't know if batch=5 performs better because bigger GPUs didn't gain when setting batch to be less than 7.

src/neural/backends/onnx/network_onnx.cc
saithepaithewhyyy: yeah, i missed this, oops

src/neural/backends/onnx/network_onnx.cc
saithepaithewhyyy: Yup, set trt_ep_context_embed_mode to "0". Also added code to remove any existing cache files with cache prefiz when dumping a new context model

I did some testing and loading the context model with the caches disabled is a tiny bit faster

```
**when caches are disabled:**
1441.582 ms  info depth 1 seldepth 1 time 1434 nodes 1 score cp 5 nps 9 tbhits 0 pv d2d4

**when caches are not disabled:**
1472.323 ms  info depth 1 seldepth 1 time 1471 nodes 1 score cp 5 nps 9 tbhits 0 pv d2d4
```

About 30 ms. Above is one instance, but on average disabling speeds up loading marginally

I checked the docs, and i did not see any mention of needing to disable cache when im running a context model. However i think it makes sense to disable them 

What do you think @Menkib64

Menkib64: I think I read the cache disabling tip from TensorRT debug message. TRT printed a recommendation of setting `trt_ep_context_embed_mode` to 0 and disabling cache when loading if I remember correctly. I didn't see anything about it in docs either. 30 ms is surprising low gain for an action where other recommended action is more like factor 3-4 faster.

src/neural/backends/onnx/network_onnx.cc
```diff
@@ -740,6 +757,18 @@ Ort::SessionOptions OnnxNetwork::GetOptions(int threads, int batch_size,
 #endif
       options.AppendExecutionProvider_TensorRT_V2(*trt_options_v2);
       api.ReleaseTensorRTProviderOptions(trt_options_v2);
+
+      if (ep_context_path && std::filesystem::exists(cache_dir)) {
+        for (const auto& entry : std::filesystem::directory_iterator(cache_dir)) {
+          if (entry.is_regular_file()) {
+            const auto& filename = entry.path().filename().string();
+            if (filename.find(cache_prefix) == 0) {
```
Menkib64: There is `std::string::starts_with` for this purpose.

src/neural/backends/onnx/network_onnx.cc
```diff
@@ -675,7 +685,7 @@ Ort::SessionOptions OnnxNetwork::GetOptions(int threads, int batch_size,
     case OnnxProvider::TRT: {
       options.SetExecutionMode(ExecutionMode::ORT_SEQUENTIAL);
 
-      std::string cache_dir = CommandLine::BinaryDirectory() + "/trt_cache";
+      std::string cache_dir = "trt_cache";
```
Menkib64: Is there a reason for cache directory change?

This reminds me that we should probably move the cache directory to user cache directory. But that should be a different pull request.

src/neural/backends/onnx/network_onnx.cc
saithepaithewhyyy: I think this may be because the context model already has the precompiled/built engine in it. On running the context model we simply just use that without having to build it again. As a result enabling/disabling the caches has no effect

This is just my understanding, but I could be very wrong

src/neural/backends/onnx/network_onnx.cc
saithepaithewhyyy: Yes. Setting the context mode to 0 fails since the cache directory was an absolute path and not relative

saithepaithewhyyy: At least that was the error I got when I ran with context mode set to 0

Menkib64: Good catch. I didn't notice the problem because I was using relative launch command which made cache directory use relative path. That is a problem which we need to solve. I guess we have to use different cache directory which is relative to the current working directory when dumping.

It would make sense to have the directory name include the output filename which user passed using `dump-embedded-weights`. We could append `_trt_cache` to the filename. It would require testing if including directories in the filename confuses TRT.  `dump-embeded-weights=../test.pb.gz` could maybe confusing for path handling.

Menkib64: Of course if user passes an absolute path as filename, then TRT would give the same error.

Menkib64: It could be possible to simplify the problem by requiring that `dump-embedded-weights` is always a file in the current working directory. Then we could use cache directory name by appending `_cache_dir` to user supplied filename. I think that should make it safe so TRT won't be able to mess up path handling. Relatively paths should mean that user can move the embedded network file and cache directory together to a different location.

saithepaithewhyyy: Done, i used this and it works fine, added a little check for current user directory
However had to do a little manipulation, moved cache_dir_ to constructor since GetOptions does not have access to opts and therefore cant get dump-embedded-weights.

saithepaithewhyyy: Tested with both non context and context files, with non context you get trt_cache and with context files you get context.pb.gz-trt-cache

src/neural/backends/onnx/network_onnx.cc
```diff
@@ -221,6 +226,7 @@ class OnnxNetwork final : public Network {
 #endif
 
  private:
+  std::string cache_dir_;
```
Menkib64: You could add a parameter to `GetOptions` like `ep_context_path`. If the parameter string isn't empty, it would override the default path.

saithepaithewhyyy: Perfect, I did this!

src/neural/backends/onnx/network_onnx.cc
```diff
@@ -673,19 +683,9 @@ Ort::SessionOptions OnnxNetwork::GetOptions(int threads, int batch_size,
...
-      iss >> major >> dot >> minor >> dot >> patch;
-      if (major == 1 && minor >= 27) {
-        CERR << "WARNING: onnxruntime 1.27 or newer has a bug with TensorRT "
-                "10 and uses FP32 compute instead of FP16. If you notice bad "
-                "performance, downgrade to onnxruntime 1.26 or earlier.";
-      }
       options.SetExecutionMode(ExecutionMode::ORT_SEQUENTIAL);
 
-      std::string cache_dir = CommandLine::BinaryDirectory() + "/trt_cache";
+      std::string cache_dir = trt_cache_dir.empty() ? (CommandLine::BinaryDirectory() + "/trt_cache") : trt_cache_dir;
```
Menkib64, line 688: We want to keep timing cache in the standard location to speed up dumping. This conditional check would apply only to `trt_engine_cache_path` and cache cleaning while `trt_timing_cache_path` keeps using the default path.

saithepaithewhyyy: Added this, this will create two caches though, one for the timing_cache (trt_cache dir) and one for engine caches (context_trt_cache)

Menkib64: Did you use the same dump filename?

I run a test with a new filename. I have timing cache in the old `test.pb.gz_cache_dir` but the new `test2.pb.gz_cache_dir` doesn't include timing cache. There is no code to remove a timing cache from the directory if it already exists.

saithepaithewhyyy: trt_timing_cache_path is set to CommandLine::BinaryDirectory() + "/trt_cache", so all the timing caches will be saved there. Regarding the cache cleaning, it only cleans the directory (conditional on trt_cache_dir) and if it has the cache prefix. Should we clean up the timing cache as well?

In that case I might as well wipe out all cache directories on new run with context model (if they exist).

Menkib64: I translated your previous message incorrectly in my head. You were talking about the default directory also being created but only one timing cache existing. I thought you wrote about two timing cache files existing. I don't think that there needs to be any more changes to make it work better. I will make a follow up path to address #2414 .
@borg323 might still have some improvement ideas. He knows the onnx converter better than I do.

saithepaithewhyyy: Sure thanks! Apologies for the misunderstanding.

@borg323 if you have some suggestions for improvements, id love to incorporate those too.

saithepaithewhyyy: Sure thanks! Apologies for the misunderstanding.

@borg323 if you have some suggestions for improvements, id love to incorporate those too.

Menkib64 (approved):

proto/net.proto
```diff
@@ -397,6 +397,11 @@ message OnnxModel {
   optional string output_wdl = 5;
   optional string output_policy = 6;
   optional string output_mlh = 7;
+  optional bool is_ep_context = 8;
+  optional uint32 trt_min_batch_size = 9;
+  optional uint32 trt_batch_size = 10;
+  optional uint32 trt_steps = 11;
+  repeated bytes step_models = 12;
```
borg323, line 404: Sorry for not noticing earlier, but I don't like adding backend specific fields in `net.proto` like that. After some discussion with other maintainers I have a suggestion for a more generic mechanism that you can see in https://github.com/LeelaChessZero/lc0/compare/997849...borg323:lc0:pr2428 (the last three commits of [this tree](https://github.com/borg323/lc0/commits/pr2428) if you want to see the intermediate stages). As it is untested (except for build tests) it is open for discussion and certainly has room for improvement.

saithepaithewhyyy: I had a look at your commits, they absolutely are valid asks. Let me do some testing (although i feel there shouldn't be any performance change or any failures but you never know).

If anything pops up, I'll add it here or if all's good I'll commit those changes.

Thanks for having a look though!

borg323: Something you can think about is being smarter with the `id` field, e.g. by expanding it with gpu/backend configuration information - but this is also fine to be left for later.

saithepaithewhyyy: @borg323, I pulled in your commits, it had a minor issue on testing so I fixed those (basically the batch,step,min_batch validation was a bit wonky and always gave me the exception that I was using the wrong config)

The caching, loading and cache dir cleanup is all working properly now

I was also thinking on your last point, regarding the id field, but I think its safer to open another PR for the same

borg323: I see you had to store integers in the protobuf instead of floats and keep the global minimum value for all batches instead of the local one:
1. The floats were for more generality across backends and
2. the local minimum for future expansion, where we may allow partial matching of the stored engines.

Neither is terribly important, but I'll try to reproduce the issue to understand what may be happening.

PR #2436, opened 2026-08-01
CUDA: vectorize FP16 attention input preprocessing

john-sp:
- Add an eight-elements-per-thread FP16 path for PE-dense attention-body preprocessing.
- Use aligned 128-bit loads and stores where channel boundaries permit.
- Retain the scalar implementation for FP32, non-dense positional encoding, and unsupported alignment.

Dispatch depends only on datatype, layout, and channel alignment; it contains no model, batch-size, device-name, or SM-count checks.

## Performance

RTX 3090 Ti, BT4, batch 84:

- Master median: 23.732 us
- Optimized median: 14.246 us
- Kernel reduction: 40.0%
- Saving: approximately 9.5 us per evaluation

AI Acknowledgement: Written by Codex

### DIFF
--- a/src/neural/backends/cuda/common_kernels.cu
+++ b/src/neural/backends/cuda/common_kernels.cu
@@ -1227,6 +1227,36 @@ __global__ void preprocess_for_attention_body_kernel(
   output[n * 64 * outputC + hw * outputC + c] = op;
 }
 
+__global__ void preprocess_for_attention_body_fp16x8_kernel(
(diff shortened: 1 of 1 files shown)


Menkib64 (approved):

borg323:
~~This will need rebasing after the latest commits.~~
I was confused by the stale cuda 11 appveyor cache, will run an fresh cycle to confirm,

PR #2442, opened 2026-08-21
add Iluvatar CoreX (ivcore11) build support

tlmn2008:
Add an opt-in `corex` build for Iluvatar CoreX (ivcore11), whose SDK ships no nvcc and uses `clang++ -x ivcore --cuda-gpu-arch=ivcore11` as its CUDA frontend. Since ivcore11 is a single fixed target, no nvcc arch/std/`__CUDA_ARCH__` probing is needed, so this is a small self-contained block in meson.build that leaves the existing nvcc logic completely untouched.

- `meson_options.txt`: new `corex` (boolean, default false) and `corex_root` (string, default `/usr/local/corex`)
- `meson.build`: a dedicated `corex` block finds clang++/libs under `corex_root` and compiles the `.cu` files directly with `clang++ -x ivcore --cuda-gpu-arch=ivcore11 -fPIC -std=c++20`; it errors out if both `corex` and `nvcc` are set (mutually exclusive)
- `fp16_kernels.cu`: gate the fp16 kernels on `__ILUVATAR__` in addition to `__CUDA_ARCH__ >= 530`, because ivcore11 reports `__CUDA_ARCH__` 300 yet fully supports fp16 (`__ILUVATAR__` is predefined by the CoreX clang frontend)

Build with `-Dcorex=true -Dnvcc=false`. No effect on existing nvcc builds (option defaults to false). 
All 52 test cases have passed on Corex.

Iluvatar Corex: https://en.wikipedia.org/wiki/Iluvatar_CoreX

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -461,6 +461,10 @@ if get_option('build_backends')
   cu_blas = cc.find_library('cublas', dirs: cudnn_libdirs, required: false)
   cu_dnn = cc.find_library('cudnn', dirs: cudnn_libdirs, required: false)
   cu_dart = cc.find_library('cudart', dirs: cudnn_libdirs, required: false)
+  if get_option('corex') and get_option('nvcc')
+    error('The corex (Iluvatar CoreX) and nvcc CUDA backends are mutually ' +
+          'exclusive; build the CoreX backend with -Dcorex=true -Dnvcc=false.')
+  endif
   nvcc = find_program(nvcc_paths,
                       required: false)
   nvcc_ok = false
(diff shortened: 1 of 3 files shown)


borg323:
I think it would be better to make it a completely separate block in `meson.build` so that future changes in the `nvcc` logic won't affect it. Also add a `corex_root` option (default `/usr/local/corex/`) instead of hardcoding paths and avoid the complicated existing path logic (which is there for historical reasons). You will need to duplicate the following block (with a simpler command) but that is OK:

```
    files += custom_target('cuda fp32 code',
      input : 'src/neural/backends/cuda/common_kernels.cu',
      output : outputname,
      depend_files: 'src/neural/backends/cuda/winograd_helper.inc',
      command : [nvcc, nvcc_extra_args, nvcc_arguments, nvcc_io_arguments]
    )

    files += custom_target('cuda fp16 code',
      input : 'src/neural/backends/cuda/fp16_kernels.cu',
      output : outputname,
      depend_files: 'src/neural/backends/cuda/winograd_helper.inc',
      command : [nvcc, nvcc_extra_args, nvcc_arguments, nvcc_io_arguments]
    )
    has_backends = true
```

You can also add an explicit check to error out if someone tries to build with both corex sdk and nvcc.

meson.build
```diff
@@ -461,6 +461,10 @@ if get_option('build_backends')
   cu_blas = cc.find_library('cublas', dirs: cudnn_libdirs, required: false)
   cu_dnn = cc.find_library('cudnn', dirs: cudnn_libdirs, required: false)
   cu_dart = cc.find_library('cudart', dirs: cudnn_libdirs, required: false)
+  if get_option('corex') and get_option('nvcc')
+    error('The corex (Iluvatar CoreX) and nvcc CUDA backends are mutually ' +
+          'exclusive; build the CoreX backend with -Dcorex=true -Dnvcc=false.')
+  endif
```
borg323: I think this block should go inside the `corex` section so that all the changes are localized.

meson.build
```diff
@@ -609,6 +613,58 @@ if get_option('build_backends')
...
+                  '-I', corex_root / 'include']
+    corex_cmd = [corex_clang, corex_args, '-c', '@INPUT@', '-o', '@OUTPUT@']
+
+    deps += [cx_blas, cx_dart]
+    cuda_files = ['src/neural/backends/cuda/layers.cc']
+    if get_option('cudnn') and cx_dnn.found()
+      deps += cx_dnn
+      cuda_files += 'src/neural/backends/cuda/network_cudnn.cc'
+      cuda_files += 'src/neural/backends/cuda/network_cuda.cc'
+      add_project_arguments('-DUSE_CUDNN', language : 'cpp')
```
borg323: Is there a reason to allow building with cudnn?  It only supports convolutional nets (the attention code is not affected) and AFAIK it is only advantageous on older nvidia cards.

tlmn2008: Removed cudnn building logic.

meson.build
tlmn2008: Moved to corex section.

borg323 (approved):

borg323:
This will need rebasing after the latest commits.

borg323:
Is this sufficient for bf16 support?
```diff
diff --git a/src/neural/backends/cuda/cuda_common.h b/src/neural/backends/cuda/cuda_common.h
index b4b3d1e8..7084640a 100644
--- a/src/neural/backends/cuda/cuda_common.h
+++ b/src/neural/backends/cuda/cuda_common.h
@@ -36,7 +36,7 @@
 #define BACKEND_NAME "CUDA"
 #define BACKEND_NAME_LC "cuda"
 
-#if CUDART_VERSION >= 11000
+#if CUDART_VERSION >= 11000 || defined(__ILUVATAR__)
 #include <cuda_bf16.h>
 #define LC0_CUDA_BF16_SUPPORTED 1
 #else
```

Moreover, the following (untested) patch should be all that is needed to allow the backend (renamed corex) to coexist with cuda:
```diff
diff --git a/meson.build b/meson.build
index 32a4b528..294d2fde 100644
--- a/meson.build
+++ b/meson.build
@@ -618,10 +618,6 @@ if get_option('build_backends')
   # A single fixed target needs no arch/std probing, so this block is
   # self-contained and independent of the nvcc logic above.
   if get_option('corex')
-    if get_option('nvcc')
-      error('The corex (Iluvatar CoreX) and nvcc CUDA backends are mutually ' +
-            'exclusive; build the CoreX backend with -Dcorex=true -Dnvcc=false.')
-    endif
     corex_root = get_option('corex_root')
     corex_clang = find_program(corex_root / 'bin' / 'clang++', required : true)
     corex_lib = corex_root / 'lib64'
diff --git a/src/neural/backends/cuda/cuda_common.h b/src/neural/backends/cuda/cuda_common.h
index b4b3d1e8..3687b92f 100644
--- a/src/neural/backends/cuda/cuda_common.h
+++ b/src/neural/backends/cuda/cuda_common.h
@@ -32,9 +32,15 @@
 #include <cublas_v2.h>
 #include <cuda_fp16.h>
 #include <cuda_runtime.h>
+#if !defined(__ILUVATAR__)
 #define NS_BACKEND cudnn_backend
 #define BACKEND_NAME "CUDA"
 #define BACKEND_NAME_LC "cuda"
+#else
+#define NS_BACKEND corex_backend
+#define BACKEND_NAME "COREX"
+#define BACKEND_NAME_LC "corex"
+#endif
 
 #if CUDART_VERSION >= 11000
 #include <cuda_bf16.h>
```

PR #2455, opened 2026-09-10
onnx: relax driver version check to warning and fix session cleanup order

jaipkapoor99:
### Summary
Separated from the BF16 PR as requested by @borg323.

### Changes
1. **Relax Driver Version Check**: Converts the hard fatal error on driver mismatch to a non-fatal warning so compatible runtime/driver configurations are not rejected prematurely.
2. **Session Cleanup Order**: Ensures Ort::Session resources and memory allocators are destroyed in the correct sequence to prevent shutdown segmentation faults.

### Testing
- Verified clean build under GCC 14/15.
- Tested ONNX backend initialization and shutdown cleanly without destructor faults.

### DIFF
--- a/src/neural/backends/onnx/network_onnx.cc
+++ b/src/neural/backends/onnx/network_onnx.cc
@@ -308,9 +308,9 @@ InputsOutputs::InputsOutputs(OnnxNetwork* network)
 OnnxNetwork::~OnnxNetwork() {
 #ifdef USE_ONNX_CUDART
   if (provider_ == OnnxProvider::TRT || provider_ == OnnxProvider::CUDA) {
-    if (compute_stream_) cudaStreamDestroy(compute_stream_);
-    if (upload_stream_) cudaStreamDestroy(upload_stream_);
-    if (download_stream_) cudaStreamDestroy(download_stream_);
+    if (compute_stream_) ReportCUDAErrors(cudaStreamDestroy(compute_stream_));
+    if (upload_stream_) ReportCUDAErrors(cudaStreamDestroy(upload_stream_));
+    if (download_stream_) ReportCUDAErrors(cudaStreamDestroy(download_stream_));
   }
 #endif
 }
--- a/src/neural/backends/onnx/network_onnx.cc
+++ b/src/neural/backends/onnx/network_onnx.cc
@@ -306,11 +306,13 @@ InputsOutputs::InputsOutputs(OnnxNetwork* network)
 }
 
 OnnxNetwork::~OnnxNetwork() {
+  free_inputs_outputs_.clear();
+  session_.clear();
 #ifdef USE_ONNX_CUDART
   if (provider_ == OnnxProvider::TRT || provider_ == OnnxProvider::CUDA) {
-    ReportCUDAErrors(cudaStreamDestroy(compute_stream_));
-    ReportCUDAErrors(cudaStreamDestroy(upload_stream_));
-    ReportCUDAErrors(cudaStreamDestroy(download_stream_));
+    if (compute_stream_) cudaStreamDestroy(compute_stream_);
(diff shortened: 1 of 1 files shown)


src/neural/backends/onnx/network_onnx.cc
```diff
@@ -306,11 +306,13 @@ InputsOutputs::InputsOutputs(OnnxNetwork* network)
 }
 
 OnnxNetwork::~OnnxNetwork() {
+  free_inputs_outputs_.clear();
+  session_.clear();
```
Menkib64: Why?

src/neural/backends/onnx/network_onnx.cc
```diff
@@ -810,8 +812,7 @@ OnnxNetwork::OnnxNetwork(const WeightsFile& file, const OptionsDict& opts,
     CERR << "Latest version of CUDA supported by the driver: "
          << nv_version(driver_version);
     if (driver_version < runtime_version) {
-      throw Exception(
-          "ERROR: The CUDA driver version is older than the runtime version.");
+      CERR << "WARNING: The CUDA driver version is older than the runtime version.";
```
Menkib64: I don't want to support backward compatibility because it is very likely to cause problems with TensorRT. onnx-cuda would be able to use older driver but it is only useful for experimenting with new network architectures. It is better make sure users have driver with all features that TensorRT expects.

src/neural/backends/onnx/network_onnx.cc
jaipkapoor99: In OnnxNetwork, member variables are destroyed in reverse order of declaration (onnx_env_, session_, ... free_inputs_outputs_).

The objects in free_inputs_outputs_ hold device allocations and Ort::MemoryInfo handles bound to the underlying CUDA streams (compute_stream_, etc.). If we simply let ~OnnxNetwork() run, cudaStreamDestroy was being called while active Ort::Session contexts and InputsOutputs resources were still alive and referencing those streams, which triggered intermittent shutdown segfaults / illegal memory access during engine teardown.

Explicitly clearing free_inputs_outputs_ and session_ first ensures all ONNX session bindings, memory handles, and pending evaluations are completely released before the CUDA streams themselves are destroyed.

Menkib64: Cuda memory allocations are bound to context instead of stream. Ort::MemoryInfo doesn't know about streams. Streams should be idle when OnnxNetwork destructor is called. If they aren't idle, it is a threading bug somewhere or an unhandled exception.

src/neural/backends/onnx/network_onnx.cc
```diff
@@ -308,9 +308,9 @@ InputsOutputs::InputsOutputs(OnnxNetwork* network)
 OnnxNetwork::~OnnxNetwork() {
 #ifdef USE_ONNX_CUDART
   if (provider_ == OnnxProvider::TRT || provider_ == OnnxProvider::CUDA) {
-    ReportCUDAErrors(cudaStreamDestroy(compute_stream_));
-    ReportCUDAErrors(cudaStreamDestroy(upload_stream_));
-    ReportCUDAErrors(cudaStreamDestroy(download_stream_));
+    if (compute_stream_) cudaStreamDestroy(compute_stream_);
```
Menkib64: cudaStreamDestroy can report errors which could be asynchronous errors from previous cuda calls. We want to check for them and report the problem to user too. Errors would indicate some kind of bug or GPU problem.

jaipkapoor99: Good catch. Restored ReportCUDAErrors around the cudaStreamDestroy calls while keeping the null checks.

Menkib64 (approved):

PR #2458, opened 2026-09-11
Add warning about onnxruntime version

Menkib64:


Menkib64:
If the fix gets merged, I will add upper limit to the version check. Something like ` && major < 30 && (major != 29  || patch < 1))`

borg323 (approved):

PR #2461, opened 2026-09-13
Move onnx caches to user cache directory

### LINKED ISSUE #2414: lc0 with ONNX backend has hardcoded cache directories in the binary directory

Menkib64:
Fixes: #2414


Menkib64:
This is on top of #2428 which modifies trt cache behavior if it gets merged first.

PR #2462, opened 2026-09-17
dag_classic: remove unused precached_node_ field

ContradNamiseb:
Dead since it was introduced in d3faf5e (add dag-preview search algorithm, upstream LeelaChessZero#2155) -- never read or written anywhere, confirmed against both this branch and upstream/master.


borg323 (approved):
This was unused since the introduction of multigather, and it was missed in an earlier cleanup. It is also in #2318 which will need rebasing anyway, so I'm accepting this.

PR #2465, opened 2026-09-19
build: oneAPI 2026.1 mkl_sycl_blas_dll lookup fallback (sycl= l0)

ContradNamiseb:
## Summary

Configuring with `-Dsycl=l0` fails on oneAPI 2026.1 because the SYCL/MKL block hard-requires the unified `mkl_sycl` library:

```
meson.build:888:19: ERROR: C++ shared or static library 'mkl_sycl' not found
```

oneMKL 2024+ split the unified `mkl_sycl` into per-domain libraries, and oneAPI 2026.1 no longer ships `mkl_sycl` at all. On this installation nothing named `mkl_sycl.lib` exists anywhere under the oneAPI tree, while `mkl_sycl_blas_dll.lib` and `mkl_sycl_dll.lib` do.

## Change

Only the `sycl=l0` block of `meson.build` is touched:

- Try `mkl_sycl_blas_dll`, then `mkl_sycl`, then `mkl_sycl_dll`. Toolkits that still ship the unified library keep working; the domain library is preferred only when present, and it exports the same BLAS interface the engine uses.
- Pass the configured `mkl_libdirs` / `opencl_libdirs` to the respective `find_library` calls.
- Add the configured `mkl_include` directories: `layers.cc.dp.cpp` includes `<oneapi/mkl.hpp>` directly, outside the `-Dmkl` BLAS-backend path, so it needs its own include dirs when `setvars` doesn't populate `INCLUDE` with MKL headers.

## AI disclosure

This change was developed with the assistance of AI coding agents under my direction, and reviewed in detail by me before submission. All findings and test results were verified against the actual build.


meson.build
```diff
@@ -885,11 +885,32 @@ if get_option('build_backends')
...
-        deps += cc.find_library('mkl_core', required: true)
-        deps += cc.find_library('OpenCL', required: true)
+        # layers.cc.dp.cpp includes <oneapi/mkl.hpp> directly, outside the
+        # -Dmkl BLAS-backend path, so it needs its own include directories
+        # (setvars.bat doesn't always populate INCLUDE with MKL headers).
+        # oneMKL 2024+ split the unified mkl_sycl into domain libraries, and
+        # oneAPI 2026.1 no longer ships mkl_sycl at all -- the GEMM symbols
+        # live in mkl_sycl_blas_dll there; the fallbacks keep older toolkits
+        # working.
+        foreach mkl_inc_dir : get_option('mkl_include')
```
borg323, line 895: Are multiple include directories needed?

ContradNamiseb: No, not everyone keeps multiple oneapi dirs like me. Ill change this now in a few hours. Well I will delete the old oneapi versions aswell and only keep only the latest one.

borg323: Can you also use the meson filesystem module instead of `checkdir.py`? It was needed for old versions of meson that we no longer support. The module is already imported, so using `fs.is_dir()` should do it. See the onnx section for an example.

meson.build
```diff
@@ -885,11 +885,32 @@ if get_option('build_backends')
...
-        deps += cc.find_library('mkl_sequential', required: true)
-        deps += cc.find_library('mkl_core', required: true)
-        deps += cc.find_library('OpenCL', required: true)
+        # layers.cc.dp.cpp includes <oneapi/mkl.hpp> directly, outside the
+        # -Dmkl BLAS-backend path, so it needs its own include directories
+        # (setvars.bat doesn't always populate INCLUDE with MKL headers).
+        # oneMKL 2024+ split the unified mkl_sycl into domain libraries, and
+        # oneAPI 2026.1 no longer ships mkl_sycl at all -- the GEMM symbols
+        # live in mkl_sycl_blas_dll there; the fallbacks keep older toolkits
+        # working.
```
borg323, line 894: This comment is too large. Two single line ones further down should do it. Something like:
```shell
# oneapi/mkl.hpp is not always found in the default icx headers.
...
# Various sycl versions ship with different libraries.
```

meson.build
ContradNamiseb: That was my plan aswell, with a drop-in replacement with something like so:
```
     mkl_include = get_option('mkl_include')
     if not fs.is_dir(mkl_include)
       error('mkl_include path does not exist: ' + mkl_include)
     endif
     includes += include_directories(mkl_include)
```

ContradNamiseb:
Addressed the review in 7ae8f45: `mkl_include` is now a single path checked with `fs.is_dir()`, and the long comment is replaced by the two short ones. I also switched the plain `-Dmkl=true` BLAS build from `checkdir.py` to `fs.is_dir()`, since I could test that configuration locally too (it builds and runs).

meson.build
```diff
@@ -885,11 +885,27 @@ if get_option('build_backends')
...
         add_project_arguments('-DMKL_ILP64', language : 'cpp')
-        deps += cc.find_library('mkl_sycl', required: true)
-        deps += cc.find_library('mkl_intel_ilp64', required: true)
-        deps += cc.find_library('mkl_sequential', required: true)
-        deps += cc.find_library('mkl_core', required: true)
-        deps += cc.find_library('OpenCL', required: true)
+        # oneapi/mkl.hpp is not always found in the default icx headers.
+        mkl_include = get_option('mkl_include')
+        if not fs.is_dir(mkl_include)
+          error('mkl_include path does not exist: ' + mkl_include)
```
borg323, line 891: This is not a good idea, a bad path should not break the build. Use a check like
```
cc.has_header('oneapi/mkl.hpp', required: true, include_directories: includes)
```
after adding `mkl_include` to `includes`

meson.build
```diff
@@ -885,11 +885,27 @@ if get_option('build_backends')
...
+        endif
+        includes += include_directories(mkl_include)
+        mkl_libdirs = get_option('mkl_libdirs')
+        opencl_libdirs = get_option('opencl_libdirs')
```
borg323, line 895: Since this is only used one just use `get_option()` directly in `find_library()`, then it is clear that it is only used once and everything else is searched through `mkl_libdirs`.

That being said, is any of this needed? The old version found everything without a `dirs:` option in `find_library()`, did something change?

meson.build
```diff
@@ -278,7 +278,7 @@ if get_option('build_backends')
   if get_option('blas')
     if get_option('mkl') and mkl_lib.found()
       mkl_inc = get_option('mkl_include')
-      if run_command('scripts/checkdir.py', mkl_inc, check : false).returncode() == 0
+      if fs.is_dir(mkl_inc)
         includes += include_directories(mkl_inc)
       endif
       if cc.has_header('mkl.h')
```
borg323, line 284: This probably needs to be `cc.has_header('mkl.h', required: true, include_directories: includes)` but it isn't important.

borg323:
> I also switched the plain `-Dmkl=true` BLAS build from `checkdir.py` to `fs.is_dir()`, since I could test that configuration locally too (it builds and runs).

Did you test whether you can build both blas with mkl and sycl at the same time? If you didn't test both, did you have to change `mkl_include` for the blas build?

ContradNamiseb:
> Did you test whether you can build both blas with mkl and sycl at the same time? If you didn't test both, did you have to change `mkl_include` for the blas build?

Yes, I built `-Dmkl=true` together with `-Dsycl=l0` (and each on its own). All three configure and build, and the BLAS (MKL and DNNL) and SYCL backends give identical evals on 744021. `mkl_include` changing to a single string was driven by the SYCL block. The BLAS build didn't need it, but it works unchanged with a string.

The other suggestions are in a1140d9: a bad `mkl_include` no longer breaks the build, and `has_header(required: true)` now checks `oneapi/mkl.hpp` and `mkl.h` using the include directories. I also dropped all the `dirs:` arguments. You're right that nothing changed: `setvars` puts the MKL and compiler lib directories on `LIB`, and the only 2026.1 change is that the static `mkl_sycl.lib` is gone, which the fallback covers.

borg323 (approved):

